PR #5, opened 2016-07-22
Update README with steps to update local pods repo

daehn:
Update README with steps to update local pods repo if necessary.
Remove CocoaPods version check from setup script as it is installed via bundler.


marcoconti83:
Looks good!

PR #10, opened 2016-07-24
Added .editorconfig

lipis:


marcoconti83:
Thanks for the contribution @lipis , what is this file for?

lipis:
To have some common rules for different editors for consistency :) Read more in http://editorconfig.org

If there different rules for different file types let me know and I'll update it

PR #11, opened 2016-07-24
Display and open originalURL if permanentURL can't be opened by the OS

daehn:
What's in this PR:

* This PR includes changes to the handling of link preview urls.
* There was an issue with some websites providing permanent urls in their Open Graph data that couldn't be opened by the OS (e.g leading `//`), to circumvent this we now fallback to displaying and opening the original URL contained in the text message in that case.
* Remove leading `www.` and trailing slashes from the displayed URL in `ArticleView`.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/ArticleView.swift
```diff
@@ -220,5 +222,39 @@ extension ArticleView : UIGestureRecognizerDelegate {
     func gestureRecognizer(gestureRecognizer: UIGestureRecognizer, shouldReceiveTouch touch: UITouch) -> Bool {
         return !messageLabel.containslinkAtPoint(touch.locationInView(messageLabel))
     }
-    
+
+}
+
+extension LinkPreview {
+
+    var validURL: NSURL? {
```
marcoconti83: Would be nice to add a comment explaining what we mean by "valid".

daehn:
@marcoconti83 re-review please

PR #20, opened 2016-07-26
Enforce line endings and keep binary data

bennyn:
If you're using Git to collaborate with others on GitHub, ensure that Git is properly configured to handle line endings. This is done by enforcing line endings via [gitattributes](https://git-scm.com/docs/gitattributes). 

To not touch line endings of images, they are declared as binary data.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,15 +1,12 @@
 # Reference: https://help.github.com/articles/dealing-with-line-endings/
+# Auto detect text files
+* text=auto
 
-# Enforce LF line endings
-* text eol=lf
+# Files that are truly text-based
+*.h text eol=lf
+*.m text eol=lf
+*.swift text eol=lf
 
-# Files that are truly binary and should not be modified
-*.eot binary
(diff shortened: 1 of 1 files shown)


marcoconti83:
Thanks for the contribution @bennyn . 
We probably have a lot more binary files than those listed (e.g. `m4a`or `.tar.bz`). We should check before accepting this PR.

marcoconti83:
Pasting this here for reference, try running this in the root folder:
`find . -type f -not -path "./Carthage/*" -not -path "./Pods/*" -not -path "./.git/*" | egrep -v "\\.(m|h|swift|json|plist|entitlements|strings)$"`

daehn:
@bennyn please have a look

bennyn:
I updated my pull request. It autodetects now text files and has declarations for pure text/code files and image files (you only use jpg & png). This should keep all line endings to LF.

bennyn:
...but since you only have OS X collaborators, this PR is not necessary.

PR #23, opened 2016-07-26
Camera keyboard follow ups

mikeger:
- Removed unnecessary second confirmation when taking picture/video with integrated camera
- Do not ask for sound recording permissions immediately when opening camera keyboard
- Open camera keyboard even if scrolled back in conversation on tap
- Close keyboard when doing file sharing
- Fetch new images when app becomes active

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
@@ -650,7 +650,6 @@
 {
     if (! self.contentViewController.isScrolledToBottom) {
         [self.contentViewController scrollToBottomAnimated:YES];
-        return YES;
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -650,7 +650,7 @@ - (BOOL)conversationInputBarViewControllerShouldBeginEditing:(ConversationInputB
 {
     if (! self.contentViewController.isScrolledToBottom) {
         [self.contentViewController scrollToBottomAnimated:YES];
-        return controller.inputBar.textView.text.length != 0;
+        return YES;
```
typfel: Can also just delete `return`

PR #24, opened 2016-07-26
Add script for copying extra audio notifications

typfel:
This PR adds an additional script phase to the build steps which copies additional audio notifications files if present.


marcoconti83:
Is `alarm.m4a` supposed to be committed? It feels like it should be with the other files.

typfel:
I'm deleting it here

marcoconti83:
My bad, thought it was added. Thanks, looks good!

PR #25, opened 2016-07-27
Update renamed dependencies

typfel:
iOS repositories were renamed and update to the Cartfile is necessary to be able to fetch binaries.


daehn:
Dependency cycle on `carthage bootstrap` on Travis

typfel:
Yep, I guess we need to update all dependencies first.

marcoconti83:
All dependencies should be updated already, other than `wire-ios-syncengine`

typfel:
Ok, I'm updating `wire-ios-sync-engine`

daehn:
Travis reports success but the tests never ran 🤔

typfel:
Hmm... I'll restart

PR #29, opened 2016-07-27
Update camera keyboard tests

mikeger:
- Due to the issue that Simulator must request permissions for our app to access camera roll testing on CI with real camera roll does not work
- Some tests that where relaying on real camera roll are reverted to fix CI
- As next we need to think on how this issue can be avoided


daehn:
Maybe something like this can help: https://github.com/plu/JPSimulatorHacks?

mikeger:
@daehn thanks, I'll check

Wire-iOS Tests/CameraKeyboardViewControllerTests.swift
```diff
@@ -57,10 +58,12 @@ final class CameraKeyboardViewControllerTests: ZMSnapshotTestCase {
     var sut: CameraKeyboardViewController!
     var splitView: SplitLayoutObservableMock!
     var delegateMock: CameraKeyboardViewControllerDelegateMock!
-    let assetLibrary = AssetLibrary(synchronous: true)
+    var assetLibrary: AssetLibrary!
     
     override func setUp() {
         super.setUp()
+        JPSimulatorHacks.grantAccessToPhotos()
```
daehn: 👍

mikeger:
@daehn @typfel we had a bug with mocking `ZMUser` class that caused crash when spinning run loop in tests. Now it is fixed (`JPSimulatorHacks` spins run loop)

mikeger:
Ok bob is still having issues with it, let's wait until tests can be run on Bob

PR #39, opened 2016-07-28
Remove pixelation view and show regular loading view instead

daehn:
**What's in this PR?**

* As we move to the new assets API there will be no more preview images inlined (web already does not send previews anymore). Because of that the pixelation view has been removed and replaced with the regular loading view (three dots).


mikeger:
@daehn would you consider adding screenshot tests for `ImageMessageCell`? So we can see how it would look now  :)

daehn:
@mikeger Yes, forgot to mention that, I am currently doing this on a separate branch. Will come in a separate PR.

PR #42, opened 2016-07-28
Update the script to download the AVS binary from multiple sources

marcoconti83:
Update the script to download the AVS binary from multiple sources

If the right environment variables are set, the AVS binary can be downloaded from a custom location (GitHub repository that has a release with an attachment).

The variables to set are:
- `AVS_REPO`: the custom "organization/repo" where to find AVS binaries
- `GITHUB_USER `: the GitHub user that has access to that repository (if auth is needed)
- `GITHUB_TOKEN `: the GitHub password or personal access token (if auth is needed)

More variables can be set to override the version, if needed:
- `AVS_VERSION` to download a specific version

### DIFF
--- a/Scripts/download-avs.sh
+++ b/Scripts/download-avs.sh
@@ -97,7 +97,11 @@ else
 	echo "ℹ️  Downloading ${AVS_RELEASE_TAG_PATH}..."
 	
 	# Get tag json: need to parse json to get assed URL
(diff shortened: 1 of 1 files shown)


marcoconti83:
Seems to fail on Travis, will check

marcoconti83:
It seems to downloads successfully now on Travis (see https://travis-ci.org/wireapp/wire-ios/builds/148117922), it can be merged now.

PR #49, opened 2016-07-29
Check if animation is needed when camera change notification arrives

mikeger:
- Camera keyboard observes current camera change
- When user clicks back button with camera keyboard it is not being deallocated
- When user starts to pick image in self profile camera sends current camera change notification, that is received also by camera keyboard
- Camera keyboard tries to animate the camera change and fails unboxing the videoSnapshot
- To avoid this behaviour we verify if animation is needed


daehn:
Why is the camera keyboard not being deallocated?

mikeger:
@daehn whole conversation is not deallocated, and keyboard camera is reatined by conversation. Potentially user can open conversation and continue posting pics, then we also keep the offset in photo stream collection.

daehn:
@mikeger makes sense, I thought it would also be retained when the keyboard is dismissed, but it sounds like it is only the case if it is still opened when the user returns to the conversation list.

PR #52, opened 2016-07-29
Handle the case when NSDataDetector detects URLs wrong gracefully

typfel:
`NSDataDetector` gets confused by URLs prefixed with `abc:` which led to a crash. We avoid this by checking that any URL attachment is in range.

### DIFF
--- a/Wire-iOS Tests/Message+FormattingTests.swift
+++ b/Wire-iOS Tests/Message+FormattingTests.swift
@@ -141,5 +141,16 @@ class Message_FormattingTests: XCTestCase {
         // then
         XCTAssertEqual(formattedText.string, "")
     }
+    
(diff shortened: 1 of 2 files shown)


Wire-iOS Tests/Message+FormattingTests.swift
```diff
@@ -141,5 +141,16 @@ class Message_FormattingTests: XCTestCase {
         // then
         XCTAssertEqual(formattedText.string, "")
     }
+    
+    func testTextWithInvalidLinkAttachment() {
+        // given
+        let textMessageData = createTextMessageData(withMessageTemplate: "hello:{preview-url}") // NSDataDetector gets confused by the this text
```
daehn: Typo in the comment `the this`

PR #67, opened 2016-08-02
Print full xcodebuild logs if build fails

daehn:
**What's in this PR?**

* Write the full xcodebuild output to `build_log.log`.
* Uploads build artifacts to our private S3 bucket, this can be useful if test break and we need to see the full xcodebuild logs.
* Unfortunately travis doesn't support the uploading on PRs yet.


mikeger:
@daehn conflict

daehn:
@mikeger fixed

PR #77, opened 2016-08-05
Download Dangerfile from wire-ios-shared-resources

daehn:
**What's in this PR?**

* As we have quite a couple of repositories and the `Dangerfile` will be the same in all of them we now download the file from https://github.com/wireapp/wire-ios-shared-resources, that way we can add new PR verification steps and they will be applied to all our repositories.


daehn:
Travis-CI timed out while building `wire-ios-sync-engine`, we should attach the framework binary of the new sync-engine release.

PR #78, opened 2016-08-05
Fix/crash picture confirmation

daehn:
**What's in this PR?**

* Fixes [7049](https://wearezeta.atlassian.net/browse/ZIOS-7049), crash when querying the `cameraDevice` when the source type was not camera


daehn:
Recreated against `release/2.12` with https://github.com/wireapp/wire-ios/pull/79

PR #90, opened 2016-08-16
Add UI to edit messages

daehn:
**What's in this PR?**

* Adds a new `UIView` subclass `InputBarEditView` which displays the buttons to confirm, cancel or undo an ongoing message edit.
* The layout logic of the input bar buttons has been extracted in a separate class as the `InputBar` grew quite big with the new editing view also in there.
* If the edit is empty the message will be deleted for everyone.

‼️ This PR depends on:
- [x] A new `wire-ios-data-model` release
- [x] A new `wire-iso-sync-engine` release


basine:
This PR still needs ZMCDataModel and zmessaging release.

Wire-iOS/Sources/Helpers/syncengine/Message.m
```diff
@@ -98,43 +98,50 @@ + (NSString *)formattedReceivedDateForMessage:(id<ZMConversationMessage>)message
...
-    
-    NSString *timestampString = nil;
+
+    [longVersionDateFormatter setDateStyle:NSDateFormatterFullStyle];
+    [longVersionDateFormatter setTimeStyle:NSDateFormatterNoStyle];
+    NSString *dateString = [longVersionDateFormatter stringFromDate:timestamp];
+
+    [longVersionDateFormatter setDateStyle:NSDateFormatterNoStyle];
+    [longVersionDateFormatter setTimeStyle:NSDateFormatterShortStyle];
+    NSString *timeString = [longVersionDateFormatter stringFromDate:timestamp];
```
Einh06: Why not doing the date formatter configuration in the dispatch once? It cannot be used outside of the scope of this function, so the configuration can effectively happen only once.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
```diff
@@ -199,7 +216,8 @@ - (void)configureForMessage:(id<ZMConversationMessage>)message layoutProperties:
     [self.messageTextView layoutIfNeeded];
     self.messageTimestampView.timestampLabel.text = [Message formattedReceivedDateLongVersion:self.message];
     self.textViewHeightConstraint.active = attributedMessageText.length == 0;
-    
+    self.editedImageView.hidden = nil == self.message.updatedAt;
```
Einh06: Sorry, have to express my OCD. This expression would be clearer with brackets I think. The `... = nil ==` caught me, had to rethink twice the expression.
Additionally, it isn't consistent with the above statement as well.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
```diff
@@ -341,6 +372,15 @@ - (void)copy:(id)sender
     }
 }
 
+- (void)edit:(id)sender;
+{
+    if([self.delegate respondsToSelector:@selector(conversationCell:didSelectAction:)]) {
+        self.contentEditing = YES;
+        [self.delegate conversationCell:self didSelectAction:ConversationCellActionEdit];
+        // TODO: Add tracking
```
Einh06: Is this TODO still up-to-date

Wire-iOS/Sources/Helpers/syncengine/Message.m
daehn: We reuse the same date formatter for the `dateString` and `timeString` and modify the styles in between.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
daehn: It's still up to date, @basine is working on it ATM (will come in a separate PR), so the warning from @wire-bot can be ignored.

mikeger:
@daehn please check tests on Travis

Wire-iOS/Sources/Helpers/syncengine/Message.m
mikeger: Is it expensive to create formatter or to set style?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.h
```diff
@@ -85,6 +88,7 @@ typedef void (^SelectedMenuBlock)(BOOL selected, BOOL animated);
 
 @property (nonatomic) CGFloat burstTimestampSpacing;
 @property (nonatomic) BOOL showsMenu;
+@property (nonatomic) BOOL contentEditing;
```
mikeger: Sounds a little bit hard to understand on my English, how about `beingEdited`? May be @marcoconti83 can propose a variant

Wire-iOS/Sources/UserInterface/Conversation/ConversationRootViewController.swift
```diff
@@ -140,11 +140,13 @@ public extension ConversationViewController {
     }
     
     func voiceCallItemTapped(sender: UIBarButtonItem) {
+        NSNotificationCenter.defaultCenter().postNotificationName(ConversationInputBarViewControllerEndEditingNotification, object: nil)
```
mikeger: I guess it is intended to do the same in case of incoming call. I think you can observe voice channel state.

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -577,6 +578,15 @@ - (void)conversationContentViewController:(ConversationContentViewController *)c
     }];
 }
 
+- (void)conversationContentViewController:(ConversationContentViewController *)contentViewController didTriggerEditingMessage:(ZMMessage *)message
+{
+    NSString *text = message.textMessageData.messageText;
+    
+    if (nil != text) {
```
mikeger: I am not sure about the logic here. Should it check for zero length?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -19,44 +19,66 @@
...
+    var isEditing: Bool {
+        if case .Editing(_) = inputbarState { return true }
+        return false
+    }
     
-    private let contentLeftMargin = CGFloat(WAZUIMagic.floatForIdentifier("content.left_margin"))
-    private let contentRightMargin = CGFloat(WAZUIMagic.floatForIdentifier("content.right_margin"))
-    private let iconSize = UIImage.sizeForZetaIconSize(.Tiny)
-    private let buttonMargin = (CGFloat(WAZUIMagic.floatForIdentifier("content.left_margin")) / 2) - (UIImage.sizeForZetaIconSize(.Tiny) / 2)
+    var inputbarState: InputBarState = .Writing {
```
mikeger: I think `Bar` in `inputbarState` should be capitalised

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -110,6 +110,10 @@ @interface  ConversationInputBarViewController (UIGestureRecognizerDelegate) <UI
 
 @end
 
+
+NSString * const ConversationInputBarViewControllerEndEditingNotification = @"ConversationInputBarViewControllerEndEditingNotification";
```
mikeger: From the name of the notification it is not clear if `ConversationInputBarViewController` sends it as a result of it's state change or it is a public API to stop editing. I would vote to dependency-inject `ConversationInputBarViewController` instance to objects that can terminate editing, or, alternatively making a static method on `ConversationInputBarViewController` to stop editing

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBarButtonsView.swift
```diff
@@ -0,0 +1,233 @@
...
+    }
+    
+    func createConstraints() {
+        constrain(self, buttonContainer)  { view, buttonRow in
+            self.buttonRowTopInset = view.top == buttonRow.top
+            buttonRow.left == view.left
+            buttonRow.right == view.right
+            buttonRowHeight = buttonRow.height == 0
+            view.height == constants.buttonsBarHeight
+            view.width == 414 ~ 750
```
mikeger: Just wondering, why the default width constraint is needed?

mikeger:
@daehn great job, really looking forward to start using it

Wire-iOS/Sources/Helpers/syncengine/Message.m
daehn: Yes, creating the formatter is slow, and we use it quite often here, we could even create two with the different styles so we would't need to reconfigure them.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.h
daehn: `beingEdited` sounds good to me 👍

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
daehn: This should just be a sanity check that caller did not pass a non text message. Maybe asserting makes more sense. In theory there should never be a text message with an empty string, but still might make sense to check.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBarButtonsView.swift
daehn: Was there before, yes. I assume it's iPad related.

Wire-iOS/Sources/Helpers/syncengine/Message.m
daehn: Just read up on this and apparently setting the format is very expensive as well, so we should create two static formatters.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.h
marcoconti83: Or just switch it around, 'editingContent'. 'beingEdited' works too though.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBarButtonsView.swift
mikeger: Indeed!

daehn:
Somehow GitHub is stuck checking the mergability and travis does not pick up the recent pushes to the PR...

PR #91, opened 2016-08-16
Only show the delete for everyone option if self is still a member of the conversation

daehn:
**What's in this PR?**

* [7085](https://wearezeta.atlassian.net/browse/ZIOS-7085): Message should only be deleted if the selfUser is still an active member of the conversation

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
@@ -20,7 +20,7 @@
 extension ConversationContentViewController {
 
     func presentDeletionAlertController(forMessage message: ZMConversationMessage) {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
```diff
@@ -20,7 +20,7 @@
 extension ConversationContentViewController {
 
     func presentDeletionAlertController(forMessage message: ZMConversationMessage) {
-        let showDelete = message.sender?.isSelfUser ?? false
+        let showDelete = message.sender?.isSelfUser ?? false && conversation.isSelfAnActiveMember
```
marcoconti83: I know that the precedence of operators is well defined and it works in this case, but just for safety and readability I would add () around expressions.
`(message.sender?.isSelfUser ?? false) && conversation.isSelfAnActiveMember`

daehn:
@marcoconti83 re-review please

PR #94, opened 2016-08-17
Add second curtain

daehn:
**What's in this PR?**

* Adds [second curtain](https://github.com/AshFurrow/second_curtain), to upload failing snapshots to S3.


marcoconti83:
Cool! Let's wait until the internal is out before merging this in

daehn:
This works now and uploads the failing snapshots to our S3 bucket. As the bucket currently is still private the content can be synced and looked at locally. If we can make the bucket world readable  we could post the link to the generated HTML using Danger.

PR #96, opened 2016-08-17
Updated open source avs version to 2.8.50

sorenskak:


mikeger:
@sorenskak if you check in open-source repo https://github.com/wireapp/avs-ios-binaries/releases the new release does not follow the old-new versioning, adding `2.8.` prefix to the version number. This requires the change also in our scripts. I changed it so it works.

PR #98, opened 2016-08-17
Add tracking for deleting and editing

basine:


ConfigurationSets/Development.xcconfig
```diff
@@ -18,7 +18,7 @@
 
 
 #include "Base.xcconfig"
-#include "Private/Development_private.xcconfig"
+#include "Development_private.xcconfig"
```
mikeger: Why this stuff moved?

mikeger:
@basine please check with @daehn if the PR should point to `develop` or `release/2.14`

ConfigurationSets/Development.xcconfig
basine: There is no such directory. My guess is that it used to be there and we removed it with open sourcing, but forgot to update the file. I added a new file manually, but we want to copy the file from the other depository in with an internal setup script.

mikeger: Yep sure no such directory in open-source repo because it contains xcconfigs with app store configuration and API keys stuff. I think the original idea was that if you need one of those configs you create `Private` folder in xcconfigs and copy them from https://github.com/wireapp/wire-ios-build-assets so I guess there is no need to change the configs location so they are not mixed together with other configs

PR #101, opened 2016-08-17
Review localytics implementation

basine:
**What's in this PR**

We were not checking if we were currently in the background when opening the session on initialisation. Other than that there were no critical implementation errors.


marcoconti83:
Thanks for checking 👍

PR #102, opened 2016-08-17
Track editing and deleting messages

basine:


.gitignore
```diff
@@ -21,3 +21,4 @@ Carthage/
 Wire-iOS/ComponentsVersions.plist
 /Libraries
 /ConfigurationSets/Private
+/ConfigurationSets/Development_private.xcconfig
```
mikeger: I guess this is also not necessary for the moment

mikeger:
@basine weird, seems no UI changes but some tests are failing @daehn do we have tests failing on `release/2.14`?

marcoconti83:
@daehn any comment on this?

daehn:
I needed to add an exception for the release branch to build on pushes on Travis, which it now does. But https://github.com/wireapp/wire-ios/pull/106 which is branched of the release branch succeeds.

mikeger:
@marcoconti83 @basine what about redundant gitignore changes?

.gitignore
marcoconti83: This was added because we do not want to commit this by accident

marcoconti83:
@mikeger that ignore was done because we want to avoid committing that file accidentally after a session of debugging

mikeger:
@marcoconti83 we cannot commit it by accident because the import declaration for this file is looking on the different location.

PR #105, opened 2016-08-17
update avs version to 2.8.56

sorenskak:


marcoconti83:
This needs to be ported to release branch eventually

PR #108, opened 2016-08-18
Fix edit view width on iPad

daehn:
**What's in this PR?**

* The editing view had the wrong width on tablets.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
@@ -23,7 +23,6 @@ import Classy
 import WireExtensionComponents
 
 
-
 public enum InputBarState: Equatable {
     case Writing
     case Editing(originalText: String)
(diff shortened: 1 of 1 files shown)


mikeger:
@daehn should we have a screenshot test for it?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -192,20 +191,20 @@ private struct InputBarConstants {
...
+            container.left == container.superview!.left
+            container.right == container.superview!.right
+            container.height == constants.buttonsBarHeight
 
-            container.leading == buttonBox.leading
-            container.trailing == buttonBox.trailing
-            self.rowTopInsetConstraint = container.top == buttonBox.top - constants.buttonsBarHeight
+            innerContainer.leading == container.leading
+            innerContainer.trailing == container.trailing
+            self.rowTopInsetConstraint = innerContainer.top == container.top - constants.buttonsBarHeight
```
mikeger: So self contains `buttonContainer` that contains `buttonInnerContainer` that contains `buttonsView`. 
Then in first constrain block `buttonInnerContainer` is mentioned as `container`, and in the second as `innerContainer`, whereas `buttonContainer` is now a `container` (also in first block `buttonRow` is `buttonContainer`, but this is a leftover from old copy-paste).
For my understanding makes it unnecessary complex for understanding, two considerations here:
- Strictly name elements the same in constrain block same as properties
- Rename `buttonContainer` and `buttonInnerContainer` by their purpose: I believe `buttonInnerContainer` is used as a clipping view, and `buttonContainer` itself is used for something else

daehn: Good point, I renamed a couple of views there and forgot about the constaint closure parameters.

The `buttonContainer` contains and clips the `buttonInnerContainer`. The inner container contains the edit view and the buttons view.

PR #109, opened 2016-08-18
Don't scroll to the bottom of the conversation when editing

daehn:
**What's in this PR?**

* When the input bar becomes first responder we scroll the conversation to the bottom, we don't want this behavior when editing a message.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
@@ -586,14 +586,11 @@
 
 - (BOOL)textViewShouldBeginEditing:(UITextView *)textView
 {
-    BOOL isEditing = (nil != self.editingMessage);
-    
     if (self.mode == ConversationInputBarViewControllerModeAudioRecord) {
         return YES;
     }
-    else if (!isEditing && [self.delegate respondsToSelector:@selector(conversationInputBarViewControllerShouldBeginEditing:)]) {
-        // This call scrolls to the bottom of the conversation, we don't want this behaviour if we are editing a message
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -586,10 +586,13 @@ - (BOOL)textView:(UITextView *)textView shouldChangeTextInRange:(NSRange)range r
...
 - (BOOL)textViewShouldBeginEditing:(UITextView *)textView
 {
+    BOOL isEditing = (nil != self.editingMessage);
+    
     if (self.mode == ConversationInputBarViewControllerModeAudioRecord) {
         return YES;
     }
-    else if ([self.delegate respondsToSelector:@selector(conversationInputBarViewControllerShouldBeginEditing:)]) {
+    else if (!isEditing && [self.delegate respondsToSelector:@selector(conversationInputBarViewControllerShouldBeginEditing:)]) {
+        // This call scrolls to the bottom of the conversation, we don't want this behaviour if we are editing a message
```
mikeger: Would be better in my opinion to move the logic to `ConversationViewController.m` where the delegate method is implemented, here two logical points
- It's up to the delegate what to do when editing starts, and comment can become irrelevant when implementation in `ConversationViewController` change
- It's not defined in protocol that when you start editing the existing message the delegate call would not be send

daehn:
@mikeger please re-review

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
daehn: Yes, good point. I moved the logic and am passing the editing state to the `ConversationViewController` now as it does not have access to the `editingMessage`.

PR #112, opened 2016-08-18
Fix being able to send empty edits and sending unedited content from return key

daehn:
**What's in this PR?**

* [7093](https://wearezeta.atlassian.net/browse/ZIOS-7093): "Message edited with spaces isn't shown as deleted"
* [7094](https://wearezeta.atlassian.net/browse/ZIOS-7094): "Message edited with space is impossible to send by tapping on send button on the keyboard" 
* [7095](https://wearezeta.atlassian.net/browse/ZIOS-7095): "Not edited but submited message is shown as edited"

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
@@ -561,10 +561,11 @@
 
         NSString *candidateText = textView.text;
         candidateText = [candidateText stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
-        BOOL conversationWasNotDeleted  = self.conversation.managedObjectContext != nil;
+        BOOL conversationWasNotDeleted = self.conversation.managedObjectContext != nil;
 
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -558,10 +558,19 @@ - (void)textViewDidChange:(UITextView *)textView
...
 {
     if ([text isEqualToString:@"\n"]) {
-        
+
         NSString *candidateText = textView.text;
         candidateText = [candidateText stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
         BOOL conversationWasNotDeleted  = self.conversation.managedObjectContext != nil;
+
+        if (self.inputBar.isEditing && nil != self.editingMessage) {
+            if (![candidateText isEqualToString:self.editingMessage.textMessageData.messageText]) {
```
marcoconti83: I would also trim the `self.editingMessage.textMessageData.messageText` otherwise it might not match if someone sent you a text with a space at the end

daehn:
@marcoconti83 please re-review, I also added trimming of the edited text to avoid enabling the confirm button in case the only changes are trailing spaces

PR #137, opened 2016-08-29
Delivery state basic UI

mikeger:
- Initial release, contains logic to show the undelivered message state with timestamp
- Also contains some code for likes

Main changes:
- Removed current unsent indicator together with frame animation and `FrameAnimationView`
- Timestamp lives now in root message cell class `ConversationCell`
- Timestamp is always shown for last message in conversation
- Resend is done with the link in timestamp / status area


daehn:
bac6f507190259aefbf8ccb4739f9f72b6da4245ca22e93f90d3bb3391500f54

mikeger:
TODO's are by design, would be removed with likes implementation

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -730,7 +730,12 @@ - (NSIndexPath *)tableView:(UITableView *)tableView willSelectRowAtIndexPath:(NS
 
 - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
 {
-    [self presentDetailsForMessageAtIndexPath:indexPath];
+    id<ZMConversationMessage>message = [self.messageWindow.messages objectAtIndex:indexPath.row];
+    BOOL isFile = [Message isFileTransferMessage:message];
+    
+    if (! isFile) {
+        [self presentDetailsForMessageAtIndexPath:indexPath];
```
daehn: There is a check if this message is a file message inside `presentDetailsForMessageAtIndexPath` that could be removed or could be inverted now.

mikeger: True, I also disliked it when wrote, but `presentDetailsForMessageAtIndexPath` is called from different places, here we don't need to call if for files, but when user clicks button on the cell we would call it for file or other media.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
```diff
@@ -138,7 +138,7 @@ - (void)conversationWindowDidChange:(MessageWindowChangeInfo *)change
         // Unless a message was appended to the bottom of the conversation we need re-configure the cells
         // since they might (not) need to display an name & avatar after the insertion/deletion/move.
         BOOL messageWasInsertedAtBottom = change.insertedIndexes.count == 1 && change.insertedIndexes.firstIndex == 0;
-        if (! messageWasInsertedAtBottom || change.deletedIndexes.count > 0 || change.movedIndexPairs.count > 0) {
+        if (messageWasInsertedAtBottom || change.deletedIndexes.count > 0 || change.movedIndexPairs.count > 0) {
```
typfel: This logic will only re-configure cells which are added to the bottom, which is not what we want. I think you should replace `messageWasInsertedAtBottom` with `change.insertedIndexes.count > 0`

mikeger:
Currently does not compile, waiting for sync engine release

marcoconti83:
Conflict!

PR #138, opened 2016-08-30
Feature/address book automation helper

marcoconti83:
- Make sure we use the new sync-engine upload API properly
- Command-line injectable search delay (for automation testing)
- Removing dead code

# Changes
- more Swift!
- Using the "search" terminology instead of "upload" everywhere (as the back end uses the contacts as a search query, not to store them)
- `addressBookUploadWasPostponed` was never used anywhere, so it was removed
- `isAddressBookSupported` was removed, as this is only relevant for iOS 6 and early, we support only 8+
- decoupling `AddressBookHelper` from `AutomationHelper`, now the link between the two is done externally (property injection)

# Testing
No testing is present for this on the UI. Will consider adding in future.


Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
```diff
@@ -198,19 +197,16 @@ - (void)handleUploadAddressBookLogicIfNeeded
     
     self.addressBookUploadLogicHandled = YES;
     
-    if (! [[AddressBookHelper sharedHelper] isAddressBookSupported]) {
-        return;
-    }
-    else if ([[AddressBookHelper sharedHelper] isAddressBookAccessGranted]) {
```
marcoconti83: This `isAddressBookAccessGranted` was removed as it is now checked inside the `startRemoteSearch` method

WireExtensionComponents/Utilities/AutomationHelper.swift
```diff
@@ -24,70 +24,112 @@ import zmessaging
...
+        }
+        return NSTimeInterval(delay)
     }
 }
+
+// MARK: - Helpers
+extension String {
+    
+    /// Returns the value of the command line argument if it is in the form of "--ARGUMENT=VALUE"
+    func valueIfPresent(commandLineArgument: String) -> String? {
```
mikeger: I don't think this method is called

WireExtensionComponents/Utilities/AutomationHelper.swift
```diff
@@ -24,70 +24,112 @@ import zmessaging
...
+    }
+}
+
+/// Command line arguments
+private struct Arguments {
+    
+    let flagPrefix = "--"
+    
+    /// Argument strings
+    let commandLineArguments : Set<String>
```
mikeger: I think it would be more convenient to keep it as a key-value, but it's not crucial.

Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
```diff
@@ -198,19 +197,16 @@ - (void)handleUploadAddressBookLogicIfNeeded
...
+        // Re-check if we need to start AB search
+        [[AddressBookHelper sharedHelper] startRemoteSearch:true];
     }
     else if ([[AddressBookHelper sharedHelper] isAddressBookAccessUnknown]) {
         [[AddressBookHelper sharedHelper] requestPermissions:^(BOOL success) {
             [self.analyticsTracker tagAddressBookSystemPermissions:success];
             if (success) {
                 dispatch_async(dispatch_get_main_queue(), ^{
-                    [[AddressBookHelper sharedHelper] uploadAddressBook];
+                    [[AddressBookHelper sharedHelper] startRemoteSearch:true];
```
mikeger: I think in Obj-C it is more common to use `YES/NO` for `BOOL` as value

Wire-iOS/Sources/Managers/AddressBookHelper.swift
```diff
@@ -0,0 +1,143 @@
...
+            NSUserDefaults.standardUserDefaults().setBool(newValue, forKey: addressBookSearchWasProposedKey)
+        }
+    }
+}
+
+// MARK: - Upload
+extension AddressBookHelper {
+    
+    /// Starts an address book search, if enough time has passed since last search
+    public func startRemoteSearch(onlyIfEnoughTimeSinceLast: Bool) {
```
mikeger: I think when this method is exported to Obj-C it misses the `onlyIfEnoughTimeSinceLast` part, therefore it is hard to understand what first `BOOL` parameter means.

WireExtensionComponents/Utilities/AutomationHelper.swift
marcoconti83: You mean `[String: AnyObject]`?
I thought about it too but it means that I will try upfront to parse all them as optional arguments (--arg=1) even if they are just flags (--arg) just to fill up this dict, and that later I would need to cast anyway from AnyObject to bool or string.

Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
marcoconti83: True... I mean, YES

WireExtensionComponents/Utilities/AutomationHelper.swift
mikeger: I also thought now about possible anonymous arguments, which would not really work with regular key-value

marcoconti83:
- Removed dead code
- Use YES/NO
- Renamed method for Obj-c

@mikeger please re-review

PR #147, opened 2016-09-02
Revert wrapping liking in enqueue changes, as this is already done a level higher

daehn:
# What's in this PR?

* Revert https://github.com/wireapp/wire-ios/pull/144 as the action already is enqueued on level higher


marcoconti83:
The hunt for the bug continues... 🔍

PR #149, opened 2016-09-02
Only show options to like a message if the message is sent or delivered

daehn:
# What's in this Pull Request?

* Only show the `likeButton` when the message has been sent or delivered, similarly also only show the `UIMenuItem` to like / unlike and let the user double tap to like a message if it is in one of these states.
* Only let the user open the `ReactionsViewController` when the message has been liked by users.
* I am preparing a separate Pull Request on `wire-ios-data-model` to ensure a message can not be liked unless it is at least sent on the model level.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ZMMessage+Likeable.swift
```diff
@@ -0,0 +1,37 @@
...
+}
+
+
+public extension ZMMessage {
+
+    class func messageCanBeLiked(message: ZMMessage) -> Bool {
+        return message.canBeLiked
+    }
+
+}
```
typfel: 😞

daehn:
@typfel @marcoconti83 please re-review

daehn:
The mentioned PR for `wire-iso-data-model`: https://github.com/wireapp/wire-ios-data-model/pull/33

PR #151, opened 2016-09-05
Restrict liking to `normal` message types

daehn:
# What's in this PR?

* Restrict liking to `normal` message types as liking of the "deleted" system message was possible and the like button was shown.


mikeger:
@daehn correction, you cannot like a knock

daehn:
But `isNormalMessage` checks if it is a knock?

PR #159, opened 2016-09-05
Don't show like icon on your last sent message

typfel:
## What's in this PR

We don't want to show the heart icon for the last message that you sent since it causes visual clutter. You should of course still be able to see likes on the last message so the following rule determine when the icon should be visible:

We want to hide the icon if it's the last message you sent <b>and</b> it isn't already like by someone <b>and</b> you haven't selected the message.


daehn:
Would be nice to have snapshot tests for the different cases when it should be visible as well.

typfel:
@daehn added tests for the selected/non-selected cases. The remaining cases  requires mocking of likes.

PR #163, opened 2016-09-05
Timestamp change animation for reactions

mikeger:
- Because animations are fun

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Helpers/UIView+SlideInState.swift
+++ b/Wire-iOS/Sources/UserInterface/Helpers/UIView+SlideInState.swift
@@ -24,7 +24,7 @@ public enum SlideDirection: UInt {
 }
 
 public extension UIView {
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Helpers/UIView+SlideInState.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import Foundation
+
+public enum SlideDirection: UInt {
+    case Up
+    case Down
+}
+
+public extension UIView {
+    func wr_animateTo(direction: SlideDirection = .Down, newState: ()->()) {
```
daehn: When reading the call to this method I didn't understand what it does because the `direction` was omitted as it was using the default parameter value, potentially it would be clearer to rename it to something containing slide or sliding.

mikeger:
@daehn updated

PR #168, opened 2016-09-05
Re-enable snapshot tests for reactions

daehn:
# What's in this PR?

* Re-enable snapshot tests for reactions / likes


typfel:
Just noticed that the tiny avatars initials are too big

PR #169, opened 2016-09-05
Track like actions

basine:
... and some refactoring

### DIFF
--- a/Wire-iOS.xcodeproj/project.pbxproj
+++ b/Wire-iOS.xcodeproj/project.pbxproj
@@ -783,6 +783,9 @@
 		F517A3BE1B31340700073990 /* SmoothLineView.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F517A3BD1B31340700073990 /* SmoothLineView.framework */; };
 		F5EAB12B1B120665008DB66A /* SketchBottomView.m in Sources */ = {isa = PBXBuildFile; fileRef = F5EAB12A1B120665008DB66A /* SketchBottomView.m */; };
 		F5F52F721B11D7050053F1C8 /* SketchViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = F5F52F711B11D7050053F1C8 /* SketchViewController.m */; };
+		F93D58431D7D9184003AE972 /* Analytics+ConversationEvents+Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = F93D58421D7D9184003AE972 /* Analytics+ConversationEvents+Swift.swift */; };
+		F93D58451D7DA865003AE972 /* ZMConversation+Analytics.swift in Sources */ = {isa = PBXBuildFile; fileRef = F93D58441D7DA865003AE972 /* ZMConversation+Analytics.swift */; };
+		F93D58471D7DA876003AE972 /* ZMMessage+Analytics.swift in Sources */ = {isa = PBXBuildFile; fileRef = F93D58461D7DA876003AE972 /* ZMMessage+Analytics.swift */; };
 		F95E08B41D0EAFE80098A9EE /* TTTAttributedLabel+Links.swift in Sources */ = {isa = PBXBuildFile; fileRef = F95E08B31D0EAFE80098A9EE /* TTTAttributedLabel+Links.swift */; };
 /* End PBXBuildFile section */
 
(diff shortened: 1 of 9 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
```diff
@@ -38,16 +38,32 @@ public extension ConversationCell {
...
-    @objc public func likeMessage(button: AnyObject!) {
+    @objc public func likeMessage(sender: AnyObject!) {
         guard message.canBeLiked else { return }
         self.messageToolboxView.forceShowTimestamp = false
+        let reactionType : ReactionType = message.liked ? .Unlike : .Like
+        trackReaction(sender, reaction: reactionType)
 
-        ZMUserSession.sharedSession().performChanges {
-            self.message.liked = !self.message.liked
+        ZMUserSession.sharedSession().performChanges { [weak self] in
```
mikeger: Curious, why it is necessary to weakify self here?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -2079,6 +2082,9 @@
 		F5EAB12A1B120665008DB66A /* SketchBottomView.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SketchBottomView.m; path = Sketchpad/SketchBottomView.m; sourceTree = "<group>"; };
 		F5F52F701B11D7050053F1C8 /* SketchViewController.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = SketchViewController.h; path = Sketchpad/SketchViewController.h; sourceTree = "<group>"; };
 		F5F52F711B11D7050053F1C8 /* SketchViewController.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; name = SketchViewController.m; path = Sketchpad/SketchViewController.m; sourceTree = "<group>"; };
+		F93D58421D7D9184003AE972 /* Analytics+ConversationEvents+Swift.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "Analytics+ConversationEvents+Swift.swift"; sourceTree = "<group>"; };
```
mikeger: I think that the file extension is already implying that the file is a swift addition.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
marcoconti83: It's not necessary, but we talked about this in general and said that it does not hurt to put weakify in blocks as a habit, even if it's not needed in this point specifically.

Wire-iOS/Sources/Analytics/Events/ZMMessage+Analytics.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+extension ZMMessage {
+    
+    public class func analyticsAttribute(message: ZMConversationMessage) -> String {
+        let messageType = Message.messageType(message)
+        return analyticsAttribute(withMessageType: messageType)
+    }
+    
+    public class func analyticsAttribute(withMessageType messageType: MessageType) -> String {
```
mikeger: For me `analyticsAttribute` sounds too broad, I would prefer to have a more specific name, like `analyticsType`.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
mikeger: Talked -- but not with me, I see no point in this.

Wire-iOS/Sources/Analytics/Events/ZMMessage+Analytics.swift
```diff
@@ -0,0 +1,50 @@
...
+        case .Unknown:      return "unknown"
+        case .Text:         return "text"
+        case .Image:        return "image"
+        case .Audio:        return "audio"
+        case .Video:        return "video"
+        case .RichMedia:    return "rich_media"
+        case .Ping:         return "ping"
+        case .File:         return "file"
+        case .System:       return "system"
+        case .Location:     return "location"
```
mikeger: Would be a bit more elegant to have a mapping dict MessageType->String, but already a good improvement.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
typfel: I thought the outcome was the we should make it clear in the documentation of `performChanges` that it's fine not use weak.

Wire-iOS/Sources/Analytics/Events/ZMMessage+Analytics.swift
basine: Yeah, but we return a String, so `type` is also weird.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
daehn: The comment was added for `NSManagedObjectContext.performGroupedBlock(_:)`, but not yet for the equivalent user session methods.

Wire-iOS/Sources/Analytics/Events/ZMMessage+Analytics.swift
mikeger: You are right, then may be `analyticsTypeString` ?

daehn:
💣 conflicts

Wire-iOS/Sources/Analytics/Events/ZMMessage+Analytics.swift
basine: Well in the optimal case we would only use Swift and use a String enum for `MessageType`. I think that would make sense to use `MessageType` only internally in Analytics, call the tracking methods with the message instead and do the mapping in a private method.

PR #173, opened 2016-09-05
Enable voice logging on Release builds

mikeger:
- Currently when building for AppStore we use `Release.xcconfig`
- The voice logs is using flag `1<<10` = bit 10 set
- Current log level mask is set to `0x11` = bit 0 and 4 set
- Solution: set debug log mask to `0x411` setting bits 0, 4 and 10


lipis:
I

lipis:
So my pocket decided to close a few issues last night.. :) Maybe I'll switch to Nokia again..

PR #175, opened 2016-09-05
Animate image moving down into the photo button on save

daehn:
# What's in this PR?

* Animate the image moving into the photo button when saving (the animation is the same as previously in `FullScreenImageViewController`).
* Fix close button alignment in `FullScreenImageViewController`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
@@ -65,7 +65,6 @@
 #import "AddContactsViewController.h"
 #import "ContactsDataSource.h"
 #import "VerticalTransition.h"
-#import "UIView+MTAnimation.h"
 
 #import "UIColor+WAZExtensions.h"
 #import "KeyboardFrameObserver+iOS.h"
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -610,9 +611,22 @@ - (BOOL)conversationContentViewController:(ConversationContentViewController *)c
...
+- (void)conversationContentViewController:(ConversationContentViewController *)contentViewController performImageSaveAnimation:(UIView *)snapshotView sourceRect:(CGRect)sourceRect
 {
-    [self.inputBarController bounceCameraIcon];
+    [self.view addSubview:snapshotView];
+    snapshotView.frame = [self.view convertRect:sourceRect fromView:contentViewController.view];
+
+    UIView *targetView = self.inputBarController.photoButton;
+    CGPoint targetCenter = [self.view convertPoint:targetView.center fromView:targetView.superview];
+    
+    [UIView mt_animateWithViews:@[snapshotView] duration:0.55 timingFunction:kMTEaseInExpo animations:^{
```
typfel: Why are you using MT animations? Would like to remove that dependency at some point.

daehn: Ah makes sense, its a relict of the previous implementation, I will replace it.

daehn:
@typfel updated

PR #176, opened 2016-09-05
Reduce number of times we configure the message tool box

typfel:
## What does this solve?
Currently when a cell is prepared to be rendered or when we like it, we configure the message toolbox view multiple times along different code paths. This makes it hard to reason about the code. This PR attempts to solve this by re-factoring code to only have one code path for updating the message tool box.

## What changed
There's only one method for configuring the tool box:
```
configureForMessage(message: ZMConversationMessage, forceShowTimestamp: Bool, animated: Bool = false)
```
which is called from `updateToolboxVisibilityAnimated:`


mikeger:
@typfel great job! We can now potentially remove checks of is the label value new in the `MessageToolboxView` : `if let currentText = self.statusLabel.attributedText where currentText.string == attributedText.string`

mikeger:
I found a small bug: if you have a liked message and re-enter the conversation (cells are reloaded) the cell does not show the people who liked list, only heart

mikeger:
@typfel please merge when ready

typfel:
There also an issue with this PR in how the selection handled when liking a message.

typfel:
@daehn @mikeger changed the selection logic when liking/unliking a message.

On like we make sure the cell is unselected to on order to show the list of likes.
*and*
On unlike we make sure the cell is selected to prevent it from collapsing.

PR #178, opened 2016-09-06
Remove delivered state

mikeger:
- Delivered disabled until the majority would send the receipts


mikeger:
@typfel I think I forgot tests 🤔

typfel:
😱

PR #181, opened 2016-09-06
Added like tooltip

mikeger:
- Like tooltip is supposed to help users understand that there is a new like button
- Also decreased initials size in small people who liked list


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -275,8 +276,8 @@ - (void)createBaseConstraints
         self.toolboxHeightConstraint = [self.messageToolboxView autoSetDimension:ALDimensionHeight toSize:0];
     }];
     [self.messageToolboxView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.messageContentView];
-    [self.messageToolboxView autoPinEdgeToSuperviewMargin:ALEdgeRight];
-    [self.messageToolboxView autoPinEdgeToSuperviewMargin:ALEdgeLeft];
+    [self.messageToolboxView autoPinEdgeToSuperviewEdge:ALEdgeRight];
```
typfel: Why was this change necessary?

mikeger: It is because the arrow on left is overflowing the timestamp to the left, and `MessageToolboxView` is clipping subviews, so it is not visible if out of bounds. The solution is to make toolbox go from side to side, but adopt the margins inside of it.

typfel: Ah right! 👍

PR #190, opened 2016-09-07
Use existing modal top bar component to ensure iPad bar has correct height

daehn:
# What's in this PR?

* Use existing `ModalTopBar` component to ensure the top bar in the `ReactionsListViewController` has the correct height on all devices.


mikeger:
@daehn should the screenshot tests be updated?

daehn:
@mikeger good point, on it.

marcoconti83:
Is this for RC?

daehn:
@marcoconti83 not necessarily, just a visual fix, could also go into `develop`

daehn:
@mikeger updated, should change the base or are we putting it in the RC?

PR #191, opened 2016-09-07
Expect correct sender for menu action

basine:


daehn:
@basine should this go into the RC?

basine:
@daehn I guess so

daehn:
@basine If I change the base here on GitHub then there a a couple of old commits in the PR

PR #194, opened 2016-09-07
Fix for travis: Sync engine that tolerates no shared directory

mikeger:
Requires https://github.com/wireapp/wire-ios-data-model/pull/43
Requires https://github.com/wireapp/wire-ios-sync-engine/pull/51

# Issue

On Travis there is no signing identity, and Xcode forbids using the shared container for our app during the tests, this makes tests crash. This is also the issue for external people who have no signing identity.

# Solution

Solution is to allow storing database in Application Data for simulator and builds that are non enterprise and non-appstore.


marcoconti83:
Conflict

mikeger:
@marcoconti83 does not matter, conflict in Cartfile, but it need to be updated before merging anyway

marcoconti83:
Data model released as 17.0.4

mikeger:
@marcoconti83 can be merged now

PR #202, opened 2016-09-13
Do not show alert together with camera keyboard

mikeger:
# Problem
The UI seems broken if user did not allowed the camera permissions in camera keyboard: the camera alert is overlapped by the keyboard.

# Solution
We do not show those dialogues at the same time. The alert is shown first, then the camera keyboard is coming into focus.

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
+++ b/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
@@ -20,6 +20,7 @@
 #import "UIApplication+Permissions.h"
 #import "AppDelegate.h"
 #import "UIAlertController+Wire.h"
+#import "UIResponder+FirstResponder.h"
 
 #import <AVFoundation/AVFoundation.h>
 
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
```diff
@@ -42,22 +43,43 @@ + (void)wr_requestOrWarnAboutVideoAccess:(void(^)(BOOL granted))grantedHandler
...
-    }]];
+        if (nil != completion) completion();
+    }];
+    
+    [noVideoAlert addAction:actionSettings];
+    
+    UIAlertAction *actionOK = [UIAlertAction actionWithTitle:NSLocalizedString(@"general.ok", nil)
+                                                       style:UIAlertActionStyleCancel
+                                                     handler:^(UIAlertAction * _Nonnull action) {
+         [[AppDelegate sharedAppDelegate].notificationsWindow.rootViewController dismissViewControllerAnimated:YES completion:nil];
```
typfel: Is this for dismissing the `UIAlertViewController`? Couldn't we just pass in the `noVideoAlert` then?

mikeger: It would be a reference circle, so you need to `weak` it. Otherwise yes.

PR #223, opened 2016-09-16
Remove self profile, move profile to settings

mikeger:
# Issue
The other mobile clients are moving to use the user profile integrated to the settings. We need to catch up in order to be able to start with other features that require richer settings. Also with the new settings the navigation to the devices list is 2 clicks closer to user.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -879,9 +888,34 @@ - (void)conversationListBottomBar:(ConversationListBottomBarController *)bar did
             break;
             
         case ConversationListButtonTypeSettings:
-            [[ZClientViewController sharedZClientViewController] openSelfProfileAnimated:YES];
-            break;
+        {
+            SettingsNavigationController *settingsViewController = [self createSettingsViewController];
             
+            if (IS_IPAD) {
```
typfel: I suspect this won't work well with split screen when in iphone size.

mikeger: Right! I updated to check with traits

Wire-iOS/Sources/UserInterface/Helpers/UIFont+MonoSpaced.swift
```diff
@@ -28,6 +28,20 @@ private let monospaceAttribute = [
     UIFontDescriptorFeatureSettingsAttribute: [monospacedFeatureSettingsAttribute]
 ]
 
+private let allCapsGeatureSettingsAttributeLowerCase = [
```
daehn: Typo `Geature` → `Feature`

Wire-iOS/Sources/UserInterface/Helpers/UIFont+MonoSpaced.swift
```diff
@@ -28,6 +28,20 @@ private let monospaceAttribute = [
...
+    UIFontFeatureTypeIdentifierKey: kLowerCaseType,
+    UIFontFeatureSelectorIdentifierKey: kLowerCaseSmallCapsSelector,
+]
+
+private let allCapsGeatureSettingsAttributeUpperCase = [
```
daehn: Typo `Geature` → `Feature`

mikeger: Updated! Thanks

PR #230, opened 2016-09-21
New domain for account pages

lipis:

PR #231, opened 2016-09-21
Add new password reset url

daehn:
# What's in this PR?

* https://github.com/wireapp/wire-ios/pull/230 against `develop`

### DIFF
--- a/Wire-iOS/Sources/Helpers/NSURL+WireURLs.m
+++ b/Wire-iOS/Sources/Helpers/NSURL+WireURLs.m
@@ -55,7 +55,7 @@
 
 + (instancetype)wr_passwordResetURL
 {
-    return [self URLWithString:@"https://wire.com/forgot/"];
+    return [self URLWithString:@"https://account.wire.com/forgot/"];
 }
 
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Helpers/NSURL+WireURLs.m
```diff
@@ -55,7 +55,7 @@ + (instancetype)wr_websiteURL
 
 + (instancetype)wr_passwordResetURL
 {
-    return [self URLWithString:@"https://wire.com/forgot/"];
+    return [self URLWithString:@"https://account.wire.com/forgot/"];
```
KVTaniguchi: have y'all considered adding a plist with different environments and stashing all the app urls in there?

KVTaniguchi:
👍

Wire-iOS/Sources/Helpers/NSURL+WireURLs.m
daehn: Good idea, however I personally would prefer an Swift `enum`, would make it easier to use and to collaborate than a plist.

PR #235, opened 2016-09-22
Feature/swift3

mikeger:


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/ArticleView.swift
```diff
@@ -27,25 +27,25 @@ import WireExtensionComponents
...
-    required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
-    }
-    
-    func configure(withTextMessageData textMessageData: ZMTextMessageData) {
-
-    }
-}
-/*
+//
```
daehn: This can be deleted I guess

PR #241, opened 2016-09-23
Fix all screenshot tests

mikeger:
- Missing fix for `ParticipantDeviceCellTests`


Wire-iOS Tests/VideoMessageCellTests.swift
```diff
@@ -65,15 +61,15 @@ class VideoMessageCellTests: ZMSnapshotTestCase {
...
         })
         verify(view: cell)
     }
     
     func testUploadedCell_fromOtherUser() {
-        let cell = self.cellWithConfig({
+        let cell = self.wrappedCellWithConfig({
             $0.fileMessageData?.transferState = .uploaded
             $0.backingFileMessageData.fileURL = .none
             $0.sender = (MockUser.mockUsers()[0] as! ZMUser)
```
KVTaniguchi: .first here?

Wire-iOS Tests/ZMSnapshotTestCase+Swift.swift
```diff
@@ -19,6 +19,48 @@
 
 import Foundation
 
+extension UITableViewCell: UITableViewDelegate, UITableViewDataSource {
+    func wrapInTableView() -> UITableView {
+        let tableView = UITableView(frame: self.bounds, style: .plain)
+        
+        tableView.delegate = self
+        tableView.dataSource = self
+        tableView.backgroundColor = .clear
```
KVTaniguchi: is this clear by default?

mikeger: Yep we use clear color per default so we can see both white and black text in the cells

Wire-iOS Tests/ParticipantDeviceCellTests.m
```diff
@@ -40,37 +40,37 @@ - (void)setUp
...
 }
 
-- (void)disabled_testThatItRendersTheCellUnverifiedFullWidthIdentifierLongerThan_16_Characters
+- (void)testThatItRendersTheCellUnverifiedFullWidthIdentifierLongerThan_16_Characters
 {
     UserClient *client = [UserClient fetchUserClientWithRemoteId:@"102030405060708090" forUser:self.user createIfNeeded:YES];
     client.deviceClass = @"tablet";
     [self.sut configureForClient:client];
-    ZMVerifyViewInAllIPhoneWidths([self.sut wrapInTableView]);
+    ZMVerifyView([self.sut wrapInTableView]);
```
typfel: We are not checking different sizes anymore?

mikeger: It does not work for cells; the table view cell's contentView on iOS 10 has an implicit constraint that sizes it to height 0. This constraint is created by table view, so I had to wrap all cells into table view for screenshot-ing. When the cell is wrapped in the table view verification in different sizes becomes complicated. On the other hand, for the cells given the verification in different sizes does not seem to be valuable.

PR #247, opened 2016-09-26
Updates for settings review

mikeger:
- Show version in last row
- Name validation
- Small copy changes

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
@@ -44,7 +44,7 @@ extension ZMUserSession: ZMUserSessionInterface {
 }
 
 protocol ValidatorType {
-    func validateValue(_ ioValue: AutoreleasingUnsafeMutablePointer<AnyObject?>, forKey inKey: String) throws
+    static func validateName(_ ioName: AutoreleasingUnsafeMutablePointer<NSString?>!) throws
 }
 
 extension ZMUser: ValidatorType {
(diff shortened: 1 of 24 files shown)


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -79,14 +92,17 @@ class SettingsPropertyFactory {
             let getAction : GetAction = { [unowned self] (property: SettingsBlockProperty) -> SettingsPropertyValue in
                 return SettingsPropertyValue.string(value: self.selfUser.name)
             }
-            let setAction : SetAction = { (property: SettingsBlockProperty, value: SettingsPropertyValue) -> () in
+            let setAction : SetAction = { (property: SettingsBlockProperty, value: SettingsPropertyValue) throws -> () in
                 switch(value) {
                 case .string(let stringValue):
+                    var stringValueAsAnyObject: AnyObject? = stringValue as AnyObject?
```
daehn: Maybe the `ValidatorType` could have an associated type so the inout value doesn't need to be `AnyObject`

mikeger: @daehn it's not that simple, I updated it the other way around. Please check

daehn: This seems to work, but it looks like this is only used with a `String` anyways and this might not work if called from Objc.

```swift
protocol Validator {
    associatedtype ValidatingType
    static func validate(value: inout ValidatingType) throws
}
```

PR #255, opened 2016-09-29
Align permission request copy

infotexture:
Use parallel constructions: “Allow Wire to access…”

@teller This fixes the issue you raised with the “You would be able to…” copy, which was introduced on September 23 in fe24cba with Feature/swift3 (#235).


daehn:
@infotexture please recreate this PR against `release/2.17` if these changes should be in the next release

infotexture:
@daehn If there's still time to port this for the next release, please do.

Not a showstopper, but certainly nice to have if possible.

PR #258, opened 2016-09-30
Disable Hockey when localytics is disabled

mikeger:
# Issue

We continue to use hockey when user selected not to track.

# Solution

Check the proper settings and sync the state between hockey and localytics.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
@@ -56,12 +56,20 @@ enum SettingsPropertyError: Error {
     case WrongValue(String)
 }
 
+protocol CrashlogManager {
+   var isCrashManagerDisabled: Bool { get set }
+}
+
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -175,12 +184,15 @@ class SettingsPropertyFactory {
                 }
             }
             let setAction : SetAction = { (property: SettingsBlockProperty, value: SettingsPropertyValue) throws -> () in
-                if var analytics = self.analytics {
+                if var analytics = self.analytics,
```
typfel: Since this are class instances I think they can be `let`

mikeger: I tried, I think there is a new rule in swift 3 that if object is let reference you also forbidden to modify fields (pretty cool)

typfel: Oh :)

PR #259, opened 2016-09-30
Crowdin update for 2.17

mikeger:


Wire-iOS/Resources/de.lproj/Localizable.strings
```diff
@@ -688,7 +703,8 @@
 "giphy.error.no_result" = "kein gif gefunden";
 
 "giphy.confirm" = "senden";
-"giphy.try_another" = "Nächstes";
+"giphy.cancel" = "abbrechen";
```
typfel: Why are these strings here? That's for next release

mikeger: Crowdin picks stuff from `develop`, we should have updated the localisation when we branched `release/2.17`.

PR #260, opened 2016-09-30
Add recently used emoji section

daehn:
# What's in this PR?

* Adds a recently used emoji section (limited to 15 emoji), the emoji are stored in the application support directory
* Fix section controller selection bugs
* Fixes a couple of emoji in the plists


Wire-iOS/Sources/UserInterface/Conversation/InputBar/Emoji/Emojis.swift
```diff
@@ -23,17 +23,25 @@ import Foundation
 typealias Emoji = String
 
 class EmojiDataSource: NSObject, UICollectionViewDataSource {
-    
+
+    enum Update {
+        case Insert(Int)
```
typfel: should not be capitalized

PR #261, opened 2016-09-30
Fix crash when searching in giphy

typfel:
Fix crash when searching in giphy since the request cancelation was not implemented. 

Depends on www.github.com/wireapp/wire-ios-sync-engine/pull/88 and www.github.com/wireapp/wire-ios-ziphy/pull/5/files


Wire-iOS/Sources/Helpers/syncengine/ZMUserSession+RequestProxy.m
```diff
@@ -24,19 +24,26 @@
 
 @implementation ZMUserSession (RequestProxy)
 
-- (void)doRequest:(NSURLRequest * __nonnull)request completionHandler:(void (^ __nonnull)(NSData * __nullable, NSURLResponse * __nullable, NSError * __nullable))completionHandler
+- (id)doRequest:(NSURLRequest *)request completionHandler:(void (^)(NSData * _Nullable, NSURLResponse * _Nullable, NSError * _Nullable))completionHandler
```
KVTaniguchi: ```instancetype``` here?

marcoconti83: I think this is supposed to be an opaque token, in fact it is a `ZMProxyRequest`, not a `ZMUserSession` (so can't be `instancetype`). 
It would be nice to have a stronger type than `id` though, maybe an empty protocol that just wraps `ZMProxyRequest`?

marcoconti83:
@typfel SE 55.0 has all the changes you need merged in.

Wire-iOS/Sources/Helpers/syncengine/ZMUserSession+RequestProxy.m
typfel: The issue is that the protocol I'm implementing here comes from `wire-ios-ziphy`which doesn't have any dependency on `wire-ios-sync-engine` which therefore  can't reference `ZMProxyRequest`.

mikeger:
@marcoconti83 please re-review

PR #269, opened 2016-10-05
Update emoji size on iPad and right margin of new send button

daehn:
# What's in this PR?

* Larger emoji in the keyboard on iPad.
* Increase right margin of the new send button.
* Update default keyboard sizes with values for different device types.
* Don't show send button when editing
* Do not allow switching from emoji to text input with tap on input field, only tapping the text input button should show the alphabetic keyboard again.
* Whitespace only text should not trigger the send button to be shown in the input bar.


daehn:
A couple of more adjustments will be done and added to this PR.

PR #276, opened 2016-10-06
Fix send button staying visible when starting edit with text input

daehn:
# What's in this PR?

* If an edit of a message was started while there was text input in the input bar / the send button was visible, then we did not update the state of the send button correctly.


mikeger:
@daehn can be covered with a screenshot test?

daehn:
@mikeger we currently only have snapshot tests for the `InputBar` and not for the `ConversationInputBarViewController` which handles this logic, we might need to add a couple of mock methods to our `MockConversation` to make it work.

PR #280, opened 2016-10-07
Show correctly name set error on iPad

mikeger:
# Issue

The alert was shown over controller that has modal presented. 

# Solution

Now alert is shown over current topmost controller.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
@@ -65,7 +65,7 @@ public extension UIApplication {
                 return .none
         }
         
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
```diff
@@ -65,8 +65,7 @@ public extension UIApplication {
                 return .none
         }
         
-        while let presentedController = topController.presentedViewController
-                , presentedController.modalPresentationStyle == .fullScreen {
+        while let presentedController = topController.presentedViewController, ((onlyFullScreen && presentedController.modalPresentationStyle == .fullScreen) || !onlyFullScreen) {
```
typfel: If you reverse the order of the `||` statements you could drop the `onlyFullScreen &&` part?

mikeger: `true`!

PR #284, opened 2016-10-10
Adjust send button color in highlighted state

daehn:
# What's in this PR?

* The send button icon color was dimmed in the highlighted state which it shouldn't be.


typfel:
Should go to release/2.18 I guess

daehn:
@typfel adjusted the base

PR #289, opened 2016-10-10
Fix/remove observer on deinit

marcoconti83:
Crash because the observer was not unregistered


marcoconti83:
Wrong branch, sorry

PR #291, opened 2016-10-10
Giphy design fixes

typfel:
- align navigation bar buttons
- add margins between buttons and giphy preview
- add no-results label
- use correct font for the title


marcoconti83:
@daehn since we have the CLA assistant now, the "Internal" label is becoming obsolete. Maybe we should disable the danger rule?

daehn:
@marcoconti83 definitely 👍 I will disable it

PR #298, opened 2016-10-12
Add first ephemeral messages UI

daehn:
# What's in this PR?

* Adds the input bar option to send _ephemeral messages_
* All cells that display messages that can be ephemeral now have a way to be obfuscated, e.g. show an orange view instead of an image, change the font to https://github.com/christiannaths/Redacted-Font in case of text messages...
* Receivers of ephemeral messages now call `startSelfDestructionIfNeeded` when the messages is visible on screen.
* Adds snapshot tests for the obfuscated cells.

### What is still missing?

* This PR depends on releases of `wire-ios-data-model`, `wire-iso-sync-engine`, `wire-ios-protos` and `wire-iso-message-strategy`.
* ~~The content of link previews does not yet correctly use the new font when obfuscated, there is an issue with `TTTAttributedLabel` that needs to worked around somehow.~~

### Coming in a separate PR:

* The cursor of the `InputBar` does not change it's color when changing the state of a conversation between sending ephemeral messages and sending non-ephemeral ones.
* The state handling  code in `InputBar` needs to be refactored.


mikeger:
@daehn conflict 👀⚠️

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.h
```diff
@@ -106,7 +108,7 @@ typedef void (^SelectedMenuBlock)(BOOL selected, BOOL animated);
 - (void)configureForMessage:(id<ZMConversationMessage>)message layoutProperties:(ConversationCellLayoutProperties *)layoutProperties;
 /// Update cell due since the message content has changed. Return True if the change requires the cell to be re-sized.
 - (BOOL)updateForMessage:(MessageChangeInfo *)changeInfo;
-- (void)willDisplayInTableView;
+- (void)willDisplayInTableView:(BOOL)onScreen;
```
mikeger: I think it needs `OnScreen` also in method name, like `willDisplayInTableViewOnScreen:(BOOL)onScreen` because when you call it the meaning of first parameter would be obscure otherwise:
`[... willDisplayInTableView:YES]`.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/AudioMessageCell.swift
```diff
@@ -67,7 +69,9 @@ open class AudioMessageCell: ConversationCell {
         self.loadingView.translatesAutoresizingMaskIntoConstraints = false
         self.loadingView.isHidden = true
 
-        self.allViews = [self.playButton, self.timeLabel, self.downloadProgressView, self.playerProgressView, self.waveformProgressView, self.loadingView]
+        obfuscationView.backgroundColor = UIColor(for: .brightOrange)
+
+        allViews = [playButton, timeLabel, downloadProgressView, playerProgressView, waveformProgressView, loadingView, obfuscationView]
```
mikeger: I think we need to discuss on the next architectural meeting about suppressing `self.` for instance variables. There are different opinions about it and in general I don't think we should change some code based on personal preferences.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -222,21 +234,31 @@ - (void)prepareForReuse
     self.topMarginConstraint.constant = 0;
     self.authorImageTopMarginConstraint.constant = 0;
     self.beingEdited = NO;
+    [self updateCountdownView];
 }
 
-- (void)willDisplayInTableView
+- (void)willDisplayInTableView:(BOOL)onScreen
```
mikeger: It looks like the parameter `onScreen` is not used.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -222,21 +234,31 @@ - (void)prepareForReuse
...
 {
     if (self.layoutProperties.showBurstTimestamp) {
         self.burstTimestampTimer = [NSTimer scheduledTimerWithTimeInterval:60 target:self selector:@selector(updateBurstTimestamp) userInfo:nil repeats:YES];
     }
     
     [self.contentView bringSubviewToFront:self.likeButton];
+
+    if ([self.delegate respondsToSelector:@selector(conversationCellShouldStartDestructionTimer:)] &&
+        [self.delegate conversationCellShouldStartDestructionTimer:self]) {
```
mikeger: I think the cell itself does not need to know if it's on or off screen, it is in my opinion the job of the cell's manager, in our case the `ConversationContentViewController`. I admit we have quite some logic in the cell right now, but I would vote against adding more here.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -571,6 +641,20 @@ - (void)setInputController:(UIViewController *)inputController
...
+    void (^cleanup)(UIViewController *) = ^(UIViewController *vc) {
+        if (! [vc isEqual:self.inputController]) {
+            vc = nil;
+        }
+    };
+
+    cleanup(self.cameraKeyboardViewController);
+    cleanup(self.audioRecordKeyboardViewController);
+    cleanup(self.emojiKeyboardViewController);
+    cleanup(self.ephemeralKeyboardViewController);
```
mikeger: I think you set to `nil` the local closure parameter to nil, not the ivar. You can achieve your goal with the macro (but don't tell anyone that I recommended that to you!)

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -0,0 +1,177 @@
...
+
+        guard let index = timeouts.index(of: conversation.destructionTimeout) else { return }
+        picker.selectRow(index, inComponent: 0, animated: false)
+
+    }
+
+    private func setupViews() {
+        CASStyler.default().styleItem(self)
+        picker.delegate = self
+        picker.delegate = self
```
mikeger: Delegate is assigned twice (copy/paste?)

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -0,0 +1,177 @@
...
+}
+
+
+class PickerView: UIPickerView {
+
+    var selectorColor: UIColor? = nil
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        for subview in subviews where subview.bounds.height <= 1.0 {
```
mikeger: I think it would be nice to have a comment on the code, right now it's a little bit obscure. I would also argue to insert this kind of "hacks" since apple can change the internals of `UIPickerView` and we would paint some views that we ain't expecting to paint in our fancy color. But is there a way to do it without hacks? May be we are fine with the default control UI?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -0,0 +1,177 @@
...
+            subview.backgroundColor = selectorColor
+        }
+    }
+
+}
+
+
+extension EphemeralKeyboardViewController: UIPickerViewDelegate, UIPickerViewDataSource {
```
mikeger: I wonder how stuff works if `dataSource` is not set?! ❓

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -337,7 +391,7 @@ private struct InputBarConstants {
     }
 
     open func undo() {
-        guard inputBarState != .writing else { return }
+        guard inputBarState.isWriting else { return }
```
mikeger: I am not sure about the change, since it looks like the logic is inverted now. Intended change?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
daehn: Yes, definitely will add a comment. Unfortunately I couldn't find a way without the hack, but I can check with design about the default UI.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
daehn: Yes, that makes sense

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
daehn: Yes, should have been the `dataSource`

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
daehn: Currently the cell is asking the `ConversationContentViewController` if it is on screen, so itself does not know it. Another option would be to start the destruction timer in the `ConversationContentViewController` directly, is that what you mean?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
daehn: Should also be possible in Swift (as we can put `nil` in collections there)

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
daehn: giphy

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
daehn: And will add a snapshot test for this ;)

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
mikeger: Yep, I was thinking to start timer in the controller.

daehn:
@mikeger please re-review

PR #300, opened 2016-10-13
Dismiss ephemeral keyboard on selected row tap

daehn:
# What's in this PR?

* Add logic to dismiss the ephemeral keyboard view controller on tap on the selected row, unfortunately the `UIPickerView` does not provide callbacks for this, which is why we now check if the tap occurred in the location of the center row.
* Add tests for the ephemeral keyboard view controller.


Wire-iOS Tests/Mocks/MockConversation.m
```diff
@@ -76,5 +76,10 @@ - (BOOL)isCallingSupported;
     return NO;
 }
 
+- (ZMConversationMessageDestructionTimeout)destructionTimeout
+{
+    return ZMConversationMessageDestructionTimeoutFifteenSeconds;
```
mikeger: `ZMConversationMessageDestructionTimeoutFifteenSeconds` is too damn long in my opinion! Not related to PR, sorry.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -140,9 +152,22 @@ extension ZMConversationMessageDestructionTimeout {
 /// This class is a workaround to make the selector color
 /// of a `UIPickerView` changeable. It relies on the height of the selector
 /// views, which means that the behaviour could break in future iOS updates.
-class PickerView: UIPickerView {
+class PickerView: UIPickerView, UIGestureRecognizerDelegate {
 
     var selectorColor: UIColor? = nil
+    var tapRecognizer: UIGestureRecognizer! = nil
+    var didTapViewClosure: ((UIGestureRecognizer) -> Void)? = nil
```
mikeger: I think passing out the gesture recogniser is making the API a little bit un-flexible; how about passing the tap coordinates in sender bounds?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -115,12 +115,24 @@ extension ZMConversationMessageDestructionTimeout {
...
         picker.selectorColor = separatorColor
+        picker.didTapViewClosure = dismissKeyboardIfNeeded
 
         titleLabel.textAlignment = .center
         titleLabel.text = "input.ephemeral.title".localized.uppercased()
         [titleLabel, picker].forEach(view.addSubview)
     }
 
+    func dismissKeyboardIfNeeded(recognizer: UIGestureRecognizer) {
+        let selectedRow = picker.selectedRow(inComponent: 0)
```
mikeger: let `selectedRow` is not really used, may be put `picker.selectedRow(inComponent: 0)` into `guard`?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -115,12 +115,24 @@ extension ZMConversationMessageDestructionTimeout {
...
+        guard selectedRow != -1 else { return }
+        let height = pickerView(picker, rowHeightForComponent: 0)
+        let rect = picker.frame.insetBy(dx: 0, dy: picker.bounds.midY - height / 2)
```
mikeger: I think it would be cool to move the logic into `PickerView`, also would be cool to have a comment on what's happening, i.e. tap on which element we are trying to detect

daehn:
@mikeger re-review please!

PR #304, opened 2016-10-14
Add new ephemeral destruction time indicator

daehn:
# What's in this PR?

* Adds the new indicator design showing the ephemeral destruction interval instead of the hourglass in case the conversation is ephemeral.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
@@ -46,6 +46,7 @@ fileprivate let shortStyleFormatter: DateComponentsFormatter = {
     return formatter
 }()
 
+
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/EphemeralKeyboardViewController.swift
```diff
@@ -63,13 +64,42 @@ extension ZMConversationMessageDestructionTimeout {
...
+        if isMinutes { return String(Int(rawValue / 60)) }
+        return nil
     }
 
 }
 
 
+extension ZMConversationMessageDestructionTimeout {
+
+    var isSeconds: Bool {
```
mikeger: You can change it now to check if the rawValue is more than 1 minute.

daehn: Good catch, updated!

PR #307, opened 2016-10-17
Fixes for ephemeral design feedback and bugs

daehn:
# What's in this PR?

* [#7458](https://wearezeta.atlassian.net/browse/ZIOS-7458): Remove old style link attachments when the underlying message gets obfuscated (in contrast to regular link previews we don't know if the message contained for example a Soundcloud link anymore after the obfuscation)
* Stop the audio player when an audio message gets deleted or obfuscated.
* Enable the snapshot test for the `EphemeralKeyboardViewController` again.
* [#7461](https://wearezeta.atlassian.net/browse/ZIOS-7461): Disable copying and saving ephemeral images in the `FullScreenImageViewController`.
* [#7457](https://wearezeta.atlassian.net/browse/ZIOS-7457): Fix wrong background color of the `UIPickerView` in dark mode.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/AudioMessageCell.swift
```diff
@@ -193,10 +197,15 @@ public final class AudioMessageCell: ConversationCell {
         }
     }
     
-    open override func prepareForReuse() {
+    public override func prepareForReuse() {
```
mikeger: Hm, why not `open`, I thought in parent class it must be open if the child would like to override the implementation.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ImageMessageCell.m
```diff
@@ -156,6 +156,8 @@ - (void)prepareForReuse
     
     self.originalImageSize = CGSizeZero;
     self.obfuscationView.hidden = YES;
+    self.fullScreenButton.hidden = NO;
+    self.sketchButton.hidden = NO;
```
mikeger: I thought that the buttons are hidden per default; design change?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/AudioMessageCell.swift
daehn: The parent class has to be `open` in order for subclasses to override the method, but we never subclass `AudioMessageCell` so it can be `public` here

mikeger: Yeah, but also `AudioMessageCell` is public, so no sense to have it `open` anyways.

PR #309, opened 2016-10-17
Add font size adjustments to MessageToolBoxView

daehn:
# What's in this PR?

* Font size adjustments to increase the delivery state status as requested by design.


daehn:
Need to re-record the snapshots

PR #310, opened 2016-10-17
Fix recently used Emoji persistence

daehn:
# What's in this PR?

* The recently emoji were not loaded from disk correctly and have been out of order
* Fix vertical alignment of the `UIPickerView` inside the `EphemeralKeyboardViewController`.


daehn:
Need to re-record the snapshot tests

PR #314, opened 2016-10-18
Fix transparent device information screen

Einh06:
**What's in this PR**
This PR fixes the personal fingerprint screen shown from the user device screen.
The screen also respect light / dark  mode.


daehn:
@Einh06 did you check if we have any snapshot tests for this that need to be updated?

Einh06:
@daehn I checked and nothing need to be updated.

PR #320, opened 2016-10-19
Add options to open links in default applications

daehn:
# What's in this PR?

This PR adds the ability to select default applications to open locations, tweets and generally links in. The currently supported applications are Google Maps for locations, Tweetbot and Twitterrific for tweets and Chrome and Firefox for other kinds of links ([#7159](https://wearezeta.atlassian.net/browse/ZIOS-7159)). 

The user can select these options under `Settings/Advanced`, but they will only be visible if the user has the option installed, there will be no option to select a default application for opening locations if the user only has Apple Maps for example. If the user has non of the mentioned apps the whole section will not be displayed.

If the opening fails (for example when the application that has been selected has ben removed from the device again) we fallback to open the link using Safari (or Apple Maps in case of locations).


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -485,6 +485,24 @@
...
 
+// Open in
+"self.settings.external_apps.header" = "External Apps";
+
+"self.settings.link_options.twitter.title" = "Tweets";
+"self.settings.link_options.maps.title" = "Maps";
+"self.settings.link_options.browser.title" = "Browser";
+
+"open_link.twitter.option.default" = "Safari / Twitter";
+"open_link.twitter.option.tweetbot" = "Tweetbot";
```
mikeger: @raphaelrobert should this copy have (c) trademark attributes?

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -220,6 +220,39 @@ class SettingsPropertyFactory {
             
             return SettingsBlockProperty(propertyName: propertyName, getAction: getAction, setAction: setAction)
 
+        case .tweetOpeningOption:
```
mikeger: I think you can simplify your code by putting the settings keys in `userDefaultsPropertiesToKeys`, then the thing maps it automagically. Less boilerplate!

daehn: Good point, will check it

Wire-iOS/Resources/Base.lproj/Localizable.strings
mikeger: @raphaelrobert

daehn:
@mikeger @typfel updated.

PR #327, opened 2016-10-21
Add tracking attributes for ephemeral messages

daehn:
# What's in this PR?

* Add `is_ephemeral` tracking attribute and `ephemeral_time` attribute in case the message is ephemeral.


daehn:
@marcoconti83 We currently don't really unit test the tracking code on the UI side, which we should change at some point. I tested this manually for now, in order to make it unit testable more easily we might want to use protocols in the UI code to simplify the setup in tests.

daehn:
More tracking attributes for the receiving side have been added to the spec, I will update this PR

daehn:
@marcoconti83 updated

PR #330, opened 2016-10-24
Persist log level switches between runs

marcoconti83:
Need to debug some decryption errors, this will allow testers to set the log on once instead of having to set it every time they restart the app. This will also allow to get logs when we are decrypting the notification stream, which happens immediately and there is no time to switch on the log manually.

# Changes
Save list of log tags that are enabled to user defaults

### DIFF
--- a/Wire-iOS.xcodeproj/project.pbxproj
+++ b/Wire-iOS.xcodeproj/project.pbxproj
@@ -4762,12 +4762,10 @@
 				TargetAttributes = {
 					1EE85C4F1B5F96FB008E084B = {
 						CreatedOnToolsVersion = 6.4;
-						DevelopmentTeam = W5KEQBF9B5;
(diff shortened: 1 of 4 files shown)


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -4759,10 +4762,12 @@
 				TargetAttributes = {
 					1EE85C4F1B5F96FB008E084B = {
 						CreatedOnToolsVersion = 6.4;
+						DevelopmentTeam = W5KEQBF9B5;
```
mikeger: Please reset so it would not break compilation for third-parties. It's the Xcode that sets it automatically.

marcoconti83: Is there any way to stop Xcode from doing that?

PR #353, opened 2016-10-31
Sketch with Emojis

typfel:
## What's in this PR
- The sketch feature has been extended to support inserting emojis.
- When sketching on top of an existing image the resulting image will keep the size of the original      image, this allows multiple users to to edit an image without degrading the image.
- We now always try to render images at retina scale in the app.
- The sketch shortcuts in the image sending preview and in the image cell have been re-designed.


Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,377 @@
...
+    func selectDrawTool() {
+        select(editMode: .draw, animated: true)
+//        canvas.mode = .draw
+//        updateButtonSelection()
+    }
+    
+    func openEmojiKeyboard() {
+        select(editMode: .emoji, animated: true)
+//        canvas.mode = .edit
+//        updateButtonSelection()
```
daehn: Can be removed?

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,377 @@
...
+                            self.isEmojiKeyboardInTransition = false
+                            removeEmojiKeyboardViewController()
+            })
+        } else {
+            removeEmojiKeyboardViewController()
+        }
+    }
+    
+    func emojiKeyboardViewControllerDeleteTapped(_ viewController: EmojiKeyboardViewController) {
+
```
daehn: I think this is not called from anywhere and can be removed

Wire-iOS/Sources/UserInterface/Components/Controllers/DefaultNavigationBar.swift
```diff
@@ -0,0 +1,50 @@
+//
+//  DefaultNavigationBar.swift
+//  Wire-iOS
+//
```
daehn: Missing copyright headers

Wire-iOS/Sources/UserInterface/Components/Views/ImageToolbarView.swift
```diff
@@ -0,0 +1,144 @@
+//
+//  ImageToolbarView.swift
+//  Wire-iOS
+//
+//  Created by Jacob on 28/10/16.
```
daehn: Missing copyright headers

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Camera.swift
```diff
@@ -76,8 +76,8 @@ extension ConversationInputBarViewController: CameraKeyboardViewControllerDelega
             confirmVideoViewController.transitioningDelegate = FastTransitioningDelegate.sharedDelegate
             confirmVideoViewController.videoURL = videoURL as URL!
             confirmVideoViewController.previewTitle = self.conversation.displayName.uppercased()
-            confirmVideoViewController.isEditButtonVisible = false
-            confirmVideoViewController.onConfirm = { [unowned self] in
+//            confirmVideoViewController.isEditButtonVisible = false
```
daehn: Can be removed?

Wire-iOS/Sources/UserInterface/SketchToolbar.swift
```diff
@@ -0,0 +1,107 @@
+//
+//  SketchToolbar.swift
+//  Wire-iOS
+//
+//  Created by Jacob on 21/10/16.
```
daehn: Missing headers, Travis is too slow

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
typfel: I have to implement it to fulfil the the protocol.

daehn: Ah I see, then we might want to add an option to hide the delete button in the `EmojiKeyBoardViewController` or an option to add a different button to dismiss it

Wire-iOS Tests/ConfirmAssetViewControllerTests.swift
```diff
@@ -32,25 +32,23 @@ class ConfirmAssetViewControllerTests: ZMSnapshotTestCase {
...
         sut.image = image(inTestBundleNamed: "unsplash_burger.jpg")
-        sut.isEditButtonVisible = true
         sut.previewTitle = "Burger & Beer"
         verifyInAllIPhoneSizes(view: sut.prepareForSnapshot())
     }
+    
+    func testThatItRendersTheAssetViewControllerWithSmallImage() {
+        accentColor = .vividRed
+        sut.image = image(inTestBundleNamed: "unsplash_small.jpg").imageScaled(withFactor: 0.5);
+        sut.previewTitle = "See Food"
```
mikeger: Typo, may be "Sea Food"?

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,383 @@
...
+    }
+    
+    override var shouldAutorotate: Bool {
+        return false
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        view.backgroundColor = .white
```
mikeger: May be set via Classy, since there is already a selector for `CanvasViewController`

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,383 @@
...
+        if (animated) {
+            isEmojiKeyboardInTransition = true
+            
+            let offscreen = CGAffineTransform(translationX: 0, y: KeyboardHeight.current)
+            emojiKeyboardViewController.view.transform = offscreen
+            view.layoutIfNeeded()
+            
+            UIView.animate(withDuration: 0.25,
+                           delay: 0,
+                           options: UIViewAnimationOptions(rawValue: UInt(7)),
```
mikeger: I think that there should be a way to pass it as a bit flag, something like [.option, .otherOption].

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+    
+    func emojiKeyboardViewControllerDeleteTapped(_ viewController: EmojiKeyboardViewController) {
+        
+    }
+    
+    func emojiKeyboardViewController(_ viewController: EmojiKeyboardViewController, didSelectEmoji emoji: String) {
+        
+        let attributes = [NSFontAttributeName: UIFont.systemFont(ofSize: 72)]
```
mikeger: Should there be a separated rendering method on String, so it can be restructured / reused?

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
```diff
@@ -0,0 +1,383 @@
...
+    func imagePickerController(_ picker: UIImagePickerController, didFinishPickingMediaWithInfo info: [String : Any]) {
+        UIImagePickerController.loadImage(fromMediaInfo: info, result: { (image, _, _) in
+            if let image = image, let cgImage = image.cgImage {
+                self.canvas.referenceImage = UIImage(cgImage: cgImage, scale: 2, orientation: image.imageOrientation)
+                self.canvas.mode = .draw
+                self.updateButtonSelection()
+            }
+            picker.dismiss(animated: true, completion: nil)
+        }) { (error) in
+            print("error: ", error)
```
mikeger: `DDLogError` should do the job.

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController.m
```diff
@@ -235,14 +258,13 @@ - (void)createConstraints
...
     // Preview image
-    [self.imagePreviewView autoPinEdge:ALEdgeTop toEdge:ALEdgeBottom ofView:self.topPanel];
-    [self.imagePreviewView autoPinEdge:ALEdgeBottom toEdge:ALEdgeTop ofView:self.bottomPanel];
-    [self.imagePreviewView autoPinEdgeToSuperviewEdge:ALEdgeLeft];
-    [self.imagePreviewView autoPinEdgeToSuperviewEdge:ALEdgeRight];
-
-    [self.editButton autoAlignAxisToSuperviewAxis:ALAxisVertical];
-    [self.editButton autoSetDimension:ALDimensionHeight toSize:32];
-    [self.editButton autoPinEdge:ALEdgeBottom toEdge:ALEdgeBottom ofView:self.view withOffset:-112];
+    CGSize imageSize = [(UIImage *)self.image size];
```
mikeger: `self.image` can be a GIF (FLAnimatedImage). May be it also has size, may be not. To check on compile time I would add the `size` to `MediaAsset` instead of casting.

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController.m
```diff
@@ -251,14 +273,18 @@ - (void)createConstraints
...
+    [self.imageToolbarViewInsideImage autoSetDimension:ALDimensionHeight toSize:48];
 }
 
 #pragma mark - Actions
 
 - (IBAction)acceptImage:(id)sender
 {
     if (self.onConfirm) {
-        self.onConfirm();
+        self.onConfirm(nil);
```
mikeger: I would always return an image, if editing happened it should be the edited one, otherwise the original one. This way it's easier for class user in my opinion.

marcoconti83:
Looking forward to see this in action! 👍

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
mikeger: There is a method now, `backspaceHidden`. Then we can mark `emojiKeyboardViewControllerDeleteTapped` as optional method in the protocol and kill the implementation.

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
typfel: Swift only allows optional methods on objc protocols

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController.m
typfel: Hmm yes, was looking at but it required quite some re-factoring because how we load images from the media library. Will take another look.

typfel:
@mikeger I will add a `size`property to the `MediaAsset` protocol.

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController.m
typfel: Yes thanks for catching for this. Was planning on doing it but forgot  :)

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
typfel: The keyboard animation curve is not exposed in the animation options.

mikeger: Puh right, goddamn UIKit.

Wire-iOS/Sources/UserInterface/CanvasViewController.swift
mikeger: Puh right, goddamn Swift.

marcoconti83: @mikeger if you do that, and then try to use that protocol from Objective-c, it might not see the default implementation. We had a case like this before. Maybe it's fixed in 3.0 though...

mikeger: I guess it's still not visible. If this protocol is used from objc then we can mark the protocol @objc and mark method as optional -- should work as well.

PR #355, opened 2016-11-01
Change how mode selection works in sketch

typfel:
- You can now toggle between draw/edit mode by tapping the draw button
- When in edit mode no button is seleced


mikeger:
@typfel no need to update snapshots?

typfel:
No because there are no snapshot for sketch, yet ;)

PR #356, opened 2016-11-01
enable sending ephemeral in groups

basine:
- [x] Needs zmessaging release with https://github.com/wireapp/wire-ios-data-model/pull/71 and https://github.com/wireapp/wire-ios-message-strategy/pull/11


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarButtonState.swift
```diff
@@ -29,11 +29,11 @@ public final class ConversationInputBarButtonState: NSObject {
     }
 
     public var hourglassButtonHidden: Bool {
-        return hasText || conversationType != .oneOnOne || editing || ephemeral || disableEphemeralSending
+        return hasText || editing || ephemeral || disableEphemeralSending
     }
 
     public var ephemeralIndicatorButtonHidden: Bool {
-        return hasText || conversationType != .oneOnOne || editing || !ephemeral || disableEphemeralSending
```
daehn: We might not need to pass in the `conversationType` at all anymore

basine: yep, you're right. I removed it.

PR #357, opened 2016-11-01
Sketch design fixes

typfel:
- flat background on image toolbar
- sketch toolbar size/margin adjustments
- Hide colorpicker when adding emojis


Wire-iOS/Sources/UserInterface/SketchToolbar.swift
```diff
@@ -92,6 +92,7 @@ class SketchToolbar : UIView {
     func createCenterButtonConstraints() {
         guard !centerButtons.isEmpty else { return }
         
+        let buttonSpacing : CGFloat = 32
```
mikeger: Does not seem to do anything. Can be removed?

mikeger: Ah, it used to be instance var.

PR #363, opened 2016-11-03
Initial ephemeral redesign

mikeger:
# New style on ephemeral messages

- [x] Ephemeral color gone
- [x] Special icons for ephemeral state
- [x] Icons in placeholders


daehn:
The destruction view dots look like they are filled in the opposite order now:
bildschirmfoto 2016-11-07 um 11 39 07

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -370,21 +370,39 @@ private struct InputBarConstants {
         return state.isWriting ? writingColor : writingColor.mix(editingColor, amount: 0.16)
     }
 
-    fileprivate func updateColors() {
-        backgroundColor = backgroundColor(forInputBarState: inputBarState)
-
+    fileprivate var isEphemeral: Bool {
```
daehn: It might make sense to pull this inside the InputBarState itself as instance var

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ImageMessageCell.m
```diff
@@ -193,9 +194,13 @@ - (void)createImageMessageViews
     self.loadingView = [[ThreeDotsLoadingView alloc] initForAutoLayout];
     [self.imageViewContainer addSubview:self.loadingView];
 
-    self.obfuscationView = [[UIView alloc] initForAutoLayout];
+    self.obfuscationView = [[UIImageView alloc] initForAutoLayout];
```
daehn: It might make sense to extract this setup logic for the obfuscation view in a separate `UIView`, eg. `ObfuscationView` that can be initialized with a color and icon type, as we repeat this setup multiple times.

mikeger:
@daehn please re-review

PR #365, opened 2016-11-04
Clarify device removal details

infotexture:


daehn:
Why are we removing the info about the message removal?

PR #378, opened 2016-11-08
Fixed accent color selection when no yellow is shown

mikeger:
# Issue

The indexes where not aligned between `ZMAccentColor.all()` array and `self.colors` array.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
@@ -212,13 +212,17 @@ extension ColorPickerController: UITableViewDelegate, UITableViewDataSource {
 
 
 
-open class AccentColorPickerController: ColorPickerController {
+final class AccentColorPickerController: ColorPickerController {
+    fileprivate let allAccentColors: [ZMAccentColor]
+    
+    
     public init() {
-        let allColors = ZMAccentColor.all().filter { $0 != .brightYellow }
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Settings/AccentColorPickerController.swift
```diff
@@ -212,13 +212,17 @@ extension ColorPickerController: UITableViewDelegate, UITableViewDataSource {
...
 
 
-open class AccentColorPickerController: ColorPickerController {
+final class AccentColorPickerController: ColorPickerController {
+    fileprivate let allAccentColors: [ZMAccentColor]
+    
+    
     public init() {
-        let allColors = ZMAccentColor.all().filter { $0 != .brightYellow }
+        self.allAccentColors = ZMAccentColor.all().filter { $0 != .brightYellow }
```
marcoconti83: Feels a bit hacky to hardcode one out. But from what I understood we need to keep this yellow around for older clients, right?

marcoconti83: Maybe store it in a static `discontinuedColors` array makes it clearer why we are leaving out yellow

mikeger: Yep, if user used to have yellow color we are not forcing it to be changed (yet)

mikeger:
@marcoconti83 please re-review

marcoconti83:
👍

PR #383, opened 2016-11-10
Update OPEN LINKS IN copy per ZIOS-7596

infotexture:


mikeger:
@infotexture this pull request is going into `fix/move-open-in-ZIOS-7596` branch. Is that correct?

PR #395, opened 2016-11-15
Fix missing rounded corners when cells are presented initially

daehn:
# What's in this PR?

* There was an issue with the rounded corners of several cells not being present when the cells are shown for the first time.
* `layoutSubViews` is not called the first time a cell is created, which is why we need to call `setNeedsLayout()` and `layoutIfNeeded()` manually in the initializers of `AudioMessageCell`, `FileTransferCell` and `VideoMessageCell`.


mikeger:
I guess it's an iOS 10 `feature` consequence

PR #397, opened 2016-11-16
Enable deletion for the selfUser for failed to send messages and in left conversations

daehn:
# What's in this PR?
* This fixes [#7281](https://wearezeta.atlassian.net/browse/ZIOS-7448) and [#7448](https://wearezeta.atlassian.net/browse/ZIOS-7281).
* There was a bug with the delete for me option not being present if the message failed to send as well as in left conversations. The issue was that we were adding the delete `UIMenuItem` to the menu if a message can be liked. To ensure we do not show a delete for everyone if the message failed to sent an additional check has been added.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
@@ -18,6 +18,20 @@
 
 import ZMCDataModel
 
+
+private extension ZMConversationMessage {
+
+    /// Whether the `Delete for everyone` option should be allowed and shown for this message.
+    var canBeDeletedForEveryone: Bool {
+        guard let sender = sender, let conversation = conversation else { return false }
+        return sender.isSelfUser
+            && conversation.isSelfAnActiveMember
+            && (deliveryState == .sent || deliveryState == .delivered)
+    }
+
+}
+
+
 extension ConversationContentViewController {
 
     /**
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageDeletion.swift
```diff
@@ -18,6 +18,20 @@
...
 
+
+private extension ZMConversationMessage {
+
+    /// Whether the `Delete for everyone` option should be allowed and shown for this message.
+    var canBeDeletedForEveryone: Bool {
+        guard let sender = sender, let conversation = conversation else { return false }
+        return sender.isSelfUser
+            && conversation.isSelfAnActiveMember
+            && (deliveryState == .sent || deliveryState == .delivered)
```
marcoconti83: We had a discussion this morning, we should allow "delete for everyone" even for messages that failed to send (timeout). This is because currently we can't delete expired messages. We don't want to show the "delete just for me" because we have no guarantee that the message was not received (it might have been received by the BE even if the network request or the timer failed), so we should always treat a deletion on an expired message as "delete for everyone".

daehn: Yes that makes sense, so the decision is to not show "Delete for me" and only show "Delete for everyone" in that case?

daehn:
@marcoconti83 I updated the behavior to be like this now:

* If a message is failed to send only "delete for everyone" is possible as it might have it the backend before we expired it.
* If a message is sent by self and not expired it can be deleted for send or everyone.
* Ephemeral messages can not be deleted (no menu at all).
* If the user leaves a conversation he can delete previously sent messages only using "delete for me".

daehn:
@marcoconti83 please confirm the updated behaviour

PR #406, opened 2016-11-18
Fix for clipping message with multiline link preview title

mikeger:
# Issue

Link preview with multiple lines is cut because of link preview view height collapse. Fix is to force labels to be one-line in this case.


daehn:
Would be nice to have a snapshot test for this

mikeger:
@daehn right, I have it on my list.

PR #409, opened 2016-11-18
Connection status header in conversation

mikeger:
# Feature

Showing nice connection status header in conversation beginning, same as in Webapp.

# Testing approach

The idea of `CombinationTest` is to test view in possible combinations of setup values. `Mutator` represents one changing property of a SUT. `CombinationTest` creates all possible versions of SUT object according to given `Mutator`s.

To test is enough to call `combinationTest.testAll` and it would call the closure with every variant of SUT. This could be useful to make a screenshot of the SUT if it is a view and store / verify it.


Wire-iOS/Sources/Helpers/syncengine/ZMUser+Additions.h
```diff
@@ -61,6 +61,9 @@ FOUNDATION_EXPORT ZMUser *BareUserToUser(id bareUser);
 /// Returns corresponding address book contact in current session
 - (ZMAddressBookContact *)contact;
 
+/// TODO: temporary debugging username generator
+- (NSString *)autoUsername;
```
typfel: Can be deleted?

daehn:
Would be nice to have some snapshot tests for this!

Wire-iOS/Sources/UserInterface/ConnectRequests/IncomingConnectRequestView.m
```diff
@@ -98,8 +97,6 @@ - (void)setupViews
     self.userImageView.suggestedImageSize = UserImageViewSizeBig;
     [self addSubview:self.userImageView];
     
-    self.commonConnectionsController = [[CommonConnectionsViewController alloc] init];
```
typfel: Do we still use the CommonConnectionsViewController somewhere else?

mikeger: Was my thinking, but it is used in connection request view and in not connected profile view.

mikeger:
@daehn I will add tests here, good point.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -249,8 +249,23 @@ - (void)updateTableViewHeaderView
     }
     
     UIView *headerView = nil;
-    if ((self.conversation.conversationType == ZMConversationTypeConnection || self.conversation.conversationType == ZMConversationTypeOneOnOne) && self.conversation.firstActiveParticipantOtherThanSelf) {
-        headerView = [[ConnectionStatusHeader alloc] initWithUser:self.conversation.firstActiveParticipantOtherThanSelf];
+    ZMUser *otherParticipant = self.conversation.firstActiveParticipantOtherThanSelf;
+    if ((self.conversation.conversationType == ZMConversationTypeConnection || self.conversation.conversationType == ZMConversationTypeOneOnOne) && nil != otherParticipant) {
+        self.connectionViewController = [[UserConnectionViewController alloc] initWithUserSession:[ZMUserSession sharedSession] user:otherParticipant];
+        @weakify(self);
+        self.connectionViewController.onIgnore = ^() {
```
daehn: Will the `ConversationContentViewController` need to know if the action was ignore, block or connecting at some point? If not and it is only notified when any action was performed we could just have one closure `onAction` or something like that as there is a lot of repetition currently, could also pass an action type enum in the closure if we will need it at some point (maybe for tracking?).

Wire-iOS/Sources/UserInterface/ConnectRequests/IncomingConnectRequestView.m
```diff
@@ -187,6 +179,11 @@ - (void)setSenderEmail:(NSString *)senderEmail
     }
 }
 
+- (void)setShowConnectionButtons:(BOOL)showConnectionButtons
```
daehn: Looks like this overridden setter does not perform any additional work

mikeger:
@daehn @typfel please review tests

typfel:
Is he button spacing as designed?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -4821,10 +4826,12 @@
 				TargetAttributes = {
 					1EE85C4F1B5F96FB008E084B = {
 						CreatedOnToolsVersion = 6.4;
+						DevelopmentTeam = W5KEQBF9B5;
```
daehn: Can be removed

Wire-iOS Tests/CombinationTest.swift
```diff
@@ -0,0 +1,119 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+@testable import Wire
+
+struct Mutator<T: Copyable, I: Hashable> {
```
marcoconti83: I would name `T` and `I` more explicitly, I'm a bit lost on what they are.
What is `T`? The object that is generated as result of the mutation?
And are `I` the values that are used to generate a specific mutation?

Wire-iOS Tests/CombinationTest.swift
```diff
@@ -0,0 +1,119 @@
...
+    typealias Mutation = (T, I)->(T)
+    typealias CombinationPair = (combination: I, result: T)
+    let mutation: Mutation
+    let combinations: Set<I>
+    init(mutation: @escaping Mutation, combinations: Set<I>) {
```
marcoconti83: If I understand correctly, the `mutation` argument in this call is a callbacks that generate a "mutation" of the SUT starting from a list of different values, the `combinations`. I would suggest calling the parameters `generator` and `mutatingValues` or something similar, it would make it easier to understand just by looking at the function call.
`Mutator<X>(generator: viewGenerator, mutatingValues: possibleValues)`

Wire-iOS Tests/CombinationTest.swift
```diff
@@ -0,0 +1,119 @@
...
+    typealias M = Mutator<T, I>
+    typealias CombinationChainPair = (combinationChain: [I], result: T)
+    let mutators: [M]
+    let mutable: T
+    init(mutable: T, mutators: [M]) {
+        self.mutable = mutable
+        self.mutators = mutators
+    }
+    
+    func allCombinations() -> [CombinationChainPair] {
```
marcoconti83: So if I understand correctly this will yield all possible combinations given by applying all mutators, so that if we have `n` mutators with `m` possible values each, this will yield `n^m` combinations?

mikeger: Correct.

marcoconti83: image

mikeger:
@marcoconti83 @daehn @typfel updated for review 🎉

PR #410, opened 2016-11-18
Open contact support in Safari

mikeger:
# Issue

Attaching files does not work in the integrated browser that we show.


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -305,7 +305,8 @@ import Foundation
         }, previewGenerator: .none)
         
         let contactButton = SettingsExternalScreenCellDescriptor(title: "self.help_center.contact_support".localized, isDestructive: false, presentationStyle: .modal, presentationAction: { _ in
-            return BrowserViewController(url: NSURL.wr_askSupport().wr_URLByAppendingLocaleParameter() as URL!)
+            UIApplication.shared.openURL(NSURL.wr_askSupport().wr_URLByAppendingLocaleParameter() as URL!)
```
daehn: If we want to open the page in the browser selected by the user (if he selected a different default in the "Open In" settings) we could call `open()` on the `URL` instance.

mikeger: Good point, updated the PR.

PR #411, opened 2016-11-18
Fix CallKit switch copy (#408)

infotexture:
* Change group title to “CALLS”
* Change setting toggle to "Share with iOS"
* Change explanation copy to:

> Show Wire calls on the lock screen and in iOS call history.
> If iCloud is enabled, call history is shared with Apple.
> Restart Wire to apply this setting.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -482,9 +482,9 @@
 "self.settings.sound_menu.ringtones.title" = "Ringtones";
 "self.settings.sound_menu.sounds.wire_sound" = "Wire";
 
-"self.settings.callkit.title" = "CallKit";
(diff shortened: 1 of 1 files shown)


mikeger:
@infotexture this switch not only changes if the calls are shared with iOS, it also disables the whole CallKit integration, including the overtake screen.

infotexture:
@mikeger Yes, we're aware of that, but since most users have no idea what CallKit is, we've chosen to avoid that term in UI copy and instead refer to iOS features with which users are more familiar (such as the lock screen and call history). 

This version of the copy was reviewed and approved with @priiduzilmer last night.

mikeger:
@infotexture may be we can use term that Skype is using, e.g. "Native Calling" ?

infotexture:
@mikeger No — we considered that and decided against it. If you look closely at the diff, you will see that we have explicitly chosen *not* to refer to native calls, as many users won't know the difference.

PR #416, opened 2016-11-21
Add first fixes to support RTL languages

daehn:
# What's in this PR?

**This PR contains multiple fixes for layout issues present when using the application on a device with a right-to-left layout direction ([#4822](https://wearezeta.atlassian.net/browse/ZIOS-4822), [#7665](https://wearezeta.atlassian.net/browse/ZIOS-7665))**

* Align text in text messages to the right.
* Fix input bar misalignment and emoji button location.
* Reverse system message cell alignment for RTL languages.
* Update alignment of the cells in settings for RTL languages.

There are still a lot of issues when using a RTL language which are tracked in [#7665](https://wearezeta.atlassian.net/browse/ZIOS-7665).


daehn:
Unfortunately I was unable to find a way to inject or mock the layout direction in tests yet, would be nice to able to have some snapshot tests for this. But they at least assert that this does not break anything in LTR languages.

PR #419, opened 2016-11-24
Add @name take over view controller

daehn:
# What's in this PR?

* This PR adds the `UserNameTakeOverViewController` which will be used to make the user select a username.
* This PR is not used yet and this PR is not dependent on other frameworks.


daehn:
@typfel updated

PR #438, opened 2016-11-29
Usernames initial implementation

mikeger:
# Contains

- https://github.com/wireapp/wire-ios/pull/409 (already reviewed)
- Update for the new SE API


daehn:
Actually I have a question, does this handle the case that a user might not have a handle?

mikeger:
@daehn when it was implemented the working assumption was that everyone is going to have a username

daehn:
@mikeger ok, I will have a look as we can't guarantee this anymore.

daehn:
@mikeger please have a look at my last commit 161059b

PR #439, opened 2016-11-29
Show username takeover screen in conversation list

daehn:
# What's in this PR?

* This adds the first logic to present the @name takeover screen where users need to decide whether to keep the suggested username or choose their own.
* We call `UserProfile.suggestHandles` in order to receive callbacks once an available handle has been found and presents the takeover if the conversation list is shown.
* We were adding the `ArchivedListViewController`, `StartUIViewController` and the different takeovers as child view controllers to the `ConversationListViewController`, which means we would have to check if we are showing the list to avoid replacing the archive etc. with a takeover, as well as recheck if we have a suggested handle when returning to the list (as we might have wanted to show the takeover while in the archive). Due to this limitation (and the quite big custom logic to present the child view controllers), the presentation logic has been refactored and now presents the archive and start UI properly. Thus the `UserNameTakeOverViewController` can be added as child view controller to the conversation list without affecting the archive or start UI and we will not need to check for a previously suggested handle again.
* Fix the valid length of an @name, which now is `[2, 21]`.


daehn:
I need to update the copy as there have been some smaller changes.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationList+Handles.swift
```diff
@@ -0,0 +1,107 @@
...
+}
+
+
+extension ConversationListViewController: UserNameTakeOverViewControllerDelegate {
+
+    func takeOverViewController(_ viewController: UserNameTakeOverViewController, didPerformAction action: UserNameTakeOverViewControllerAction) {
+        switch action {
+        case .chooseOwn(let suggested): openChangeHandleViewController(with: suggested)
+        case .keepSuggestion(let suggested): setSuggested(handle: suggested)
+        case .learnMore: URL(string: "https://www.wire.com")?.open() // TODO: Insert correct URL
```
daehn: post-64231-this-is-fine-dog-fire-comic-im-n7mp

marcoconti83: @jenskasemets do you have the right URL?

daehn: @marcoconti83 @jenskasemets I spoke to @Yorgos-V, the final URL will follow tomorrow, but I would suggest to update it on `develop` once it is there.

mikeger:
We have some other takeover screens, for example 

- Add email
- Add phone
- Share address book
- Allow push notifications

I see you checked that this takeover should not appear over start UI or other active screen; can it interfere with other takeovers?

daehn:
@mikeger I need to double check about the add email/phone ones, but the push notifications takeover will now only be shown if the user has a username to ensure that the username takeover is always shown first.

daehn:
As discussed in person the phone and email takeovers are part of the registration flow and do not affect the conversation list takeovers.

daehn:
I am fixing the conflicts 🔥

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.h
```diff
@@ -71,7 +71,7 @@
 - (void)openDetailScreenForConversation:(ZMConversation *)conversation;
 
 /**
- * Open the user clietn list screen
+ * Open the user client list screen
```
typfel: 🙌

Wire-iOS/Sources/UserInterface/ConversationList/ConversationList+Handles.swift
jenskasemets: @daehn @marcoconti83 Late to the party, but now have final URL: https://wire.com/support/username/

PR #443, opened 2016-11-30
Make correlation formatter more generic and use it in search result cells

daehn:
# What's in this PR?

* This updates the introduces `AddressBookCorrelationFormatter` to be more generic so that it can be reused in different places.
* Update `SearchResultCell` to use a `AddressBookCorrelationFormatter` to generate the subtitle as well as the users handle to the cell if it is present.


daehn:
Fixing the conflicts 🔥

PR #447, opened 2016-12-01
Limit max width of user image in connection view, fix small layout issues

daehn:
# What's in this PR?

* Remove `CommonConnectionsViewController` from the profile details and the files and related files from the project.
* Limit the width/height of the `UserConnectionView` user image to 264pt.
* Add a pluralisation rule for the "XYZ friend{s} in common" copy.
* Only show the correlation string if the user is not yet connected.
* Reload the connection view on changes (accept, remote connection state changes etc.)
* Update invitation copy.
* Move from one label and layout using newlines to multiple labels in `UserConnectionView`.


marcoconti83:
Lots of code gone 👍

typfel:
💣conflict

daehn:
@typfel fixed!

PR #451, opened 2016-12-05
Fix user avatar tap not working anymore

daehn:
# What's in this PR?

* Tapping user avatars in the search UI was broken with https://github.com/wireapp/wire-ios/pull/439 when the change was made to present the search UI, archive list etc instead of adding them as child view controller. Due to this change the `parentViewController` property was nil and we failed to present the controller.
* Remove some unused code as the common connections have been removed from the `ProfileDetailViewController`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
+++ b/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
@@ -504,8 +504,8 @@ static NSUInteger const StartUIInitiallyShowsKeyboardConversationThreshold = 10;
     UICollectionViewCell *cell = [self.startUIView.collectionView cellForItemAtIndexPath:indexPath];
     
     [self.profilePresenter presentProfileViewControllerForUser:bareUser
-                                                  inController:self.parentViewController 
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
```diff
@@ -504,8 +504,8 @@ - (void)presentProfileViewControllerForUser:(id<ZMBareUser>)bareUser atIndexPath
     UICollectionViewCell *cell = [self.startUIView.collectionView cellForItemAtIndexPath:indexPath];
     
     [self.profilePresenter presentProfileViewControllerForUser:bareUser
-                                                  inController:self.parentViewController 
-                                                      fromRect:[self.parentViewController.view convertRect:cell.bounds fromView:cell] 
+                                                  inController:self
+                                                      fromRect:[self.view convertRect:cell.bounds fromView:cell]
```
mikeger: Would it still work to show the popover on iPad on the right side, i.e. out of left sidebar?

daehn: Yeah it shows the popover on the right side of the conversation list.

PR #453, opened 2016-12-05
Fix project compilation and crash on file send

mikeger:
# Issue 1

We experienced the crashes in file sending. Those seems to happen because not all the components where bumped correctly.

# Issue 2

Fixed basic compilation due to Contacts framework absence. @marcoconti83


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -4849,10 +4856,12 @@
 				TargetAttributes = {
 					1EE85C4F1B5F96FB008E084B = {
 						CreatedOnToolsVersion = 6.4;
+						DevelopmentTeam = W5KEQBF9B5;
```
daehn: Can be removed

daehn: Can be removed

mikeger:
@daehn removed

PR #457, opened 2016-12-06
Fix layout for users without username but with common friends

daehn:
# What's in this PR?

* Fix padding in `UserDetailView`.
* Use `UserDetailView` in `IncomingConnectionView` instead of repeating the logic there.
* Use the first label for the correlation text if the user does not have an username.
* Add accessibility identifier to `SearchResultsCell`.


Wire-iOS/Sources/UserInterface/ConnectRequests/IncomingConnectionView.swift
```diff
@@ -30,21 +30,15 @@ public final class IncomingConnectionView: UIView {
...
-    private let labelContainer = UIView()
-
+    private let userDetailView = UserNameDetailView()
     private let userImageView = UserImageView()
     private let incomingConnectionFooter = UIView()
     private let acceptButton = Button(style: .full)
     private let ignoreButton = Button(style: .empty)
 
-    public var user: User {
+    public var user: ZMUser {
```
daehn: Removing the `ZMBareUser & ZMBareUserConnection & ZMSearchableUser` was related to a change that will come with the PR using the new address book name API.

daehn: Ah I see, still works in the tests passing in a `MockUser`.

Wire-iOS/Sources/UserInterface/ConnectRequests/UserConnectionView.swift
```diff
@@ -66,48 +64,53 @@ public final class UserConnectionView: UIView, Copyable {
...
     private func setup() {
-        [handleLabel, correlationLabel].forEach {
+        [firstLabel, secondLabel].forEach {
             $0.numberOfLines = 0
             $0.textAlignment = .center
         }
 
-        handleLabel.accessibilityIdentifier = "handle"
-        correlationLabel.accessibilityIdentifier = "correlation"
+        firstLabel.accessibilityIdentifier = "handle"
```
mikeger: The accessibility label is not correct any more.

daehn: Good catch, thanks!

daehn:
@mikeger please re-review!

PR #463, opened 2016-12-07
Fix header view moving up on successive connection requests

daehn:
# What's in this PR?

* The height of the `ConnectionRequestCell`, which should neither be used nor inserted anymore was set to 0, which unfortunately will make the table view set it to the default height of 40. We now set it to 0.5 to ensure the header view does not get pushed up (as we use a rotated tableview we actually set the footer view when setting a header view, but as the default position will be scrolled to the bottom the header view will be pushed out of the visible viewport).
* The underlying issue that we still insert connection request system messages into the conversation should be fixed in `wire-ios-sync-engine` which is tracked in [#7737](https://wearezeta.atlassian.net/browse/ZIOS-7737).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConnectionRequestCell.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConnectionRequestCell.m
@@ -40,7 +40,7 @@
 - (void)createConstraints
 {
     [NSLayoutConstraint autoSetPriority:UILayoutPriorityDefaultHigh forConstraints:^{
-        [self.messageContentView autoSetDimension:ALDimensionHeight toSize:0];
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConnectionRequestCell.m
```diff
@@ -40,7 +40,7 @@ - (instancetype)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(NSStr
 - (void)createConstraints
 {
     [NSLayoutConstraint autoSetPriority:UILayoutPriorityDefaultHigh forConstraints:^{
-        [self.messageContentView autoSetDimension:ALDimensionHeight toSize:0];
+        [self.messageContentView autoSetDimension:ALDimensionHeight toSize:0.5];
```
mikeger: Table view at it's best

PR #464, opened 2016-12-08
Added privacy warning on log as discussed

mikeger:
# Issue

The calling logs are including the personal data (phone, email), so it was decided to add additional privacy warning.

Also renamed Link Opening to LinkOpening for sake of removing whitespace from path.


Wire-iOS/Sources/UserInterface/Settings/SettingsTechnicalReportViewController.swift
```diff
@@ -135,6 +140,42 @@ class SettingsTechnicalReportViewController: UITableViewController, MFMailCompos
         }
     }
     
+    override func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+        switch (section) {
+        case 1:
```
typfel: Maybe create an section enum so it more clear what having height 20

mikeger:
@typfel please check, updated!

PR #467, opened 2016-12-08
Use the search user and it's underlying user when presenting profile from search

daehn:
# What's in this PR?

* Request refreshing user data when entering a connection view.
* Use a search users underlying user (if there is one) to access an eventual address book entry.
* Use properties from bare user instead of casting it (this was causing a missing username and common connections label when the profile details were opened from search for an unconnected user).


daehn:
@marcoconti83 I added some logic in order to request the common contacts

PR #468, opened 2016-12-08
Add check if address book access status changed and upload if it became available

daehn:
# What's in this PR?

* We now persist the last address book authorisation status so we can check if it changed on app start.
* If the status changed and became available we upload the address book immediately.
* We also check if we should upload the address book every time the application enters the foreground and not only on network reachability changes.

### DIFF
--- a/Wire-iOS/Sources/Managers/AddressBookHelper.swift
+++ b/Wire-iOS/Sources/Managers/AddressBookHelper.swift
@@ -60,7 +60,9 @@ extension AddressBookHelper {
                 return
             }
             
-            ABAddressBookRequestAccessWithCompletion(addressBookRef) { (granted, error) in
+            ABAddressBookRequestAccessWithCompletion(addressBookRef) { [weak self] (granted, error) in
+                self?.persistCurrentAccessStatus()
+
                 DispatchQueue.main.async {
                     callback?(granted)
                 }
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/Managers/AddressBookHelper.swift
```diff
@@ -105,6 +107,26 @@ extension AddressBookHelper {
...
+        UserDefaults.standard.set(NSNumber(value: status), forKey: addressBookLastAccessStatusKey)
+    }
+
+    private var lastAccessStatus: ABAuthorizationStatus? {
+        guard let value = UserDefaults.standard.object(forKey: addressBookLastAccessStatusKey) as? NSNumber else { return nil }
+        return ABAuthorizationStatus(rawValue: value.intValue)
+    }
+
+    @objc public var accessStatusDidChangeToGranted: Bool {
+        guard let lastStatus = lastAccessStatus else { return false }
```
marcoconti83: If we have no last access status, but now we have access, should we say that the access status changed to granted?

daehn: I am thinking about how we could end up in this situation, we store the status after asking for permissions, before that point we will not have a `lastStatus`. But AFAIK the user will not see the option to enable AB access until we asked at least once so there should not be the possibility that we did not have a status and then it changes to `.granted` if I am not missing anything.

PR #472, opened 2016-12-09
Only show handle if it is non-empty.

daehn:
# What's in this PR?

* The BE sometimes returns an empty handle (at least on staging) until the change has been propagated, to avoid showing an @-sign without a handle we now additionale check the handle's length.


mikeger:
Screenshot test?

daehn:
@mikeger added

daehn:
@mikeger wrong base branch, will recreate

PR #474, opened 2016-12-09
Connect to people address book

mikeger:
# Issue 

Before with one-way autoconnect it was not possible to have person that is in your address book and on wire. Now it is possible, and we need to handle this case.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
+++ b/Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
@@ -65,7 +65,7 @@
                                                                         focusOnView:YES
                                                                            animated:YES];
         }];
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
```diff
@@ -81,40 +111,30 @@ - (NSArray *)actionButtonTitlesForContactsViewController:(ContactsViewController
     return @[
              NSLocalizedString(@"contacts_ui.action_button.open", @""),
              NSLocalizedString(@"contacts_ui.action_button.invite", @""),
+             NSLocalizedString(@"connection_request.send_button_title", @""),
```
typfel: add a `TODO` to make a new string for this.

PR #475, opened 2016-12-12
Fix analytics event not being triggered on phones

daehn:
# What's in this PR?

* There was a bug preventing the takeover screen event to be tracked on non `.pad` devices, guard considered harmful.


typfel:
Should go into release branch?

mikeger:
Guard is not good for usual if logic!

daehn:
@mikeger agree, very easy to miss a guard somewhere in the code that results in following lines not being executed.

PR #478, opened 2016-12-13
Only show takeover after initial sync

daehn:
# What's in this PR?

* Check and observer initial sync status before showing the takeover.
* Dismiss the takeover when the handle becomes available (when set remotely on a secondary device for example).

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/94 has been merged.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -863,3 +876,13 @@ - (void)updateArchiveButtonVisibility
 }
 
 @end
+
+@implementation ConversationListViewController (InitialSyncObserver)
```
mikeger: Not totally clear why one observer callback is in swift and another one is in ObjC, I guess you needed the protocol conformance for this one? Cannot be done in Swift?

daehn: Could be moved to Swift as well, but then we would need to expose more properties of the class. Now the initial call to present the takeover are in the main file in Objective-C, everything after that is handled in the Swift extension.

daehn:
@mikeger @typfel dependencies are bumped, please merge when ✅

PR #482, opened 2016-12-13
Only show the takeover if the self user is not in a dirty state.

daehn:
# What's in this PR?

* Only show the takeover if the self user does not need to be updated from the backend.
* Move the check into `viewWillAppear` to ensure we show the takeover at some point when the user is not in a dirty state anymore but the user does not have a username.

### Depends on:
- [ ] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/181 has been merged.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -229,7 +230,7 @@ - (void)viewDidAppear:(BOOL)animated
 
 - (void)requestSuggestedHandlesIfNeeded
 {
-    if (nil == ZMUser.selfUser.handle && self.initialSyncCompleted) {
+    if (nil == ZMUser.selfUser.handle && self.initialSyncCompleted && !ZMUser.selfUser.needsToBeUpdatedFromBackend) {
```
typfel: 🙈

daehn: dirtyhacks.com

daehn: We might want to abstract this later in a more general property to check if all hotfix related changes have been done, e.g. `ZMUser.isPendingHotFixChanges`.

Issue #332, opened 2016-10-24
Static and bigger sending button on iOS


IpsmLorem:
It would be easier to have a static all height long sending button (like a box) on iOS.
When writing a a several lines message, the button is moving up and the thumb fails to reach it.
![cap](https://cloud.githubusercontent.com/assets/1250795/19656485/7fa20928-9a20-11e6-9307-146bdf4671a1.png)

KoalaBoomka:
@IpsmLorem,
thanks for the suggestions. This is really nice to have it pinned to the bottom of the input field from the usability point of view.

jenskasemets:
Hello, @IpsmLorem,

This change was now in yesterday's release: 2.26

PR #483, opened 2016-12-14
Add input bar send and emoji button improvements

daehn:
# What's in this PR?

* Align emoji and send button sticky to the bottom instead of to the top of the input bar.
* Increase the hit area of the send button to the right.


mikeger:
Screenshots where not changed?

daehn:
@mikeger we should add a `Danger` rule that checks "if any of the modified files contains a UIView subclass and there are no snapshot tests added or changed leave a comment on the PR"

daehn:
@mikeger please re-review

daehn:
This fixes https://github.com/wireapp/wire-ios/issues/332

PR #485, opened 2016-12-14
Fix crash when using emoji as handle input

daehn:
# What's in this PR?

* There was a crash when trying to input an emoji as username.
* We now check if the handle, which we already checked to be composed out of the valid characters is equal to its transliteration to Unicode names. As alphanumerics will stay the same in this process but emoji will be transformed in a way like `🐷 → {PIG FACE}` we can know that there is an invalid character.
* I previously tried to make this more generic an detect emoji in a string in general, however as other non-emoji characters are not equal to their Unicode name this does not work (see https://github.com/wireapp/wire-ios-utilities/pull/10 for more information).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/ChangeHandleViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/ChangeHandleViewController.swift
@@ -158,7 +158,7 @@ struct HandleChangeState {
     /// This function does not update the `HandleChangeState` itself.
     func validate(_ handle: String) throws {
         let subset = CharacterSet(charactersIn: handle).isSubset(of: HandleChangeState.allowedCharacters)
-        guard subset else { throw ValidationError.invalidCharacter }
+        guard subset && handle.isEqualToUnicodeName else { throw ValidationError.invalidCharacter }
         guard handle.characters.count >= HandleChangeState.allowedLength.lowerBound else { throw ValidationError.tooShort }
         guard handle.characters.count <= HandleChangeState.allowedLength.upperBound else { throw ValidationError.tooLong }
         guard handle != currentHandle else { throw ValidationError.sameAsPrevious }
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/ChangeHandleViewController.swift
```diff
@@ -158,7 +158,7 @@ struct HandleChangeState {
     /// This function does not update the `HandleChangeState` itself.
     func validate(_ handle: String) throws {
         let subset = CharacterSet(charactersIn: handle).isSubset(of: HandleChangeState.allowedCharacters)
```
typfel: Interestingly this check work also for emoji:s if you when you create the handle character set include at least one normal character.

```
var handleSet = CharacterSet(charactersIn: "a")
handleSet.insert(charactersIn: "😎")
handleSet.isSubset(of: HandleChangeState.allowedCharacters) == false
```

daehn: 📡 Radar time

daehn: I found a solution that works similar to https://github.com/wireapp/wire-ios-utilities/pull/10 and also works for non latin characters by first normalising the String: 

```swift
extension String {

    static private var transforms: [StringTransform] {
        return [
            .toLatin,
            .stripDiacritics,
            .stripCombiningMarks
        ]
    }

    private var normalized: String? {
        return String.transforms.reduce(self) {
            $0?.applyingTransform($1, reverse: false)
        }
    }

    var containsEmoji: Bool {
        guard let normalized = normalized else { return false }
        let unicodeName = normalized.applyingTransform(.toUnicodeName, reverse: false)
        return self != unicodeName
    }
}


"a😎وتية ومرئية آمنة ".containsEmoji  // true
```

I will update the `wire-ios-utilities` PR.

daehn: I updated https://github.com/wireapp/wire-ios-utilities/pull/10

PR #489, opened 2016-12-15
View content of file transfers

mikeger:
# What's new

- Refactored conversation content cells to views in order to reuse them in the collection view
- Refactored the way we open files into `MessagePresenter`
- Added file transfer sections to `CollectionsViewController`


daehn:
How come the font color in the user image changed?

Wire-iOS/Sources/UserInterface/Collections/CollectionsView.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.constrainViews()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    private func constrainViews() {
+        constrain(self, self.navigationBar, self.blurView, self.collectionView) { (selfView: LayoutProxy, navigationBar: LayoutProxy, blurView: LayoutProxy, collectionView: LayoutProxy) -> () in
+            navigationBar.left == selfView.left
```
daehn: We could use `leading` and `trailing` instead to make it work with RTL languages in case we add navigation items at some point.

Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -0,0 +1,242 @@
...
+        self.messagePresenter.analyticsTracker = self.analyticsTracker
+        
+        self.view = self.collectionsView
+        
+        self.collectionsView.collectionView.delegate = self
+        self.collectionsView.collectionView.dataSource = self
+        self.navigationItem.titleView = ConversationTitleView(conversation: self.conversation)
+        
+        
+        let button = IconButton.iconButtonDefault()
```
daehn: I think it would make sense to move the calls unrelated to loading the view from `loadView` into `init` or `viewDidLoad`

Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.m
```diff
@@ -0,0 +1,212 @@
...
+        [self openLocationMessage:message];
+    }
+    else if ([Message isFileTransferMessage:message]) {
+        [self openFileMessage:message targetView:targetView];
+    }
+}
+
+- (void)openLocationMessage:(id<ZMConversationMessage>)message
+{
+    // TODO!!
```
daehn: TODO

Wire-iOS/Sources/UserInterface/Collections/CollectionsView.swift
mikeger: Not sure if it would add value, the close button would move from right to left?

daehn: Yes, which is the default location for RTL languages

mikeger: I checked, it works already since the content of navbar is aligned properly by the navbar itself (there is a benefit of using it, I found it).

mikeger:
@daehn thanks for the review, here is the updated version!

PR #491, opened 2016-12-16
Collections: Grouping and opening content

mikeger:
# Changes

- Added "Library" icon and updated stylekit
- Added section headers and made them interactive
- Added ability to show only certain section in the controller

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
@@ -19,11 +19,37 @@
 
 import Foundation
 
+public struct CollectionsSectionSet: OptionSet {
+    public let rawValue: UInt
+    
+    public init(rawValue: UInt) {
+        self.rawValue = rawValue
+    }
+    
+    public init?(index: UInt) {
+        self = type(of: self).allAsArray[Int(index)]
+    }
+    
+    public static let none = CollectionsSectionSet(rawValue: 0)
+    public static let images = CollectionsSectionSet(rawValue: 1)
+    public static let filesAndAudio = CollectionsSectionSet(rawValue: 1 << 1)
+    public static let videos = CollectionsSectionSet(rawValue: 1 << 2)
+    public static let links = CollectionsSectionSet(rawValue: 1 << 3)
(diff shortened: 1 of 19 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -19,11 +19,37 @@
...
+        self = type(of: self).allAsArray[Int(index)]
+    }
+    
+    public static let none = CollectionsSectionSet(rawValue: 0)
+    public static let images = CollectionsSectionSet(rawValue: 1)
+    public static let filesAndAudio = CollectionsSectionSet(rawValue: 1 << 1)
+    public static let videos = CollectionsSectionSet(rawValue: 1 << 2)
+    public static let links = CollectionsSectionSet(rawValue: 1 << 3)
+    
+    public static let all = CollectionsSectionSet(rawValue: UInt.max)
```
daehn: I think `all` should be expressed as an union of all the other types (see https://developer.apple.com/reference/swift/optionset):
```swift
public static let all: CollectionSectionSet = [images, filesAndAudio, videos, links] 
```

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -421,7 +403,7 @@ - (void)setConversation:(ZMConversation *)conversation
 
 - (void)setupNavigatiomItem
 {
-    self.titleView = [[ConversationTitleView alloc] initWithConversation:self.conversation];
+    self.titleView = [[ConversationTitleView alloc] initWithConversation:self.conversation interactive:YES];
```
daehn: The `titleView` should only be interactive if the conversation's connection is not an outgoing connection request, see https://github.com/wireapp/wire-ios/pull/491/files#diff-64b0ad5fc5c36accc32ddffdef3c09ccL50

mikeger: Right, but this logic is not changed I think, or I am missing something?

mikeger:
@daehn updated!

PR #492, opened 2016-12-16
Track button taps instead of waiting for BE results

daehn:
# What's in this PR?

* We were tracking the result of a tap on the `"Keep suggested"` button on the takeover view controller when we receive the callback from sync-engine. When selecting choose own we do not remove the observer as we want to dismiss the controller once a different username has been chosen, this lead to non matching tracking numbers as the keep suggested event was tracked in this case as well.
* The intention of the user is more important so we now track the button taps directly.


daehn:
If we are doing a hotfix then this change should go in as well. // cc @jenskasemets

PR #493, opened 2016-12-16
Show usernames in reactions list view

daehn:
# What's in this PR?

* Add username label to `ReactionCell`.


Wire-iOS/Sources/UserInterface/Conversation/ReactionsList/ReactionCell.swift
```diff
@@ -21,37 +21,60 @@ import zmessaging
...
 
+
 @objc public class ReactionCell: UICollectionViewCell {
     public let userImageView = UserImageView(magicPrefix: "people_picker.search_results_mode")
-    public let userNameLabel = UILabel()
+    public let userDisplayNameLabel = UILabel()
+    public let usernameLabel = UILabel()
+
+    var displayNameVerticalConstraint: NSLayoutConstraint?
+    var displayNameTopConstraint: NSLayoutConstraint?
```
mikeger: You can also put two labels in the container and center it vertically, but I guess this would also work.

PR #495, opened 2016-12-19
Added use of the new data model fetching API

mikeger:
Integrated the new API from Data Model.


mikeger:
PR updated with links support and autolayout for cells

Wire-iOS/Sources/UserInterface/Collections/CollectionFileCell.swift
```diff
@@ -47,6 +47,24 @@ final public class CollectionFileCell: UICollectionViewCell, Reusable {
         self.loadView()
     }
     
+    var isHeightCalculated: Bool = false
+    var containerWidth: CGFloat = 320
```
typfel: Should be hardcoded?

mikeger: Just the default value, actually set when the cell is created.

mikeger:
@typfel updated the PR

PR #497, opened 2016-12-19
Fix/conversation list resorting

marcoconti83:
Keep conversation list in order after coming to foreground

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
@@ -132,6 +132,7 @@
 
 @property (nonatomic, assign) BOOL openArchiveGestureStarted;
 @property (nonatomic) CGFloat contentControllerBottomInset;
+@property (nonatomic) IconButton *debugButton;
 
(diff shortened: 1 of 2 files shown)


marcoconti83:
⚠️ Found issue, do not merge ⚠️

typfel:
@marcoconti83  so we can close this?

marcoconti83:
So this can be merged. We will need then to test in develop build with the new data model framework to confirm that it fixes the issue. It's not easy to test.

typfel:
Oh I thought it would all be handled from SE side.

marcoconti83:
This was also addressed not on sync engine but on data model. So now in data model we resort the list when we are flushing the accumulate changes (basically when we go back to live processing of changes). 
However I would like to leave a "refresh view" in the UI (no resorting, the underlying list should be resorted by data model already), which if I understood correctly causes the entire UI view to be reloaded from the underlying list (which now should already be resorted by data model).
As I mentioned this issue is not easy to reproduce (it requires leaving the phone alone for a few hours) so better to add a bit more extra safety and then see how it behaves on real device.

marcoconti83:
@typfel any further thought on this?

typfel:
No but would nice to clean this up later if we refactor the observation

PR #498, opened 2016-12-19
Align download links

infotexture:
The ’get.wire.com’ variant is used in more places and offers platform-specific redirects that take users directly to the app stores on supported platforms.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -98,9 +98,9 @@
 
 "send_invitation.subject" = "Connect with me on Wire";
 "send_invitation.text" = "I’m on Wire, search for %@ or visit get.wire.com.";
-"send_invitation_no_email.text" = "I’m on Wire. Visit https://get.wire.com to connect with me.";
+"send_invitation_no_email.text" = "I’m on Wire. Visit get.wire.com to connect with me.";
 
-"send_personal_invitation.text" = "I’m on Wire, talk to you there. https://get.wire.com.";
+"send_personal_invitation.text" = "I’m on Wire, talk to you there: get.wire.com.";
 
 
 // +++++++++++++++++++++++++++++++++++++++++++++++++++++x
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -97,10 +97,10 @@
 "peoplepicker.hide_search_result_progress" = "Hiding…";
 
 "send_invitation.subject" = "Connect with me on Wire";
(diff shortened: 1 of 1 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -97,10 +97,10 @@
 "peoplepicker.hide_search_result_progress" = "Hiding…";
 
 "send_invitation.subject" = "Connect with me on Wire";
-"send_invitation.text" = "I’m on Wire, search for %@ or visit wire.com/download.";
+"send_invitation.text" = "I’m on Wire, search for %@ or visit get.wire.com.";
 "send_invitation_no_email.text" = "I’m on Wire. Visit https://get.wire.com to connect with me.";
 
-"send_personal_invitation.text" = "I’m on Wire, talk to you there. https://get.wire.com";
+"send_personal_invitation.text" = "I’m on Wire, talk to you there. https://get.wire.com.";
```
jenskasemets: Adding dot at the end of URL might break it in some cases, I would suggest not to do that.

infotexture: OK, so that was intentionally left out in the original string — thought it was a mistake.

In that case, feel free to omit this change when merging.

marcoconti83: Hi @infotexture, it is not possible for us to selectively merge part of the pull request and leave another part out. It's either all or nothing. Could you please edit the pull request on your side? Thanks!

infotexture: Sure, I'll push an update.

marcoconti83: @jenskasemets

jenskasemets: @infotexture Lines 100-103 are all used currently, but can be simplified and use similar copy for some of them. Because they are used in different places in the client, they need to have different identifiers

Most probably all modern browsers and email clients detect all valid format y.x.com address@es as hyperlinks, removing protocol shouldn't be a problem.
@mykola-mokhnach Do you have any experience with this?

mykola-mokhnach: unfortunately not, but I wouldn't rely on the fact, that all browser and email clients parse it properly

jenskasemets: I would merge it in as such, and change if we'll find out that it doesn't work in some conditions.

infotexture: @jenskasemets OK, I pushed an extra commit to remove the protocol from the remaining instances.

PR #516, opened 2016-12-23
Cleaned up jump to message in conversation

mikeger:
# Issue

Removed bogus delays in jump to message in conversation method.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -218,15 +218,22 @@ extension ConversationContentViewController {
...
-            let cellIndexPath = IndexPath(row: indexToShow, section: 0)
-            self.tableView.scrollToRow(at: cellIndexPath, at: .middle, animated: true)
-            delay(0.35) {
-                completion?(self.tableView.cellForRow(at: cellIndexPath) as! ConversationCell)
-            }
+
+        let indexToShow = self.messageWindow.messages.index(of: messageToShow)
+
+        if indexToShow == NSNotFound {
+            self.expectedMessageToShow = messageToShow
```
typfel: In which case does it take branch and when does it take the other?

mikeger: So right now expanding is synchronous, at least for my test conversation (which is rather small), but I think I cannot rely on it's implementation, so if it is async we have to wait until message window has the desired message.

typfel: Yeah I think it's always synchronous. But ok, I get your point.

PR #518, opened 2016-12-23
Correct title for collections view

mikeger:
# Issue

We supposed to show both title of conversation and title of the screen in navigation title item.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/TwoLineTitleView.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/TwoLineTitleView.swift
@@ -1,0 +1,64 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Collections/TwoLineTitleView.swift
```diff
@@ -0,0 +1,64 @@
...
+            subtitleLabel.trailing == selfView.trailing
+            subtitleLabel.bottom == selfView.bottom
+        }
+
+        CASStyler.default().styleItem(self)
+
+        setNeedsLayout()
+        layoutIfNeeded()
+        
+        let size = self.systemLayoutSizeFitting(CGSize(width: 320, height: 44))
```
typfel: Why is this necessary

mikeger: I think the navigation bar is not too good with autolayout, if I would not do this the title view stays with zero frame

typfel: :/

PR #527, opened 2016-12-29
Track database save failures

marcoconti83:
Send event to localytics when the database needs to be rolled back because of a save error


marcoconti83:
Tested this on simulator, it works as expected

PR #539, opened 2017-01-09
Split share extension into multiple files

marcoconti83:
All share extension VC were in one file. Split for clarity.
Also renamed the name that appears in the share extension when picking the app ("Wire Share Extension" -> "Wire")


Wire-iOS.xcodeproj/xcshareddata/xcschemes/WireExtensionComponents.xcscheme
```diff
@@ -1,80 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
typfel: This should be deleted?

marcoconti83: Not sure exactly what happened to the PR, I merged in from DEV, now it looks better, please have a look now.

PR #543, opened 2017-01-10
Update for new collections design

mikeger:
# Issue

The existing design of the collection view seems to be not that clear to the users.

# Solution

The new design reveals the connection between the elements using the insets and groups. Also sections now have the icons.

Additionally the new term, name with accent color, is added. The name accent color is generated from the user accent color with the certain opacity. The opacity is defined in the table and is different per color. Opacity also different for the light/dark color schemes.

Screenshot:

simulator screen shot jan 10 2017 17 31 19

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
@@ -285,7 +285,7 @@ extension CollectionsViewController: UICollectionViewDelegate, UICollectionViewD
             return min(self.fileAndAudioMessages.count, max)
             
         case CollectionsSectionSet.videos:
-            let max = self.inOverviewMode ? self.maxOverviewVideoElementsInTable : Int.max
+            let max = self.inOverviewMode ? self.maxOverviewElementsInGrid : Int.max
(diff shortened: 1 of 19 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -395,8 +399,8 @@ extension CollectionsViewController: UICollectionViewDelegate, UICollectionViewD
 
             let cell = collectionView.dequeueReusableCell(withReuseIdentifier: CollectionVideoCell.reuseIdentifier, for: indexPath) as! CollectionVideoCell
             cell.message = message
-            cell.desiredWidth = collectionView.bounds.size.width
-            cell.desiredHeight = collectionView.bounds.size.width * (3.0 / 4.0)
+            cell.desiredWidth = self.girdElementSize.width
```
daehn: Typo `girdElementSize` → `gridElementSize`

mikeger: https://en.wikipedia.org/wiki/Gird

mikeger:
@daehn updated

PR #544, opened 2017-01-10
Fix showing image placeholder in link previews in text messages cells.

daehn:
# What's in this PR?

* There was a bug when creating an `ArticleView` in `TextMessageCell`, for forwarding we do not want to show the image, but if there is none and the context is not forwarding we do not want to show one as well.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
@@ -201,8 +201,9 @@
     }
 
     if (linkPreview != nil && nil == self.linkAttachmentViewController && !isGiphy) {
-        ArticleView *articleView = [[ArticleView alloc] initWithImagePlaceholder:textMesssageData.hasImageData];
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m
```diff
@@ -201,8 +201,9 @@ - (void)configureForMessage:(id<ZMConversationMessage>)message layoutProperties:
     }
 
     if (linkPreview != nil && nil == self.linkAttachmentViewController && !isGiphy) {
-        ArticleView *articleView = [[ArticleView alloc] initWithImagePlaceholder:textMesssageData.hasImageData];
-        articleView.imageHeight = self.smallLinkAttachments ? 0 : 144;
```
typfel: `articleView.imageHeight` is still used elsewhere?

daehn: Yes in the context of forwarding we set it to 0 in order to hide the image (if there is one).

PR #546, opened 2017-01-11
Fix failing property tests

marcoconti83:
Tests where failing when extracting a property that was set as a boolean, because it was extracted as a number instead.

# Remarks
I just fixed the casting to the right type (number/bool). However the entire SettingsPropertyValue system seems a bit complicated and I think it could be simplified in more concise code. Let's keep it in mind for the future.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
@@ -201,17 +201,14 @@ class SettingsUserDefaultsProperty : SettingsProperty {
     }
     
     internal func value() -> SettingsPropertyValue {
-        let value : AnyObject? = self.userDefaults.object(forKey: self.userDefaultsKey) as AnyObject?
-        if let boolValue = value as? Bool {
+        switch self.userDefaults.object(forKey: self.userDefaultsKey) as AnyObject? {
+        case let boolValue as Bool:
             return SettingsPropertyValue.propertyValue(boolValue as AnyObject?)
-        }
-        if let numberValue = value as? NSNumber {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
```diff
@@ -202,10 +202,13 @@ class SettingsUserDefaultsProperty : SettingsProperty {
     
     internal func value() -> SettingsPropertyValue {
         let value : AnyObject? = self.userDefaults.object(forKey: self.userDefaultsKey) as AnyObject?
-        if let numberValue : NSNumber = value as? NSNumber {
+        if let boolValue = value as? Bool {
+            return SettingsPropertyValue.propertyValue(boolValue as AnyObject?)
+        }
+        if let numberValue = value as? NSNumber {
```
typfel: should be `else if`?

marcoconti83: The previous `if` returns in the positive case. In theory we need no `else` anywhere in this method.

marcoconti83: I can remove all `else` for consistency

mikeger: I would vote to have if-else, so the flow is clear. Also if somebody forgot to return in one of the cases if there would be if-else the compiler would catch the isssue.

marcoconti83:
I replaced with a `switch` which it's basically what the `if-else` was trying to do. I think it's less ambiguous. What do you think? 
@mikeger @typfel

PR #552, opened 2017-01-11
Added message forwarding from image view

mikeger:
# Issue

There was a menu items missing for the full screen view, I added forward and reveal in conversation.
Also some design changes here and there.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
@@ -86,6 +86,13 @@
 
 @end
 
+
+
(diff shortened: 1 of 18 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -814,3 +821,31 @@ - (void)conversationWindowDidChange:(MessageWindowChangeInfo *)note
...
+        return YES;
+    }
+    
+    return NO;
+}
+
+- (void)wantsToPerformAction:(MessageAction)action forMessage:(id<ZMConversationMessage>)message
+{
+    switch (action) {
+        case MessageActionForward:
```
daehn: Might be simpler with an `if` as long as we only care about one case.

PR #553, opened 2017-01-11
Fix/share extension titles

marcoconti83:
There were a few issues with share extension:
- files were shared as urls (`file:///storage/aasdsadasdsad/...`) instead of files
- the share interface had an empty box in case of files or URLs without a title. Now it has some "suggested test" to make it clear that you are sharing something

# Changes 
- there are a lot of async operations when processing share extension attachments. Now all callbacks are on the UI thread for safety.
- extracted common async operations to common methods
- fixed detection of files vs. URLs, based on whether the URL has a file scheme


Wire-iOS Share Extension/ShareViewController.swift
```diff
@@ -42,10 +42,7 @@ class ShareViewController: SLComposeServiceViewController {
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        
-        if let rightButtonBarItem = navigationController?.navigationBar.items?.first?.rightBarButtonItem {
```
typfel: Why is this not necessary anymore? This was replacing the default post button action which dismisses the share extension and then calls didSelectPost.

marcoconti83: That is the default behaviour, as I understood. `didSelectPost` is called automatically.

typfel: Yes but it also dismisses the share extension and we need show progress.

typfel: This code was a work around to prevent the dismissing until the message is sent.

marcoconti83: I see, then I have to put it back, thanks for the tip. However now I'm confused on why we created (before this PR) an implementation of `didSelectPost`, that was also sending a text message. Does it mean we were sending it twice? or was the `didSelectPost` ignored (dead code)?

marcoconti83:
@typfel should be fixed now. I tested with a large file, the progress view is back!

PR #555, opened 2017-01-12
Fix share extension sending cancellation

daehn:
# What's in this PR?

* Share extension sending cancellation broke due to changes made in https://github.com/wireapp/wire-ios/pull/550, this PR fixes this.

### DIFF
--- a/Wire-iOS Share Extension/ShareViewController.swift
+++ b/Wire-iOS Share Extension/ShareViewController.swift
@@ -140,7 +140,14 @@ class ShareViewController: SLComposeServiceViewController {
         let progressSendingViewController = SendingProgressViewController()
         
         progressSendingViewController.cancelHandler = { [weak self] in
-            self?.cancel()
+            guard let `self` = self else { return }
+            self.observer?.sendables.filter {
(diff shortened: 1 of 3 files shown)


daehn:
I added a missing enqueue onto the sharing engine, please re-review @mikeger

Wire-iOS Share Extension/ShareViewController.swift
```diff
@@ -141,10 +141,15 @@ class ShareViewController: SLComposeServiceViewController {
         
         progressSendingViewController.cancelHandler = { [weak self] in
             guard let `self` = self else { return }
-            self.observer?.sendables.filter {
+
+            let sendablesToCancel = self.observer?.sendables.lazy.filter {
```
mikeger: Why `lazy`?

daehn: Just to avoid iterating over the array twice, but as it should not be that big (I don't assume many people will try to share more than ~10 items at a time, and if so we might want to limit it using the activation rules) it actually should not make a difference. The main part was to enqueue the cancellations and cancel all sendables in one enqueue block to make it cancel them a bit faster as well.

PR #561, opened 2017-01-12
Lower the share extension deployment target

daehn:
# What's in this PR?

* Lower the share extension's deployment target to 8.
* Use iOS 8 compatible API.


mikeger:
Did you tried running it on iOS 8 device / simulator?

daehn:
Not yet, Xcode is still downloading them 🙈

PR #565, opened 2017-01-12
Localisation for release 2.28

jenskasemets:
What's in this PR:

- Added localisation files and strings for share extension:.
- Added localisation for collection.
- Removed protocol from invite links.
- Other updated strings from Crowdin. Some of them fall back to English because base string changed and there's no updated translation yet.

### DIFF
--- a/Wire-iOS/Resources/da.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/da.lproj/Localizable.strings
@@ -97,10 +97,10 @@
 "peoplepicker.hide_search_result_progress" = "Skjuler…";
 
 "send_invitation.subject" = "Forbind dig med mig på Wire";
-"send_invitation.text" = "Jeg bruger Wire. Søg efter %@ eller besøg get.wire.com";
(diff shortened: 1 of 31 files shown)


Wire-iOS/Resources/da.lproj/Localizable.strings
```diff
@@ -781,5 +781,5 @@
 "collections.section.files.title" = "Filer";
 "collections.section.videos.title" = "Videoer";
 "collections.section.links.title" = "Links";
-"collections.section.all.button" = "Alle %d →";
-"collections.section.no_items" = "Ingen elementer i biblioteket";
+"collections.section.all.button" = "Show all %d →";
```
typfel: This is replacing the already translated text to english?

mikeger: The original string was `All  %d ->`  in english, it was changed to `Show all %d ->` so I guess the translation system decided that it should be reset in the other languages. I am not 100% sure it is true, but the translation should change I guess.

typfel: OK makes sense

PR #572, opened 2017-01-13
Allow browsing collection of images by swiping

mikeger:
Next step would be to use the same controller in the conversation.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
@@ -1,0 +1,171 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -0,0 +1,171 @@
...
+        return self.imageMessages.count
+    }
+    
+    func pageViewController(_ pageViewController: UIPageViewController, viewControllerAfter viewController: UIViewController) -> UIViewController? {
+        guard let imageController = viewController as? FullscreenImageViewController else {
+            fatal("Unknown controller \(viewController)")
+        }
+        
+        guard let messageIndex = self.imageMessages.index(where: { (message: ZMConversationMessage) -> (Bool) in
+            message.hash == imageController.message.hash
```
daehn: In other places we use the `nonce` to check the identity  of messages, why not here (Maybe we also override the `hash` to be relying on the nonce)?

mikeger: Good point, I guess I can create equal method that casts the things to `ZMMessage`, then I can have access to the remote identifier.

PR #576, opened 2017-01-16
Collections images update

mikeger:
# Improvements

- Fixed name accent color for violet color
- Added actions toolbar to swiping images controller. Now you can save/forward/copy/reveal image from the toolbar.
- Unified message actions handling.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
@@ -513,8 +513,7 @@ extension CollectionsViewController: CollectionCellDelegate {
         }
         
         switch action {
-        case .forward: fallthrough
-        case .showInConversation:
+        case .forward, .showInConversation:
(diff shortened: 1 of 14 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -549,6 +549,8 @@ extension CollectionsViewController: MessageActionResponder {
         if Message.isImageMessage(message) {
             switch action {
             case .forward: fallthrough
+            case .copy: fallthrough
+            case .save: fallthrough
             case .showInConversation:
```
vytis: A bit more readable when it's written as ` case .forward, .copy, .save, .showInConversation`

mikeger: But less old-school 🎩 
Will update

vytis: :neckbeard:

mikeger:
@vytis updated

PR #577, opened 2017-01-16
Collections: added initial snapshot tests

vytis:
## What's in this PR?

Added snapshot tests for collections controller. The plan is to have each type of section showing on it's own, starting with files. 

Also added empty state and loading spinner snapshots


mikeger:
@vytis did you tried setting the layer speed to 0 for spinner?

vytis:
Trying it now, doesn't seem to fix the problem, but trying to investigate some more.

PR #580, opened 2017-01-16
Do not show collection icon for pending connections

mikeger:
- Do not show collection buttons for pending incoming or outgoing connections

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
@@ -90,9 +90,13 @@ public extension ConversationViewController {
             backButton.hitAreaPadding = CGSize(width: 28, height: 20)
             items.append(UIBarButtonItem(customView: backButton))
         }
-        let collectionsButton = collectionsBarButtonItem
-        collectionsButton.hitAreaPadding = CGSize(width: 0, height: 20)
-        items.append(UIBarButtonItem(customView: collectionsButton))
+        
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -90,9 +90,13 @@ public extension ConversationViewController {
             backButton.hitAreaPadding = CGSize(width: 28, height: 20)
             items.append(UIBarButtonItem(customView: backButton))
         }
-        let collectionsButton = collectionsBarButtonItem
-        collectionsButton.hitAreaPadding = CGSize(width: 0, height: 20)
-        items.append(UIBarButtonItem(customView: collectionsButton))
+        
+        if let connection = conversation.connection, connection.status != .pending && connection.status != .sent {
```
typfel: Will this reload when a connection is accepted?

mikeger: Good question, it is not reloaded, but also not likely that you would have a lot of content right after accepting the connection. I would check with QA and design.

typfel: Seems same logic would apply to calling buttons. How's that handled atm?

mikeger: I update it now the same way the calling buttons are updated: https://github.com/wireapp/wire-ios/pull/580/files#diff-d0adf7620aab7082451935f800968beaR889

PR #585, opened 2017-01-17
Immediately disable the send button when tapped. Add dismissal animation

daehn:
# What's in this PR?

* Immediately disable the share extension's send button once tapped.
* Add dismissal animation to share extension.


Wire-iOS Share Extension/ShareViewController.swift
```diff
@@ -93,18 +93,22 @@ class ShareViewController: SLComposeServiceViewController {
 
     /// invoked when the user wants to post
     func appendPostTapped() {
+        navigationController?.navigationBar.items?.first?.rightBarButtonItem?.isEnabled = false
```
mikeger: Dat key path.

PR #587, opened 2017-01-18
Use default environment for Travis CI

vytis:
## What's the reason behind this PR?

We have only Danger running on Travis CI, but it takes a while for it to finish the job. The reason that backlog of tasks for Mac is huge during the day:
mac-backlog

Compared to Linux:
linux-backlog

Since we don't use any Mac-specific things now it should work just fine with Linux image and hopefully pick up the job quicker during peak hours.


marcoconti83:
Nice catch! 👍

PR #588, opened 2017-01-18
Show progress spinner when attachment needs preparation

daehn:
# What's in this PR?

* The current way the share extension sends messages is by calling the `send` method, which prepares and appends all messages and then showing the progress view.
* As some messages, for example videos, need additional time to be prepared before being appended, the UI is stuck and the user does not get any feedback until the preparation has been done.
* This PR adds a workaround for this issue, a proposal for a better solution (which will take some refactoring and time) is appended at the end of this description.
* The workaround is to pass in an additional block that will only be invoked if there are items that need preparation into the `send` method.
* This workaround requires to keep track if the operation has been cancelled, as the messages might not yet have been appended they can not be cancelled (and the extension might still run in the background which could lead to them being appended).
* One drawback of this workaround is that it will show the preparation screen when sending an image and a video and will start sending the image while doing so.
* There also were a couple of places where we checked for a condition to be met for an early return inside of a dispatch block. These cases have been changed to return before entering the block.

### Proposed future solution

* Instead of calling a `send` method which does all the preparation and message appending I would propose to warp all sending operations in a helper class, which would basically be an unappended, unprepared message that can be constructed fast. These objects can report if they need preparation and can report their progress once appended. This way the progress controller can query if it needs to do preparation and can send one item at a time while showing the progress or preparation spinner.
* This change would require some refactoring as the current sending and preparation logic is all located in the `ShareViewController`, but would also declutter this class a lot.

### DIFF
--- a/Wire-iOS Share Extension/Base.lproj/Localizable.strings
+++ b/Wire-iOS Share Extension/Base.lproj/Localizable.strings
@@ -26,6 +26,7 @@
 "share_extension.conversation_selection.empty.value" = "Choose";
 "share_extension.input.placeholder" = "Type a message";
 "share_extension.sending_progress.title" = "Sending…";
+"share_extension.preparing.title" = "Preparing…";
 "share_extension.not_signed_in.title" = "You need to sign into Wire before you can share anything";
 "share_extension.not_signed_in.close_button" = "Close";
 
--- a/Wire-iOS Share Extension/SendingProgressViewController.swift
+++ b/Wire-iOS Share Extension/SendingProgressViewController.swift
@@ -23,6 +23,10 @@ import Cartography
 
 class SendingProgressViewController : UIViewController {
 
+    enum ProgressMode {
+        case preparing, sending
+    }
+
     var cancelHandler : (() -> Void)?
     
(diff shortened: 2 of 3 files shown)


mikeger:
Just an observation, as the loading animation can work while the main thread is busy, you can potentially always set the spinner to indefinite and begin the animation when you send the message and update it to be deterministic when you get the first progress callback.

daehn:
@mikeger I wanted to avoid it being nondeterministic only for a couple of milliseconds and only show it if we actually perform long running preparation work.

typfel:
The progress animation kicked off in `viewDidAppear` is not messing things up when in .preparing mode?

mikeger:
@daehn Yep, you need to `dispatch_after` or do on the next run loop to allow the animation to start, otherwise would be frozen.

daehn:
@typfel no, without a progress the spinning would not be visible. But I will move it into a better place as this worked accidentally ATM.

daehn:
@mikeger updated with your suggestion, please re-review.

daehn:
A first draft of the proposed solution in the description can be found here: https://github.com/wireapp/wire-ios/pull/594

PR #590, opened 2017-01-18
Collection identifiers fixes

mikeger:
- Fixed localytics events identifiers
- Added action buttons identifiers for image view

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
@@ -91,18 +91,22 @@ internal final class ConversationImagesViewController: UIViewController {
     private func createControlsBar() {
         let copyButton = IconButton.iconButtonDefault()
         copyButton.setIcon(.copy, with: .tiny, for: .normal)
+        copyButton.accessibilityLabel = "copy"
         copyButton.addTarget(self, action: #selector(ConversationImagesViewController.copyCurrent(_:)), for: .touchUpInside)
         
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -91,18 +91,22 @@ internal final class ConversationImagesViewController: UIViewController {
     private func createControlsBar() {
         let copyButton = IconButton.iconButtonDefault()
         copyButton.setIcon(.copy, with: .tiny, for: .normal)
+        copyButton.accessibilityLabel = "copy"
```
typfel: Should not be `copyButton.accessibilityIdentifier`?

mikeger: Both works for automation, I think that `label` is also visible for VoiceOver, so beneficial to disabled people. Should be localized then, but something is better than nothing.

typfel: Feel like a bit messy adding unlocalized strings to the app. Why be consistent use  accessibilityIdentifier.

mikeger: True, but on most recent controls I used label, right now its 160 `identifier` to 50 `label`, so identifier is indeed more frequent to be used.

typfel: 😔 Ok let keep this for now and create a ticket for cleaning this up.

PR #591, opened 2017-01-18
Collections: added visual feedback when image is copied to clipboard

vytis:
## Reason behind this PR

When viewing an image in collection there is a button to copy it to clipboard. As it's not obvious from the icon what it does, this PR adds a text overlay on top of toolbar with text that is shown for a bit and fades out.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -783,3 +783,4 @@
 "collections.section.links.title" = "Links";
 "collections.section.all.button" = "Show all %d →";
 "collections.section.no_items" = "No items in collection";
+"collections.image_viewer.copied.title" = "PICTURE COPIED";
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -783,4 +783,4 @@
 "collections.section.links.title" = "Links";
 "collections.section.all.button" = "Show all %d →";
 "collections.section.no_items" = "No items in collection";
-"collections.image_viewer.copied.title" = "PICTURE COPIED";
+"collections.image_viewer.copied.title" = "Picture copied";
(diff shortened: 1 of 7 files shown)


mikeger:
@vytis looks like the bot is not right, linux issue?

vytis:
Yeah not sure, need to investigate a bit more.

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -783,3 +783,4 @@
 "collections.section.links.title" = "Links";
 "collections.section.all.button" = "Show all %d →";
 "collections.section.no_items" = "No items in collection";
+"collections.image_viewer.copied.title" = "PICTURE COPIED";
```
mikeger: We usually keep localized text in normal case and call `.uppercased()` on it so the translators would not get puzzled and would not change the design.

Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -73,6 +79,16 @@ internal final class ConversationImagesViewController: UIViewController {
             buttonsBar.trailing == view.trailing
             buttonsBar.bottom == view.bottom
             buttonsBar.height == 84
+
+            overlay.height == buttonsBar.height
```
mikeger: You can use `overlay.edges == buttonsBar.edges`, this would create 4 constraints.

Wire-iOS/Sources/UserInterface/Collections/FeedbackOverlayView.swift
```diff
@@ -0,0 +1,61 @@
...
+            label.centerX == container.centerX
+            label.centerY == container.centerY
+            label.left >= container.left + 24
+            label.right <= container.right - 24
+        }
+    }
+
+    public func show(text: String) {
+        titleLabel.text = text
+        UIView.animateKeyframes(withDuration: 3, delay: 0, options: [.calculationModeCubic], animations: {
```
mikeger: So is there a way to have easing in-out for the animation?

PR #592, opened 2017-01-18
Forbid forwarding the files that are not downloaded

mikeger:
- It was possible to forward the file that is not downloaded. This caused the crash.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionAudioCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionAudioCell.swift
@@ -71,12 +71,7 @@ final public class CollectionAudioCell: CollectionCell {
     override open func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
         switch action {
         case #selector(CollectionCell.forward(_:)):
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionAudioCell.swift
```diff
@@ -68,9 +68,18 @@ final public class CollectionAudioCell: CollectionCell {
         }
     }
     
-    public override func prepareForReuse() {
-        super.prepareForReuse()
-        self.message = .none
+    override open func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        switch action {
+        case #selector(CollectionCell.forward(_:)):
+            if let _ = self.message?.fileMessageData?.fileURL {
```
vytis: Maybe worth making an extension to `ZMConversationMessage` with this?

mikeger: I think there is an extension already, let me check

mikeger:
@vytis updated

PR #594, opened 2017-01-18
Extract sending code from ShareViewController

daehn:
# What's in this PR?

* Declutter the code and extract the sending logic into separate classes.
* Objects of type `UnsentSendableType` can be constructed and can be asked if they need to do any additional work before they can be appended to a conversation.
* They expose a `prepare(completion: @escaping () -> Void)` function to prepare their content for appending and a similar `send(completion: @escaping (Sendable?) -> Void)` function to append them. If an objects needs preparation calling `send` before `prepare` will assert.
* The `SendController` encapsulates the preparation and sending of multiple `Sendables` while it does have no knowledge about conversation degradation. Conversation degradation observation is still done in the `PostContent`, which now uses a `SendController` to append messages.
* The `SendController` reports the type of the current operation (`.preparing`, `.startingSending`, `.sending(_)`, `.conversationDegraded(_)` and `.done`) and encapsulates all the sending logic.
* This ensures all messages are ready to be appended and enables a better way of progress reporting, we now show a "Preparing…" screen until all messages are ready to be appended, this change also enables to show a progress like "Sending 1 of 3" or "Preparing 2 of 3" in the future.


mikeger:
@daehn FYI has a conflict now

daehn:
@mikeger on it

daehn:
@mikeger updated

Wire-iOS Share Extension/SendController.swift
```diff
@@ -0,0 +1,141 @@
...
+//
+
+
+import Foundation
+import WireShareEngine
+import ZMCDataModel
+
+
+typealias DegradationStrategyChoice = (DegradationStrategy) -> ()
+typealias Progress = (_ type: ProgressType) -> Void
```
mikeger: I think it can be a little bit confusing to have two types `Progress` and `ProgressType`. I have a suggestion to call `Progress`, as it is essentially a closure or more particularly a callback, `ProgressCallback`. I would also think that it would be important to know to which progress it is referred. I think `ProgressType` can be renamed as `SendingProgress` to clarify about which progress it is about (knowing Swift naming, the important words should not be omitted). Therefore I would vote to have `SendingProgress` and `SendingProgressCallback` or something around it. I hope my argumentation is clear :)

Wire-iOS Share Extension/SendController.swift
```diff
@@ -0,0 +1,141 @@
...
+
+
+/// This enum specifies the current state of the sending progress and is passed
+/// as a parameter in a `Progress` closure.
+enum ProgressType {
+    case preparing // Some attachments need to be prepared, this case is not always invoked.
+    case startingSending // The messages are about to be appended, the callback will always be invoked axecatly once.
+    case sending(Float) // The progress of the sending operation.
+    case conversationDidDegrade((Set<ZMUser>, DegradationStrategyChoice)) // In case the conversation degrades this case will be passed.
```
mikeger: I would also vote to exclude this as a separate callback, since it does not really refer to the progress. But it's might be too late!

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -0,0 +1,218 @@
...
+import MobileCoreServices
+import WireExtensionComponents
+
+
+enum UnsentSendableError: Error {
+    case unsupportedAttachment
+}
+
+
+protocol UnsentSendableType {
```
mikeger: I think it is not necessary to add `Type` postfix to the name, since `Sendable` itself does not have it.

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -0,0 +1,218 @@
...
+
+
+extension UnsentSendableType {
+    func prepare(completion: @escaping () -> Void) {
+        precondition(needsPreparation, "Ensure this objects needs preparation, c.f. `needsPreparation`")
+    }
+}
+
+
+class UnsentSendableAbstract {
```
mikeger: I would call it `Base...` since usually you cannot instantiate an abstract class:
```
An abstract class is a class that cannot be instantiated, but must be inherited from. 
```

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+}
+
+
+class UnsentFileSendable: UnsentSendableAbstract, UnsentSendableType {
+
+    private let attachment: NSItemProvider
+    private var metadata: ZMFileMetadata?
+    private var fm = FileManager.default
```
mikeger: The good book says to avoid the abbreviations: https://swift.org/documentation/api-design-guidelines/ also the thing is not var, also it is used only in one method -> can be a local let with name `fileManager`

Wire-iOS Share Extension/SendController.swift
daehn: Yeah, for me the progress was meant to include all steps of the sending operation, that's why I, maybe it should be a `SendingState` or something like that, as progress callbacks usually only care about the completed percentage.

Wire-iOS Share Extension/UnsentSendable.swift
marcoconti83: I like the quoting of the "good book" :)

Wire-iOS Share Extension/SendController.swift
marcoconti83: > Write a documentation comment for every declaration. Insights gained by writing documentation can have a profound impact on your design, so don’t put it off.

when naming is not enough to capture the full nuance of a class/var/method, I think it's important to add code documentation to make more responsibilities clear.

Wire-iOS Share Extension/UnsentSendable.swift
daehn: True, I think we use `fm` for the default FileManager in a lot of places (but maybe I introduced it there as well 🙈)…

Wire-iOS Share Extension/SendController.swift
daehn: @marcoconti83 I somehow missed your last comment until now, you are right, I will add some documentation in a separate PR.

PR #598, opened 2017-01-19
Fix/share extension failed verification

marcoconti83:
Adding security takeover to share extension

# Changes
- Moved the posting logic in `PostContent`
- Use more reliable observation
- Use new datamodel API


Wire-iOS Share Extension/ShareViewController.swift
```diff
@@ -208,264 +220,54 @@ class ShareViewController: SLComposeServiceViewController {
...
-        self.loadItem(forTypeIdentifier: kUTTypeData as String, options: [:], dataCompletionHandler: { (data, error) in
-            guard let data = data, error != nil else {
-                completion(nil)
-                return
-            }
-            completion(data)
-        })
-    }
+private func titleForMissingClients(users: Set<ZMUser>) -> String {
+    let template = users.count > 1 ? "meta.degraded.degradation_reason_message.plural" : "meta.degraded.degradation_reason_message.singular"
```
vytis: This won't work in languages that have more than two variations for numbers (lithuanian, polish etc.) - moving it to `.stringsdict` is the only way to properly translate the message.

marcoconti83: Good point, I'm not a translation expert, I'll let someone else answer this, for the moment we keep the same behavior as the main app (I just copied from there).

marcoconti83: Do we have the same issue with Russian or Ukrainian?

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,310 @@
...
+
+
+/// Content that is shared on a share extension post attempt
+class PostContent {
+    
+    var conversationObserverToken : Any?
+    
+    /// Conversation to post to
+    var target : Conversation? = nil {
+        didSet {
```
daehn: Can be removed

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,310 @@
...
+    
+    /// Process data to the right format to be sent
+    private func prepareForSending(data: Data, UTIString UTI: String, name: String?, completionHandler: @escaping (URL?, Error?) -> Void) {
+        let fileExtension = UTTypeCopyPreferredTagWithClass(UTI as CFString, kUTTagClassFilenameExtension as CFString)?.takeRetainedValue() as! String
+        
+        let fileName = name ?? "\(UUID().uuidString).\(fileExtension)"
+        
+        let tempDirectory = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true).appendingPathComponent(UUID().uuidString) // temp subdir
+        if !FileManager.default.fileExists(atPath: tempDirectory.absoluteString) {
+            try! FileManager.default.createDirectory(at: tempDirectory, withIntermediateDirectories: true)
```
vytis: Force unwrap might be a bit dangerous here

Wire-iOS Share Extension/PostContent.swift
marcoconti83: Done

Wire-iOS Share Extension/ShareViewController.swift
typfel: Right we should add plural cases to the `.stringdict` file instead.

marcoconti83: My question is, I copied this from the main app, and the main app already supports Ukrainian localization. What's the strategy there? Maybe we accept that there will be some grammar errors because we did not invest time on handling all cases. @mikeger ? @jenskasemets ?

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,306 @@
...
+            try data.write(to: tempFileURL)
+        } catch {
+            completionHandler(nil, error)
+            return
+        }
+        
+        
+        if UTTypeConformsTo(UTI as CFString, kUTTypeMovie) {
+            AVAsset.wr_convertVideo(at: tempFileURL) { [weak self] (url, _, error) in
+                // Video conversaion can take a while, we need to ensure the user did not cancel
```
vytis: typo `conversaion` 👨‍🎓

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,306 @@
...
+        let queue = DispatchQueue(label: "share extension URLs queue")
+        
+        self.attachments.forEach { attachment in
+            if attachment.hasItemConformingToTypeIdentifier(kUTTypeURL as String) {
+                group.enter()
+                attachment.fetchURL { url in
+                    defer { group.leave() }
+                    guard let url = url else { return }
+                    queue.async {
+                        urls.append(url)
```
vytis: Won't it `group.leave()` when this is queued, not when an URL is actually appended to the array?

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,306 @@
...
+                    }
+                }
+            }
+        }
+        group.notify(queue: queue) { _ in callback(urls) }
```
vytis: Notifies on the background queue, but the comment above says it should happen on main queue

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,306 @@
...
+    }
+}
+
+extension NSItemProvider {
+    
+    /// Extracts the URL from the item provider
+    func fetchURL(completion: @escaping (URL?)->()) {
+        self.loadItem(forTypeIdentifier: kUTTypeURL as String, options: nil, urlCompletionHandler: { (url, error) in
+            guard let url = url, error == nil else {
+                completion(nil)
```
daehn: Actually we could just completion(url) here, as we do not care about the error and don't do anything else

Wire-iOS Share Extension/PostContent.swift
marcoconti83: In which cases would it fail? 
- Out of space? (then nothing will work and the app will crash anyway)
- Invalid directory name? We asked the system to give us a name, if they gave us an invalid one, then iOS bug I guess...

do you have another case in mind?

Wire-iOS Share Extension/PostContent.swift
```diff
@@ -0,0 +1,306 @@
...
+            }
+            completion(url)
+        })
+    }
+    
+    /// Extracts data from the item provider
+    func fetchData(completion: @escaping(Data?)->()) {
+        self.loadItem(forTypeIdentifier: kUTTypeData as String, options: [:], dataCompletionHandler: { (data, error) in
+            guard let data = data, error != nil else {
+                completion(nil)
```
daehn: Same applies here with `data`

Wire-iOS Share Extension/ShareViewController.swift
jenskasemets: Let's keep this solution for now and create separate issue for this l10n issue, as it's affects also other similar strings.

Wire-iOS Share Extension/PostContent.swift
marcoconti83: 👍

Wire-iOS Share Extension/PostContent.swift
marcoconti83: Yes, good point!

Wire-iOS Share Extension/PostContent.swift
vytis: Yeah can't think of any other plausible cases this could fail

Wire-iOS Share Extension/PostContent.swift
marcoconti83: True

Wire-iOS Share Extension/ShareViewController.swift
```diff
@@ -208,264 +220,54 @@ class ShareViewController: SLComposeServiceViewController {
         pushConfigurationViewController(notSignedInViewController)
     }
     
+    /// Unregister conversation observer
+    private func removeConversationObserver() {
```
typfel: Can't the conversation observer token be nil out  in `deinit` like the other observers. And to follow up on that why do we need to nil out them all all in the `deinit`.

Wire-iOS Share Extension/PostContent.swift
marcoconti83: Then I think crashing is a good strategy. In case of iOS bug, at least we will know straight away and react to it (we will see it in Hockey). In case of no disk space, the app is doomed anyway :)

Wire-iOS Share Extension/ShareViewController.swift
vytis: @jenskasemets when looking through CrowdIn I noticed few other issues even when string is in `.stringsdict`, but not all of it is in the pluralizeable (meta: not sure if this is even a word) part of the translation. So yeah maybe would be worth cleaning that up at some point.

Wire-iOS Share Extension/ShareViewController.swift
marcoconti83: Yep

PR #600, opened 2017-01-20
Add debug button in internal build to delete clients

marcoconti83:
This will make conversation degradation testing faster, making QA life easier

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
@@ -43,7 +43,7 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
     var resetButton: ButtonWithLargerHitArea!
     
     /// Used for debugging purposes, disabled in public builds
-    var deleteDeviceButton: ButtonWithLargerHitArea!
+    var deleteDeviceButton: ButtonWithLargerHitArea?
     
     var fromConversation: Bool = false
     
@@ -257,7 +257,6 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
     
     func createVerifiedToggle() {
         let verifiedToggle = UISwitch()
-        verifiedToggle.translatesAutoresizingMaskIntoConstraints = false
         verifiedToggle.isOn = self.userClient.verified
         verifiedToggle.addTarget(self, action: #selector(ProfileClientViewController.onTrustChanged(_:)), for: .valueChanged)
         self.contentView.addSubview(verifiedToggle)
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
```diff
@@ -278,6 +282,20 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
         self.resetButton = resetButton
     }
     
+    func createDeleteButton() {
+        let deleteButton = ButtonWithLargerHitArea()
+        deleteButton.translatesAutoresizingMaskIntoConstraints = false
+        deleteButton.setTitle("DELETE (⚠️ this will cause decryption errors ⚠️)", for: UIControlState())
+        self.contentView.addSubview(deleteButton)
+        self.deleteDeviceButton = deleteButton
+        if !DeveloperMenuState.developerMenuEnabled() {
```
mikeger: I think it would be easier to read if the logic would be inverted, i.e.
```
if something { // not if !something
...
}
else {
...
}
```

Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
```diff
@@ -278,6 +282,20 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
         self.resetButton = resetButton
     }
     
+    func createDeleteButton() {
+        let deleteButton = ButtonWithLargerHitArea()
+        deleteButton.translatesAutoresizingMaskIntoConstraints = false
```
mikeger: Not necessary with `Cartography`, it sets the flag for the views that are passed into `constrain`.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
```diff
@@ -363,10 +387,23 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
     }
     
     func onResetTapped(_ sender: AnyObject) {
-        self.userClient.resetSession()
+        ZMUserSession.shared()?.performChanges {
```
mikeger: Good catch!

Wire-iOS/Sources/UserInterface/UserProfile/ProfileClientViewController.swift
```diff
@@ -42,6 +42,9 @@ class ProfileClientViewController: UIViewController, UserClientObserver, UITextV
     var verifiedToggleLabel: UILabel!
     var resetButton: ButtonWithLargerHitArea!
     
+    /// Used for debugging purposes, disabled in public builds
+    var deleteDeviceButton: ButtonWithLargerHitArea!
```
mikeger: I would make it optional and do not create it when it is not necessary.

marcoconti83:
@mikeger please re-review

PR #614, opened 2017-01-23
Use automatic Localytics session tracking

daehn:
# What's in this PR?

* Use the automatic session tracking that Localytics provides instead of manually forwarding the calls.

#### This PR depends on:
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/218 has been merged. ~~More testing is needed before this can be merged.~~


marcoconti83:
> More testing is needed before this can be merged.

What kind of testing? Unit/integration or manual testing of the app with Localytics events?

daehn:
@marcoconti83 I was thinking manual, I don't think there is a good way to test the automated session tracking.

daehn:
@marcoconti83 I think there might be Localytics sandbox though that we can use to check correct session tracking.

PR #617, opened 2017-01-24
Update calling protocol selection in developer options

typfel:


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -240,6 +240,17 @@ class SettingsPropertyFactory {
                         throw SettingsPropertyError.WrongValue("Incorrect type \(value) for key \(propertyName)")
                     }
             })
+            
+        case .callingProtocol:
```
mikeger: Potentially should work automatically via `userDefaultsPropertiesToKeys`, did you tried it that way?

typfel: Yes, didn't work since I'm storing integers and not an object.

daehn: It's additionally assigned to the user session as well

PR #623, opened 2017-01-25
Add support to copy fingerprint label in device detail view

daehn:
# What's in this PR?

* Add a `UILabel` subclass called `CopyableLabel` which can be selected and copied.
* Clean up the `ProfileClientViewController`, most of the properties were implicitly unwrapped optionals which was not necessary. also updated the access protection levels of most of the methods which do not need to be internal.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/CopyableLabel.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/CopyableLabel.swift
@@ -1,6 +1,6 @@
 //
 // Wire
-// Copyright (C) 2016 Wire Swiss GmbH
+// Copyright (C) 2017 Wire Swiss GmbH
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


daehn:
@typfel fixed the header

Wire-iOS/Sources/UserInterface/UserProfile/CopyableLabel.swift
```diff
@@ -0,0 +1,84 @@
...
+import Foundation
+
+
+/// This class is a drop-in replacement for UILabel which can be copyied
+class CopyableLabel: UILabel {
+
+    private let dimmedAlpha: CGFloat = 0.4
+    private let dimmAnimationDuration: TimeInterval = 0.33
+
+    init() {
```
typfel: Think you remove this init. The `frame` or `coder` will always be called.

daehn: I am calling this one in the `ProfileClientViewController`, but I guess I can use `coder` instead

daehn: @typfel fixed.

PR #627, opened 2017-01-26
Fixed bug with dismissing settings as the result of color selection

mikeger:
# Issue

When user selects accent color on iPad, the whole settings controller is dismissed.

# Investigation

Recently in SplitViewController was added the code to dismiss the modal controllers over the right controller when the right controller is changed. For some reason, when anything is presented over the app the right controller also says that it is presented over it. When the accent color changes we reload the conversation controller, which triggers the dismissal of the controller that is presented over the right controller. iOS dismisses the settings, so the user is getting confused.

# Solution

I changed the conversation controller so it would keep track on it's own about the presented collection controller and dismiss it if it's being deallocated.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
@@ -435,7 +435,11 @@ NSString *SplitLayoutObservableDidChangeToLayoutSizeNotification = @"SplitLayout
         return;
     }
     
-    [self.rightViewController dismissViewControllerAnimated:NO completion:nil];
+    // To determine if self.rightViewController.presentedViewController is actually presented over it, or is it
+    // presented over one of it's parents.
+    if (self.rightViewController.presentedViewController.presentingViewController == self.rightViewController) {
+        [self.rightViewController dismissViewControllerAnimated:NO completion:nil];
+    }
     
     UIViewController *removedViewController = self.rightViewController;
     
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -435,7 +435,11 @@ - (void)setRightViewController:(nullable UIViewController *)rightViewController
         return;
     }
     
-    [self.rightViewController dismissViewControllerAnimated:NO completion:nil];
+    // To determine if self.rightViewController.presentedViewController is actually presented over it, or is it
+    // presented over one of it's parents.
+    if (self.rightViewController.presentedViewController.presentingViewController == self.rightViewController) {
```
typfel: 😕

typfel:
If the user is in the settings modifying the accent color there shouldn't be any conversation view controller loaded AFAIK?

mikeger:
@typfel on iPad there is one, since the settings are not fullscreen

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -167,30 +167,34 @@ public extension ConversationViewController {
...
         })
         
-        collections.onDismiss = {[weak self] _ in
-            guard let `self` = self else {
-                return
-            }
-            
-            self.parent?.dismiss(animated: true, completion: { 
+        collections.onDismiss = { _ in
+            ZClientViewController.shared().dismiss(animated: true, completion: {
```
typfel: could also be `self.collectionController`?

mikeger: Sure, missed it, thanks!

mikeger:
@typfel updated

PR #629, opened 2017-01-26
Add option to fetch link previews in share extension

daehn:
# What's in this PR?

* Add a new class called `ExtensionSettings`, which is similar to `Settings` but uses the shared `UserDefaults`instead.
* Add a feature flag to the developer menu to opt-out of fetching link previews (default is enabled at the moment).
* Pass the value of this flag to the conversation when appending a new text message.
* Update for use with the new `wire-ios-share-engine` API.

#### This PR depends on:
- [x] A new release of `wire-ios-share-engine` after https://github.com/wireapp/wire-ios-share-engine/pull/17 has been merged.


WireExtensionComponents/Utilities/ExtensionSettings.swift
```diff
@@ -0,0 +1,83 @@
...
+        defaults?.register(defaults: ExtensionSettingsKey.defaultValueDictionary)
+    }
+
+    public func reset() {
+        ExtensionSettingsKey.all.forEach {
+            defaults?.removeObject(forKey: $0.rawValue)
+        }
+
+        // As we purposely crash afterwards we manually call synchronize.
+        defaults?.synchronize()
```
marcoconti83: We purposely crash? I assume this is only when the developer menu is used? :)

daehn: Yes, this option is only available when the developer menu is enabled. 
Inside `ZMUserSession/resetStateAndExit()` we can `exit(0)`.

daehn: And the `reset` call is only performed at that point.

PR #639, opened 2017-01-30
instead of resorting, recreate conversationlists

basine:
Instead of resorting the list when re-entering the foreground, it should be recreated. Otherwise conversations that were inserted or deleted while the app was in the background may not be included in the list.

Note: This need a new ZMCDataModel release with https://github.com/wireapp/wire-ios-data-model/pull/151
and then retested before merging


marcoconti83:
@basine Data model has been merged, we can go ahead or do we need the "additional testing"?

daehn:
@mikeger this one still had a `DO NOT MERGE`

mikeger:
@daehn  I asked Sabine, she told that it can be merged

PR #642, opened 2017-01-30
Right to left: keeping the same conversation list toolbar buttons order

vytis:
Even with RTL layout we have the conversations list opening on the same side, so it makes sense to keep the toolbar buttons in the same order as well.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
@@ -423,12 +423,20 @@
 
 - (void)updateRightNavigationItemsButtons
 {
-    self.navigationItem.rightBarButtonItems = [self rightNavigationItemsForConversation:self.conversation];
+    if ([UIApplication isLeftToRightLayout]) {
+        self.navigationItem.rightBarButtonItems = [self rightNavigationItemsForConversation:self.conversation];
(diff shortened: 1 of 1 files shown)


typfel:
Can't we just mark the toolbar as being excluded from RTL?

vytis:
There is a way to do this, but it's iOS9+ only (`UIView semanticContentAttribute`).

vytis:
Apparently `//FIXME: bla bla` adds a nice icon in methods list
screen shot 2017-01-30 at 17 25 20

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -60,7 +60,8 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
-      allowLocationSimulation = "YES">
+      allowLocationSimulation = "YES"
+      language = "IDELaunchSchemeLanguageRightToLeftLayoutDirection">
```
typfel: Should be reverted I think.

vytis: Oops

PR #643, opened 2017-01-31
Change account email

vytis:
## What's in this PR?

Added possibility to change account email from settings. It adds a new screen where you can edit the email and then another one which is shown until user verifies the email.


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -73,20 +73,32 @@ extension SettingsCellDescriptorFactory {
...
+            isDestructive: false,
+            presentationStyle: .navigation,
+            presentationAction: { () -> (UIViewController?) in
+             if let email = ZMUser.selfUser().emailAddress, !email.isEmpty {
+                return ChangeEmailViewController()
+             } else {
+                let addEmailController = AddEmailPasswordViewController()
+                addEmailController.showsNavigationBar = false
+                let stepDelegate = DismissStepDelegate()
+                stepDelegate.strongCapture = stepDelegate
```
typfel: It has strong reference to itself. Won't this create a leak?

vytis: It was in the original code and is used throughout settings screen. Looking at the code it seems that the `.strongCapture` is set to `nil` at some point later, so it shouldn't leak, but I didn't check extensively.

Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+
+final class ChangeEmailViewController: SettingsBaseTableViewController {
+
+    fileprivate weak var userProfile = ZMUserSession.shared()?.userProfile
+    var state = ChangeEmailState()
+    private var observerToken: AnyObject?
+
+    init() {
+        super.init(style: .grouped)
+        CASStyler.default().styleItem(self)
```
mikeger: This seems unnecessary since at this point the views are not yet set up, and also there seem to be no direct reference to `ChangeEmailViewController` in classy. There are references to cells, but they are also not yet loaded at this point.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
typfel: OK

Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
vytis: Makes sense, will remove

Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        navigationItem.rightBarButtonItem = UIBarButtonItem(
+            title: "self.settings.account_section.email.change.save".localized,
+            style: .done,
+            target: self,
+            action: #selector(saveButtonTapped)
+        )
+        toggleSaveButton()
+    }
+    
+    func toggleSaveButton(enabled: Bool? = nil) {
```
mikeger: I understand something named `toggle` to change the state to the opposite, I think in this case the method actually updates the button state. If it is correct I guess it can be renamed to provide the better understanding of what method does.

Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+        
+        alert.addAction(.init(title: "general.ok".localized, style: .cancel, handler: nil))
+        present(alert, animated: true, completion: nil)
+    }
+}
+
+extension ChangeEmailViewController: ConfirmEmailDelegate {
+    func didConfirmEmail(inController controller: ConfirmEmailViewController) {
+        if let viewControllers = navigationController?.viewControllers, viewControllers.count > 2 {
+            let accountController = viewControllers[1]
```
mikeger: I would suggest to enumerate controllers to find one of the expected type, otherwise if we add some intermediate step the logic would break.

Wire-iOS/Sources/UserInterface/Settings/ConfirmEmailViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    fileprivate weak var userProfile = ZMUserSession.shared()?.userProfile
+    weak var delegate: ConfirmEmailDelegate?
+    let newEmail: String
+    fileprivate var observer: UserCollectionObserverToken?
+
+    init(newEmail: String, delegate: ConfirmEmailDelegate?) {
+        self.newEmail = newEmail
+        self.delegate = delegate
+        super.init(style: .grouped)
+        CASStyler.default().styleItem(self)
```
mikeger: Same as with other guy, seems not necessary to style.

Wire-iOS/Sources/UserInterface/Settings/ConfirmEmailViewController.swift
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    internal func setupViews() {
+        ShortLabelTableViewCell.register(in: tableView)
+        
+        title = "self.settings.account_section.email.change.verify.title".localized
+        view.backgroundColor = .clear
+        tableView.isScrollEnabled = false
+        
+        tableView.sectionHeaderHeight = UITableViewAutomaticDimension;
```
mikeger: Semicolon!! ⚠️

mikeger: :hurtrealbad:

vytis: Ooops, came back from Obj-C and forgot to wash my hands!

Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
vytis: Not sure if it would help much in this case - both the main settings controller and account details are of the same class (`SettingsTableViewController`). Maybe worth re-doing it so we would pop to one before the current.

mikeger: May be writing a note in comment would be enough.

typfel: Couldn't you store the pushed controller in a instance variable and compare that it's the expected instance here?

typfel: Maybe store the a reference to the account controller when `ChangeEmailViewController` is initially pushed.

vytis: I will just inspect the `viewControllers` array and pick the previous one. Settings are constructed by these cell descriptors and pushing controllers happens in quite generic part of code, so would need to refactor quite a bit to get that reference from outside.

vytis: @typfel tweaked the code a bit

PR #645, opened 2017-01-31
Remove config for the old static AVS library

typfel:


vytis:
So now AVS is managed through Carthage?

typfel:
Yes partially. Since we need need use a different framework when submitting to the appstore we still have script which downloads the framework and overwrites the carthage framework.

PR #647, opened 2017-02-01
Added accessibility identifier to change email text field

vytis:
To make life easier when testing


Wire-iOS/Sources/UserInterface/Settings/ChangeEmailViewController.swift
```diff
@@ -42,6 +42,7 @@ final class ChangeEmailTableViewCell: UITableViewCell {
 
     func setupViews() {
         addSubview(emailTextField)
+        emailTextField.accessibilityIdentifier = "EmailField";
```
mikeger: ;;;;;;;;;;;;;;;;;;-)

PR #648, opened 2017-02-01
Fixes for data detectors

mikeger:
1. Fixed the behavior on iOS 9 when certain kinds of detected elements where not possible to open.

The issue happened when the special iOS internal kind of URL was provided for address and other special types and we tried to open it directly.

2. The issue with iOS integrated menu with "Add to reading list" and other default elements where displayed for long press menu on the URL. Now we display the default menu instead.


vytis:
Is 3d touch used for data detectors in other apps? Or it does work in place of long tap without any extra work?

mikeger:
@vytis works automagically if data detectors are enabled. iOS at it's best.

vytis:
Nice!

PR #649, opened 2017-02-01
Email changing: fixing several small issues

vytis:
## What is fixed in this PR?

* Showing proper error message when something goes wrong
* Providing feedback when resending confirmation email
* When coming back to email edit screen after e.g. mistyping email, your original email was shown, not the edited one.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -498,8 +498,8 @@
 "self.settings.account_section.email.change.verify.description" = "Check your email inbox and follow the instructions";
 "self.settings.account_section.email.change.verify.resend" = "Resend to %@";
 
-"self.settings.account_section.email.change.failure_alert.title" = "Unable to change email";
```
mikeger: This one was not used before?

mikeger: Ah I see now

PR #650, opened 2017-02-01
Support dynamic type

mikeger:
So your granny can use Wire.


mikeger:
Please do not merge until 2.29 is out. I don't want to push more stuff into the current release.

Be-ing:
Thanks for taking care of this!

PR #655, opened 2017-02-02
Allow locking the app with passcode / touch id

mikeger:
It comes out that since iOS 9 it is possible to lock the app with the system passcode if there is no touch ID.

The app checks if the authentication is possible before showing the configuration property.

Setting is stored in the keychain.


Wire-iOS/Sources/AppController+LocalAuth.swift
```diff
@@ -0,0 +1,73 @@
...
+                                                              selfUser: ZMUser.selfUser(),
+                                                              crashlogManager: BITHockeyManager.shared())
+        
+        let lockApp = settingsPropertyFactory.property(.lockApp)
+        
+        return lockApp.value() == SettingsPropertyValue.bool(value: true)
+    }
+    
+    /// @param callback confirmation; if auth is not needed called with 'true'
+    func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool)->()) {
```
vytis: Using `guard` would make reduce the indentations in this method by quite a bit I think

mikeger: Yeah, this one is "hadooken"

typfel: Could add an optional label to to the `Bool` in the callback

Wire-iOS/Sources/AppController+LocalAuth.swift
```diff
@@ -0,0 +1,73 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import LocalAuthentication
+import CocoaLumberjackSwift
+
+extension AppController {
+    var appLockActive: Bool {
+        let settingsPropertyFactory = SettingsPropertyFactory(userDefaults: UserDefaults.standard,
```
typfel: Feels a bit heavy to have instantiate `SettingsPropertyFactory` just to check a boolean in the keychain.

Wire-iOS/Sources/AppController+LocalAuth.swift
mikeger: Tried, but the method is used from ObjC, therefore cannot use optional C type.

Wire-iOS/Sources/AppController+LocalAuth.swift
mikeger: I would make it `shared` property. The `SettingsPropertyFactory` was designed to have no implicit dependencies or shared state, so it's creation is kind of heavy as a consequence.

Wire-iOS/Sources/AppController+LocalAuth.swift
```diff
@@ -21,53 +21,49 @@ import LocalAuthentication
...
         
-        let lockApp = settingsPropertyFactory.property(.lockApp)
+        return settingsPropertyFactory
+    }
+    
+    var appLockActive: Bool {
+        let lockApp = type(of: self).settingsPropertyFactory.property(.lockApp)
         
-        return lockApp.value() == SettingsPropertyValue.bool(value: true)
+        return lockApp.value() == SettingsPropertyValue(true)
```
typfel: Was thinking we could access the keychain directly without involving `SettingsPropertyValue`

PR #658, opened 2017-02-03
Track call setup time

typfel:
Adds tracking for call setup time.

- [x] depends on https://github.com/wireapp/wire-ios-sync-engine/pull/246


Wire-iOS/Sources/Analytics/Clusterizers/TimeIntervalClusterizer.m
```diff
@@ -29,6 +29,9 @@
 IntRange MessageEditTimeIntervalRangesRanges[] = {{0,0}, {1, 60}, {61, 300}, {601, 1800}, {1801, 3600}, {3601, 86400}};
 RangeSet MessageEditTimeIntervalRanges = {MessageEditTimeIntervalRangesRanges, 6};
 
+IntRange CallSetupTimeIntervalRangesRanges[] = {{0,0}, {1, 1}, {2, 2}, {3, 3}, {4, 4}, {5, 10}, {11, 30}};
+RangeSet CallSetupTimeIntervalRanges = {CallSetupTimeIntervalRangesRanges, 7};
+
```
daehn: We actually have a `Swift` version of the clusterization logic (https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Source/Analytics/Clusterizer.swift) and should move all of the clusterizers there at some point (maybe it makes sense to do this when extracting all analytics into a separate framework).

PR #659, opened 2017-02-03
Share engine tracking

daehn:
# What's in this PR?

* Add tracking to share extension, as discussed in [#7884](https://wearezeta.atlassian.net/browse/ZIOS-7884) this is now done by storing tracking events in the shared container and tracking them in the main app. 
* As fetching an attachments url is asynchronous, getting the correct tracking data is as well. This is the main reason why we only track the attached types in case the user actual sends them. I tried to track them when cancelling as well (in the overridden `cancel` method in the `ShareViewController`).
But as this method has to be sync I needed to block the main thread until the fetch is done, which resulted in a deadlock as the `fetchURL` method was executed on the main thread for some reason.
* For now there are three events, `share_extension_opened ` will be stored every time the share extension is opened. The events `share_extension_sent` and `share_extension_cancelled` will be stored when the user chooses to send or to cancel respectively.

Example attributes for a `share_extension_sent` event when sending one video, two images and text:
```json
{
    "verified_conversation": false,
    "number_of_images": 2,
    "video": true,
    "file": 0,
    "text": true,
    "conversation_did_degrade": false
}
  ```
### 🚫 This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/157 has been merged.
- [x] A new release of `wire-ios-share-engine` after https://github.com/wireapp/wire-ios-share-engine/pull/18 has been merged.
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/245 has been merged.

### CODE REVIEW

File: Wire-iOS Share Extension/ShareExtensionAnalytics.swift [hunk 1]
```diff
@@ -0,0 +1,145 @@
...
+    func populateAttachmentTypes(completion: @escaping () -> Void) {
+        hasVideo = attachments.contains { $0.hasVideo }
+        numberOfImages = attachments.filter { $0.hasImage }.count
+
+        let group = DispatchGroup()
+        attachments.forEach { _ in
+            group.enter()
+        }
+
+        attachments.forEach {
```
mikeger: Can be done in one `forEach`?

Wire-iOS/Sources/Analytics/Analytics+StorableTrackingEvent.swift
```diff
@@ -0,0 +1,29 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import ZMCDataModel
+
+
+extension Analytics {
+
+    @objc(tagStorableEvent:) public func tag(_ storableEvent: StorableTrackingEvent) {
```
mikeger: I would mark it to be used only from share extension, I guess nothing bad would happen if it would be used from the app, just would be weird.

Wire-iOS Share Extension/ShareExtensionAnalytics.swift
daehn: I wanted to make sure that no group can be left before all of them have been entered.

Wire-iOS/Sources/Analytics/Analytics+StorableTrackingEvent.swift
daehn: This is actually intended to be used form the main application, the share extension serializes the tracking events and the main app picks it up and performs the actual tracking calls.

mikeger:
💣  Conflict

PR #666, opened 2017-02-06
Fix missing views for incoming video call

typfel:
## What was broken
The video received state handling has been changed for calling 2 and 3. Previously the UI was listening to two different states `incoming video active` and `remote is sending video`. These two states are now combined into `ReceivedVideoStateStarted`. That broke the UI which still separates the two states.

## Fix
The solution to showing the correct view layout is to take the `call status` and the `video status` into account.


daehn:
@typfel what happened to the snapshot tests? They look like audio calls now.

typfel:
@daehn what part looks like an audio call?

PR #667, opened 2017-02-06
Remove .bool case from SettingsPropertyValue

daehn:
# What's in this PR?

* When retrieving a value from the UserDefaults we were losing the type information, specifically if the value we set represents a `Bool` or for example an `Int`, this lead to bugs (e.g. [#8005](https://wearezeta.atlassian.net/browse/ZIOS-8005)).
* We now store all numbers and booleans as `NSNumbers`, the `.bool` case has been removed from the `SettingsPropertyValue`.
* Convenience initializers have been added to the `SettingsPropertyValue` to create a new value using an `Int`, `Bool` or `UInt`, this avoids duplicated `NSNumber` wrapping on the caller side.
* A test to catch the bug storing an integer value higher than 1 has been added, before this change retrieving the value would result in a `true` / `1` being returned.
* Add `@discardableResult` to `open()` method on `URL` to remove warning.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
@@ -89,21 +89,33 @@ enum SettingsPropertyName: String, CustomStringConvertible {
 }
 
 enum SettingsPropertyValue: Equatable {
-    case number(value: Int)
+    case number(value: NSNumber)
     case string(value: Swift.String)
-    case bool(value: Swift.Bool)
     case none
-    
+
+    init(_ bool: Bool) {
+        self = .number(value: NSNumber(value: bool))
+    }
+
+    init(_ uint: UInt) {
+        self = .number(value: NSNumber(value: uint))
+    }
+
+    init(_ int: Int) {
+        self = .number(value: NSNumber(value: int))
+    }
+
+    init(_ int: Int16) {
+        self = .number(value: NSNumber(value: int))
+    }
+
     static func propertyValue(_ object: Any?) -> SettingsPropertyValue {
         switch(object) {
-        case let intValue as Int:
-            return SettingsPropertyValue.number(value: intValue)
+        case let number as NSNumber:
+            return SettingsPropertyValue.number(value: number)
             
         case let stringValue as Swift.String:
             return SettingsPropertyValue.string(value: stringValue)
-            
-        case let boolValue as Swift.Bool:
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
```diff
@@ -89,22 +89,34 @@ enum SettingsPropertyName: String, CustomStringConvertible {
...
 enum SettingsPropertyValue: Equatable {
-    case number(value: Int)
+    case number(value: NSNumber)
     case string(value: Swift.String)
-    case bool(value: Swift.Bool)
     case none
-    
+
+    init(_ bool: Bool) {
+        self = .number(value: NSNumber(value: bool))
```
mikeger: Can it be a generic initializer with protocol bound?

mikeger:
@daehn so this changes are fixing the issues, then should be merged to current, i.e. 2.29, release?

jenskasemets:
@mikeger yes, this should go in to 2.29

Wire-iOS/Sources/Components/Settings/SettingsProperty.swift
daehn: I think unfortunately not as the `NSNumber` initializers are overloads for all different classes.

daehn: Would be possible to check which class the value belongs to inside of the `init`, but passing a protocol to the `NSNumber` initializer won't work.

daehn: Basically like this:
```swift 
protocol NSNumberInitializee {}

extension Bool: NSNumberInitializee {}
extension UInt: NSNumberInitializee {}
extension Int: NSNumberInitializee {}


init(_ value: NSNumberInitializee) {
    if let uint = value as? UInt {
        self = .number(value: NSNumber(value: uint))
    } else if let int = value as? Int {
        self = .number(value: NSNumber(value: int))
    } else if let bool = value as? Bool {
        self = .number(value: NSNumber(value: bool))
    }
}
```

But this would basically be the same, with the difference that now the compiler does all the work instead of it being done at runtime.

mikeger: Nah, I thought may be there is a more beautiful way than if-else. I guess the current state is OK.

PR #672, opened 2017-02-07
Add crash reporting to share extension

daehn:
# What's in this PR?

**TL;DR: This PR adds crash reporting to the share extension, in order to do so the HockeySDK framework is no longer integrated using Carthage but instead added to the project manually (twice).**

* In order to get crash data from the share extension a different version of the HockeySDK (which is suitable for the use in an application extension context) had to be integrated. Until now we were building the HockeySDK manually and creating Carthage releases. As we now need to versions of the SDK, one for the main app with all features and the extension only SDK for the share extension, this approach is not feasible anymore.
* The two frameworks have been added in `Vendor` folders in the workspace.
* Some `Settings` configurations had to be migrated to the shared container (`ExtensionSettings`) in order for the share extension to pick up if the user opted out of sharing analytics and crash data.
* The `CrashReporter` is used to wrap all crash reporting related code inside the share extension.

**I am still evaluating options not having the need to commit the SDKs to the main repo, like for example downloading them via a script like we do it for AVS.**


mikeger:
@daehn can we have two separate repos of hockey SDK with binary versions there?

daehn:
I don't think we can do it with Carthage if they have the same name. But just two repos where we have the binaries that we then pull into the application with a custom script could work, yes.

mikeger:
@daehn I am not totally following, how we can have tow frameworks with the same name in the product? They all sit in `Bundle.app/Frameworks` folder, so one would overwrite another?

daehn:
@mikeger I moved the SDK to CocoaPods and use the `CrashOnlyExtension` in the share extension target now, makes this a lot easier. Somehow the project file got super messed up now though. I will recreate this PR without the intermediate changes. 🔥

daehn:
@mikeger the new PR is https://github.com/wireapp/wire-ios/pull/676

PR #674, opened 2017-02-07
Rudimentary search implementation

mikeger:
- Implemented to test the performance of "stupid" search on conversations.
- It is enabled only on developer builds.
- Aim is to test the performance and the simple approach on searching the conversation content.

# Disclaimer

There is no stemming or any other nice technique used. The search is happening on main thread.


Wire-iOS/Sources/UserInterface/Search/TextSearchViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+                self.results = []
+                return
+            }
+            
+            let predicate = NSPredicate(block: { (object, _) -> Bool in
+                guard let object = object as? ZMTextMessageData else {
+                    return false
+                }
+                
+                return object.messageText.localizedCaseInsensitiveContains(query)
```
daehn: `messageText` is an implicitly unwrapped optional, so this will crash if it is `nil` for some reason

mikeger:
@daehn updated, please review!

PR #676, opened 2017-02-07
Add HockeySDK to share extension

daehn:
# What's in this PR?

* CocoaPods version of https://github.com/wireapp/wire-ios/pull/672.
* Move `HockeySDK` from Carthage to CocoaPods so we can specify the correct sub specs that we need in the share extension.
* Some `Settings` configurations had to be migrated to the shared container (`ExtensionSettings`) in order for the share extension to pick up if the user opted out of sharing analytics and crash data.
* The `CrashReporter` is used to wrap all crash reporting related code inside the share extension.


ConfigurationSets/Wire-Share-Extension.release.xcconfig
```diff
@@ -18,3 +18,4 @@
 
 
 #include "Wire-Share-Extension.base.xcconfig"
+#include "Pods/Target Support Files/Pods-Wire Share Extension/Pods-Wire Share Extension.release.xcconfig"
```
mikeger: I hope the cocoapods guys are not going to change the path in the next version (suspicious)

Podfile
```diff
@@ -1,6 +1,7 @@
 source 'https://github.com/CocoaPods/Specs.git'
 
 def ios_pods
+    pod "HockeySDK"
```
mikeger: Should we clone it as other cocoapods we have? It makes it more stable in case the foreign repo gets corrupted / removed / tag deleted etc

daehn: Good Idea!

Wire-iOS Share Extension/CrashReporter.swift
```diff
@@ -0,0 +1,55 @@
...
+        let manager = BITHockeyManager.shared()
+        manager.isCrashManagerDisabled = ExtensionSettings.shared.disableCrashAndAnalyticsSharing // Whether the user opted out of sharing analytics and crash data
+        manager.configure(withIdentifier: hockeyIdentifier)
+        manager.crashManager.crashManagerStatus = .autoSend
+        manager.start()
+    }
+
+    private static var hockeyEnabled: Bool {
+        let configUseHockey = useHockey() // The preprocessor macro USE_HOCKEY (from the .xcconfig files)
+        let automationUseHockey = AutomationHelper.sharedHelper.useHockey // Command line argument used by automation
```
mikeger: Can share extension have the command line arguments?

daehn: You can at least provide them as environment arguments in the Scheme → Run Settings which can be useful.

Wire-iOS Share Extension/PreprocessorHelper.m
```diff
@@ -0,0 +1,38 @@
...
+
+#define STRINGIZE_INTERNAL(x) #x
+#define STRINGIZE(x) STRINGIZE_INTERNAL(x)
+
+
+NSString * hockeyAppId() {
+    return @STRINGIZE(HOCKEY_APP_ID_KEY);
+}
+
+BOOL useHockey() {
```
mikeger: I would suggest to namespace the functions, like WR_useHockey, since the C functions namespace is pretty overused

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -5378,6 +5418,21 @@
...
+			files = (
+			);
+			inputPaths = (
+			);
+			name = "[CP] Check Pods Manifest.lock";
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "diff \"${PODS_ROOT}/../Podfile.lock\" \"${PODS_ROOT}/Manifest.lock\" > /dev/null\nif [[ $? != 0 ]] ; then\n    cat << EOM\nerror: The sandbox is not in sync with the Podfile.lock. Run 'pod install' or update your CocoaPods installation.\nEOM\n    exit 1\nfi\n";
```
mikeger: Goddamnit, cocoapods, why not create a shell script for it? (Rhetoric question)

Podfile
daehn: @mikeger Seems to me that the releases are not copied over in a fork, which means we would have to do it manually.

mikeger: @daehn yeah the releases are not copied, but I think cocoapods does not download the binary releases anyways?

daehn: Looks like it does, see https://github.com/bitstadium/HockeySDK-iOS/blob/develop/HockeySDK.podspec#L40

daehn: bildschirmfoto 2017-02-07 um 16 18 26

mikeger: @daehn 🙌

mikeger: Ok, I guess let's keep it with original repo then (until it breaks once)

mikeger:
💣 you know

PR #678, opened 2017-02-08
Improvements for app locking

mikeger:
- Setting is moved to designed location
- Added contents dimming
- Added the grace period of 10 seconds until the next auth is needed (so if you briefly go out of the app the auth is not needed)
- Replaced full dim with blur


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -280,6 +280,36 @@ class SettingsPropertyFactory {
                     default: throw SettingsPropertyError.WrongValue("Incorrect type \(value) for key \(propertyName)")
                     }
             })
+        case .lockAppLastDate:
+            return SettingsBlockProperty(
+                propertyName: .lockAppLastDate,
+                getAction: { _ in
+                    guard let data = ZMKeychain.data(forAccount: SettingsPropertyName.lockAppLastDate.rawValue),
```
typfel: Is really necessary to store the lockAppLastDate in the keychain? Couldn't we store it in memory. 

Isn't one issue with the lastDate implementation is that you could circumvent it by changing the date on the phone to the past. Persisting it makes this even easier.

mikeger: @typfel I can change it in the way that it's checking the time distance to last saved date, I guess then changing clock would be way too precise to be done.

Keeping it in memory would be ok for short grace period, for longer one it is possible that app would be killed, not sure if I also want to remove the keychain code now as it works :)

typfel: hehe :) I'd say it also a security decision about what should happen if you manually kill the app.

Wire-iOS/Sources/AppController+LocalAuth.swift
```diff
@@ -39,13 +39,35 @@ extension AppController {
...
     func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool)->()) {
         guard #available(iOS 9.0, *), self.appLockActive else {
             callback(true)
             return
         }
         
+        let lastAuthDate = self.lastUnlockedDate
+        
+        // The app was authenticated at least N seconds ago
+        if (Date().timeIntervalSinceReferenceDate - lastAuthDate.timeIntervalSinceReferenceDate) < type(of: self).authenticationPersistancePeriod {
```
typfel: This can be simplified to:
`if lastAuthData.timeIntervalSinceNow < type(of: self).authenticationPersistancePeriod`

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
typfel: `off topic:` why is it important to store `lockAppLastDate` and `lockApp.` in the keychain? A boolean and date is not sensitive information, or?

mikeger: Well, you can put it to user defaults, but they can be modified from the computer, before it was possible to change them with helper apps.

mikeger:
@typfel you might want to re-review this one, there are a lot of changes:
1. If user cancels the auth we show the special screen where you can restart the auth
2. Dimming is not showing the app content

Wire-iOS/Sources/AppController.m
```diff
@@ -67,8 +67,8 @@ @interface AppController ()
 
 @property (nonatomic) ClassyCache *classyCache;
 
-@property (nonatomic) BOOL localAuthenticationSucceeded;
-
+@property (nonatomic) BOOL localAuthenticationCancelled;
+@property (nonatomic) BOOL localAuthenticationNeeded;
```
typfel: Isn't the  `lastUnlockedDate` making `localAuthenticationNeeded` superfluous?

Wire-iOS/Sources/AppController.m
```diff
@@ -142,9 +144,20 @@ - (void)applicationWillEnterForeground:(UIApplication *)application
...
+}
+
 - (void)applicationDidEnterBackground:(UIApplication *)application
 {
-    self.localAuthenticationSucceeded = NO;
+    self.localAuthenticationNeeded = YES;
+    if ([self appLockActive]) {
+        self.notificationWindowController.dimContents = YES;
+    }
+    self.lastUnlockedDate = [[NSDate alloc] init];
```
typfel: Shouldn't there be a check here to only update `lastUnlockedDate` if the user unlocked the app.

Wire-iOS/Sources/AppController.m
mikeger: No, you need to know when you are going to the background if the app is unlocked now

Wire-iOS/Sources/AppController.m
mikeger: @typfel sorry forgot to push

mikeger:
@typfel good point, I would create the view controller for logic, right now it's kind of spread.

mikeger:
@typfel updated as suggested. Please check.

PR #681, opened 2017-02-09
Display unknown message warning for messages types we don't understand

typfel:
### Background
When we introduce new messages types in future versions of Wire, we want old clients to display a warning to the user when it receives a a message type it doesn't understand.

needs:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/258


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/UnknownMessageCell.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import Foundation
+import Cartography
+import TTTAttributedLabel
+import Classy
+
+public final class UnknownMessageCell : ConversationCell {
+    
+    private let supportURL = URL(string: "http://support.wire.com/error/unknown-message") // TODO update final link
```
mikeger: We have a directory of links in `NSURL+WireURLs.m`

jenskasemets: Final link is https://wire.com/compatibility/unknown-message

marcoconti83:
SE merged but not released

typfel:
@mikeger changed a few things. Please re-check.

PR #686, opened 2017-02-11
Add three new emoticon shortcuts;

mat1th:
I've added three new emoticon short cuts. The 🎉 , 👍 and 👎; These are often used so it would be nice to auto correct them.


mat1th:
I see that the Github name I was using was not my github name. So I will close this pull request and create a new one.

PR #688, opened 2017-02-13
Disable app lock if passcode is removed

mikeger:
- Fixed layout on iPad
- Dismiss keyboard if needed

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
@@ -107,13 +107,7 @@ import HockeySDK.BITHockeyManager
                 self.lockView.showReauth = false
                 self.requireLocalAuthenticationIfNeeded { grantedOptional in
                     
-                    let granted: Bool
-                    if let grantedUnwrap = grantedOptional {
-                        granted = grantedUnwrap
-                    }
-                    else {
-                        granted = true
-                    }
+                    let granted = grantedOptional ?? true
                     
                     self.dimContents = !granted
                     self.localAuthenticationCancelled = !granted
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -109,9 +128,9 @@ import HockeySDK.BITHockeyManager
     }
 
     /// @param callback confirmation; if auth is not needed called with 'true'
-    func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool)->()) {
+    func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool?)->()) {
```
typfel: What's the reasoning behind making this optional?

mikeger: If the auth is not needed or device does not support it, it supposed to return .none (also I need to update the comment)

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -170,5 +182,6 @@ extension AppLockViewController: UIApplicationDelegate {
     
     func applicationDidBecomeActive(_ application: UIApplication) {
         self.showUnlockIfNeeded()
+        self.resignKeyboardIfNeeded()
```
typfel: Shouldn't we dismiss the keyboard when going to the background as well? Isn't the keyboard briefly visible when only dismissing it in `applicationDidBecomeActive`?

mikeger: I updated to dismiss the keyboard before going offscreen too

PR #694, opened 2017-02-14
Add command line argument for logging tags

basine:
We would like to log tags that we log through ZMLog via command line. This should enable us to add a comma separated list of tags to log.

### DIFF
--- a/WireExtensionComponents/Utilities/AutomationHelper.swift
+++ b/WireExtensionComponents/Utilities/AutomationHelper.swift
@@ -63,6 +63,9 @@ import zmessaging
         if arguments.hasFlag(AutomationKey.LogNetwork.rawValue) {
             ZMSLog.set(level: .debug, tag: "Network")
         }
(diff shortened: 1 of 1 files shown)


WireExtensionComponents/Utilities/AutomationHelper.swift
```diff
@@ -85,6 +89,13 @@ import zmessaging
         return ZMEmailCredentials(email: email, password: password)
     }
     
+    // Switches on all flags that you would like to log listed after `--debug-log=` tags should be separated by comma
+    fileprivate static func enableLogTags(_ arguments: ArgumentsType) {
+        guard let tagsString = arguments.flagValueIfPresent(AutomationKey.LogTags.rawValue) else { return }
+        let tags = tagsString.replacingOccurrences(of: " ", with: "").components(separatedBy: ",")
```
daehn: Some tags might have spaces in them

basine: too late! :P Will make another PR

PR #703, opened 2017-02-15
Fix potential crash in app lock

mikeger:
- ZMUserSession.shared can be nil, we seen the crash like that


daehn:
There also is a way to enqueue blocks to be performed after the `ZMUserSession` has been initialised.

mikeger:
@daehn good point, also what bugs me is that AppLockViewController does not really need the user session. I guess it would be cool to rethink `SettingsPropertyFactory` in the way that it only needs the arguments it is going to use.

mikeger:
@daehn updated

PR #704, opened 2017-02-15
Inverse images order in fullscreen images view

mikeger:
When the swiping view is opened from the conversation the images order must be inverted to match the logical images stream direction (top-down to left-right).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
@@ -37,12 +37,15 @@ internal final class ConversationImagesViewController: UIViewController {
     internal var buttonsBar: InputBarButtonsView!
     internal let overlay = FeedbackOverlayView()
     internal let separator = UIView()
-
+    
+    internal let inverse: Bool
+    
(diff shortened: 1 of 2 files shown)


typfel:
I don't get it. Do I need to swipe to the left after this PR?

mikeger:
@typfel when you are looking on the collection view, the images are going from the top to the bottom, and when you are looking on the conversation then they are going from the bottom to the top. Now when you open last image you would need to swipe left to reveal pre-last one.

daehn:
So when opened from the conversation swiping from left to right will reveal the previous image, whereas when opened from collections swiping from right to left will reveal the previous (in regards to the timestamp) image if I understand correctly?

mikeger:
@daehn correct

typfel:
Is that easier to understand?

mikeger:
@typfel it is more natural I guess, Marke requested the change.

daehn:
To me it makes sense from a users perspective, when the view containing multiple elements is newest at the top, then the detail should be horizontally newest at the most left (and vice versa when the order is inversed).

mikeger:
I think you expect that the previous message is to the left.

PR #709, opened 2017-02-16
Add :+1: , :-1: and :tada: to the autocorrect

mat1th:
@daehn @mikeger  This is a new pull request with only my name as pull request. So the auto CLAassistant accepts it.  The old one was #686;


jenskasemets:
Hi @mat1th, really big thank you for your effort to create a pull request. Sadly we will not merge in your pull request now. Reason for that is that we want all our platforms to be in sync feature and functionality-wise — it's not always possible because of platform specifics (e.g. Force touch on newer iPhones, no touchscreen on desktop etc.), but what comes content rendering (aka. user generated content), it should be always similar no matter on which device or screen you look on it. 
Currently emoticons ("autocorrect") are not supported on our desktop clients because of some technical limitations/complexity. Until this issue is not solved, we don't want to make any changes to this part of the code on other platforms, to diverge it even more.

We will keep your request in mind and once we have clarity on the next steps, we for sure will come back to this change. Thank you again.

mat1th:
Okay, good to know for the next time. 👍

PR #716, opened 2017-02-16
Display popup in case of AVS I/O error on internal builds

marcoconti83:
Display popup in case of AVS I/O error on internal builds. We want to monitor how often this happens on internal builds

### DIFF
--- a/Wire-iOS/Sources/Analytics/AnalyticsVoiceChannelTracker.m
+++ b/Wire-iOS/Sources/Analytics/AnalyticsVoiceChannelTracker.m
@@ -20,6 +20,7 @@
 #import "AnalyticsVoiceChannelTracker.h"
 #import "Analytics.h"
 #import "zmessaging+iOS.h"
+#import "DeveloperMenuState.h"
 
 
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Analytics/AnalyticsVoiceChannelTracker.m
```diff
@@ -83,6 +84,14 @@ - (void)callCenterDidFailToJoinVoiceChannelWithError:(NSError *)error conversati
 
 - (void)callCenterDidEndCallWithReason:(VoiceChannelV2CallEndReason)reason conversation:(ZMConversation *)conversation callingProtocol:(enum CallingProtocol)callingProtocol
 {
+    if (reason == VoiceChannelV2CallEndReasonInputOutputError && [DeveloperMenuState developerMenuEnabled]) {
```
jenskasemets: Does this mean it's only caught for v2 calls?

typfel: No, this error is actually only reported for v3 calls by AVS.

PR #717, opened 2017-02-16
Fix conversation list visible in conversation (voiceover)

mikeger:
# Issue

We where not hiding the conv list view when the conversation was open, therefore drawing the voiceover quite useless.

- Also removed "expanded" mode for left view controller when it used to occupy the whole viewport, for example in self profile;


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -350,34 +346,11 @@ - (NSArray *)constraintsInactiveForCurrentLayout
         [constraints addObjectsFromArray:@[self.pinLeftViewOffsetConstraint, self.sideBySideConstraint]];
     }
     
-    if (self.leftViewControllerExpanded) {
-        [constraints addObjectsFromArray:@[self.leftViewWidthConstraint]];
-    } else {
-        [constraints addObjectsFromArray:@[self.expandLeftViewConstraint]];
-    }
+    [constraints addObjectsFromArray:@[self.expandLeftViewConstraint]];
```
daehn: Do we still need this constraint at all?

mikeger:
@daehn you are right, please check again.

PR #718, opened 2017-02-17
Search UI improvements

mikeger:
- Updated icons in search UI
- Fixed black status bar on black theme
- Hide placeholder when the input is selected		
- Fixed accessiblity labels not visible		
- Fix for random black window bug


typfel:
@mikeger snapshots doesn't need to be updated?

mikeger:
@typfel checked, the button with this icon is not visible per default

PR #726, opened 2017-02-20
Fix search spinner design

daehn:
# What's in this PR?

* Adapt the new search spinner design.
* The new spinner location is left to the input fields cancel button, the spinner gets shown after 2 seconds.


daehn:
Need to rerecord some snapshots

Wire-iOS/Sources/UserInterface/Components/Views/TextSearchInputView.swift
```diff
@@ -86,14 +96,24 @@ public final class TextSearchInputView: UIView {
...
             searchInput.top == selfView.top
             searchInput.bottom == selfView.bottom
-            searchInput.trailing == cancelButton.leading
-            
+
+            placeholderLabel.leading == searchInput.leading
+            placeholderLabel.top == searchInput.top
+            placeholderLabel.bottom == searchInput.bottom
+            placeholderLabel.trailing == cancelButton.leading
             placeholderLabel.edges == searchInput.edges
```
mikeger: Wouldn't those constraints create the conflict?

daehn: Oh good catch, the edges one was supposed to be removed

daehn:
@mikeger fixed!

PR #727, opened 2017-02-20
Collection + text search fixes

mikeger:
- Fix for wrong layout after reopening of collection
- Fix for wrong keyboard color on dark theme
- Fix for collection not refetching new items


Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -147,13 +158,19 @@ final public class CollectionsViewController: UIViewController {
...
+
     }
     
     override public func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         self.textSearchController.teardown()
     }
     
+    override public func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
```
vytis: I guess we don't really need it if it only calls `super`

mikeger:
@vytis good point, removed it.

PR #728, opened 2017-02-20
Unify order and options in message long press menus

daehn:
## What's in this PR?

* Unify the order of the menu items in the `UIMenuController` that is presented when long pressing a message in the conversation view and in the collection of a conversation.
* Add an `open` item for files which presents the share sheet.
* Add the `like` item to the collection and the full screen image view.

## What's not in this PR?

* The `delete` item will be added to the collection message menus in a separate PR.


daehn:
@mikeger I added a couple of other commits, please re-review. The additions are:

* Add `delete` menu item to the collection, for now it just refetches after the deletion.
* Add `delete` option to the single image view.
* Add `like` / `unlike` as well as the `delete` option to the horizontal detail view bottom bar.

Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -684,6 +686,10 @@ extension CollectionsViewController: CollectionCellDelegate {
         switch action {
         case .forward, .showInConversation:
             self.delegate?.collectionsViewController(self, performAction: action, onMessage: message)
+        case .delete:
+            deletionDialogPresenter?.presentDeletionAlertController(forMessage: message, source: cell) { [weak self] in
+                self?.refetchCollection()
```
mikeger: According to ZIOS-7962 there should be a special deleted state for the message.

Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.m
```diff
@@ -552,6 +556,11 @@ - (void)likeImage
     [self.delegate wantsToPerformAction:MessageActionLike forMessage:self.message];
 }
 
+-(void)deleteImage
+{
+    [self.delegate wantsToPerformAction:MessageActionDelete forMessage:self.message];
```
mikeger: So would the view be closed after the deletion?

Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
daehn: Yes, like I mentioned in the comment for now it just refetches after the deletion. I need to investigate how much work this deletion state is, if I think about it it doesn't really make sense as the message will already have been deleted when the collection with the special state is shown, which means we need to keep track of deleted messages there separately somehow. I will first check with design what is preferred there.

Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.m
daehn: Yes, see https://github.com/wireapp/wire-ios/pull/728/files#diff-8ef41015046c0abf5c7336ac214dcf0fR742

daehn: When opened from the main app we dismiss the controller before all actions at the moment.

mikeger:
@daehn ok, looks good to me. Shall we merge?

Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
mikeger: I think we can just allow fetching the deleted image and video messages, then we can show the special cell for them. We would not have any image and sender, but I guess this is fine. We would also then have to filter out the deleted from horizontal swipe view. Anyways not the most useful feature if you ask me.

daehn:
@mikeger sure, let's merge it in!

PR #737, opened 2017-02-20
Align invitation copy

infotexture:
* Use most recent (username) variant for all three
* Drop trailing punctuation per YV preference

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -100,9 +100,6 @@
 "send_invitation.text" = "I’m on Wire, search for %@ or visit get.wire.com";
 "send_invitation_no_email.text" = "I’m on Wire, search for %@ or visit get.wire.com";
 
-"send_personal_invitation.text" = "I’m on Wire, search for %@ or visit get.wire.com";
-
-
 // +++++++++++++++++++++++++++++++++++++++++++++++++++++x
 // In-app notifications (chat heads)
 "notifications.shared_a_photo" = "shared a picture";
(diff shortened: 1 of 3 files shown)


infotexture:
Now that the copy has been aligned for all 3 cases, it might make sense to check whether it's still necessary to have 3 redundant copies of the same string, or whether code can be refactored to re-use one instance and remove the others.

infotexture:
If that was the reason for the extra variants, we should still align the remaining copy as closely as possible.

daehn:
@infotexture I checked and the logic is as follows:

1. If the user has a username we use the string with the key `send_invitation.text` with the username as substitution.
2. If the user does not have an username we check the email and use the same string as above but substitute with the email instead of the username.
3. As a last resort we fall back and use the string with the key `send_invitation_no_email.text`, which should not contain any substitutions.

daehn:
Note that 2. is actually a bug, as the search with email has been disabled. Currently this would render as `I’m on Wire, search email@example.com or visit get.wire.com.` which will result in an empty search result if the receiving user tries to do so (cc @jenskasemets ).

jenskasemets:
@daehn As scenario 2. can't be accessed anymore, we can also remove copy string and all related functionality

daehn:
@jenskasemets @infotexture I will do these changes on this branch then.

daehn:
@infotexture I updated this PR, there are only the following cases now:

1. The user has an username, `"send_invitation.text"` with the username (prefixed with @) as substitution is used.
2. The user does not have an username, `"send_invitation_no_email.text"` is used (no substitutions).

infotexture:
OK. For the variant without substitutions, I suggest we revert `send_invitation_no_email.text` to it's original form:

> I’m on Wire. Visit get.wire.com to connect with me.

infotexture:
@daehn Would you like me to do this ↑ — or will you?

infotexture:
👍 Thanks.

daehn:
@infotexture should the trailing punctuation be removed?

infotexture:
Not in this case, since there are 2 distinct sentences.

PR #738, opened 2017-02-20
Logic to exclude emoji cache from the backup

mikeger:
- Logic can be used for other random files that iOS creates


mikeger:
@daehn ¯\_(ツ)_/¯

PR #739, opened 2017-02-20
Remove AdSupport.framework

mikeger:
- Customers complained about us using the permanent ad ids
- Localytics needs this to track FB installations
- Removing it and using localytics seems to work out of the box


mat1th:
Do you mean Facebook by FB? Why do you actually use that in your app. You have also added the [FacebookAppID](https://github.com/wireapp/wire-ios/blob/663edefed93c3ce4796cc346c3ed9de4cd803b41/Wire-iOS/Wire-Info.plist#L75) in your app. Thats common used for Facebook tracking. I'll hope not because Facebook is the baddest company about privacy.  For what kind of service do you use it?

mikeger:
@mat1th We are using Localytics for anonymous usage data. Localytics could track FB as a source, but since this is not needed we are removing it.

FacebookAppID will also be removed in the scope of this pull request. Thanks for bringing it up!

mat1th:
@mikeger Okay, thanks for the explanation of the usage. And thanks for the FB removal as a source!

PR #742, opened 2017-02-21
Remove media metadata when sending

mikeger:
# Issue

For privacy reasons it is necessary to remove the metadata information from the images and videos that are being transferred via Wire.

# Requirements

- [x] zimages bump with https://github.com/wireapp/wire-ios-images/pull/5
- [x] wire-ios-data-model bump with https://github.com/wireapp/wire-ios-data-model/pull/189


daehn:
So are we stripping the metadata in all cases now or only when sending from camera roll?

mikeger:
@daehn the video metadata is always stripped on conversion; images metadata is stripped when image is sent as an image, when it's sent as a file we are not removing the metadata.

Cartfile.resolved
```diff
@@ -25,7 +25,7 @@ github "wireapp/wire-ios-utilities" "13.2.7"
 github "wireapp/wire-ios-images" "7.2.0"
 github "wireapp/wire-ios-transport" "18.0.2"
 github "wireapp/wire-ios-protos" "9.3.0"
-github "wireapp/wire-ios-data-model" "43.0.3"
```
daehn: This will break compilation as there were breaking changes in the data-model which first need https://github.com/wireapp/wire-ios/pull/746

mikeger: @daehn thanks, I know, waiting for your PR

PR #744, opened 2017-02-22
Add tracking for all settings

mikeger:
Knowing that all the settings are centralized, we can easily add the tracking events on any setting change.


typfel:
Shouldn't we remove tracking of settings which is now tracked twice?

- theme color
- something else?

mikeger:
I removed tracking for disableSendButton. Will check the theme color

mikeger:
- Fixed setting the profile picture event
- Removed the unused code
- Do not track `lockAppLastDate`

PR #749, opened 2017-02-23
Do not animate the archive update

mikeger:
- Temporary measure to limit the crash rate
- Should not break any functionality


Wire-iOS/Sources/UserInterface/Conversation/ArchivedListViewController.swift
```diff
@@ -144,21 +144,8 @@ extension ArchivedListViewController: ArchivedListViewModelDelegate {
     internal func archivedListViewModel(_ model: ArchivedListViewModel, didUpdateArchivedConversationsWithChange change: ConversationListChangeInfo, usingBlock: @escaping () -> ()) {
   
         guard initialSyncCompleted else { return }
-        let indexPathForItem: (Int) -> IndexPath = { return IndexPath(item: $0, section: 0) }
```
daehn: The passed in block is no longer executed, IIRC that one was used to update the data source or not?

mikeger:
@daehn right, missed that part

PR #750, opened 2017-02-23
Only fallback to underlying user if current one does not have a handle

daehn:
# What's in this PR?

* Only fallback to underlying user if current one does not have a handle.


mikeger:
@daehn it was just changed to work as it currently does. Does this behavior introduced the bug?

daehn:
I recently changed it to only consider the underlying user, which did't fully fix the issue.

mikeger:
OK, let's try this one then.

PR #753, opened 2017-02-24
Verification system messages when adding / removing participants in verified conversations

mikeger:
# Issue

When conversation gets verifier or degraded because of the participants change we would like to show the system message immediately.

Additionally, the verification system message should chronologically follow the changed participants message.

Additionally the system messages should be correctly added when the action is performed remotely.

# Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/195


Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Additions.m
```diff
@@ -116,10 +114,8 @@ - (ZMUser *)lastMessageSender
 
 - (void)removeParticipants:(NSArray *)participants
 {
-    for (ZMUser *user in participants) {
-        NSAssert([user isKindOfClass:ZMUser.class], @"Trying to remove a participant that is not a ZMUser!");
-        [self removeParticipant:user];
-    }
+    NSAssert([user isKindOfClass:ZMUser.class], @"Trying to remove a participant that is not a ZMUser!");
```
daehn: Where is `user` coming from?

mikeger: What Zoidberg says to Mike? 
1kjvl7

mikeger:
@daehn updated, thanks for spotting the error!

PR #757, opened 2017-02-27
Adjust swipe to reveal constants

daehn:
# What's in this PR?

* Adjust the constants used for the context menu revelation in the conversation list. Swiping very fast should trigger the menu now.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
```diff
@@ -46,7 +46,8 @@
 
 
 static const CGFloat MaxVisualDrawerOffsetRevealDistance = 48;
-static const NSTimeInterval IgnoreOverscrollTimeInterval = 0.1;
+static const NSTimeInterval IgnoreOverscrollTimeInterval = 0.005;
```
daehn: I tested this value manually and when swiping very fast values under 0.01 seconds are not unlikely.

PR #763, opened 2017-02-28
Open images and files by one tap

mikeger:
- With the current design it is hard to understand that tapping on file icon opens it
- Already working like that in collection
- It was decided to unify with images too: now tapping on image once is opening it


typfel:
Does this mean double tap to like is a back?

mikeger:
@typfel yes, but in trial mode, we need to check if people understand how it works. Some are got used to double-tap to open from other apps (like Twitter) and they accidentally like when they want to open the message.

Issue #370, opened 2016-11-06
Cannot edit long multiline messages


IpsmLorem:
On iOS, one cannot edit long multiline sent messages. Only the first lines are displayed without being able to edit further.

KoalaBoomka:
Hi @IpsmLorem! Thanks for the report.
I'm trying to reproduce it, but works fine for me. Sent 7 lines message, long tapped on it and chose Edit. I could see whole message in the edit block and could edit any line.

IpsmLorem:
It was on a "small" screen : iPhone 5. This could be the origin.

KoalaBoomka:
I see a problem on the small screens: there is a weird transition with a keyboard after choosing edit option, but still it's possible to edit the message. 
How long is a message what you tried to edit?

IpsmLorem:
Here is the shape of the text:

`xxxxx xxxxxxxxxxxx xxxx xx xxxxxxxxx xx xxx xxxxxx xxxxxxxx xxxxx. xxxxxx xx xxxxxxxx xx xxxxx xxxx xxxx x xxxx xx xxxxxxxx xxx xxx xx xx xxxxxx xxx xxxxxxxxxx 
xxx xxxxxx xx xxxxxxxxxxxx xxxxxxxx xxx xx xxxxx xx xx xxxxx xx xxxxxxxx xxxxxxxxx xxxxxx x xxxx`

KoalaBoomka:
Unfortunately, still works for me.

IpsmLorem:
With word wraps, the edit control displays:

`xxxxx xxxxxxxxxxxx xxxx xx
xxxxxxxxx xx xxx xxxxxx
xxxxxxxx xxxxx. xxxxxx xx 
xxxxxxxx xx xxxxx xxxx
xxxx x xxxx xx xxxxxxxx xxx
xxx xx xx xxxxxx xxx xxxx
xxxxxx.
xxx xxxxxx xx xxxxxxxxxxxx`

and doesn't displays the remaining:

`xxxxxxxx xxx xx xxxxx xx xx xxxxx xx xxxxxxxx xxxxxxxxx xxxxxx x xxxx`.

I've found something. This happend when you long tap on the text and select modfify.
The input box displays the croped text as previously explained.
But if you scroll down the keyboard to colaps it, the input control display the full text to edit.
If I scroll up back the keyboard the full text is still displayed to edit. It's a work around.

daehn:
[...]

PR #764, opened 2017-02-28
Add workaround to fix UITextView truncation

daehn:
# What's in this PR?

* Workaround for `UITextView` truncating the text in some cases.
* Fixes https://github.com/wireapp/wire-ios/issues/370.


mikeger:
@daehn old-good UIKit

Issue #756, opened 2017-02-27
UI: unread messages counter on the "back to contacts list" button


anatoli26:
Showing unread messages counter as a badge on the "back to contacts list" button (the top left arrow when in "chat with a contact" window) would give a hint that there are unread messages with other contacts.

Now I should click this button to open the contacts list so I can see if there are new messages for some contacts.

mikeger:
Thanks for the suggestion @anatoli26 . Our team found this idea interesting, initial implementation can be found here https://github.com/wireapp/wire-ios/pull/767

jenskasemets:
Hi @anatoli26, thank you for this nice small improvement suggestion. It's now available in release 2.32.

anatoli26:
Jens, thank you guys for taking into account and implementing so quickly the suggestions from the community!

PR #767, opened 2017-02-28
Unread status dot to the back button

mikeger:
Inspired by @anatoli26
Thanks @Yorgos-V and @markesaaremets 

https://github.com/wireapp/wire-ios/issues/756

screen shot 2017-02-28 at 15 49 04


typfel:
Is the button icon hooked up to a observer of the unread count?

daehn:
@mikeger do we observe the conversation list to update this when receiving a message while in a different conversation?

mikeger:
@typfel @daehn good point about possible incoming messages, I would need to add the observing

mikeger:
@daehn @typfel updated, now with observer! I hope it would not make the whole app 2x slower 🐌

Issue #725, opened 2017-02-17
Voice messages are not audible


rbieb:
iOS is not playing voice messages properly. This is a recent issue which has only been happening since  two or three weeks maybe.

When I play a voice message on ios the file is being played but there is just no sound. When I play the same file on the desktop version all is fine.

anatoli26:
Could it be that it is played to the ear speaker, not the loudspeaker? It was happening to me too. Putting the phone to the ear and off a couple of times fixed the problem. But this is something that should be checked by devs, though I know other apps (e.g. Signal) on both iOS and Android happen to have this problem too, but there are also apps that don't have this problem (Telegram, wapp).

daehn:
This issue is being addressed in https://github.com/wireapp/wire-ios/pull/769

PR #769, opened 2017-02-28
Set output device to earpiece initially when playing audio message

daehn:
# What's in this PR?

* Set output device to earpiece initially when playing audio message.
* Fixes https://github.com/wireapp/wire-ios/issues/725.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
```diff
@@ -370,14 +370,14 @@ final class AudioMessageView: UIView, TransferView {
     
     // MARK: - Audio state observer
     
-    func audioProgressChanged(_ change: NSDictionary) {
+    @objc private func audioProgressChanged(_ change: NSDictionary) {
```
mikeger: Just a note: the combination of @objc and private makes it practically public

daehn: Yeah, but somehow this still feels better than public. But as the `objc` should only be needed to  invoke dynamic dispatch it might be possible to mark it as `dynamic private` instead...

PR #770, opened 2017-02-28
Pinch to zoom in conversation

mikeger:
Adds support to look on the images in detail.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+PinchZoom.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+PinchZoom.m
@@ -1,0 +1,107 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+PinchZoom.m
```diff
@@ -0,0 +1,107 @@
...
+{
+    NSIndexPath *indexPath = [self.tableView indexPathForRowAtPoint:point];
+    id<ZMConversationMessage> message = [self.messageWindow.messages objectAtIndex:indexPath.row];
+    return message;
+}
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch
+{
+    CGPoint locationOfTouch = [touch locationInView:self.tableView];
+    id<ZMConversationMessage> message = [self messageAtPoint:locationOfTouch];
```
daehn: I would maybe add a check if the message is non-nil. The `Message` class has been rewritten in Swift and does not expect an Optional IIRC.

mikeger:
@daehn updated! Good point about swift, I automatically started to write code directly in the main file, and then realized that it is good to have a category :) Would be too much to rewrite it then.

PR #777, opened 2017-03-01
Disable text view workaround i os 8

daehn:
# What's in this PR?

* https://github.com/wireapp/wire-ios/pull/764 added a workaround to fix `UITextView` truncating its content. However, this workaround broke the text view on iOS 8. This PR makes the workaround conditional.
* This PR also fixes the access control modifiers in `AudioMessageView`.


typfel:
This should not go into the release?

daehn:
Yes, on it.

daehn:
@typfel https://github.com/wireapp/wire-ios/pull/778

PR #779, opened 2017-03-01
Only make links in tweet tappable in link previews

daehn:
# What's in this PR?

* Only make links in tweet tappable in link previews.


mikeger:
What was there for an issue?

daehn:
That links were tappable when they were contained in the `og:title` property. As this link was tapped accidentally by users trying to open the link preview we now limit this behaviour to tweets (see [ZIOS-7797](https://wearezeta.atlassian.net/browse/ZIOS-7797)).

PR #787, opened 2017-03-02
Vertically align title view containing text attachments

daehn:
# What's in this PR?

* Adjust vertical alignment inside `ConversationTitleView` to respect `NSTextAttachment`s which  change the title labels height.


Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
```diff
@@ -68,16 +78,21 @@ public final class ConversationTitleView: UIView {
...
     }
     
-    private func createConstraints() {
+    private func createConstraints(_ hasAttachment: Bool) {
         constrain(self, titleButton) { view, button in
-            button.edges == view.edges
+            button.leading == view.leading
+            button.trailing == view.trailing
+            button.top == view.top
+            button.bottom == view.bottom - (hasAttachment ? 4 : 0)
```
mikeger: I guess having `4` there is a necessary evil? Can you check how it works with dynamic type?

daehn: As the offset should only be affected by the text attachments (which are fixed height) dynamic type should not affect this magic number.

PR #788, opened 2017-03-02
Fix top alignment in BrowserViewController when shown with statusbar

daehn:
# What's in this PR?

* Fix top alignment in BrowserViewController when shown with statusbar.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.h
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.h
@@ -25,6 +25,7 @@
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder NS_UNAVAILABLE;
 - (instancetype)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil NS_UNAVAILABLE;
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.h
```diff
@@ -25,6 +25,7 @@
 
 - (instancetype)initWithCoder:(NSCoder *)aDecoder NS_UNAVAILABLE;
 - (instancetype)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil NS_UNAVAILABLE;
-- (instancetype)initWithURL:(NSURL *)URL NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithURL:(NSURL *)URL forUseWithStatusBar:(BOOL)statusBar NS_DESIGNATED_INITIALIZER;
```
mikeger: Wouldn't it be easier to have a bool property?

daehn: I thought it's cleaner to initialise it with the property (it should be clear by then if there will be a statusbar or not). Otherwise the constraints need to be updated which, as the status bar visibility doesn't change in this change, seemed unnecessary.

PR #790, opened 2017-03-02
Fixed cogwheel dot disappearing bug

mikeger:
- The observers change caused some logical changes


mikeger:
@daehn this is partially true, since right now we are not having any separate logic for marking the clients as not requiring the user attention. But I was more concerned about coupling the logic with settings controller, since it could happen that the device popup logic would change, and it's not necessary would be reflected on the cogwheel logic.

I would improve it in the way of adding the UserClient observers inside the user observer, but then I thought that it is actually only needed for this use case and would slow down all the other cases.

mikeger:
@daehn as you pointed out, there is a client observer, but it observes the relation, i.e. when the clients are added/removed, and not the properties on the clients inside.

PR #795, opened 2017-03-06
Fixed pull to dismiss

mikeger:
- Issue happened mostly because of the changes for styling the controllers before showing them. This changed the loading sequence of the views of the controllers.
- Other issue happened because of the screenshot view was shared between several fullscreen image views.
- Also removed some sketchy force unwraps.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
@@ -64,7 +64,7 @@ internal final class ConversationImagesViewController: UIViewController {
             self.createNavigationTitle()
         }
     }
-    internal var pageViewController: UIPageViewController?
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -104,15 +134,17 @@ internal final class ConversationImagesViewController: UIViewController {
     }
     
     private func createPageController() {
-        self.pageViewController = UIPageViewController(transitionStyle:.scroll, navigationOrientation:.horizontal, options: [:])
+        let pageViewController = UIPageViewController(transitionStyle:.scroll, navigationOrientation:.horizontal, options: [:])
```
daehn: Could be moved to the declaration to remove the optionality and the force unwrap when creating the constraints.

mikeger:
@daehn updated

PR #806, opened 2017-03-07
Add new conversation list calling indicators

daehn:
# What's in this PR?

* Add new conversation list calling icons.
* Add `"JOIN"` button to conversation list items with an ongoing group call.
* Update & rewrite `MissedCallCell` in Swift with new design and add `PerformedCallCell`. 
* Together with a new `wire-ios-date-model` and `wire-ios-sync-engine` release this adds the logic to show the collapsed cells for multiple subsequent calls.

### This PR depends on:
- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/204 has been merged.
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/301 has been merged.
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/304 has been merged.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CallCells.swift
```diff
@@ -0,0 +1,126 @@
...
+            let sender = message.sender,
+            let labelFont = font,
+            let labelBoldFont = boldFont,
+            let labelTextColor = textColor,
+            systemMessageData.systemMessageType == systemMessageType
+            else { return nil }
+
+        let senderString = string(for: sender) && labelBoldFont
+        let calledString = " \(string(for: "called"))" && labelFont
+        var title = senderString + calledString
```
mikeger: I know that we are doing this in other places, but can it be an issue for localization, where in the language the first and second parts should be swapped? Fortunately in this case not an issue for German (I think!). We can also do it the way that one string includes the parameter anchor %@, so it would be "%@ called" and then the translator can move who called around? Just an improvement idea.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -293,7 +293,7 @@ - (void)createBaseConstraints
         [self.authorImageView autoSetDimension:ALDimensionHeight toSize:authorImageDiameter];
     }];
     
-    [NSLayoutConstraint autoSetPriority:UILayoutPriorityDefaultHigh forConstraints:^{
+    [NSLayoutConstraint autoSetPriority:UILayoutPriorityRequired forConstraints:^{
```
mikeger: I think required is the default priority, so `autoSetPriority` is not necessary. What is the reasoning for changing the priorities by the way?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CallCells.swift
daehn: Yeah good point!

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
daehn: The intrinsic content size of the `MessageToolBoxView` was removed, instead it infers its height through the contained labels height now.

typfel: So if this becomes required don't we need to lessen the priority of the top / bottom constraints of the constrained label to avoid broken constraints when collapsing the toolbox?

daehn: They are lowered to `750` (which is `.defaultHigh`)

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CallCells.swift
daehn: @mikeger updated!

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
```diff
@@ -114,9 +123,13 @@ extension ZMConversationMessage {
     private func createConstraints() {
         constrain(self, reactionsView, statusLabel, labelClipView, likeTooltipArrow) { selfView, reactionsView, statusLabel, labelClipView, likeTooltipArrow in
             labelClipView.leading == selfView.leadingMargin
-            labelClipView.centerY == selfView.centerY
             labelClipView.trailing == selfView.trailingMargin
-            
+
+            selfView.height >= 28 ~ LayoutPriority(750)
```
typfel: @daehn  So this height constraint got lowered priority. But since the label is connected to the top and bottom instead of being centered wouldn't we intrinsic height constraint of the label cause a problem?

typfel: But again if you tried it and didn't see any warnings I guess it's ok :)

daehn: Didn't notice any warnings when testing, but I can check again. We don't want to restrict the labels height as it can grow larger, but as this constraint is `>=` a larger label should not be an issue.

typfel: I was thinking about the case when the toolbox gets collapsed to 0 height and the label gets squeezed since it's now connected to the top and bottom of the toolbox.

daehn: Good point. The height constraint should break as it has a lower priority, but the label will get squeezed instead of clipped as before I guess, but to be honest I don't see a difference in the animation in comparison to before. We need the edge constraints to size the view according to the label size, so I guess another solution would be to re-add the centerY constraints with a higher priority and lower the priorities of the edge constraints.

daehn: @typfel All snapshot tests testing `ConversationCell`s are failing at the moment because the content moved half a pixel down. I will need to check the reason for that (it doubt anyone will notice not looking at the snapshot diff, but still a bit suspicious).

daehn:
@mikeger @typfel I re-recorded the old (failing) snapshots for now until I found the root cause (4134049). I also added snapshot tests for the new cells (4134049). The added tests do not use the existing `MockMessage` setup but instead use a `ZMTestSession` and real messages etc. I think this is a more stable approach that we should consider using in the future (and maybe also convert the old tests at some point).

mikeger:
@daehn is there a test for the batched messages?

mikeger:
@daehn also what was the issue with 1px-off tests?

daehn:
@mikeger :

1. No there is none (yet). 
2. As mentioned I was not able to find the root cause yet, that's why I re-recorded the snapshots. If I get to find the cause I will update them again.

daehn:
@mikeger updated!

PR #809, opened 2017-03-08
Another call in progress: added base translation

vytis:
We didn't have any translations for this, maybe @infotexture could take a look at this?


jenskasemets:
Let's merge it for now and can later review copy if needed.

PR #814, opened 2017-03-09
Fix: video permissions dialog is shown on audio calls

vytis:
The underlying reason was that the video view is created when setting `videoViewFullscreen = true`. In the Obj-C version it was assigning it to ivar, bypassing the setter.


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
```diff
@@ -35,6 +35,21 @@ let GroupCallAvatarLabelHeight: CGFloat = 30.0;
     var shieldOverlay: DegradationOverlayView!
     var degradationTopConstraint: NSLayoutConstraint!
     var degradationBottomConstraint: NSLayoutConstraint!
+    var videoView: AVSVideoView!
```
mikeger: Is it wise to keep those force-unwrap, knowing that they are optional?

vytis: good point, optional it is!

vytis:
@mikeger updated with both video views optional

PR #817, opened 2017-03-10
Set input bar state correctly

daehn:
# What's in this PR?

* Since we removed the animation in https://github.com/wireapp/wire-ios/pull/800 we failed to actually set the input bar state.
* Fixes [8221](https://wearezeta.atlassian.net/browse/ZIOS-8221) and [8243](https://wearezeta.atlassian.net/browse/ZIOS-8243).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
@@ -106,11 +106,7 @@ private struct InputBarConstants {
         return inputBarState.isEditing
     }
     
-    var inputBarState: InputBarState = .writing(ephemeral: false) {
-        didSet(oldValue) {
-            updateInputBar(withState: inputBarState, oldState: oldValue)
-        }
-    }
+    private var inputBarState: InputBarState = .writing(ephemeral: false)
     
     fileprivate var textIsOverflowing = false {
         didSet {
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
@@ -340,7 +340,13 @@ private struct InputBarConstants {
 
     // MARK: - InputBarState
 
-    func updateInputBar(withState state: InputBarState, oldState: InputBarState? = nil, animated: Bool = true) {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -340,7 +340,13 @@ private struct InputBarConstants {
 
     // MARK: - InputBarState
 
-    func updateInputBar(withState state: InputBarState, oldState: InputBarState? = nil, animated: Bool = true) {
+    public func setInputBarState(_ state: InputBarState, animated: Bool) {
+        let oldState = inputBarState
```
typfel: Couldn't this logic of sometimes be handled inside the `updateInputBar()` method? Like if it's not supplied `oldState = state`.

the name `setInputBarState` is not very swifty.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -340,7 +340,13 @@ private struct InputBarConstants {
...
+        inputBarState = state
```
typfel: Setting `inputBarState` will trigger `didSet` which in turn calls `updateInputBar()`?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -340,7 +340,13 @@ private struct InputBarConstants {
...
+        updateInputBar(withState: state, oldState: oldState, animated: animated)
+    }
+
+    private func updateInputBar(withState state: InputBarState, oldState: InputBarState? = nil, animated: Bool = true) {
```
typfel: Maybe this `animated: Bool` should default to`false`, then you can just get the behaviour you want by setting the `inputBarState` property.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
daehn: Yeah good point, will restructure this.

daehn:
@typfel updated, now all external calls to update the `inputBarState` have to use the `setInputBarState` function and settings the instance variable does not have any side effects anymore.

daehn:
@typfel please re-review

PR #818, opened 2017-03-10
Call overlay: moved to swift

vytis:
Finished re-writing calling overlay in Swift. It's still a bit of a mess with around 1 kLOC, but didn't want to spend too much time on refactoring it. Currently it's almost a direct translation from Obj-C version.


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
```diff
@@ -18,25 +18,132 @@
...
+    func leaveButtonTapped()
+    func muteButtonTapped()
+    func speakerButtonTapped()
+    func videoButtonTapped()
+    func switchCameraButtonTapped()
+}
+
+@objc enum VoiceChannelOverlayState: Int {
+    case invalid
+    case incomingCall
```
mikeger: Could be done as an enumeration with associated value: incomingCall(Bool) meaning inactive (same for outgoing). Not sure if would be too helpful, but will be the translation to Swift :)

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
```diff
@@ -18,25 +18,132 @@
...
-    var degradationTopLabel: UILabel!
-    var degradationBottomLabel: UILabel!
-    var shieldOverlay: DegradationOverlayView!
-    var degradationTopConstraint: NSLayoutConstraint!
-    var degradationBottomConstraint: NSLayoutConstraint!
-    var videoView: AVSVideoView?
-    var videoPreview: AVSVideoPreview?
+fileprivate let VoiceChannelOverlayVideoFeedPositionKey = "VideoFeedPosition"
+
+@objc protocol VoiceChannelOverlayDelegate: NSObjectProtocol {
```
mikeger: I would implement it in the way of having one callback with action enum parameter, but I guess changing it now is too much of a hassle.

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
vytis: Unfortunately the `VoiceChannelOverlayController` that assigns the state to this view is still in ObjC so we can't do anything like that without more refactoring on that side.

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
```diff
@@ -562,65 +763,150 @@ extension VoiceChannelOverlay {
...
+        case .incomingCallDegraded:
+            return "OverlayIncomingCallDegraded"
+        case .joiningCall:
+            return "OverlayJoiningCall"
+        case .outgoingCall:
+            return "OverlayOutgoingCall"
+        case .outgoingCallDegraded:
+            return "OverlayOutgoingCallDegraded"
+        case .connected:
+            return "OverlayConnected"
```
mikeger: The file is still huge, I would consider splitting the categories in the separate files, or thinking on how to change the state of affairs on higher level, i.e. splitting the class into several, or making an MVVM-like solution.

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
vytis: That would be an option as well, but I tried to avoid changing `VoiceChannelOverlayController` as much as possible

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
vytis: Yeah I initially thought it will shrink substantially, but it went down only by 10% or so. I had to add some more functionality at the same time though. 
Well it's now in Swift so it will make it easier to work on it in the future. Probably splitting it into several classes along the `VoiceChannelOverlayState` lines would work best.

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlay.swift
```diff
@@ -52,7 +159,54 @@ let GroupCallAvatarLabelHeight: CGFloat = 30.0;
...
+    let makeDegradedCallButton = IconLabelButton(icon: .phone, label: "voice.call_button.title".localized, accessibilityIdentifier: "SecurityCallButton")
+    let acceptButton = IconLabelButton(icon: .phone, label: "voice.accept_button.title".localized, accessibilityIdentifier: "AcceptButton")
+    let acceptVideoButton = IconLabelButton(icon: .videoCall, label: "voice.accept_button.title".localized, accessibilityIdentifier: "AcceptVideoButton")
+    let ignoreButton = IconLabelButton(icon: .endCall, label: "voice.decline_button.title".localized, accessibilityIdentifier: "IgnoreButton")
+    let leaveButton = IconLabelButton(icon: .endCall, label: "voice.hang_up_button.title".localized, accessibilityIdentifier: "LeaveCallButton")
+    let muteButton = IconLabelButton(icon: .microphoneWithStrikethrough, label: "voice.mute_button.title".localized, accessibilityIdentifier: "CallMuteButton")
+    let speakerButton = IconLabelButton(icon: .speaker, label: "voice.speaker_button.title".localized, accessibilityIdentifier: "CallSpeakerButton")
+    let videoButton = IconLabelButton(icon: .videoCall, label: "voice.video_button.title".localized, accessibilityIdentifier: "CallVideoButton")
```
daehn: We should think about splitting this class into multiple views at some point, the amount of subviews is **a bit** high.

daehn: E.g. one view with the buttons, one with all the degradation related views / state and so on (not now, but at some point in the future™).

PR #824, opened 2017-03-14
Add new conversation renamed system message

daehn:
# What's in this PR?

* Remove `NameChangedCell` and add `ConversationRenamedCell`.
* Add `CoreDataSnapshotTestCase` helper to test target. This class creates a `ZMTestSession` and inserts a selfUser and one other user into the context. This class should be used to create objects for snapshot tests to avoid using the mock objects (which was very error prone) and is itself a subclass of `ZMSnapshotTestCase`.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationRenamedCell.swift
```diff
@@ -0,0 +1,99 @@
...
+            let labelTextColor = labelTextColor,
+            let senderString = sender(for: message) else { return nil }
+
+        let title = key(with: "title").localized(args: senderString).uppercased() && labelFont
+        return title.adding(font: labelBoldFont, to: senderString) && labelTextColor
+    }
+
+    private func attributedName(with systemMessage: ZMSystemMessageData) -> NSAttributedString? {
+        guard let name = systemMessage.text, let font = nameLabelFont, let color = labelTextColor  else { return nil }
+        return name && font && color
```
vytis: What magic is this? O_O

daehn: Custom operator magic 😬…

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationRenamedCell.swift
```diff
@@ -0,0 +1,99 @@
...
+        guard let color = labelTextColor else { return }
+        leftIconView.image = UIImage(for: .pencil, fontSize: 16, color: color)
+    }
+
+    private func updateLabel() {
+        guard let systemMessage = message.systemMessageData,
+            systemMessage.systemMessageType == .conversationNameChanged else { return }
+
+        labelView.attributedText = attributedTitle(for: message)
+        nameLabel.attributedText = attributedName(with: systemMessage)
```
mikeger: So one is `for: message` and the other is `with: systemMessage`, I guess it's the same message, would be cool to have a more similar approach, just to keep it simple?

daehn: @mikeger It's the same message, but the `systemMessageData` is optional and needs to be accessed in the outer method anyways, so this avoids another optional binding in `attributedName`. I updated the naming to be consistent.

PR #829, opened 2017-03-15
Add new burst design and day burst message.

daehn:
# What's in this PR?

* Increase the necessary time difference for the burst timestamp to be displayed between two subsequent messages to 45 minutes.
* Adopt new design for the burst timestamp.
* Show a day burst timestamp when the previous message is from a different day.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -256,16 +267,28 @@ - (void)didEndDisplayingInTableView
...
-    [self.burstTimestampLabel autoAlignAxisToSuperviewAxis:ALAxisVertical];
-    self.burstTimestampHeightConstraint = [self.burstTimestampLabel autoSetDimension:ALDimensionHeight toSize:0];
+
+    self.topMarginConstraint = [self.burstLabelContainer autoPinEdgeToSuperviewEdge:ALEdgeTop];
+    [self.burstLabelContainer autoPinEdgeToSuperviewEdge:ALEdgeLeading];
+    [self.burstLabelContainer autoPinEdgeToSuperviewEdge:ALEdgeTrailing];
+
+    [self.burstTimestampLabel autoPinEdgesToSuperviewMargins];
+
+    [NSLayoutConstraint autoSetPriority:UILayoutPriorityRequired forConstraints:^{
```
daehn: Necessary in order to avoid the container height to increase during the `MessageToolBoxView` show/hide animation.

mikeger: Hm, I thought the default priority is required?

daehn: I think you're right:

```objective-c
static const UILayoutPriority UILayoutPriorityRequired NS_AVAILABLE_IOS(6_0) = 1000; // A required constraint.  Do not exceed this.
```

Odd that it fixed the issue then 🤔

PR #834, opened 2017-03-16
Add new participants cells

daehn:
# What's in this PR?

* Add `ParticipantsCell`, which is used to display participant changes when a group conversation is started, participants have been removed or added.
* Remove the old `ParticipantsChangedView` and `ConversationParticipantsCell` (as well as related test classes).
* This PR also includes a change to the test setup, the app no longer starts the sync-engine when tests are running, which avoids random crashes in test. Before that change strategies could for example request an image download and do other unnecessary work while the tests are running.
* The user images (cells) are not tappable at the moment, but the logic for user interaction is there and just has to be enabled.


vytis:
Noticed in other snapshots tests as well - maybe it would be worth adding the view to be snapshotted into a fixed size container? The view size is added to the filename so if it changes by 1px we see it as new image instead of the actual change that happened in the view.

daehn:
We add the view size manually to the test name in `ZMSnapshotTestCase` as we have some tests that call `verifyInAllDeviceWidths` which will then compare the snapshots for all common sizes. But we do not use this method that often (at least recently) and might want to think about this.

vytis:
Size is useful for full-screen tests, but probably not much so in plain view snapshot...

daehn:
Otherwise they have the same name when using `verifyInAllDeviceWidths` as it uses the test name but takes several screenshots per tests (which would otherwise collide), but we could change it to append `_iPhone5` instead of the size.

daehn:
@vytis I will change the snapshot naming in a separate PR.

Wire-iOS/Resources/Classy/stylesheet.cas
```diff
@@ -753,37 +753,6 @@ VideoMessageCell {
...
-    }
-    
-    separatorView: @{
-        backgroundColor: $color-separator;
-    }
-}
-
 MessageToolboxView {
     statusLabel: @{
         font: $font-small-medium;
```
markesaaremets: $font-small-semibold

daehn: The conversion from using `semibold` instead of `medium` will also be done in a separate PR (tracked in [#8283](https://wearezeta.atlassian.net/browse/ZIOS-8283)).

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsCollectionViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        collectionViewLayout.prepare()
+        return collectionViewLayout.collectionViewContentSize
+    }
+
+}
+
+
+final class ParticipantsCollectionViewController<Cell: UICollectionViewCell>: UIViewController, UICollectionViewDelegate, UICollectionViewDataSource {
+
+    typealias SelectAction = (ZMUser, UIView) -> Void
```
mikeger: Should be `ZMUser, Cell` ?

daehn: @mikeger updated!

PR #838, opened 2017-03-16
Adjust conversation margins

daehn:
# What's in this PR?

* Adjust conversation margins and reduce author image size.
* Delete & rerecord all snapshots to clean up.


mikeger:
@daehn I see some system messages with the upper case in the screenshots. Would it be OK, or you would create another PR to make them lowercase (if it is the plan) ?

daehn:
I will merge back develop, where they are lowercased already.

daehn:
@mikeger updated!

daehn:
@mikeger found one snapshot where the title is still uppercased, will check.

daehn:
@mikeger updated (again)!

PR #843, opened 2017-03-17
Remove avatar border + image notification extension (not active)

mikeger:
- Removed avatars border.
- Added image notification extension.
- Deactivated image notification extension.
- FLAnimatedImage is moved to Carthage 🏟 .


daehn:
Would be good if someone else would also have a look (cc @typfel @vytis )

Podfile
```diff
@@ -37,3 +36,10 @@ end
 target 'WireExtensionComponents' do
     platform :ios, '8.0'
 end
+
+target 'ImageNotificationExtension' do
+    platform :ios, '8.0'
+    pod 'PureLayout', :git => 'https://github.com/wireapp/PureLayout', :tag => 'v3.0.0'
```
vytis: So what's the current consensus of what we have in CocoaPods and what in Carthage? I understand we like Carthage more, so maybe worth moving PureLayout there as well?

WireExtensionComponents/Views/ImageMessageView.swift
```diff
@@ -0,0 +1,165 @@
...
+        
+        self.imageView.contentMode = .scaleAspectFit
+        self.imageView.clipsToBounds = true
+        
+        self.userNameLabel.font = UIFont.systemFont(ofSize: 12, contentSizeCategory: .small, weight: .medium)
+        
+        self.userImageView.initials.font = UIFont.systemFont(ofSize: 11, contentSizeCategory: .small, weight: .light)
+        
+        constrain(self, self.imageView, self.userImageView, self.userImageViewContainer, self.userNameLabel) { selfView, imageView, userImageView, userImageViewContainer, userNameLabel in
+            userImageViewContainer.left == selfView.left
```
vytis: Probably not too important, but I tend to use leading/trailing to work better with RTL languages

Podfile
mikeger: @vytis we try to push everything to Carthage, I think last time we checked PureLayout used not to have the Carthage version. Might be that it do have the Carthage version now. But let me check why I added the purelayout to image notification extension, seems not necessary since it's all swift-only.

mikeger:
@vytis @daehn updated. Should we merge it to current release or to the next one? New target is not active, so signing should not break by merging this.

vytis:
Looks good, can't comment on when to merge this though:)

mikeger:
Merge now, cry later.

PR #844, opened 2017-03-20
Show sender if previous message is ping

daehn:
# What's in this PR?

* The ping cell no longer shows the author image, which is why we now always show it in case the previous message is a ping.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ZMConversationMessageWindow+Formatting.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ZMConversationMessageWindow+Formatting.m
@@ -91,7 +91,9 @@ static NSTimeInterval const BurstSeparatorTimeDifference = 60 * 45; // 45 minute
     }
     
     if (!systemMessage) {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ZMConversationMessageWindow+Formatting.m
```diff
@@ -91,7 +91,9 @@ - (BOOL)shouldShowSenderForMessage:(id<ZMConversationMessage>)message
     }
     
     if (!systemMessage) {
-        return ![self isPreviousSenderSameForMessage:message] || message.updatedAt != nil;
+        return  ![self isPreviousSenderSameForMessage:message]
+              || [Message isKnockMessage:[self messagePreviousToMessage:message]]
```
mikeger: Could be an issue, since `-messagePreviousToMessage` gives optional, whereas `-isKnockMessage:` expects `nonnull` and is implemented in Swift.

daehn:
@mikeger updated

PR #845, opened 2017-03-20
Updating font weights, new launch screen, unifying footer and header margins

markesaaremets:
- New launcscreen image
- Unifying footer and header margins to be 16pt from the edges (as Conversation)
- Medium font changed to Semibold to increase the contrast beween different font weights


Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileHeaderView.swift
```diff
@@ -82,7 +82,8 @@ final class ProfileHeaderView: UIView {
         }
 
         constrain(self, dismissButton, verifiedImageView, detailView.titleLabel) { (view: LayoutProxy, dismiss: LayoutProxy, verified: LayoutProxy, title: LayoutProxy) -> () in
-            dismiss.top == view.top + 24
+            dismiss.top == view.top + 26
+            dismiss.right == view.right - 8
```
daehn: The `dismiss.right` constraint might break the view in some contexts. Depending on the `headerStyle` the dismiss button is displayed on the left or the right side.

daehn: @markesaaremets see https://github.com/wireapp/wire-ios/pull/845/files#diff-1b8505afca416dff9fcb34ed891d1201R97

daehn: It looks like you can modify the `horizontalMargin` instead (`profile_temp.content_left_margin`).

jenskasemets: @daehn would it make sense to switch to `leading-trailing` here?

daehn:
@markesaaremets The newly recorded call system message has a different timestamp? The time changed one hour 🤔

Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileHeaderView.swift
mikeger: @jenskasemets may be yes, but Marke would not do it. Not like he cannot do it, but would take too much time to dig in this old ObjC code.

daehn: I think updating the magic constant is enough, as the constraint is already created here: https://github.com/wireapp/wire-ios/pull/845/files#diff-1b8505afca416dff9fcb34ed891d1201R97

PR #846, opened 2017-03-20
Add missing case to participants system message

daehn:
# What's in this PR?

* Add missing `"XYZ left"` case to participants system message.
* Update `accessibilityLabel `s for new system messages.


typfel:
No need to update snapshot tests?

daehn:
@typfel I will add one for the new case. Good point.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsCellViewModel.swift
```diff
@@ -44,14 +50,23 @@ struct ParticipantsCellViewModel {
...
-            let sender = message.sender.map(name),
+            let sender = message.sender,
             let labelFont = font,
             let labelBoldFont = boldFont,
             let labelTextColor = textColor else { return nil }
 
+        let senderName = name(for: sender)
+
+        // `XYZ` left
+        if systemMessage.users == [sender] && systemMessage.systemMessageType == .participantsRemoved {
```
mikeger: I think it would be somewhat better to implement it with `switch`, so it would be visible that first and second case (left and removed) are exclusive.

daehn:
@typfel @mikeger updated

PR #849, opened 2017-03-20
Small UI testing framework for beginners

mikeger:
- Allows to show arbitrary view fullscreen.
- Could be useful when you develop the view that is hard to show otherwise and for iOS development beginners @markesaaremets .


Wire-iOS/Sources/AppController.m
```diff
@@ -246,6 +246,8 @@ - (void)loadLaunchController
     
     [self.window makeKeyAndVisible];
     
+    //[TestView wr_testShowInstanceWithFullscreen:false];
```
daehn: 😱 `false`

mikeger: Oh `NO`es, AGAIN?

PR #851, opened 2017-03-20
Use OS blur, remove option of un-blur

mikeger:
- Removed BackgroundViewImageProcessor.
- Load image synchronously.
- Use iOS integrated blur view.


mikeger:
Hold on guys, I would rewrite it in Swift. It is too simple to keep the old code.

mikeger:
@daehn @typfel @vytis please re-review.
- Rewrote in swift
- Added tests
- Moved to live inside split view controller left panel

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+import UIKit
+import Cartography
+import zmessaging
+import WireExtensionComponents
+
+final public class BackgroundViewController: UIViewController {
+    fileprivate let imageView = UIImageView()
+    private let cropView = UIView()
+    private var blurView: UIVisualEffectView!
+    private var userObserverToken: NSObjectProtocol! = .none
```
vytis: `nil` would be more obvious here

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+    private let user: ZMBareUser
+    private let userSession: ZMUserSession?
+    
+    public init(user: ZMBareUser, userSession: ZMUserSession?) {
+        self.user = user
+        self.userSession = userSession
+        super.init(nibName: .none, bundle: .none)
```
vytis: same here

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+        super.viewDidLoad()
+        
+        self.configureViews()
+        self.createConstraints()
+        
+        self.updateForUser()
+    }
+    
+    override open var prefersStatusBarHidden: Bool {
+        get {
```
vytis: `return true` is enough, no need for getter

mikeger: No, compiler error, because `super` is not readonly.

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+            self.setBackground(color: user.accentColorValue.color)
+        }
+        else {
+            self.setBackground(imageData: user.imageMediumData)
+        }
+    }
+    
+    internal func updateFor(imageMediumDataChanged: Bool, accentColorValueChanged: Bool) {
+        switch (imageMediumDataChanged, accentColorValueChanged) {
+        case (true, _):
```
daehn: I think this might be easier to follow like this, but maybe I'm missing something here:
```swift 
if let data = imageMediumData, imageDataChanged {
    self.setBackground(imageData: data)
} else {
    self.setBackground(color: user.accentColorValue.color)
}
````

Comment (vytis) [inline Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift]: or another way is a 3-way switch:
```
     switch (imageMediumDataChanged, accentColorValueChanged, user.imageMediumData) {
     case let(true, _, .some(data)):
		self.setBackground(imageData: data)
     case let(true, _, nil):
		self.setBackground(color: user.accentColorValue.color)
     case (_, true, nil):        
         self.setBackground(color: user.accentColorValue.color)
     default: break
     }
```

File: Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m [hunk 5]
```diff
@@ -260,7 +259,7 @@ - (void)selectConversation:(ZMConversation *)conversation focusOnView:(BOOL)focu
     [stopWatch restartEvent:[NSString stringWithFormat:@"ConversationSelect%@", conversation.displayName]];
     
     @weakify(self);
-    [self.splitViewController setLeftViewController:self.conversationListViewController animated:animated completion:^{
+    [self.splitViewController setLeftViewController:self.backgroundViewController animated:animated completion:^{
```
typfel: Why do we need to set the left view controller here, In theory it should never change anymore, no?

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
vytis: on the other hand if it's force-unwrapped it's nil by default

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
vytis: @daehn way is much better!

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
mikeger: What would be the argumentation pro nil? 

My argument is that using `nil` is not precise, while nil can refer to any NilLiteralConvertible. Also nil has a certain aura of memory-related stuff which has no place in swift. So when I have optional I always use .none.

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
daehn: But also updates the accent color when neither `imageDataChanged`, nor `accentColorChanged` is `true`, but an `else if imageDataChanged ||  accentColorChanged` would fix that.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
mikeger: Good point, I think before we used to set the start UI as left controller.

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
typfel: Why can't the `userObserverToken` be optional?

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
mikeger: Yep, should work this way. Also would help to make sure not to reload the image when it's not changed.

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
mikeger: User is not optional, so token should always exist too.

mikeger: It's not like I really need it to exist :)

vytis: Well I think `.none` is more of an implementation detail of the optional enum than signifying an absence of value. 

Not sure if we would ever use `NilLiteralConvertible` anywhere, but then I see it more as an argument of using nil, not the other way around. If we would ever need this functionality then we would not need to change the code. Analogy with automatic type inference - we can freely change the underlying types of variables and don't worry about updating stuff everywhere because they will be inferred automatically as much as possible.

On the other hand it's like tabs-vs-spaces discussion, we can agree to disagree:)

mikeger:
2all: updated for the requested changes. Tests are passing.

PR #855, opened 2017-03-21
Make sure email is available when deleting devices, and add logs

marcoconti83:
There is a crash when trying to delete a device. We don't know why this happens. I wanted to fix it by showing an alert and allow for sending logs. However sending logs is a bit messy now, so it needs refactoring.

# Changes
- Show a debug alert when there is no email and we are trying to delete a device
- Refactor debug alerts and log sending


Wire-iOS/Sources/Components/Settings+Logging.swift
```diff
@@ -58,7 +58,7 @@ extension Settings {
     /// Sets whether recording is enabled
     private func setLogRecording(enabled: Bool) {
         if enabled && DeveloperMenuState.developerMenuEnabled() { // never record on non-internal
-            ZMSLog.startRecording(size: 20000)
+            ZMSLog.startRecording(size: 100000)
```
mikeger: It's over 9000!

PR #857, opened 2017-03-21
Initial VBR/CBR implementation

mikeger:
- Added switch to disable VBR in settings
- Added UI to show CBR state during the audio call
- Added tests

- [x] Requires https://github.com/wireapp/wire-ios-sync-engine/pull/313


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelOverlayController.m
```diff
@@ -532,4 +536,11 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
     return YES;
 }
 
+#pragma mark - CBR State Observer
+
+- (void)currentCBRChanged:(NSNotification *)notification
```
typfel: This name implies that the CBR could change to `on` or `off`. Maybe name it could be named `voiceChannelEnabledCBR:`.

Going beyond this. It would be nicer to expose this as a property on the voice channel at some point. The UI isn't supposed to know about the `WireCallCenterV3`.

mikeger: True, also currently the callback is not informing about which call it is firing, so Søren would add the proper information there.

mikeger:
@typfel updated!

PR #858, opened 2017-03-21
Add conversation list avatar view

daehn:
# What's in this PR?

* Add `ConversationListAvatarView`, which will be used to display a user image in the conversation list next to the conversation name. In case of a group conversation the view displays the number of participants.
* Update missed call icon color and re-record related snapshots.


vytis:
The snapshots for more than one user avatar seems like a squashed square instead of circle. Is that expected? 
(it seems you can't comment on binary files on GitHub)

daehn:
@vytis Yes that's on purpose.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    public var conversation: ZMConversation? {
+        didSet {
+            updateViewVisibility()
+            updateAlpha()
+        }
+    }
+
+    public let userImageView = UserImageView()
```
mikeger: Would be cool to have the `UserImageView` loaded on demand.

PR #859, opened 2017-03-21
Use objectID instead of remoteIdentifier for mapping conversation call state

marcoconti83:
Using `remoteIdentifier` for mapping conversation call state was causing a crash because remote identifier might be nil. `ObjectID` will never be nil (worst case: it will be a temporary one).

### DIFF
--- a/Wire-iOS/Sources/Managers/SoundEventListener.m
+++ b/Wire-iOS/Sources/Managers/SoundEventListener.m
@@ -41,7 +41,7 @@ static NSTimeInterval const SoundEventListenerIgnoreTimeForPushStart = 2.0;
 @property (nonatomic) id voiceChannelStateObserverToken;
 
 @property (nonatomic) ZMConversation *currentlyActiveVoiceChannelConversation;
-@property (nonatomic) NSMutableDictionary<NSUUID *, NSNumber *> *previousVoiceChannelState;
+@property (nonatomic) NSMutableDictionary<NSManagedObjectID *, NSNumber *> *previousVoiceChannelState;
 @property (nonatomic) SoundEventRulesWatchDog *watchDog;
 
 @end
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Managers/SoundEventListener.m
```diff
@@ -159,8 +159,8 @@ - (void)didReceiveNewUnreadKnockMessages:(NewUnreadKnockMessagesChangeInfo *)cha
 - (void)callCenterDidChangeVoiceChannelState:(VoiceChannelV2State)voiceChannelState conversation:(ZMConversation *)conversation callingProtocol:(enum CallingProtocol)callingProtocol
 {
     VoiceChannelV2State state = voiceChannelState;
-    VoiceChannelV2State previousState = self.previousVoiceChannelState[conversation.remoteIdentifier].integerValue ?: VoiceChannelV2StateInvalid;
-    self.previousVoiceChannelState[conversation.remoteIdentifier] = @(voiceChannelState);
+    VoiceChannelV2State previousState = self.previousVoiceChannelState[conversation.objectID].integerValue ?: VoiceChannelV2StateInvalid;
+    self.previousVoiceChannelState[conversation.objectID] = @(voiceChannelState);
```
mikeger: Can the call happen in the conversation without the remote identifier?

marcoconti83: Yes, the bug repro steps are: 
- Go in airplane mode
- Create conversation
- Make a call

mikeger: Right, I thought we have the guards agains starting the call in offline: 
https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Source/Calling/VoiceChannelV2+CallFlow.m#L216
https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Source/Calling/VoiceChannelV3.swift#L84

So whenever we call `join` it should check first if the conversation is valid. Which is I guess having a flaw somewhere, if we are getting to the state where it is crashing. But I think it's worth checking why the checks are passing in the `join` calls and we try to start the call. The thing is also if the conversation does not have the remote identifier then we are practically unable to have the call there.

marcoconti83: There is a race condition, if you disable airplane mode and then try to make the call before we managed to sync / complete the request to insert the conversation on the BE, in this case you will not be in offline mode but you still don't have a remote ID.
I agree, we should fix the fact that you should not be able to start a call in a conversation that does not have a remote identifier. I would do that in a separate PR.

PR #862, opened 2017-03-22
Pin line view to xHeight of the first line of text in IconSystemCell

daehn:
# What's in this PR?

* The line to the right of the `UILabel` in `IconSystemCell` was pinned to the top with a hardcoded distance. With dynamic type the font size can change and the label will not be aligned to the center of the first line of text in the label anymore. 
* We now calculate the y-Position of the center of the first line of text to take changing fonts into account.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/IconSystemCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/IconSystemCell.swift
@@ -34,7 +34,14 @@ open class IconSystemCell: ConversationCell, TTTAttributedLabelDelegate {
     
     var labelTextColor: UIColor?
     var labelTextBlendedColor: UIColor?
-    var labelFont: UIFont?
+
+    var lineBaseLineConstraint: NSLayoutConstraint?
+
+    var labelFont: UIFont? {
+        didSet {
+            updateLineBaseLineConstraint()
+        }
+    }
     var labelBoldFont: UIFont?
 
     var verticalInset: CGFloat {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/IconSystemCell.swift
```diff
@@ -99,13 +105,33 @@ open class IconSystemCell: ConversationCell, TTTAttributedLabelDelegate {
...
+
+        constrain(lineView, labelView, leftIconContainer) { (lineView: LayoutProxy, labelView: LayoutProxy, icon: LayoutProxy) -> () in
+            lineBaseLineConstraint = lineView.centerY == labelView.top + self.labelView.font.median - 2
+            icon.centerY == lineView.centerY
         }
     }
 
+    private func updateLineBaseLineConstraint() {
+        guard let font = labelFont else { return }
+        lineBaseLineConstraint?.constant = font.median - 2
```
mikeger: Would be cool to extract the `-2` calculation into the separate method and call from here and `createConstraints`.

daehn: Yeah, was thinking about that. But then maybe into a different method and not into the `median`, as the 2 is kind of a magic number here and not related to the median of a font.

mikeger: Right, I am just afraid that someone would change -2 in one place and not in the other one!

PR #863, opened 2017-03-22
Add avatar view to conversation list cell

daehn:
* This temporarily removes the indicator view (which displays missed calls, unread indicator etc).
* Fix archive top bar margins now that the status bar is always visible.
* Restrict subtitle label to 1 line.
* Fix tests and update snapshot.


Wire-iOS/Sources/UserInterface/ConversationList/ArchivedNavigationBar.swift
```diff
@@ -75,7 +77,7 @@ import Cartography
     }
     
     override var intrinsicContentSize : CGSize {
-        return CGSize(width: UIViewNoIntrinsicMetric, height: barHeight)
+        return CGSize(width: UIViewNoIntrinsicMetric, height: barHeight + statusbarHeight)
```
mikeger: Should `statusbarHeight` be 0 for iPad? We need to discuss how to handle it, may be add the inset into the background view controller, since I already have the inset logic there and in the conv list.

mikeger:
@daehn I forgot to move the new devices dot to the top bar :(

Wire-iOS/Sources/UserInterface/ConversationList/ArchivedNavigationBar.swift
daehn: Good point, we actually have another top bar class called `ModalTopBar` which gets initialised with a `Bool` specifying whether or not there should be extra padding for the status bar. Maybe it makes sense to use that one instead. Then we decide to hide it on iPad, or better when having a non-compact horizontal trait collection maybe?

PR #866, opened 2017-03-22
Fix crash in UserImageView (missing UserSession)

mikeger:
UserSession has to be injected now, since the UserImageView is ported to WireExtensionComponents, and there the user session is not possible to have.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ReactionsList/ReactionCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ReactionsList/ReactionCell.swift
@@ -55,6 +55,8 @@ import Classy
     public override init(frame: CGRect) {
         super.init(frame: frame)
         
+        self.userImageView.userSession = ZMUserSession.shared()
+        
         self.contentView.addSubview(self.userDisplayNameLabel)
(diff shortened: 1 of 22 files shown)


Wire-iOS/Sources/UserInterface/Conversation/ReactionsList/ReactionCell.swift
```diff
@@ -55,6 +55,8 @@ import Classy
     public override init(frame: CGRect) {
         super.init(frame: frame)
         
+        self.userImageView.userSession = ZMUserSession.shared()
```
daehn: If the `UserImageView` needs the user session in order to request downloading pictures it might make sense to add is as a parameter to the `init`. Could be optional as is won't be there in an extension, but I fear people in the future not knowing they have to set a user session in order for the image downloading to work.

mikeger: @daehn agree, was also thinking how to inject it properly, but we have several `init`s and also sometimes we have magic involved, so I would propose to make an another PR that would remove magic and add the parameter.

PR #868, opened 2017-03-22
Feature/set profile picture v3 zios 8302

daehn:
# What's in this PR?

* Use `UserProfileImageUpdateStatus` to update the user profile image everywhere.

### This PR depends on 
- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/315

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/ProfilePictureStepViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Registration/ProfilePictureStepViewController.m
@@ -295,10 +295,10 @@ NSString * const UnsplashRandomImageLowQualityURL = @"https://source.unsplash.co
         [[ZMUserSession sharedSession] checkNetworkAndFlashIndicatorIfNecessary];
         
         if ([self.editabledUser isKindOfClass:[ZMIncompleteRegistrationUser class]]) {
-            self.editabledUser.originalProfileImageData = imageData;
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Registration/ProfilePictureStepViewController.m
```diff
@@ -295,10 +295,10 @@ - (void)setPictureImageData:(NSData *)imageData
         [[ZMUserSession sharedSession] checkNetworkAndFlashIndicatorIfNecessary];
         
         if ([self.editabledUser isKindOfClass:[ZMIncompleteRegistrationUser class]]) {
-            self.editabledUser.originalProfileImageData = imageData;
+            [[ZMUserSession sharedSession].profileUpdate updateImageWithImageData:imageData];
```
daehn: I am not sure why the call is not wrapped in an `enqueueChanges` in case the user is an `ZMIncompleteRegistrationUser`. I assume we do not want to save & process changes, but why not? (cc @mikeger @typfel)

mikeger: I think it should be wrapped. Most likely working by accident.

vytis:
It's being merged to a temporary branch - we can verify everything when the final PR is ready

PR #870, opened 2017-03-23
Initial conversation list second line implementation

mikeger:
- With tests.
- Not final implementation, designers wanted to play with it first.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
```diff
@@ -19,126 +19,193 @@
...
-    public required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+    public func rand<ContentType>() -> ContentType {
+        let currentStep = self.step
+        let result = source.withUnsafeBytes { (pointer: UnsafePointer<ContentType>) -> ContentType in
+            return pointer.advanced(by: currentStep % source.count).pointee
+        }
+        step = step + 1
+        
+        return result// ^ ContentType(Int(step / source.count))
```
daehn: The commented code can be removed?

WireExtensionComponents/Utilities/Enum.swift
```diff
@@ -0,0 +1,29 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public func iterateEnum<T: Hashable>(_: T.Type) -> AnyIterator<T> {
+    var i = 0
+    return AnyIterator {
+        let next = withUnsafeBytes(of: &i) { $0.load(as: T.self) }
+        if next.hashValue != i { return nil }
```
daehn: Does this work for all kind of `Enum`s (even if the base type is not an `Int` for example)? 
If not it could be done as an extension on `RawRepresentable` by restricting the `RawValue` to be an Int or something like that. Also I think it would make more sense to add this as a static variable holding all possible enum cases (removing the need to return an iterator here, e.g. `SomeEnum.allValue.forEach`).

daehn: I saw an implementation which doesn't rely on `hashValue` similar to this somewhere recently (don't remember where it's from):
```swift
extension RawRepresentable where RawValue == Int {
    static var all: [Self] {
        var index = 0
        var values = [Self]()
        while let value = Self(rawValue: index) {
            values.append(value)
            index += 1
        }

        return values
    }
}
```

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
mikeger: Yeah, I am still thinking how to do it properly, when it would start rolling over the edge of data given the values would start to repeat, but I cannot yet figure out how to do the generic xor properly. Easiest is to create the data of size of the input and xor two datas, but it's not optimal IMO.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
```diff
@@ -19,126 +19,193 @@
...
+    public let source: Data
+    private var step: Int = 0
+    
+    init(data: Data) {
+        source = data
     }
     
-    public required init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+    public func rand<ContentType>() -> ContentType {
```
typfel: Could explain this function? Iterates over a an array of UUIDs?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
daehn: This is used to get the same 2 (or 4) random users for the image view at the moment, right? What speaks against ordering the participants by name and taking the first ones, or ordering them and then splitting the array in 2 (or 4) parts of equal length and taking the first user of each. The images might change once participants are added / removed in any case, or not?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
mikeger: I would add the comment, the main idea is that it uses the UUID plain data to generate the random values, so the 2x2 avatars in the group conversations with 4+ participants would be showing the same users on different platforms.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
daehn: @mikeger saw you comment regarding cross platform behaviour. Makes sense to show the same pictures on all platforms, was just thinking if there might a a simpler way to achieve this.

WireExtensionComponents/Utilities/Enum.swift
mikeger: I now remember that we have this code somewhere in SE, but anyways.

mikeger:
@typfel @daehn I have the new design for group avatar :) would do another PR, this one is already too big.

Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -0,0 +1,416 @@
...
+        if self.messages.count < 10 {
+            hasMessages = self.messages.flatMap {
+                StatusMessageType(message: $0 as! ZMConversationMessage)
+            }.count > 0
+        }
+        else {
+            hasMessages = true
+        }
+        
+        let isOngoingCall: Bool = self.voiceChannel?.state ?? .noActiveUsers != .noActiveUsers
```
vytis: Maybe some brackets would be good to make the logic more obvious.

mikeger: Good point!

WireExtensionComponents/Utilities/AttributedStringOperators.swift
```diff
@@ -160,7 +160,25 @@ public extension NSAttributedString {
...
+        
+        var first = true
+        
+        self.forEach { string in
+            if !first {
+                result.append(separator)
+            }
+            result.append(string)
+            
+            first = false
```
vytis: A bit weird to have `.forEach` with side-effects, maybe better to use `for in` in this case then.

WireExtensionComponents/Utilities/RawRepresentable+AllValues.swift
```diff
@@ -0,0 +1,32 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension RawRepresentable where RawValue == Int {
+    public static var allValues: [Self] {
```
vytis: A small caveat that this expects all enum cases to have sequential values without gaps - will not work if we decide to skip a number or few.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAvatarView.swift
```diff
@@ -19,126 +19,208 @@
...
+    /// / AB \
+    /// \ CD /
+    case four
+}
 
-    required public init?(coder aDecoder: NSCoder) {
-        fatalError("init(coder:) has not been implemented")
+extension Mode {
+    fileprivate init(conversation: ZMConversation) {
+        if conversation.activeParticipants.count <= 2 || conversation.conversationType != .group {
```
vytis: Would be slightly cleaner with a `switch` on `conversation.activeParticipants.count` and using ranges

WireExtensionComponents/Utilities/AttributedStringOperators.swift
mikeger: Right, I would add it to the comment.

WireExtensionComponents/Utilities/RawRepresentable+AllValues.swift
mikeger: Right, I would add it to the comment.

PR #872, opened 2017-03-24
Removed double space in added device system message

mikeger:
- It was also visible on the snapshot.


Wire-iOS/Resources/Base.lproj/Localizable.stringsdict
```diff
@@ -82,72 +82,72 @@
...
+                <key>many</key>
+                <string>%d others</string>
+                <key>other</key>
+                <string>%d others</string>
+            </dict>
+        </dict>
         <key>content.system.people_started_using</key>
         <dict>
             <key>NSStringLocalizedFormatKey</key>
-            <string>%@ %#@d_number_of_others@ started using %#@d_new_devices@</string>
```
mikeger: This is the real change + line 160

typfel:
Is the double space only present in the english translation?

mikeger:
@typfel in all, but @jenskasemets usually tells not to touch the other translations with manual changes

jenskasemets:
@typfel with this fix it should also fix most of the issues with translations as well. In some language we need manually modify localization strings in Crowdin, small tweaks.

PR #876, opened 2017-03-27
New conversation list

mikeger:
- Contains already reviewed changes.

Pending tests for accessory view and new design for the group avatar + design and QA reviews. This would be done as the separate PRs.

Depends on 

- [x] https://github.com/wireapp/wire-ios/pull/879


Wire-iOS Tests/ConversationListAvatarViewTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+import XCTest
+import Cartography
+@testable import Wire
+
+private let accentColors: [ZMAccentColor] = [.vividRed, .softPink, .brightYellow, .strongBlue, .strongLimeGreen]
+
+extension Array where Element: ZMUser {
+
+    func assignRandomAccentColors() {
```
daehn: I would maybe change the name as it's not random (and we rely on the fact that it is deterministic in the tests).

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -117,9 +117,43 @@
...
 "list.archived_conversations" = "ARCHIVE";
 "list.archived_conversations_close" = "Close archive";
 
+"conversation.status.call" = "Ongoing call";
+"conversation.status.typing" = "Typing a message...";
+"conversation.status.typing.group" = "%@: Typing a message...";
+"conversation.status.silenced" = "Muted";
+"conversation.status.blocked" = "Blocked";
+
+"conversation.silenced.status.message.text" = "%d new text messages";
```
daehn: I think these strings should have pluralisation rules and be in the `.stringsDict` file, e.g. `"1 new message"` vs. `"3 new messages" (or there could be one string for singular one for plural, but the dict might be cleaner).

mikeger: Sure, but I thought that localization people can do it?

daehn: But this also affects the code, as the number has to be passed in as argument to the `localized()` function in case of the strings dict or a different localization key has to be used in case of separate singular/plural localizations.

mikeger: Ah right, good point.

Wire-iOS Tests/ConversationListAvatarViewTests.swift
mikeger: There is always an XKCD for it https://xkcd.com/221/

daehn:
The typing indicator looks misaligned in this snapshot, intentional?

bildschirmfoto 2017-03-29 um 15 53 49

mikeger:
@daehn was had to notice on alpha background, now fixed.

daehn:
@mikeger good to merge from my side!

PR #878, opened 2017-03-28
Add message to logs, and more warnings

marcoconti83:
We get many log reports where it's not clear why they were sent.

# Changes
If the log comes from a popup, include the message of the popup in the report


Wire-iOS/Sources/Developer/DeveloperOptionsController.swift
```diff
@@ -84,7 +84,15 @@ extension DeveloperOptionsController {
     /// Creates a cell to forward logs
     func forwardLogCell() -> UITableViewCell {
         return self.createCellWithButton(labelText: "Forward log records") {
-            DebugLogSender.sendLogsByEmail()
+            let alert = UIAlertController(title: "Add explaination", message: "Please explain the problem that made you send the logs", preferredStyle: .alert)
```
daehn: Typo `explaination` → `explanation`

marcoconti83:
@daehn fixed

PR #879, opened 2017-03-28
New conversation list improvemens

mikeger:
- Removed unused code.
- Calculate cells size dynamically.
- Fixed logic with showing "1" for one new message.
- Fixed logic with muted state subtitle.
- Fixed font color for subtitle.
- Increased background saturation.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController.m
```diff
@@ -466,6 +468,24 @@ - (UIEdgeInsets)collectionView:(UICollectionView *)collectionView layout:(UIColl
     return UIEdgeInsetsZero;
 }
 
+- (CGSize)collectionView:(UICollectionView *)collectionView
+                  layout:(UICollectionViewLayout *)collectionViewLayout
+  sizeForItemAtIndexPath:(NSIndexPath *)indexPath
+{
+    self.layoutCell.itemView.titleText = @"Ü";
```
daehn: Did you check if these characters need the maximum line height? Might be that some Emoji would make the label expand more vertically. Another solution might be to compute this using the font directly.

Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -393,6 +408,20 @@ extension ConversationStatus {
         return [topMatcher] + topMatcher.combinesWith.filter { $0.isMatching(with: self) }
     }
     
+    func appliedMatcherForIcon(for conversation: ZMConversation) -> ConversationStatusMatcher? {
+        for matcher in appliedMatchers {
+            let icon = matcher.icon(with: self, conversation: conversation)
+            switch icon {
+            case .none:
+                break
+                default:
```
daehn: The indentation is off here.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController.m
mikeger: This is to check the size with the current font, I want all cells to be the same height, with and without the subtitle, hence I need to put some data into the view to have the correct layout.

daehn: I understand that, what I meant is if the letters `Ü` and `Ä` use the full lineHeight. If there are Emoji in the text the label might need to stretch even more vertically, which might cut off the Emoji if the height calculation used a different letter, but I might be wrong.

mikeger: True, you are thinking to set the extra constraint on label with line height for layout cell?

daehn: Not sure, maybe it's enough to get the `lineHeight` or rather to compute some maximum line height from the font itself.

daehn: But we can merge it and optimise this later!

PR #882, opened 2017-03-29
Visual cleanup of conv list, archive, settgings, start UI

mikeger:
- Same row height.
- Same font.
- Same separators.
- Group avatars in Start UI.


typfel:
The separators are gone from the new snapshots, intentional?

mikeger:
@typfel Bug!

PR #883, opened 2017-03-29
New avatar design

mikeger:
- Also changed background image saturation to 2.
- Updates for big PR review ( 2 @daehn ).


Wire-iOS/Sources/Helpers/String+WireLocales.swift
```diff
@@ -18,6 +18,13 @@
 
 import Foundation
 
+extension String {
+    func capitalizingFirstLetter() -> String {
```
daehn: You can use `capitalized` on `String` instead.

mikeger: It Capitalizes All The Words

daehn: Ah yeah, was too fast commenting 👍

PR #888, opened 2017-03-30
Fixed stable random, added tests

mikeger:
- Unsafe data pointers also unsafe in Swift: it does not check the memory bounds, therefore the rubbish data was added to the random output.
- Also renamed `ConversationAvatarView` and moved to common controls.

### DIFF
--- a/Wire-iOS Tests/ConversationAvatarViewTests.swift
+++ b/Wire-iOS Tests/ConversationAvatarViewTests.swift
@@ -37,13 +37,13 @@ extension Array where Element: ZMUser {
     }
 }
 
-class ConversationListAvatarViewTests: CoreDataSnapshotTestCase {
+class ConversationAvatarViewTests: CoreDataSnapshotTestCase {
 
-    var sut: ConversationListAvatarView!
+    var sut: ConversationAvatarView!
 
     override func setUp() {
         super.setUp()
-        sut = ConversationListAvatarView()
+        sut = ConversationAvatarView()
     }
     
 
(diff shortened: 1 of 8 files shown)


daehn:
@mikeger it's hard to see the `rand` fix as the file has been renamed, can you explain it here instead?

mikeger:
@daehn as I mentioned in the PR description, I was walking over some random memory that was after the data pointer:
Imagine UUID binary in memory is 
```
0                                        32
XXXXXXXXXXXXXXXXXXXRRRRRRRR
```

Where X is an actual UUID binary value and R is rubbish or the memory that belongs to some other objects/plain data.
I my random source has an index where to get the next value: 

```
I=0
0                                        32
XXXXXXXXXXXXXXXXXXXRRRRRRRR
```
Then when the next value was requested, let's say of size 8 bytes, it was getting first 8 bytes, but moved the I only on 1 byte. It's not an issue until you get to the offset of 32-8, because then it would start to read the rubbish values that are laying over the 32rd byte. Therefore the values where not really stable random.

The fix is to advance by the requested value size and move index to 0 already when next value would start reading from the rubbish memory. Nicer fix would be to use the ring buffer, but I decided to patch the current approach.

mikeger:
@daehn I also just renamed and moved the group avatar view file, this was for some reason not recognized as move by git.

marcoconti83:
What is the randomization used for?

mikeger:
@marcoconti83 to pick the avatars for the group conversation image based on conversation backend ID (if there are more than 5 participants)

PR #889, opened 2017-03-30
List filtering demo

mikeger:
- Initial PoC implementation.
- Logic is: in work space show conversations that have "ω" in the name.


Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -210,34 +210,43 @@ final public class ConversationAvatarView: UIView {
...
-                yPosition = yPosition + size.height + inset
+        else {
+            clippingView.frame = self.bounds.insetBy(dx: 2, dy: 2)
+            
+            let size: CGSize
+            let inset: CGFloat = 2
+            let containerSize = self.clippingView.bounds.size
+            
+            switch mode {
+            case .one:
```
daehn: Can this all be moved into one `switch`?  Then the `fatal` wouldn't be needed.

mikeger: Unfortunately no, because all the other code does not make sense for "one" mode :(

mikeger:
@daehn don't merge, I found a bug

mikeger:
@daehn I got excited and added other list fixes in this PR, sorry

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
daehn: Hmmm, but maybe the code common for `.two` and `.four` could be extracted in a separate method and only called from the `case .two` and `case .four` then. Or the switch could have a `case .two, .four` and get the size depending on the mode in that case.

mikeger: Yeah, but then inside this separate method I would also need to have switch that would have .one as an option. I would create the method that takes size as the parameter.

mikeger:
@daehn also I think I forgot something: run tests!!

mikeger:
@daehn please re-review

PR #894, opened 2017-03-31
Fix accessibilityValue value not updating for IconSystemCell

daehn:
# What's in this PR?
* The `accessibilityLabel` of the `MissedCallCell` was not changed when setting a new one in some cases, using a regular `UILabel`  instead of a `TTTAttributedLabel` fixes this issue.
* Remove setting a default `accessibilityLabel` in `IconSystemCell` as the labels text will be used in by default.
* The magic y-inset we used for the `TTTAttributedLabel` is not needed when using a `UILabel`.


daehn:
@mikeger updated!

PR #896, opened 2017-03-31
Fix previewing view controller frame when previewing images

daehn:


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+PeekPop.swift
```diff
@@ -30,6 +30,9 @@ extension ConversationContentViewController: UIViewControllerPreviewingDelegate
         
         if message.isImage {
             let controller = self.messagePresenter.viewController(forImageMessage: message, actionResponder: self)
+            if let rect = (tableView.cellForRow(at: cellIndexPath) as? ConversationCell)?.selectionRect, rect != .zero {
+                previewingContext.sourceRect = rect
```
mikeger: Is `selectionRect` in cell's coordinates? Would be useful to convert to the coordinate space that Peek&Pop is expecting even if it's the same now, because we can change the way either is working and it would break the code here.

daehn: @mikeger updated

PR #900, opened 2017-04-03
Profile picture using assets V3

vytis:
Changes on UI are tiny, all functionality is in:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/328
- [x] https://github.com/wireapp/wire-ios-data-model/pull/218


Wire-iOS/Sources/UserInterface/Registration/ProfilePictureStepViewController.m
```diff
@@ -295,10 +295,10 @@ - (void)setPictureImageData:(NSData *)imageData
         [[ZMUserSession sharedSession] checkNetworkAndFlashIndicatorIfNecessary];
         
         if ([self.editabledUser isKindOfClass:[ZMIncompleteRegistrationUser class]]) {
-            self.editabledUser.originalProfileImageData = imageData;
+            [[ZMUserSession sharedSession].profileUpdate updateImageWithImageData:imageData];
```
typfel: It still makes sense to have a `if` `else` here?

vytis: Yeah probably not, will update

vytis:
@typfel updated. Or should I wrap it in `enqueueChanges` instead?

typfel:
@vytis if it modifies a core data object then yes.

vytis:
Updated

PR #918, opened 2017-04-06
Observe active player state in the list cells

mikeger:
- Added the new notification on when the player state is changed.


daehn:
@mikeger how was this done before? Or was it not done before?

mikeger:
@daehn there is a KVO on player, which we wanted to avoid. Otherwise the list was reloaded when controller appeared before.

PR #922, opened 2017-04-10
"Started a conversation" status

mikeger:
- Was missing before.


mikeger:
@daehn there are tests (actually only one for this particular case, but still)

daehn:
@mikeger I meant UI tests for the status line (to check which parts are emphasised and which are not), but I am unsure if we have UI tests for the conversation list cell, yet.

mikeger:
Yeah, having the tests for the whole cells is also doable.

PR #923, opened 2017-04-10
Set -setUiStartsAudio: on the media manager on start up

daehn:
# What's in this PR?

* See https://github.com/wireapp/wire-ios/pull/825.
* This flag should be set to `YES` when CallKit is enabled, then AVS won't start the audio before the audio session is active.


mikeger:
@daehn would it work with open-source AVS (from Cartfile)?

typfel:
Will this work correctly on simulator? I think callkit is disabled there but via `#if`

typfel:
Another option is this set this flag in the init method of `ZMCallKitDelegate`

daehn:
@typfel interesting, will double check that. And the open source version as well @mikeger.

daehn:
@mikeger it's in the open source version as well.

daehn:
I'll move this call to sync-engine.

daehn:
Closed in favour of https://github.com/wireapp/wire-ios-sync-engine/pull/337

PR #926, opened 2017-04-10
Fix audio message temporary urls for users with non latin names

daehn:
# What's in this PR?

* Audio messages were temporarily stored in a file with a name created from the users name, this logic didn't work for users with emoji in their name (or other non-latin characters I assume). This PR fixes this issue and uses the `remoteIdentifier` instead of the users name.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
+++ b/Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
@@ -23,12 +23,12 @@ extension NSString {
     
     fileprivate static let dateFormatter : DateFormatter = {
         let formatter = DateFormatter()
-        formatter.dateFormat = "yyyy MM dd hh.mm.ss"
+        formatter.dateFormat = "yyyy-MM-dd-hh.mm.ss"
         return formatter
     }()
-    
+
     static func filenameForSelfUser() -> NSString {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
```diff
@@ -23,12 +23,12 @@ extension NSString {
...
+        formatter.dateFormat = "yyyy-MM-dd-hh.mm.ss"
         return formatter
     }()
-    
+
     static func filenameForSelfUser() -> NSString {
-        let filename = "\(ZMUser.selfUser().name!) \(dateFormatter.string(from: Date()))"
-        return filename.replacingOccurrences(of: " ", with: "-") as NSString
+        let identifier = ZMUser.selfUser().remoteIdentifier?.transportString() ?? ""
+        return "\(identifier)-\(dateFormatter.string(from: Date()))" as NSString
```
mikeger: I think this would "break" the audio file name for the media player. Would be better to preserve the user name in some way. You see the file name when you start playing the audio file and then scroll it away.

daehn: Then we might need to normalise it / convert it to latin, but better than seeing your `remoteIdentifier` I guess.

daehn:
@mikeger updated, now the user name is normalized.

PR #927, opened 2017-04-10
Enable some logs by default when on internal

marcoconti83:
We want to force internal users to enable the right logs for debug reports

### DIFF
--- a/WireExtensionComponents/Utilities/DeveloperMenuState+Debugging.swift
+++ b/WireExtensionComponents/Utilities/DeveloperMenuState+Debugging.swift
@@ -1,9 +1,19 @@
 //
-//  DeveloperMenuState+Debugging.swift
-//  Wire-iOS
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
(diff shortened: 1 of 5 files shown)


WireExtensionComponents/Utilities/DeveloperMenuState+Debugging.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Created by Marco Conti on 10/04/2017.
+//  Copyright © 2017 Zeta Project Germany GmbH. All rights reserved.
+//
+
+import Foundation
+import WireSystem
+
+extension DeveloperMenuState {
+    
+    public static func prepareForDebugging() {
```
mikeger: I think it would be easier to have one swift method that sets the logs if necessary. Right now there is one objc and 2 swift methods. It can be puzzling.

marcoconti83:
@mikeger good point. I removed the intermediate Obj-c one.

marcoconti83:
Please have a look again

PR #931, opened 2017-04-10
Check for correct voice channel state and enqueue changes

daehn:
# What's in this PR?

* We were checking for the wrong voice channel state. 
* Warp the call to join the ongoing call in an `enqueueChanges` block.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
@@ -191,7 +191,7 @@ static const NSTimeInterval OverscrollRatio = 2.5;
         [self toggleMediaPlayer];
     }
     else if (self.conversation.voiceChannel.state == VoiceChannelV2StateIncomingCallInactive) {
-        [self.conversation startAudioCallWithCompletionHandler:nil];
+        [self.conversation acceptIncomingCall];
     }
 }
     
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
@@ -190,10 +190,10 @@ static const NSTimeInterval OverscrollRatio = 2.5;
         mediaPlaybackManager.activeMediaPlayer.sourceMessage.conversation == self.conversation) {
         [self toggleMediaPlayer];
     }
-    else {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
```diff
@@ -190,10 +190,10 @@ - (void)onRightAccessorySelected:(UIButton *)sender
         mediaPlaybackManager.activeMediaPlayer.sourceMessage.conversation == self.conversation) {
         [self toggleMediaPlayer];
     }
-    else {
-        if (self.conversation.voiceChannel.state == VoiceChannelV2StateIncomingCall) {
+    else if (self.conversation.voiceChannel.state == VoiceChannelV2StateIncomingCallInactive) {
+        [ZMUserSession.sharedSession enqueueChanges:^{
             (void)[self.conversation.voiceChannel joinWithVideo:NO userSession:[ZMUserSession sharedSession]];
```
typfel: Hmm don't we have some UI wrapper which checks permissions before joining a call

daehn: Good point, will check.

daehn: Yes, will update.

daehn: @typfel updated!

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
```diff
@@ -190,10 +190,8 @@ - (void)onRightAccessorySelected:(UIButton *)sender
         mediaPlaybackManager.activeMediaPlayer.sourceMessage.conversation == self.conversation) {
         [self toggleMediaPlayer];
     }
-    else {
-        if (self.conversation.voiceChannel.state == VoiceChannelV2StateIncomingCall) {
-            (void)[self.conversation.voiceChannel joinWithVideo:NO userSession:[ZMUserSession sharedSession]];
-        }
+    else if (self.conversation.voiceChannel.state == VoiceChannelV2StateIncomingCallInactive) {
+        [self.conversation startAudioCallWithCompletionHandler:nil];
```
typfel: maybe we should use `acceptIncomingCall` then it also works for video calls

PR #937, opened 2017-04-13
Self user profile

mikeger:
- First page of settings is transformed to self user profile that shows the user's name, image and handle.
- Settings can be accessed from the profile.
- Devices are also moved to the self profile.

Some snapshots changed while the like icon was adjusted.
Also the conversation list badge is updated.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListAccessoryView.swift
```diff
@@ -149,13 +151,15 @@ final internal class ConversationListAccessoryView: UIView {
...
             
-        case .missedCall:
-            self.badgeView.backgroundColor = ZMAccentColor.vividRed.color
-            
         case .typing:
             self.badgeView.isHidden = true
             self.typingView.isHidden = false
             self.typingView.image = UIImage(for: .pencil, fontSize: 12.0, color: .white)
+            
+        case .unreadMessages(_):
```
daehn: 🎉

Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+@objc internal class ProfileView: UIView {
+    public let imageView = UserImageView(size: .big)
+    public let nameLabel = UILabel()
+    public let handleLabel = UILabel()
+    
+    init(user: ZMUser) {
+        super.init(frame: .zero)
+        self.imageView.user = user
+        self.nameLabel.text = user.displayName
+        self.handleLabel.text = "@" + user.handle
```
daehn: `handle` can be nil for old users. I think we should hide the `handleLabel`in that case.

mikeger:
@daehn updated

mikeger:
@daehn I would add some tests for the top bar

mikeger:
@daehn added tests

PR #943, opened 2017-04-18
You called vs someone called in German (and others)

mikeger:
- Changed the way the cell is localized for cases:
  - You vs someone called.
  - You vs someone renamed the conversation.

screen shot 2017-04-19 at 11 25 34

### DIFF
--- a/Wire-iOS/Resources/de.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/de.lproj/Localizable.strings
@@ -252,9 +252,11 @@
 "content.system.this_device" = "dieses Gerät";
 
 "content.system.call.you" = "Du";
-"content.system.call.called" = "%@ hast angerufen";
+"content.system.call.called" = "%@ hat angerufen";
+"content.system.call.called-you" = "%@ hast angerufen";
 
 "content.system.renamed_conv.title" = "%@ hat das Gespräch umbenannt";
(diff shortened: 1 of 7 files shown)


Wire-iOS/Resources/de.lproj/Localizable.strings
```diff
@@ -252,9 +252,11 @@
 "content.system.this_device" = "dieses Gerät";
 
 "content.system.call.you" = "Du";
-"content.system.call.called" = "%@ hast angerufen";
+"content.system.call.called" = "%@ hat angerufen";
+"content.system.call.called-you" = "%@ hast angerufen";
 
 "content.system.renamed_conv.title" = "%@ hat das Gespräch umbenannt";
+"content.system.renamed_conv.title-you" = "%@ hast das Gespräch umbenannt";
```
typfel: For consistency the `-` should be a `_`

mikeger: I wanted to have it separated from the string name, so I used the "new" character. Let's see if the localization people would understand it. The logic is done in the way that if you ask for the 2nd person string it would try to get one with the postfix, i.e. "name-you" but if it's not there it would go for "name". If we would use the regular char then the localization can get confusing, for example you can use "name.you" and "name" without an intention of having it automatically selected, which can lead to the bug.

typfel: Ok

PR #944, opened 2017-04-19
Re-add initial sync observer to conversation list controller

daehn:
# What's in this PR?

* The `ConversationListViewController` somehow no longer was a `ZMInitialSyncCompletionObserver`, which it needs to be in order to only show the username takeover screen once the initial sync is done. This is due to the fact that the user might already have set a username on another device and we need to wait for the user to be refetched.


mikeger:
@daehn I think I broke it with the new list, or even earlier.

PR #948, opened 2017-04-20
Message composition and drafts

daehn:
# What's in this PR?

* Add logic to compose message drafts.
* This adds a (small) CoreData stack with only one context to the UI which only has one entity called `MessageDraft`, using CoreData to store drafts might seem a bit excessive at first, but comes with a couple of nice advantages. One of those advantages is making use of an `NSFetchedResultsController` which is hooked up to a table view in order to keep the stored drafts in the list up to date (see `CoreDataTableViewController`, `MessageDraftStorage`, `MessageDraft` and `DraftListViewController`).
* Some `TextMessageCell` tests had to be rerecorded as the like icon seems to have 

### This PR depends on:
- [x] A feature flag has to be introduced to hide / show the new UI.
- [x] The `ComposeEntryViewController` animations should use auto layout instead of manual layout to work correctly with RTL languages (and the button labels have to added here as well).


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -164,11 +167,8 @@ extension ConversationContentViewController: UIAdaptivePresentationControllerDel
         let conversations = SessionObjectCache.shared().allConversations.shareableConversations(excluding: message.conversation!)
         
         let shareViewController: ShareViewController<ZMConversation, ZMMessage> = ShareViewController(shareable: message as! ZMMessage, destinations: conversations)
-        
-        let displayInPopover: Bool = self.traitCollection.horizontalSizeClass == .regular &&
-                                     self.traitCollection.horizontalSizeClass == .regular
```
mikeger: surprised-koala
Just to make sure 🥇

daehn: The check just moved down a line, it was checking the same thing twice so I put it into the `if` directly.

mikeger: Yeah, I mostly reacted on the old `if` that was the victim of the C-c C-v code.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+Private.h
```diff
@@ -31,6 +32,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) ConversationListContentController *listContentController;
 @property (nonatomic, weak, readonly) id<UserProfile> userProfile;
 @property (nonatomic) ConversationListTopBar *topBar;
+@property (nonatomic, nullable) SpaceSelectorView *spacesView;
```
daehn: @mikeger I think this happened during the merge and `spacesView` and the flag should no longer be here, right?

Wire-iOS/Sources/UserInterface/ConversationList/ComposeEntryViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+}
+
+
+@objc class ComposeEntryViewController: UIViewController, UIViewControllerTransitioningDelegate {
+
+    fileprivate let plusButton = IconButton()
+    fileprivate let plusButtonContainer = UIView()
+    fileprivate let contactsButton = IconButton()
+    fileprivate let composeButton = IconButton.iconButtonCircularLight()
+    fileprivate let dimmView = UIView()
```
mikeger: Should be `dim` or `dimming` ?

Wire-iOS/Sources/UserInterface/ConversationList/ComposeEntryViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private func setupViews() {
+        dimmView.backgroundColor = dimmViewColor
+        dimmView.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(dismissController)))
+        contactsButton.backgroundColor = .white
+        contactsButton.circular = true
+        contactsButton.setIcon(.person, with: .tiny, for: .normal)
+        contactsButton.setIconColor(UIColor(for: .strongBlue), for: .normal)
```
mikeger: It is blue or should be an accent color?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+Private.h
mikeger: 💯 this is not necessary to have any more.

Wire-iOS/Sources/UserInterface/ConversationList/ComposeEntryViewController.swift
daehn: Good catch!

Wire-iOS/Sources/UserInterface/DraftListViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+        super.init(fetchedResultsController: persistence.resultsController)
+        configureCell = { (cell, draft) in
+            cell.configure(with: draft)
+        }
+
+        onCellSelection = { (_, draft) in
+            self.showDraft(draft)
+        }
+
+        onDelete = { (_, draft) in
```
mikeger: I personally find un-cool to have the blocks instead of the open methods or delegate, mostly because they implementation can be changed on the runtime. With `override` and `final` you can also avoid the dynamic dispatch.
Also visually it looks not that elegant to assign the blocks on init. Also it is potentially hard to find where the blocks are actually assigned, and which implementation is called now.

Not asking to rewrite it here, just a comment.

Wire-iOS/Sources/UserInterface/DraftMessageCell.swift
```diff
@@ -0,0 +1,44 @@
...
+
+import Foundation
+
+
+final class DraftMessageCell: UITableViewCell {
+
+    override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
+        super.init(style: .subtitle, reuseIdentifier: reuseIdentifier)
+        textLabel?.font = FontSpec(.normal, .light).font!
+        textLabel?.textColor = ColorScheme.default().color(withName: ColorSchemeColorTextForeground)
```
mikeger: The new UI's: should they follow the dark/light theme switch?

daehn: I assumed so for now.

daehn: But will have to double check in the cell with the `backgroundColor`.

Wire-iOS/Sources/UserInterface/ConversationList/ComposeEntryViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+        contactsButton.addTarget(self, action: #selector(contactsTapped), for: .touchUpInside)
+        composeButton.backgroundColor = UIColor(for: .strongBlue)
+        composeButton.addTarget(self, action: #selector(draftsTapped), for: .touchUpInside)
+        composeButton.setIcon(.compose, with: .tiny, for: .normal)
+        plusButton.setIcon(.plus, with: .tiny, for: .normal)
+        plusButton.setIconColor(.white, for: .normal)
+        plusButton.addTarget(self, action: #selector(dismissController), for: .touchUpInside)
+        plusButtonContainer.addSubview(plusButton)
+        [dimmView, plusButtonContainer, contactsButton, composeButton].forEach(view.addSubview)
```
mikeger: Should the buttons here have the accessibility ids?

daehn: Yes, definitely. Will add.

Wire-iOS/Sources/UserInterface/MessageDraftStorage.swift
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+        super.init()
+    }
+
+    private static func createDirectoryIfNeeded(at url: URL) throws {
+        if !FileManager.default.fileExists(atPath: url.path) {
+            try FileManager.default.createDirectory(at: url, withIntermediateDirectories: true, attributes: nil)
+        }
+        try url.wr_excludeFromBackup()
```
typfel: 💫

mikeger:
@daehn did you check about the dark/light theme?

daehn:
@mikeger I fixed the cell, was using the wrong background.

PR #950, opened 2017-04-20
Exclude whole library dir from backup

daehn:
# What's in this PR?

* Try to exclude the whole Library directory from the iCloud backup. We currently exclude all relevant folders manually.

### DIFF
--- a/Wire-iOS/Sources/Managers/FileBackupExcluder.swift
+++ b/Wire-iOS/Sources/Managers/FileBackupExcluder.swift
@@ -34,7 +34,10 @@ public extension URL {
 
 final internal class FileBackupExcluder: NSObject {
     typealias FileInDirectory = (NSFileManager.SearchPathDirectory, String)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Managers/FileBackupExcluder.swift
```diff
@@ -34,7 +34,10 @@ public extension URL {
 
 final internal class FileBackupExcluder: NSObject {
     typealias FileInDirectory = (NSFileManager.SearchPathDirectory, String)
-    private static let filesToExclude: [FileInDirectory] = [(.libraryDirectory, "Preferences/com.apple.EmojiCache.plist")]
+    private static let filesToExclude: [FileInDirectory] = [
+        (.libraryDirectory, "Preferences/com.apple.EmojiCache.plist"),
+        (.libraryDirectory, ".")
```
mikeger: I don't think the current implementation is recursive, but there is also no clear documentation if it should be. Let's try this for now and see how it would work.

PR #952, opened 2017-04-21
Fix for spaces not working

mikeger:
- The update notification was throttled to avoid recreation of the spaces view.
- Spaces view was looking on the wrong instances of Spaces object.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/Space.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/Space.swift
@@ -118,35 +118,38 @@ internal class Space: NSObject {
             let workspaceName = factory.property(.workspaceName).rawValue() as? String,
             !workspaceName.isEmpty {
             
-            let privateSpace: Space = {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/Space.swift
```diff
@@ -118,35 +118,38 @@ internal class Space: NSObject {
...
             
-            let privateSpace: Space = {
-                let selfUser = ZMUser.selfUser()
-                
-                var image: UIImage? = .none
-                
-                if let imageData = selfUser?.imageMediumData {
-                    image = UIImage(from: imageData, withMaxSize: 100)
-                }
+            if spaces.isEmpty || spaces[1].name != workspaceName {
```
daehn: Are we always guaranteed to have 2 spaces if the array is non-empty?

mikeger: For the time being yes. The function is not for production, too. I would add the check just to be sure.

PR #953, opened 2017-04-21
Drafts design review

daehn:
# What's in this PR?

* Adds the first changes requested by design.
* The custom view controller transition for the 2 buttons grew quite big but is very easy to fine tune that way. However we can think of using Cartography constraint groups instead of having arrays for the different states.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
@@ -20,6 +20,7 @@
 import Foundation
 import Cartography
 
+
 protocol MessageComposeViewControllerDelegate: class {
     func composeViewController(_ controller: MessageComposeViewController, wantsToSendDraft: MessageDraft)
     func composeViewControllerWantsToDismiss(_ controller: MessageComposeViewController)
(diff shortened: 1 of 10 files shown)


Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
```diff
@@ -66,39 +64,48 @@ final class MessageComposeViewController: UIViewController {
...
+        navigationItem.leftBarButtonItem = splitViewController?.displayModeButtonItem
+        navigationItem.leftItemsSupplementBackButton = true
+
+        subjectTextField.delegate = self
+        subjectTextField.textColor = color(ColorSchemeColorTextForeground)
+        subjectTextField.tintColor = .accent()
+        subjectTextField.textAlignment = .center
+        let placeholder = "compose.drafts.compose.subject.placeholder".localized.uppercased()
+        subjectTextField.attributedPlaceholder = placeholder && color(ColorSchemeColorSeparator) && FontSpec(.normal, .none).font!
+        subjectTextField.bounds = CGRect(x: 0, y: 0, width: 200, height: 44)
```
mikeger: May be would make sense to ask the label to lay out to match the text length?

daehn: Yeah, but I guess as it's editable we will need to update it constantly, will need to check if `UITextField` exposes its contentSize though, then it might be possible.

PR #961, opened 2017-04-24
Enable group calling V3

basine:


vytis:
Tests failed for some reason...

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
```diff
@@ -205,7 +210,7 @@ - (void)updateActiveCallConversation
 - (void)callCenterDidChangeVoiceChannelState:(VoiceChannelV2State)voiceChannelState conversation:(ZMConversation *)conversation callingProtocol:(enum CallingProtocol)callingProtocol
 {
     DDLogVoice(@"SE: Voice channel state change to %d (%@)", voiceChannelState, StringFromVoiceChannelV2State(voiceChannelState));
-    [self updateOverlays];
+    [self updateOverlaysWithOutgoingCallInConversation: voiceChannelState == VoiceChannelV2StateOutgoingCall ? conversation : nil];
```
typfel: Is this some kind work-around?

mikeger:
@basine as always, some mock in tests expired 💥

mikeger:
I guess we better switch the old tests to use the real CoreData

Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
basine: Yeah, for outgoing calls, AVS does not return the outgoing call (yet) when we notify the UI - This is probably due to threading issues. Looking for other solutions for this, too. :-/

PR #965, opened 2017-04-25
Drafts design review (3)

daehn:
# What's in this PR?

* Adjust fonts.
* Add empty state for `DraftsListViewController`.
* Moved from the assigned closures to overridden functions in `CoreDataTableViewController` / `DraftsListViewController`.
* Adjusted the navigation bar in drafts.
* Further animation tuning.


Wire-iOS/Sources/UserInterface/CoreDataTableViewController.swift
```diff
@@ -18,22 +18,24 @@
 
 
 import Foundation
+import Cartography
 
 
-class CoreDataTableViewController<Model: NSFetchRequestResult, Cell: UITableViewCell>: UITableViewController, NSFetchedResultsControllerDelegate {
+class CoreDataTableViewController<Model: NSFetchRequestResult, Cell: UITableViewCell>: UIViewController, UITableViewDelegate, UITableViewDataSource, NSFetchedResultsControllerDelegate {
```
mikeger: That is always the case whenever I use `UITableViewController`

daehn: Yes, but it's no longer a `UITableViewController` but instead is a `UIViewController` now so it doesn't have a `UITableView` as its `view`. This is done so we can easily hide the `tableView` and display other subviews.

mikeger: Right, that is what I mean, whenever I use `UITableViewController` next step is to stop using it :)

PR #968, opened 2017-04-25
Strip leading whitespace from draft messages

daehn:
# What's in this PR?

* It shouldn't be possible to send whitespace only message drafts.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
@@ -174,9 +174,10 @@ final class MessageComposeViewController: UIViewController {
         if let draft = draft {
             persistence.enqueue(block: {
                 if self.hasDraftContent {
-                    guard draft.subject != self.subjectTextField.text || draft.message != self.messageTextView.text else { return }
-                    draft.subject = self.subjectTextField.text
-                    draft.message = self.messageTextView.text
+                    let (subject, message) = (self.subjectTextField.text, self.messageTextView.text)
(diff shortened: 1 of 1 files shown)


daehn:
Alternatively we could also strip trailing and leading whitespace?

Wire-iOS/Sources/UserInterface/MessageComposeViewController.swift
```diff
@@ -243,3 +261,12 @@ extension MessageComposeViewController: UITextFieldDelegate {
     }
 
 }
+
+
+extension String {
+
+    var onlyWhitespace: Bool {
+        return CharacterSet(charactersIn: self).isSubset(of: .whitespaces)
```
mikeger: This would be slightly suboptimal, since it would always have to create the set from all the characters, I think it would be faster to look for the first occurrence of the character from the inverted `.whitespaces` set, then it would fail quickly on the first char that is not a whitespace.
Imagine you are typing a really large text, and for each key stroke it would recreate the character set from all the chars.

daehn: Yeah good point, alternatively we can just check if the replacement text is a character in `shouldReplaceCharactersInRange` (and the location is 0).

daehn: But I double checked and for messages send not using the drafts UI we just trimm leading and trailing white spaces and newlines, so I will do that as well.

daehn:
@mikeger updated, also fixed the interactive swipe gesture which always breaks because we highjack the `leftNavigationItem` and use it as a `backButtonItem` replacement.

PR #974, opened 2017-04-27
Present the debug alert on top of the topmost controller

mikeger:
- Fixes issue with sending the logs on iPad.


mikeger:
@daehn I think it would make sense to merge if you approve the PR 👍

PR #975, opened 2017-04-27
Developer option to share database

typfel:
Many times it's could be useful to look at db when debugging. This PR adds a option to the develop settings to for easily exporting the DB via a share sheet.


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsShareDatabaseCellDescriptor.swift
```diff
@@ -0,0 +1,56 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import ZipArchive
+
+class DocumentDelegate : NSObject, UIDocumentInteractionControllerDelegate {
+    
+    func documentInteractionControllerViewControllerForPreview(_ controller: UIDocumentInteractionController) -> UIViewController {
+        return UIApplication.shared.keyWindow!.rootViewController!
```
mikeger: Let's use `UIApplication.shared.wr_topmostController(onlyFullScreen: false)`, then it would work on iPad

PR #979, opened 2017-04-28
Do not show incoming calls for silenced conversations

basine:
- [x] needs release of sync engine with https://github.com/wireapp/wire-ios-sync-engine/pull/357


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
```diff
@@ -172,14 +172,19 @@ - (VoiceChannelOverlayController *)voiceChannelOverlayControllerForConversation:
...
+              (conversation.voiceChannel.state == VoiceChannelV2StateIncomingCall ||
+               conversation.voiceChannel.state == VoiceChannelV2StateIncomingCallDegraded);
     }];
     
     if (incomingCallConversations.count > 0) {
         return incomingCallConversations.lastObject;
     }
     else if (self.activeCallConversation) {
+        VoiceChannelV2State state = self.activeCallConversation.voiceChannel.state;
+        if (self.activeCallConversation.isSilenced && (state == VoiceChannelV2StateIncomingCallDegraded || state == VoiceChannelV2StateIncomingCall)) {
```
typfel: Can you explain this statement? Why should `silenced` flag affect active calls?

basine: It only affects it if the callState is IncomingCall or incomingDegraded which I guess should not happen.

typfel: yes, If I remember correctly an active calls are only outgoing or calls you've connected to. So we can remove this check then?

PR #987, opened 2017-05-03
Forbid deletion of unsent image

mikeger:
- It was possible to delete the not completely sent image from the full screen view toolbar.
- This can mess up the SE logic.


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -61,11 +61,13 @@ internal final class ConversationImagesViewController: UIViewController {
     fileprivate var imageMessages: [ZMConversationMessage] = []
     internal var currentMessage: ZMConversationMessage {
         didSet {
+            self.updateButtonsForMessage()
             self.createNavigationTitle()
         }
     }
     internal var pageViewController: UIPageViewController = UIPageViewController(transitionStyle:.scroll, navigationOrientation:.horizontal, options: [:])
     internal var buttonsBar: InputBarButtonsView!
+    internal var deleteButton: IconButton!
```
typfel: Maybe create the delete button here then we can remove the `!`

mikeger:
@typfel updated

PR #988, opened 2017-05-03
Pin share view controller send bar to keyboard

daehn:
# What's in this PR?

* Pin `ShareViewController ` send bar to keyboard.
* Adjust return key type.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
@@ -167,15 +167,16 @@ extension ConversationContentViewController: UIAdaptivePresentationControllerDel
         let conversations = SessionObjectCache.shared().allConversations.shareableConversations(excluding: message.conversation!)
         
         let shareViewController: ShareViewController<ZMConversation, ZMMessage> = ShareViewController(shareable: message as! ZMMessage, destinations: conversations)
+        let keyboardAvoiding = KeyboardAvoidingViewController(viewController: shareViewController)!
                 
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -167,15 +167,16 @@ extension ConversationContentViewController: UIAdaptivePresentationControllerDel
...
         let shareViewController: ShareViewController<ZMConversation, ZMMessage> = ShareViewController(shareable: message as! ZMMessage, destinations: conversations)
+        let keyboardAvoiding = KeyboardAvoidingViewController(viewController: shareViewController)!
                 
         if traitCollection.horizontalSizeClass == .regular {
             shareViewController.showPreview = false
         }
         
-        shareViewController.preferredContentSize = CGSize(width: 320, height: 568)
-        shareViewController.modalPresentationStyle = .popover
+        keyboardAvoiding.preferredContentSize = CGSize(width: 320, height: 568)
```
vytis: Will it work fine for smaller devices and/or ipad?

daehn: On smaller devices it is presented as a modal, on iPad as a foresheet, the only thing that changed is that it now moves the content up when the keyboard appears. That could get a bit cramped on very small devices, which is why we hide the autocorrection bar (which should be about the same size as the sending bar which is moved up now).

vytis: Sounds reasonable:)

PR #989, opened 2017-05-03
set isTyping to no when entering background or sending a message

basine:


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -175,6 +176,14 @@ - (void)dealloc
     }
 }
 
+- (void)didEnterBackground:(NSNotification *)notification
+{
+    NOT_USED(notification);
+    if([self.conversation shouldDisplayIsTyping] && self.inputBar.textView.text.length > 0) {
```
mikeger: I think that `shouldDisplayIsTyping` is a leftover from back in the day when we used to support typing only on 1-1, so it is basically a build-time feature flag. As we always show it, `shouldDisplayIsTyping` can be removed all together with `IsTypingInGroupConversationsEnabled` and `IsTypingEnabled`.

basine: OK, I removed it

PR #996, opened 2017-05-05
Only track data base size when logged in

daehn:
# What's in this PR?

* Only track data base size when logged in.
* Bump components.


typfel:
So there's another place where we call `trackLegacyMessageCount`?

daehn:
@typfel yes, when the app comes into the foreground we check if we're logged in and call it form there as well.

PR #997, opened 2017-05-05
Fix constraints when hiding the message toolbox

typfel:
Lower priority on collapsing constraint so it will animate properly. Remove top & bottom constraints for the `labelClipView` so that the message toolbox can collapse without breaking any constraints.


typfel:
Not working for batched calls

typfel:
@daehn updated

PR #1001, opened 2017-05-05
Adjust empty label visuals in drafts list

daehn:
# What's in this PR?

* Adjust empty label visuals in drafts list.
* Fix status bar style change timing when dismissing `DraftsRootViewController`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/DraftsRootViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/DraftsRootViewController.swift
@@ -56,9 +56,9 @@ final class DraftsRootViewController: UISplitViewController {
         UIApplication.shared.wr_updateStatusBarForCurrentControllerAnimated(true)
     }
 
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/DraftsRootViewController.swift
```diff
@@ -56,9 +56,9 @@ final class DraftsRootViewController: UISplitViewController {
         UIApplication.shared.wr_updateStatusBarForCurrentControllerAnimated(true)
     }
 
-    override func viewDidDisappear(_ animated: Bool) {
-        super.viewDidDisappear(animated)
-        UIApplication.shared.wr_updateStatusBarForCurrentControllerAnimated(false)
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        UIApplication.shared.setStatusBarStyle(.lightContent, animated: animated)
```
daehn: Not the cleanest way to hard code this here, but we always dismiss to the conversation list which always has a light status bar. I think we should overthink our status bar approach at some point and maybe move to `UIViewControllerBasedStatusBarAppearance`.

PR #1011, opened 2017-05-10
Reuse existing VoiceChannelOverlayController

basine:
When we receive a call while we are already in a call and reject the new incoming call we are currently creating a new `VoiceChannelOverlayController` for the already ongoing call instead of reusing the old one. Furthermore we never deallocate the previously added `VoiceChannelOverlayController` except the last one we created. This leads to crashes when we receive another call in one of the conversations already observed by the left-over controllers.

This PR includes two changes: 
(1) It removes the previous childViewController
(2) It adds the voiceChannelParticipantController on viewDidLoad when the call is in callState connected

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
+++ b/Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
@@ -144,7 +144,9 @@
         
         [self transitionFromViewController:fromController toViewController:toController duration:0.35 options:UIViewAnimationOptionTransitionCrossDissolve animations:^{
             [self.primaryVoiceChannelOverlay.view autoPinEdgesToSuperviewEdgesWithInsets:UIEdgeInsetsZero];
-        } completion:nil];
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelController.m
```diff
@@ -134,14 +134,17 @@ - (void)transitionFromVoiceChannelOverlayController:(VoiceChannelOverlayControll
             fromController.blurEffectView.effect = nil;
             fromController.overlayView.alpha = 0;
         } completion:^(BOOL finished) {
-            [fromController.view removeFromSuperview];
-            [fromController removeFromParentViewController];
+            // Remove all view controllers also those from interrupting calls
+            for (VoiceChannelOverlayController *controller in self.childViewControllers) {
```
typfel: Something must have gone wrong earlier if we have two VoiceChannelOverlayControllers in view hierarchy the at this point.

PR #1013, opened 2017-05-11
Increase conversation list cells height to 64 points

mikeger:
- Looking better on crowded screen.


daehn:
@mikeger no tests affected by this change?

mikeger:
@daehn no, but let's give CircleCI some minutes to confirm it.

PR #1014, opened 2017-05-11
Show  label in the upper left corner when on a v2 call

typfel:
https://wearezeta.atlassian.net/browse/ZIOS-8556

depends on:
 - [x] https://github.com/wireapp/wire-ios-sync-engine/pull/378


daehn:
So we are prefixing the name and adding a label?

typfel:
@daehn yes, if you use callkit you might never open the app.

PR #1027, opened 2017-05-23
Removed SessionObjectCache, use new conv lists

mikeger:
- New API for conversation lists in teams.
- Removed SessionObjectCache, since the conversation lists are already cached in the data model.
- Added a convenience API `ZMUser.activeTeam() -> Team?` assuming for the first release we would have one active team.


Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.m
```diff
@@ -75,11 +74,11 @@ - (void)viewDidLoad
     self.tableView.delegate = self;
     self.tableView.dataSource = self;
     
-    ZMConversationList *pendingConnectionsList = [[SessionObjectCache sharedCache] pendingConnectionRequests];
+    ZMConversationList *pendingConnectionsList = [ZMConversationList pendingConnectionConversationsInUserSession:[ZMUserSession sharedSession] team:nil];
```
vytis: So maybe it's worth removing `team` parameter from `pendingConnectionConversations` altogether? Could anything go wrong if active team is passed there instead of nil?

daehn: It should always return an empty list when a team is passed in

vytis: It seems like an easy way to introduce bugs by just writing similar code as in all other conversation lists...

typfel:
@mikeger @vytis @daehn  so should we merge this? It's blocking other UI development

PR #1029, opened 2017-05-24
Added new observers support

mikeger:
- Observe list of teams.
- Observe individual teams.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBar.swift
```diff
@@ -102,6 +107,21 @@ final class ConversationListTopBar: TopBar {
     }
 }
 
+extension ConversationListTopBar: ZMUserObserver {
+    public func userDidChange(_ changeInfo: UserChangeInfo) {
+        if ZMUser.selfUser().teams.count > 0 {
```
daehn: We could check if the teams changed, should be in the `changeInfo`.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ConversationListViewModel+Private.h
```diff
@@ -0,0 +1,12 @@
+//
+//  ConversationListViewModel+Private.h
+//  Wire-iOS
+//
+//  Created by Mihail Gerasimenko on 5/24/17.
```
daehn: Missing copyright header

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ConversationListViewModel+Teams.swift
```diff
@@ -17,20 +17,32 @@
...
+    
+    func createTeamsObservers() {
+        self.teamsObservers = ZMUser.selfUser().teams.map { TeamChangeInfo.add(observer: self, for: $0) }
+    }
+}
+
+extension ConversationListViewModel: ZMUserObserver {
+    public func userDidChange(_ note: UserChangeInfo) {
+        if note.teamsChanged {
+            createTeamsObservers()
```
daehn: Do we need to remove the previous observers or are they removed automagically when they get released?

Wire-iOS/Sources/UserInterface/ConversationList/TeamSelectorView.swift
```diff
@@ -70,22 +70,27 @@ internal class LineView: UIView {
...
-                    self.teams.filter { $0.remoteIdentifier != selectedTeam.remoteIdentifier }.forEach { $0.isActive = false }
-                }
-                else {
-                    self.teams.forEach { $0.isActive = false }
-                }
                 
-                selectedTeam?.isActive = true
+                ZMUserSession.shared()?.performChanges {
+                    if let selectedTeam = selectedTeam {
+                        self.teams.filter { $0.remoteIdentifier != selectedTeam.remoteIdentifier }.forEach { $0.isActive = false }
```
daehn: Is this fork necessary here? Looks like it should also work with a `nil` selected team, e.g. `$0.remoteIdentifier != selectedTeam?.remoteIdentifier`

Wire-iOS/Sources/UserInterface/ConversationList/TeamSelectorView.swift
```diff
@@ -124,15 +129,16 @@ final internal class TeamSelectorView: UIView {
     
     init() {
         super.init(frame: .zero)
-        self.update(with: Array(ZMUser.selfUser()?.teams ?? Set()))
         self.clipsToBounds = true
+        
+        selfUserObserverToken = UserChangeInfo.add(observer: self, forBareUser: ZMUser.selfUser())
```
mikeger: This view shows the list of teams, so it needs to know when the teams relation is changing on self user, therefore I thought to use here the ZMUser observer. The smaller views that are inserted into this one observe the teams (See `TeamView.swift`).

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ConversationListViewModel+Teams.swift
mikeger: They should be removed automatically when token is deallocated, but thanks to @basine I changed the logic there.

Wire-iOS/Sources/UserInterface/ConversationList/TeamSelectorView.swift
basine: ah ok, do we need to animate moves? do you need a setChangeInfo or do you just reload?

mikeger: For the moment it's not necessary to animate the change, honestly saying I think it would not be necessary looking forward too.

mikeger:
@daehn @basine updated.

PR #1031, opened 2017-05-24
Phase 1

mikeger:
- Created for tracking the general overview.


Wire-iOS Tests/ConversationListTopBarTests.swift
```diff
@@ -143,62 +152,61 @@ class ConversationListTopBarTests: ZMSnapshotTestCase {
...
-        // WHEN
-        _ = sut.snapshotView()
-        
-        // AND WHEN
-        sut.setShowSpaces(to: false)
-        
-        // THEN
-        self.verify(view: sut.snapshotView())
-    }
+//    func testThatItRendersSpacesBarAfterDefaultBar() {
```
vytis: Commented out code

Wire-iOS/Sources/UserInterface/StartUI/Common/InputField/PeopleInputController.m
```diff
@@ -128,14 +141,34 @@ - (BOOL)userDidConfirmInput
...
-        NSArray *users = [tokens valueForKeyPath:@"@distinctUnionOfObjects.representedObject"];
-        [self.selectionDelegate peopleInputController:self changedPresentedDirectoryResultsTo:[NSSet setWithArray:users]];
-    }
+    NSArray *users = [tokens valueForKeyPath:@"@distinctUnionOfObjects.representedObject"];
+    [self.userSelection replace:users];
+    
+//    if ([self.selectionDelegate respondsToSelector:@selector(peopleInputController:changedPresentedDirectoryResultsTo:)]) {
+//        NSArray *users = [tokens valueForKeyPath:@"@distinctUnionOfObjects.representedObject"];
+//        [self.selectionDelegate peopleInputController:self changedPresentedDirectoryResultsTo:[NSSet setWithArray:users]];
+//    }
```
vytis: Commented out code here as well, not sure if still needed

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsController.swift
```diff
@@ -0,0 +1,193 @@
...
+        pendingSearchTask = task
+    }
+    
+    func updateVisibleSections() {
+        var sections : [CollectionViewSectionController]
+        
+        switch (mode, team != nil) {
+        case (.search, false):
+            sections = [contactsSection, conversationsSection, directorySection]
+            break
```
vytis: `break` not really needed

vytis:
I also noticed that some of the snapshot tests are cropped - probably the size was hardcoded before labels were added. Maybe worth increasing the height by a bit so that everything fits? 

On the same topic, the size of the image is appended to the filename which means we loose diff capability when size changes. Is that done by FBSnapshot or by us? Could we change this behaviour?

daehn:
@vytis size is appended by us (see `ZMSnapshotTestCase`), that was don ein order to have multiple reference files for the same test but different sizes, but I guess we could do that only for these special methods (e.g. `verifyInAlliPhoneWidths`) and not append the size but the width there instead.

PR #1034, opened 2017-05-26
Fix connection requests in teams

daehn:
# What's in this PR?

* Fix connection requests showing up in teams lists.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ConversationListViewModel+Teams.swift
```diff
@@ -37,7 +37,7 @@ extension ConversationListViewModel: ZMUserObserver {
 extension ConversationListViewModel: TeamObserver {
     public func teamDidChange(_ changeInfo: TeamChangeInfo) {
         if changeInfo.isActiveChanged {
-            self.updateConversationListAnimated()
+            reload()
```
typfel: So it wont' animate anymore?

daehn: No, not for now. It still fades out the old list and fades in the new one though.

PR #1037, opened 2017-05-29
Update section name for teams and add title to start UI

typfel:


Wire-iOS/Sources/UserInterface/ConversationList/SearchView/SearchViewController.m
```diff
@@ -23,53 +23,87 @@
 #import "IconButton.h"
 #import "WireSyncEngine+iOS.h"
 #import "WAZUIMagicIOS.h"
+#import "UIColor+WAZExtensions.h"
 
-@interface SearchViewController ()
+@interface SearchViewController () <ZMUserObserver>
 @property (nonatomic, readwrite) PeopleInputController *peopleInputController;
-@property (nonatomic, readwrite) SearchView *searchView;
+@property (nonatomic, readwrite) NSString *searchTitle;
```
mikeger: I think the `UIViewController` has an integrated property called `title` of type `NSString`, so I guess this one can be used?

Wire-iOS/Sources/UserInterface/ConversationList/SearchView/SearchViewController.m
```diff
@@ -79,4 +113,13 @@ - (void)onCloseButtonPressed:(id)sender;
     }
 }
 
+#pragma mark - ZMUserObserver
+
+- (void)userDidChange:(UserChangeInfo *)note
+{
+    if ([note.user isSelfUser] && note.accentColorValueChanged) {
```
mikeger: Can probably be done through classy without the observation?

typfel: Maybe, just moved the code.

typfel: using `accent-current-color` doesn't update the view the accent color changes.

mikeger: Sounds like a bug in our Classy implementation. Let's merge it as is then!

PR #1040, opened 2017-05-29
Update list observer

basine:
When changing teams we were not updating the observers for the new team.

after reverting and reverting the revert :D

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
@@ -180,25 +180,12 @@
     [self updateNoConversationVisibility];
     [self updateArchiveButtonVisibility];
     
-    [self registerListObservers];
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -180,13 +180,19 @@ - (void)viewDidLoad
...
+    [self updateObserverTokensForActiveTeam];
     self.teamsObserver = [TeamChangeInfo addTeamObserver:self forTeam:nil];
     
     [self showPushPermissionDeniedDialogIfNeeded];
 }
 
+- (void)updateObserverTokensForActiveTeam
+{
+    self.allConversationsObserverToken = [ConversationListChangeInfo addObserver:self forList:[ZMConversationList conversationsIncludingArchivedInUserSession:[ZMUserSession sharedSession] team:[[ZMUser selfUser] activeTeam]]];
+    self.connectionRequestsObserverToken = [ConversationListChangeInfo addObserver:self forList:[ZMConversationList pendingConnectionConversationsInUserSession:[ZMUserSession sharedSession] team:[[ZMUser selfUser] activeTeam]]];
```
daehn: We could always pass `nil` for the `pendingConnectionConversations`, as the team will be ignored internally anyways in that case (there are no connection requests inside of a team).

PR #1043, opened 2017-05-29
Improved performance on conversation list avatars

mikeger:
- Hiding the unnecessary views is significantly more performant than removing / adding due to autolayout.
- `regularStyle` and `emphasisStyle` to use cached values.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
@@ -139,15 +139,19 @@ extension ConversationStatusMatcher {
     }
 }
 
+fileprivate let defaultRegularStyle = [NSFontAttributeName: FontSpec(.medium, .none).font!,
+                                       NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+
+fileprivate let defaultEmphasisStyle = [NSFontAttributeName: FontSpec(.medium, .medium).font!,
+                                        NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+
 extension ConversationStatusMatcher {
     static func regularStyle() -> [String: AnyObject] {
-        return [NSFontAttributeName: FontSpec(.medium, .none).font!,
-                NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+        return defaultRegularStyle
     }
     
     static func emphasisStyle() -> [String: AnyObject] {
-        return [NSFontAttributeName: FontSpec(.medium, .medium).font!,
-                NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+        return defaultEmphasisStyle
     }
 }
 
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
@@ -159,9 +159,9 @@ final class ContentSizeCategoryUpdater {
 }
 
 final class ConversationStatusStyle {
-    private(set) var defaultEmphasisStyle: [String: AnyObject] = [:]
-    private(set) var defaultRegularStyle: [String: AnyObject] = [:]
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -139,15 +139,19 @@ extension ConversationStatusMatcher {
...
 }
 
+fileprivate let defaultRegularStyle = [NSFontAttributeName: FontSpec(.medium, .none).font!,
+                                       NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+
+fileprivate let defaultEmphasisStyle = [NSFontAttributeName: FontSpec(.medium, .medium).font!,
+                                        NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
+
 extension ConversationStatusMatcher {
     static func regularStyle() -> [String: AnyObject] {
```
daehn: Why not make it a constant? e.g. `static let regularStyle`

mikeger: Was my first idea, but extensions cannot contain the stored properties.

mikeger: (`ConversationStatusMatcher` is a protocol)

daehn: Ah you're right, but it should be possible to do it like this then:
```swift 
static var title: String = {
        return "Title"
    }()
```

daehn: Will only be executed once

vytis: I think you should be able to do (`var` instead of `let`):
```
static var title: String {
        return "Title"
    }
```

Will be recreated every time but same as original implementation

daehn: I think reusing the same values was intended to improve the performance.

vytis: Ah true, didn't read through the whole diff!

mikeger: This would actually break the dynamic type if user would not restart the app, need to think about a more robust solution.

mikeger:
@typfel @daehn updated.

Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -135,34 +135,68 @@ extension ConversationStatusMatcher {
...
+    private let callback: () -> ()
+    private var observer: NSObjectProtocol!
+    
+    init(callback: @escaping () -> ()) {
+        self.callback = callback
+        callback()
+        self.observer = NotificationCenter.default.addObserver(forName: Notification.Name.UIContentSizeCategoryDidChange,
+                                                               object: nil,
+                                                               queue: nil) { [weak self] _ in
+                                                                guard let `self` = self else {
```
daehn: Could get rid of the guard and us optional chaining like `self?.callback()`

Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -135,34 +135,68 @@ extension ConversationStatusMatcher {
...
     }
 }
 
-fileprivate let defaultRegularStyle = [NSFontAttributeName: FontSpec(.medium, .none).font!,
-                                       NSForegroundColorAttributeName: UIColor(white:1.0, alpha:0.64)]
 
-fileprivate let defaultEmphasisStyle = [NSFontAttributeName: FontSpec(.medium, .medium).font!,
+final class ContentSizeCategoryUpdater {
+    private let callback: () -> ()
+    private var observer: NSObjectProtocol!
```
daehn: Currently the token is never being removed, I think this should be done in `deinit`.

PR #1059, opened 2017-05-31
Start UI: not possible to make a call to person without existing conversation

vytis:
The fix is to initiate the call only after the conversation is created. However, @typfel found that there is a race condition because the conversation needs to exist remotely for this to work.

This PR partly fixes the issue, more robust fix coming soon.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
@@ -117,16 +117,22 @@
     if (call) {
         [self dismissPeoplePickerWithCompletionBlock:^{
             if (users.count == 1) {
+                __block ZMConversation *conversation = nil;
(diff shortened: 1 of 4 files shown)


mikeger:
@vytis would video call succeed for group in this case? Or should the button be hidden? For another PR?

vytis:
@mikeger it starts the video call but from my limited testing it seems that on the receiving end it is an audio call. So you see yourself in the call overlay and it says "Video off". Would be good to remove this button actually, will update the PR.

typfel:
@vytis hiding the button is one many case where we need to treat group conversations as one-on-one. Maybe we should discuss a strategy before fixing this everywhere

vytis:
Well we never supported video calls in groups. I think what changed is the action when button is tapped, now it opens group conversation instead of 1to1 when selecting user. So just adding a way to switch on/off buttons depending on type of conversation would do the job

vytis:
@mikeger @typfel updated

PR #1060, opened 2017-05-31
Not asking for address book permissions when in a team

vytis:


mikeger:
@vytis why do not merge? Should be fine, since the PR is pointing to `develop`. Or any other issue you might think of before merging?

vytis:
Oh yeah, we have release branch by now anyway, good to go then

PR #1065, opened 2017-06-01
fix cases in conversation system message

basine:
German cases for creating conversations / adding / removing members where gramatically wrong.


basine:
@jenskasemets

jenskasemets:
I think it's better just to modify base file (English) and handle German strings directly in Crowdin, otherwise there might be conflict during building.

PR #1066, opened 2017-06-01
Add empty result label to add participants screen

typfel:
- Turn `SearchResultsController` into a view controller in order to share more code with the start UI.
- Delete unused classes


typfel:
The fonts is messed up in the test some how. Need to investigate.

vytis:
It seems like Classy didn't do its job, maybe rerunning the tests would sort it out ¯\_(ツ)_/¯

typfel:
Thanks! I know what wrong. Classy breaks because I removed the classes but they are still references from the `.cas` file.

PR #1071, opened 2017-06-06
Use new SearchDirectory in ContactsDataSource

typfel:
Also bump SE which fixes profile images for search users

depends on:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/410


Wire-iOS/Sources/UserInterface/ConnectRequests/IncomingConnectionView.swift
```diff
@@ -95,7 +89,7 @@ public final class IncomingConnectionView: UIView {
                 user: user,
                 fallbackName: "",
                 addressBookName: BareUserToUser(user)?.addressBookEntry?.cachedName,
-                commonConnections: Int(commonConnectionsCount)
+                commonConnections: 0
```
vytis: Maybe worth removing the parameter from `UserNameDetailViewModel` and others as well? Is not used anymore I assume?

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsDataSource+Searching.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    @objc
+    public func search(withQuery query: String) {
+        let request = SearchRequest(query: query, searchOptions: [.contacts, .addressBook])
+        let task = self.searchDirectory.perform(request)
+        
+        task.onResult { [weak self] (searchResult, _) in
+            guard let `self` = self else { return }
+            self.ungroupedSearchResults = searchResult.addressBook
+            self.delegate?.dataSource!(self, didReceiveSearchResult: self.ungroupedSearchResults!)
```
vytis: `?` instead of `!` would work the same probably. Also using `searchResult.addressBook` instead of local var would get rid of scary ❗️

Wire-iOS/Sources/UserInterface/ConnectRequests/IncomingConnectionView.swift
typfel: Yes was was considering it. Was unsure if `AddressBookCorrelationFormatter` behaved the same without common connections. But let's try

PR #1077, opened 2017-06-09
Add extra case for when everyone has been added to conversation

typfel:
https://wearezeta.atlassian.net/browse/ZIOS-8661


typfel:
Check with Yorgos before merging.

PR #1079, opened 2017-06-09
Add separate string for analytics option

infotexture:
Remove redundancy between section title & option


infotexture:
As discussed w/ @mikeger

PR #1085, opened 2017-06-13
StartUI: design tweaks

vytis:
Tweaked padding, colors and added clear button to search field


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchHeaderViewController.swift
```diff
@@ -120,6 +134,13 @@ public class SearchHeaderViewController : UIViewController {
         delegate?.searchHeaderViewControllerDidCancelAction(self)
     }
     
+    @objc
```
daehn: Could be marked `dynamic` instead of `@objc`

vytis: Makes sense, added not for ObjC compatibility anyway

PR #1095, opened 2017-06-21
Remove V2 Calling

typfel:
depends on:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/417


Podfile.lock
```diff
@@ -62,4 +62,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 635304be06eca03999f2ecf7da5a56db48a69f1f
 
-COCOAPODS: 1.2.1
+COCOAPODS: 1.2.0
```
vytis: Might be worth updating CocoaPods - there some changes to the project because of that as well

Wire-iOS/Sources/UserInterface/Settings/SettingsTechnicalReportViewController.swift
```diff
@@ -71,7 +71,7 @@ class SettingsTechnicalReportViewController: UITableViewController, MFMailCompos
     }
     
     lazy private var lastCallSessionReports: [TechnicalReport] = {
-        let voiceChannelDebugString = VoiceChannelV2.voiceChannelDebugInformation().string.trimmingCharacters(in: .whitespaces)
+        let voiceChannelDebugString = ""
```
daehn: Looks like there won't be any records anymore, maybe this method could be completely removed?

PR #1096, opened 2017-06-21
Disable address book upload when there is a team

daehn:
# What's in this PR?

* There already was a check to not show the alert to share contacts, this adds additional checks to ensure we never upload contacts when there is a team.

### This PR depends on:
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/420 has been merged.


mikeger:
@daehn we upload the address book in other places too, search for `startRemoteSearch` in UI project. Should we may be disable the upload in sync engine, in `ZMUserSession uploadAddressBook` ?

daehn:
@mikeger all calls to `startRemoteSearchWithCheckingIfEnoughTimeSinceLast` are behind a check now.

daehn:
@mikeger As a caller I would expect the address book (or the next chunk) to be upload when calling `ZMUserSession.uploadAddressBook()`, but we could add a convenience method e.g. `uploadIfNoTeam` or something like that. But still, if we reach this point it means we prompted a user with a team to share contacts in the first place, which we do not want to do, so we need earlier checks either way.

mikeger:
@daehn not sure, if it's checked for example in https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/Permissions/ShareContactsViewController.m#L223 ?

Agree that if user calls `uploadAddressBook` then it should be uploaded.

daehn:
Right, there is a check to skip presenting the `ShareContactsViewController` when there is a team though, but maybe worth adding a check inside as well. Then again I would argue that this controller should never be presented in the first place if there is a team. But if this happens we should still ensure that we don't share.

mikeger:
@daehn as I am thinking more about it, I would remove `uploadAddressBook` method and replace it with `uploadIfNeeded` or something similar, while it is going to be a serious issue for business user if one would start getting autoconnects. You can check the details with Yorgos. Thanks!

daehn:
Yeah, we definitely need to avoid that. I might also add preconditions to the view controller inits or something like that as well, as it would be a programmer error if we show a dialog and ask for permissions to share and then do nothing.

daehn:
Closed in favour of https://github.com/wireapp/wire-ios/pull/1098

PR #1101, opened 2017-06-23
Show team name in self profile

mikeger:
- Handle label is moved above the user image.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
@@ -23,37 +23,58 @@ import Cartography
     public let imageView = UserImageView(size: .big)
     public let nameLabel = UILabel()
     public let handleLabel = UILabel()
+    public let teamNameLabel = UILabel()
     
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -23,50 +23,71 @@ import Cartography
...
+        nameLabel.accessibilityLabel = "profile_view.accessibility.name".localized
+        nameLabel.accessibilityIdentifier = "name"
+        handleLabel.accessibilityLabel = "profile_view.accessibility.handle".localized
+        handleLabel.accessibilityIdentifier = "username"
+        
+        nameLabel.text = user.name
+        nameLabel.accessibilityValue = nameLabel.text
+        
+        if let teamName = user.team?.name {
+            teamNameLabel.text = String(format: "profile_view.team_name.in".localized, teamName)
```
daehn: Could also do `"profile_view.team_name.in".localized(args: teamName)`.

mikeger:
@daehn updated, thanks, this way it's more to the point.

PR #1104, opened 2017-06-26
Autocorrect last word in the message

mikeger:
# Issue

Users complained that when writing a message, the last word is not being autocorrected. The issue happens because the iOS does not know when we are taking the text from the input text view and therefore is unable to figure out when to trigger the autocorrect.

@alanddd

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Helpers/UITextView+Autocorrect.swift
+++ b/Wire-iOS/Sources/UserInterface/Helpers/UITextView+Autocorrect.swift
@@ -22,7 +22,7 @@ import UIKit
 extension UITextView {
     // Autocorrects the last word, if necessary.
     func autocorrectLastWord() {
-        text = text + " "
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Helpers/UITextView+Autocorrect.swift
```diff
@@ -0,0 +1,28 @@
...
+//
+
+import Foundation
+import UIKit
+
+extension UITextView {
+    // Autocorrects the last word, if necessary.
+    func autocorrectLastWord() {
+        text = text + " "
+        text = text.substring(to: text.index(before: text.endIndex))
```
daehn: Did you try `resignFirstResponder()` followed by `becomeFirstResponder()`? Seems like the common approach to accept the current autocorrection.

mikeger:
@daehn updated, looks cleaner now, I hope some day Apple would add the API for it.

PR #1105, opened 2017-06-26
Joined contacts and team members in start UI

mikeger:
# Issue

The team members section should contain the contacts of user, labeled as guests.


typfel:
@mikeger did you try it in the add participants screen as well? Uses the same code.

mikeger:
@typfel yes, works there for free 👍

PR #1110, opened 2017-06-30
Hide team selector for single account mode

daehn:
# What's in this PR?

* In the first stage it won't be possible to log into multiple accounts, the team selector will be hidden until that's possible.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBar.swift
```diff
@@ -76,20 +79,13 @@ final class ConversationListTopBar: TopBar {
...
-            if !showTeams {
-                self.setShowTeams(to: true)
-            }
-        }
-        else {
-            if showTeams {
-                self.setShowTeams(to: false)
-            }
-        }
+        guard showTeams != self.showTeams else { return }
```
mikeger: I am not 100% sure, but it looks like we compare self.showTeams to itself here.

daehn: Yeah, good point, had that logic in `setShowTeams` before where it's passed as a parameter.

daehn: @mikeger updated!

PR #1117, opened 2017-07-05
Add join call button to conversation navigation bar if there is an ongoing call

daehn:
# What's in this PR?

Add join call button to conversation navigation bar if there is an ongoing call:

simulator screen shot 05 07 2017 10 36 43

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
@@ -54,7 +54,7 @@ public extension ConversationViewController {
         return button
     }
     
-    var audioCallBarButtonItem: IconButton {
+    var audioCallButton: IconButton {
         let button = barButtonItem(withType: .callAudio,
                                    target: self,
                                    action: #selector(ConversationViewController.voiceCallItemTapped(_:)),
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -174,6 +175,22 @@ public extension ConversationViewController {
         
         return items
     }
+
+    private func shouldShowCollectionsButton() -> Bool {
+        guard #available(iOS 8.3, *) else { return false }
```
vytis: Why it is disabled for iOS 8.0-8.2?

daehn: I only moved this part to have all button related code next to each other, but as far as I can recall there were some issues with the collection view. However I will remove this check as we dropped iOS 8 altogether now.

vytis: Oh yeah, I always forget that it really happened 😢

daehn: @vytis updated!

PR #1118, opened 2017-07-05
Adjust ephemeral timeout formatting

daehn:
# What's in this PR?

Adjust ephemeral timeout formatting. Unfortunately `ZeroFormattingBehavior.pad`  does not add a leading 0 to the first unit (e.g. 5:21 instead of 05:21), which is the reason we need to fallback to manual formatting in some cases.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
@@ -1,6 +1,6 @@
 //
 // Wire
-// Copyright (C) 2016 Wire Swiss GmbH
+// Copyright (C) 2017 Wire Swiss GmbH
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
```diff
@@ -0,0 +1,54 @@
...
+        return formatter
+    }()
+
+    func string(from interval: TimeInterval) -> String? {
+        return timeString(from: interval).map {
+            "content.system.ephemeral_time_remaining".localized(args: $0)
+        }
+    }
+
+    private func timeString(from interval: TimeInterval) -> String? {
```
vytis: But this will break all the internationalization that happens for time formatting... Maybe we can work around this issue by using `NSDateFormatter`?

daehn: See the comment, the dissed behaviour is to have a leading zero, which is not possible even using `DateComponentsFormatter`s zero formatting behaviour `.pad`, which should append leading zeros and does so, but does not do it for the first unit.

daehn: I will try hacking it together using DateFormatter tough.

daehn: @vytis I double checked with @Yorgos-V and he is ok with the zero formatting behaviour. I will change it to use `DateComponentsFormatter` again.

daehn:
@vytis updated, but let's wait for the tests to complete on Circle-CI

PR #1120, opened 2017-07-05
Don't show subtitle for left members in a conversation in list

daehn:
# What's in this PR?

* Don't show subtitle for left members in a conversation in list.


mikeger:
@daehn looks like the tests are failing, should I restart them?

daehn:
@mikeger there are missing symbols after bumping all components (running `carthage update`), I am on it.

daehn:
@mikeger fixed, tests should run now!

PR #1121, opened 2017-07-05
Remove use of the deprecated methods

mikeger:
- Initial PR to check the CI compilation


ConfigurationSets/Warnings.xcconfig
```diff
@@ -16,7 +16,7 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 // 
 
-
+GCC_TREAT_WARNINGS_AS_ERRORS = YES
```
vytis: 🥇

vytis:
I guess we will still have the "unsafe in extension" warnings? Will that stop compilation as well?

ConfigurationSets/Warnings.xcconfig
mikeger: It was on for Objective-C before, just moved it to `xcconfig` :)

vytis: Ah I see, not very ground-breaking then:)

mikeger:
@vytis the warnings about frameworks have to be fixed in the frameworks, this would be the next step. Those are linker warnings and they are not treated as errors (however they should be).

vytis:
Yeah, this could be done separately, fixing UI project warnings is more important anyway.

mikeger:
This one is good to go now.

PR #1122, opened 2017-07-06
CircleCI: using unified configuration

vytis:
There are some things that we need to do slightly differently to build UI project, but this was added in https://github.com/wireapp/wire-ios-shared-resources/pull/11


vytis:
Some tests are failing for some reason, closing this PR for now.

PR #1127, opened 2017-07-11
Add debug button to identify unread conversations

marcoconti83:
In some situation it’s not possible to find which conversations are causing the badge count to be non-zero, even if they all look read. This option in the debug menu will allow to identify which one is the offending conversation. Hopefully this will allow us to understand the pattern of this issue.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
@@ -192,23 +192,6 @@ import Foundation
             return DeveloperOptionsController()
         }
         
-        let sendBrokenMessage = { (type: SettingsCellDescriptorType) -> Void in
-            guard
-                let userSession = ZMUserSession.shared(),
-                let conversation = ZMConversationList.conversationsIncludingArchived(inUserSession: userSession).firstObject as? ZMConversation
-            else {
-                return
-            }
-
-            let builder = ZMExternal.builder()
-            _ = builder?.setOtrKey("broken_key".data(using: .utf8))
-            let genericMessage = ZMGenericMessage.genericMessage(pbMessage: builder!.build(), messageID: UUID().transportString(), expiresAfter: nil)
-            
-            userSession.enqueueChanges {
-                conversation.append(genericMessage, expires: false, hidden: false)
-            }
-        }
-        
         developerCellDescriptors.append(devController)
         
         let diableAVSSetting = SettingsPropertyToggleCellDescriptor(settingsProperty: self.settingsPropertyFactory.property(.disableAVS))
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -310,4 +295,39 @@ import Foundation
...
     }
     
+    private static func findUnreadConversation(_ type: SettingsCellDescriptorType) {
+        guard let userSession = ZMUserSession.shared() else { return }
+        let predicate = ZMConversation.predicateForConversationConsideredUnread()!
+        
+        let message: String
+        if let convo = (ZMConversationList.conversations(inUserSession: userSession) as! [ZMConversation])
+            .first(where: { predicate.evaluate(with: $0) }) {
+            message = ["Found an unread conversation:",
```
mikeger: Good start, may be you can add the button to copy the conversation description in the alert view?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -310,4 +295,39 @@ import Foundation
...
+                       "\(convo.displayName)",
+                        "<\(convo.remoteIdentifier?.uuidString ?? "n/a")>"
+            ].joined(separator: "\n")
+        } else {
+            message = "No unread conversation"
+        }
+        UIAlertView(title: nil, message: message, delegate: nil, cancelButtonTitle: "OK").show()
```
mikeger: We are moving from UIAlertView (it produces the deprecation warning). Please use `UIAlertController`.

marcoconti83: Using `UIAlertController` requires having a reference to a VC to call `UIViewController.presentViewController(alert)`. Is there a shared instance I should use, or do I need to inject one into this `SettingsCellDescriptorFactory`?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
marcoconti83: What do you mean by "to copy the conversation description"? You mean copy to the clipboard, to be pasted somewhere else?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
marcoconti83: Found this:
```
        guard let controller = UIApplication.shared.wr_topmostController(onlyFullScreen: false) else { return }
```

mikeger:
Yes, this should work! O was thinking to be able to copy the description so it can be pasted to the text editor.

Sent from a Mobile Device

marcoconti83:
Updated

PR #1133, opened 2017-07-17
WIP: Feature/markdown

johnxnguyen:
This is a basic implementation of the Markdown feature. 

Some things I will continue to work on are:
- markdown button highlighting -> e.g. if selection or caret is within bold text, the bold icon will be highlighted
- disabling certain buttons depending current markdown tokens -> e.g. if we're in a heading, we should not be able to insert any other markdown tokens.
- toggling syntax -> tapping on button once inserts token, tapping again removes it
- deleting opening syntax also deletes closing syntax
- incremental numbered lists -> after inserting token for first item, return key should insert the second token automatically. Tapping return key twice should exit the list
- tweaking the styling attributes


johnxnguyen:
markdown is available in playground build #991819

olivereitalu:
- Code renders in accent, but should be the same as text
- When editing a previously sent message the markdown icon is still active, but the tools are no longer there
- Numbered list numbers should be as per the spec (light grey, small type)
- The longpress Heading menu should have the selections in the typeface the text will be formatted in

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -159,7 +159,7 @@
       </AdditionalOptions>
    </LaunchAction>
    <ProfileAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
```
mikeger: Please reset the build configuration for profiling:
screen shot 2017-07-17 at 15 46 37

Wire-iOS/Sources/UserInterface/Helpers/MarklightCache.swift
```diff
@@ -0,0 +1,45 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+import Marklight
+
+open class MarklightCache: NSObject, MarklightCacheDelegate {
```
mikeger: I think that `MarklightCache` class is not necessary, since you can just pass the instance of `NSCache`.

johnxnguyen: ok I'll pull it out

Wire-iOS/Sources/UserInterface/Components/Views/ResizingTextView.h
```diff
@@ -19,6 +19,7 @@
 
 #import <UIKit/UIKit.h>
 #import "TextView.h"
+@import UIKit;
```
mikeger: I think UIKit is already imported some lines higher.

Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
```diff
@@ -0,0 +1,192 @@
...
+                offset = 2
+            }
+            
+            // preserve relative caret position
+            let newPos = position(from: start, offset: offset)!
+            selectedTextRange = textRange(from: newPos, to: newPos)
+            
+        case .bold:
+            // wrap syntax around selection
+            if !selection.isEmpty {
```
vytis: The logic for both `.bold`, `.italic` and `.code` seems to be very similar, maybe worth factoring it out to a separate method that could handle both cases?

Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
```diff
@@ -0,0 +1,192 @@
...
+        textStorage.fontTextStyle = UIFontTextStyle.subheadline.rawValue
+        textStorage.hideSyntax = hideSyntax
+        
+        textStorage.defaultAttributes = [
+            NSForegroundColorAttributeName: colorScheme.color(withName: ColorSchemeColorTextForeground),
+            NSFontAttributeName: FontSpec(.normal, .none).font!
+        ]
+    }
+    
+    public func insertSyntaxForMarkdownElement(type: MarkdownElementType) {
```
vytis: It would be really useful to add some tests for the logic in this method!

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -188,7 +189,8 @@ - (void)viewDidLoad
     [self createInputBar]; // Creates all input bar buttons
     [self createSendButton];
     [self createEphemeralIndicatorButton];
-    [self createEmojiButton];
+//    [self createEmojiButton];
```
vytis: Was the emoji button removed for testing purposes only? Could be easy to forget!

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -106,6 +119,10 @@ private struct InputBarConstants {
         return inputBarState.isEditing
     }
     
+    var isMarkingDown: Bool {
```
vytis: 👍 for name

Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
johnxnguyen: true, I put it on my to do list :)

daehn: I think in general the switch cases could be extracted in separate methods.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
mikeger: The emoji keyboard has to be removed in order to host the button for the markdown, unfortunately. You can check the details with Oliver. We decided to retain the keyboard code for the moment, while it is still possible that we would find the place for emoji keyboard.

daehn:
@johnxnguyen while using I noticed quite some delay between pressing the markdown button and the markdown UI being presented.

johnxnguyen:
@daehn do you mean when the keyboard appears or when the markdown icons slide down?

daehn:
@johnxnguyen when the container with the icons slides down.

johnxnguyen:
@daehn strange I haven't noticed that, I'll have a look at it though

daehn:
@johnxnguyen I am running iOS 11, so if it doesn't happen for other users let's assume that's the reason.

Wire-iOS/Sources/UserInterface/MarkdownBarView.swift
```diff
@@ -0,0 +1,148 @@
...
+        }
+        
+        addSubview(stackView)
+        
+        constrain(self, stackView) { view, stackView in
+            stackView.edges == view.edges
+        }
+        
+        headerButton.itemIcons = [.markdownH1, .markdownH2, .markdownH3]
+        headerButton.expandDirection = .right
```
daehn: Did you try this in a language using right-to-left layout? In assume the `UIStackView` will layout its arranged subviews mirrored, so maybe this should be dynamically set (you can check RTL layout easily by setting `Application Language` in scheme settings to `Right to Left PseudoLanguage`).

Wire-iOS/Sources/UserInterface/MarkdownBarView.swift
```diff
@@ -0,0 +1,148 @@
...
+    @objc private func buttonTapped(sender: IconButton) {
+        
+        let elementType: MarkdownElementType
+        
+        switch sender {
+        case headerButton:
+            switch headerButton.iconType(for: .normal) {
+            case .markdownH1:       elementType = .header(.h1)
+            case .markdownH2:       elementType = .header(.h2)
+            default:                elementType = .header(.h3)
```
daehn: I would suggest removing the `default` case and being explicit (`.markdownH3`), in order to profit from compiler errors in case the enum changes in the future.

Wire-iOS/Sources/UserInterface/MarkdownBarView.swift
johnxnguyen: ahh I see, it expands off the screen. Nice spotting! I'll change it so the button checks if the popup fits onscreen and if it doesn't, reverse the direction

daehn: Even better you should be able to check the `semanticContentAttribute` of any UIView and see if you should layout everything RTL.

johnxnguyen: cool, I didn't know about this attribute. So, you think I should keep the expand direction, but just reverse it if the layout is RTL?

johnxnguyen:
@daehn @vytis @mikeger I've been looking into the failed CI tests (snapshot tests on the input bar) and I'm having trouble finding a solution. Basically, the text in the input bar is slightly narrower than the  text in the snapshot, and the RTL test is failing too.

I've been able to narrow down the problem to the Marklight framework, particularly at the the point where the default attributes are applied to the string, just before the markdown attributes are applied on top. Each time the text changes, the string is parsed and the markdown styling is applied, so we need to 'reset' the attributes to the defaults first so that non marked down text appears normal.

The default attributes are simply the foreground colour (using the `ColorScheme`) and the font (using the `FontSpec`). Do you have any idea why the text is rendering slightly narrower?

PR #1136, opened 2017-07-20
Update Localizable.strings

zecakeh:
Fixed typos in the translation

### DIFF
--- a/Wire-iOS/Resources/fr.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/fr.lproj/Localizable.strings
@@ -78,7 +78,7 @@
 "contacts_ui.title" = "Inviter des personnes";
 
 "contacts_ui.no_contact.title" = "Aucune conversation";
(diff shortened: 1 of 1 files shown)


mikeger:
Hey @zecakeh ! Thanks for the corrections for the french translation, we at Wire are really focused on continuous improvement for the translations for the variety of languages we are supporting. 

The process we have for translations is the following: we at Wire are using the [Crowdin](https://crowdin.com/) to localize the clients, so the translations are being reviewed and approved from there. Whenever we are doing a release, the newest translation is pulled from Crowdin and merged into the client. Therefore if we would merge your pull request now it is going to be overwritten by the Crowdin version. I would like therefore to ask you to register on Crowdin and request the translator permissions on wire-ios project. I would give you the permissions and you would be able to apply the same changes there, and going forward being able to review and contribute to our future translations!

With kind regards, Mike.

PR #1137, opened 2017-07-21
Adjust join call button layout

daehn:
# What's in this PR?

* Small join button layout adjustments.


daehn:
@mikeger unfortunately not. We do not have any screenshot tests for the conversation navbar yet. I looked into it and it is pretty tightly coupled with the `ConversationViewController` at the moment. However I agree that it would make sense to extract some of the components here at some point and make it easily testable.

PR #1141, opened 2017-07-24
Design fixes/background color change

johnxnguyen:
**Changes:**
- renamed `ColorSchemeColorBackgroundNew` to `ColorSchemeColorConversationBackground`
- updated snapshot test references

The failing tests are the VoiceChannelOverlayTests, which were present before these changes

**Notes**:

The typing indicator background is now visible as it sits over both the conversation background and the input text view, is this ok? See the screenshot:

conversation background color typing indicator


typfel:
The cell tests doesn't look correct. The gray color is only behind the text and not the whole cell.

johnxnguyen:
I think because the `ConversationContentViewController` background isn't included in the snapshot test, which is the same colour as the `TextMessageCell` background

WireExtensionComponents/Utilities/ColorScheme.m
```diff
@@ -29,7 +29,7 @@
 
 NSString * const ColorSchemeColorSeparator = @"separator";
 NSString * const ColorSchemeColorBackground = @"background";
-NSString * const ColorSchemeColorBackgroundNew = @"background-new";
+NSString * const ColorSchemeColorConversationBackground = @"background-new";
```
mikeger: I guess we can also rename the key, check `stylesheet.cas` line 1632 for where it needs to be renamed 👍

mikeger:
🔥 Conflict to fix.

johnxnguyen:
hey @daehn I've added the changes you requested

PR #1145, opened 2017-07-25
Design fixes/add people screen

johnxnguyen:
## Fixed:

• Search field on the Add People screen: since the field has a white opacity, it is not visible on the white mode. 
• Search field bottom margins when you scroll on the Start UI list.
 
add participants screen


johnxnguyen:
all tests passed locally, looking into why CircleCI failed

typfel:
CI go stuck on cocoapods. I'd restart it.

PR #1149, opened 2017-07-25
Design fixes/drafts

johnxnguyen:
## Changes:
- move separator view between send button & markdown tool bar
- extend text view to the top of the markdown bar, but set the bottom content inset above the send button
- replace delete button with an alert when dismiss button tapped

img_5833

img_5832

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -945,3 +945,8 @@
 "compose.drafts.compose.delete.confirm.title" = "Confirm Deletion";
 "compose.drafts.compose.delete.confirm.message" = "This action will permamently delete this draft and cannot be undone.";
 "compose.drafts.compose.delete.confirm.action.title" = "Delete";
+
+"compose.drafts.compose.dismiss.confirm.title" = "Confirm Changes";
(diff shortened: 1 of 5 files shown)


johnxnguyen:
Do you think we need to show the alert when the dismiss button is tapped from the draft lists? On one hand it might be unclear which draft the alert is referring to, but on the other hand, in iPad the dismiss button isn't displayed in the compose message view at all...

mikeger:
@johnxnguyen you mean when the dismiss button is tapped inside the list? I think then the draft is not getting deleted? If this is correct, then no need for confirmation, user can come back and continue typing.

johnxnguyen:
@mikeger yes when it's tapped within the list. My concern is that since the alert view is only displayed when the user dismisses the draft (from the compose view), and since there is no dismiss button within this view for the iPad because it's sitting in a split view, the user won't be able to trigger the alert view at all. But, they can always delete the draft by swiping the list cell

mikeger:
Ah, I see, it is important since we removed the delete button. I think swiping in the list is good enough, since it's so common in other apps.

johnxnguyen:
@mikeger ok so we're good to merge?

mikeger:
👍

PR #1150, opened 2017-07-25
Teams UI

mikeger:
- Added create team button.
- Added team image on profile.
- Added add account button.
- Removed invite button.


mikeger:
@daehn, no changes, since I added the new mode, but did not changed the old one.

PR #1157, opened 2017-07-26
Design fix tweaks

johnxnguyen:
## Changes:
- Dismissing draft now brings up an alert sheet and deleting a draft returns the user back to the drafts list
- conversation background colour is slightly darker for light mode
- conversation navigation bar and input bar background colour are slightly lighter for dark mode

## Fixes:
- badge colour for the devices cell in settings appears correctly when the cell is highlighted
- cell highlight color in drafts list for dark mode


typfel:
screen shot 2017-07-26 at 15 12 30

Cells have almost the same color as the background now. Has the designed tried this out them self?

johnxnguyen:
@typfel I assume so but I'll check with them

PR #1162, opened 2017-07-27
Unify log enabling on start-up to one place

typfel:
- also don't enable any logs by default when running on the simulator.

https://wearezeta.atlassian.net/browse/ZIOS-8702


daehn:
@typfel I personally like having network log enabled by default when running on Simulator.

typfel:
@daehn we can always enable them by default in the *xcscheme

PR #1165, opened 2017-07-28
Update background color

daehn:
# What's in this PR?

* Update conversation background color, new color is `#F8F8F8`.


mikeger:
@daehn CircleCI says that the screenshot comparison failed, may be recorded on the wrong simulator?

PR #1172, opened 2017-08-01
Markdown/design fix

johnxnguyen:
## What's in this PR?

Some minor design fixes for the markdown feature.

### Fixed:
- italic font weight should be `light`
- header popup button overlay renders incorrect color for dark mode

### DIFF
--- a/WireExtensionComponents/Utilities/FontScheme.swift
+++ b/WireExtensionComponents/Utilities/FontScheme.swift
@@ -95,7 +95,7 @@ extension UIFont {
         return fontDescriptor.symbolicTraits.contains(.traitItalic)
     }
     
-    public func setItalic() -> UIFont {
+    public func italicFont() -> UIFont {
         
         if isItalic {
             return self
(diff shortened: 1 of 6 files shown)


WireExtensionComponents/Utilities/FontScheme.swift
```diff
@@ -90,6 +90,28 @@ extension UIFont {
     }
 }
 
+extension UIFont {
+    public var isItalic: Bool {
+        return fontDescriptor.symbolicTraits.contains(.traitItalic)
+    }
+    
+    public func setItalic() -> UIFont {
```
vytis: Maybe it's possible to add a single variable with getter and setter overriden?

WireExtensionComponents/Utilities/FontScheme.swift
```diff
@@ -90,6 +90,28 @@ extension UIFont {
...
+        return fontDescriptor.symbolicTraits.contains(.traitItalic)
+    }
+    
+    public func italicFont() -> UIFont {
+        
+        if isItalic {
+            return self
+        } else {
+            var symbolicTraits = fontDescriptor.symbolicTraits
+            symbolicTraits.insert([.traitItalic])
```
daehn: @johnxnguyen as `symbolicTraits` is a bit mask you don't need to warp it in `[]` when only inserting one value, e.g. `symbolicTraits.insert(.traitItalic)` should work as well.

PR #1177, opened 2017-08-04
Fix/markdown

johnxnguyen:
# What's in this PR?
This PR addresses three issues:
- Sending empty markdown inserts an empty message [ZIOS-8860]
- Choosing second markdown doesn't erase the first [ZIOS-8859]
- Punctuation is eaten by markdown [ZIOS-8863]

## Solutions
- [ZIOS-8860]: Empty markdown (valid syntax matches with zero content or space only content) is now stripped from the text before being sent. This stripped text is also used to determine when the message is sendable or not; by disabling the send button if the message only contains empty markdown. If the send button is disabled, then attempting to send empty markdown will not do anything
- [ZIOS-8859] & [ZIOS-8863]: Solved by updating the regular expression for italic markdown (in the Marklight framework)


mikeger:
@johnxnguyen one test failed:
```
  testThatItStripsEmptyMarkdownInMessageContainingNonEmptyMarkdown, XCTAssertEqual failed: ("# header
  /Users/distiller/wire-ios/Wire-iOS Tests/MarklightTextViewTests.swift:274
        // then
        XCTAssertEqual(result, expectation)
    }
```

johnxnguyen:
@mikeger yeah I'm looking into it, might be a big in the test because simulating the test case in the app seems to work fine

PR #1190, opened 2017-08-11
Multi account support

typfel:
Updates the UI for the multi account support added in SE.


Wire-iOS/Sources/AppController.m
```diff
@@ -435,8 +435,11 @@ - (void)setupUserSession:(ZMUserSession *)userSession
     (void)[Settings sharedSettings];
     
     _zetaUserSession = userSession;
-
-    self.analyticsEventPersistence = [[ShareExtensionAnalyticsPersistence alloc] initWithSharedContainerURL:userSession.sharedContainerURL];
+    
+    NSString *appGroupIdentifier = NSBundle.mainBundle.appGroupIdentifier;
+    NSURL *sharedContainerURL = [NSFileManager sharedContainerDirectoryForAppGroupIdentifier:appGroupIdentifier]; // TODO jacob fallback
```
daehn: AFAIK this method already handles the fallback internally.

Wire-iOS/Sources/AppController.m
```diff
@@ -447,10 +450,28 @@ - (void)setupUserSession:(ZMUserSession *)userSession
     // Singletons
     AddressBookHelper.sharedHelper.configuration = AutomationHelper.sharedHelper;
     
-    [MessageDraftStorage setupSharedStorageAtURL:userSession.sharedContainerURL error:nil];
+    [MessageDraftStorage setupSharedStorageAtURL:sharedContainerURL error:nil];
```
daehn: This means we will share drafts between accounts, but I personally think that's good (as drafts are excluded from the next release anyway this does not matter at the moment, but should be considered when the time has come).

Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -106,12 +106,19 @@ class ShareExtensionViewController: SLComposeServiceViewController {
 
     private func recreateSharingSession() {
         let infoDict = Bundle.main.infoDictionary
-
-        guard let applicationGroupIdentifier = infoDict?["ApplicationGroupIdentifier"] as? String,
-            let hostBundleIdentifier = infoDict?["HostBundleIdentifier"] as? String else { return }
+    
+        guard
```
daehn: We have an extension on `Bundle` somewhere directly exposing these infoDict values as optional properties, but maybe not accessible from the share extension at the moment.

Wire-iOS/Sources/AppController.m
typfel: Ah ok

Wire-iOS Share Extension/ShareExtensionViewController.swift
typfel: I think it's accessible. I'll update

typfel: It's not

PR #1191, opened 2017-08-11
Removing special links to Vimeo

vytis:
We would like to sunset treating Vimeo links in a special way. This would enable us to remove problematic `VIMNetworking` framework that prevents us from cleanly building all dependencies from scratch.

Needs release with:

- [x] https://github.com/wireapp/wire-ios-link-preview/pull/16


typfel:
What's stopping us from merging this? Need to clear it with marketing? @vytis

mikeger:
Approved by design.

PR #1192, opened 2017-08-14
Fix/markdown emoji zios 8862

johnxnguyen:
## Problems
- when sending markdown that only contains emojis, the emojis are rendered as normal text size, however they should appear large.
- trying to add an emoji to header or list markdown causes the app to crash

## Solutions:
- before sending the message, we search the string for any markdown that contains only emojis and whitespace, then we strip the markdown syntax from this token
- the app was crashing because of incompatibilities between Swift `String` ranges and `NSRange`; an emoji that is considered to be of length 1 by `String` is considered to be of length 2 by `NSString`. Thus, when the text contains at least one emoji and we convert an `NSRange` to a `String` range, we sometime encounter an out of bounds error. The solution is rely on casting the `String` to an `NSString` when deleting characters or extracting substrings.

NOTE: Swift 4 closes the bridge between these range incompatibilities, so I will refactor this code once we migrate.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
@@ -256,7 +256,7 @@ extension MarklightTextView {
     }
 }
 
-// MARK: - Range calculations
+// MARK: - Range Calculations
 
 extension MarklightTextView {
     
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/MarklightTextView.swift
```diff
@@ -235,8 +236,8 @@ extension MarklightTextView {
             else { return }
         
         // remove postRange first so preRange is still valid
-        text.removeSubrange(text.rangeFrom(range: postRange))
-        text.removeSubrange(text.rangeFrom(range: preRange))
+        text.deleteCharactersIn(range: postRange)
```
typfel: What's the difference between `text.removeSubrange ` and `text.removeSubrange`?

typfel: Ah found the definition at the bottom.

typfel: So this is the NSRange <>  Range conversation fix.

johnxnguyen: yeah, so I'm not converting `NSRange` to `Range` anymore, rather just using `NSString`'s methods to delete characters

PR #1196, opened 2017-08-16
Add logout logic

daehn:
# What's in this PR?

* Adapt new logout API from `wire-ios-sync-engine`.
* Fix logout related crashes and leaks.

### This PR depends on:
- [x] A new release of `wire-ios-sync-engine` after https://github.com/wireapp/wire-ios-sync-engine/pull/496 has been merged.


Wire-iOS/Sources/Analytics/AnalyticsConversationListObserver.m
```diff
@@ -92,6 +92,10 @@ - (void)setObserving:(BOOL)observing
 
 - (void)probablyReceivedFullConversationList
 {
+    if (nil == ZMUserSession.sharedSession.managedObjectContext) {
```
typfel: We might have user session  without a managedObjectContext?

daehn: Yeah, we need to investigate who retains it (as I weakified the `zetaUserSession` on `AppControler`).

daehn: Actually it might be that we just don't nil it in `SessionManager` when logging out at the moment...

PR #1199, opened 2017-08-22
Refactor AppController

typfel:
Extract the app state machine into a separate class for easier readbility/testing
and move objects into the ZClientViewController since they are only relevant when
we have a user session.

depends on:
- [x] new SE release


Wire-iOS/Sources/AppState.swift
```diff
@@ -0,0 +1,49 @@
...
+    case unauthenticated(error : Error?)
+    case blacklisted
+    case migrating
+    case suspended
+
+    public static func ==(lhs: AppState, rhs: AppState) -> Bool {
+        
+        switch (lhs, rhs) {
+        case (.headless, .headless):
+            fallthrough
```
vytis: `fallthrough` police  🚔

typfel: In this case I think it looks for readably since otherwise it's all on long line.

typfel:
Forgot I did some small SE changes

Wire-iOS/Sources/AppState.swift
vytis: You can stack cases:
```
case (.headless, .headless)
case (.authenticated, .authenticated):
...
```

typfel: Oh :)

PR #1200, opened 2017-08-24
Enable account switcher

typfel:
This re-enables the team switcher for switching accounts. This is work-in-progress and many things still needs to be done.

depends on:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/500


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -167,8 +185,8 @@ class AppRootViewController : UIViewController {
             MagicConfig.shared()
             self.installOverlayWindow()
         } else {
-            overlayWindow?.removeFromSuperview()
-            overlayWindow = nil
+//            overlayWindow?.removeFromSuperview()
+//            overlayWindow = nil
```
mikeger: Not necessary? May be consider removing / adding a TODO

typfel: Ah no this is proper TODO actually. Let me take another look at that.

typfel: It's unnecessary to re-load magic when switching accounts but we probably want to remove the `NotificationStatusViewController` when logging out.

mikeger:
FYI: I started the SE build

mikeger:
wire-ios-sync-engine 105.1.1 is released

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -87,11 +101,41 @@ class AppRootViewController : UIViewController {
         sessionManager = SessionManager(appVersion: appVersion!, mediaManager: mediaManager!, analytics: analytics, delegate: appStateController, application: UIApplication.shared, launchOptions: launchOptions, blacklistDownloadInterval: Settings.shared().blacklistDownloadInterval)
     }
     
+    func enqueueTransition(to appState: AppState) {
+        
+        transitionQueue.async {
```
daehn: Is it possible to put the `prepare` and `transition` calls into the same dispatch block?

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -76,6 +77,10 @@ class AppStateController : NSObject {
             return .blacklisted
         }
         
+        if (isSuspended) {
```
daehn: `()` 😱

Wire-iOS/Sources/AppRootViewController.swift
typfel: Would we always enqueue the two block after another?

daehn: Yeah, you're right, mixed something up.

PR #1202, opened 2017-08-25
Reduce audio message size

johnxnguyen:
## Problem
Audio messages are being recorded at a relatively high sample rate (44.1khz, the same used for CDs) with a variable bit rate that averages around 200kbps. This leads to large file sizes and impacts upload/download performance.

## Solution
Lowering the sample rate to only 32khz generates files with an average bitrate of 47kbps, effectively reducing the file size to a quarter of the original, while at the same time maintaining acceptable audio quality for voice recorded messages.


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -54,7 +54,7 @@ class AppRootViewController : UIViewController {
         
         let isCallkitEnabled = !Settings.shared().disableCallKit
         var isCallkitSupported = false
-        if #available(iOS 10, *), TARGET_OS_SIMULATOR != 0 {
```
mikeger: Not sure if this should be removed, I think the simulator does not support the CallKit.

mikeger:
@johnxnguyen I am not sure,  I think the PR contains some unnecessary changes for AVS and SE versions, may be those can be separated into another PR for consistency.

Wire-iOS/Sources/AppRootViewController.swift
typfel: We need to find another way to detect. This compiled with an error against a device.

johnxnguyen:
@mikeger do you mean making a PR for only the component bumps?

Wire-iOS/Sources/AppRootViewController.swift
johnxnguyen: ok I'll take a look :)

mikeger: We can do `#if arch(i386) || arch(x86_64)`

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -105,6 +105,10 @@ class AppStateController : NSObject {
 
 extension AppStateController : SessionManagerDelegate {
     
+    func sessionManagerWillSuspendSession() {
```
typfel: Don't don't know why this didn't conflict. But you should take the implementation from develop.

mikeger:
@johnxnguyen would be cool, just for the consistency, so later in the commit tree we would not have the commit to reduce the audio message size that also bumps the AVS and updates some internal behavior, certainly if AVS update is not necessary for the audio message size reduction,

johnxnguyen:
@mikeger yeah that makes sense, I just ran `carthage update` to get the new versions.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -77,7 +77,7 @@ class AppStateController : NSObject {
             return .blacklisted
         }
         
-        if (isSuspended) {
+        if isSuspended {
```
typfel: 🙌

PR #1206, opened 2017-08-28
Fix shifted UI

mikeger:
- New AppRootViewController updated to host itself in the window correctly.

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -220,7 +220,7 @@ class AppRootViewController : UIViewController {
             overlayWindow.rootViewController = NotificationWindowRootViewController()
         }
         
-        if !isClassyInitialized {
+        if !isClassyInitialized && (appState == .authenticated(completedRegistration: false) || appState == .unauthenticated(error: nil)) {
             isClassyInitialized = true
             MagicConfig.shared()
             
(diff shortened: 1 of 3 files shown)


mikeger:
Also initializes classy once (seems like it should work like that).

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -213,7 +220,7 @@ class AppRootViewController : UIViewController {
             overlayWindow.rootViewController = NotificationWindowRootViewController()
         }
         
-        if isClassyInitialized || appState == .authenticated(completedRegistration: false) || appState == .unauthenticated(error: nil) {
+        if !isClassyInitialized {
```
typfel: The idea was not initialise classy when launching in the background but only when entering the authenticated or unauthenticated state.

mikeger:
@typfel updated, there was another bug that app could have an inactive state, that was considered as app is running in the foreground.

typfel:
@mikeger isn't the app inactive when the app is in foreground but some system message gets displayed?

mikeger:
I was following this post: https://stackoverflow.com/questions/7937400/need-clarification-about-uiapplicationstate

Also when the app is starting due to push event it's inactive for me, so the classy was loaded, I had to change it to avoid this effect.

typfel:
@mikeger ah ok 👍

PR #1209, opened 2017-08-28
Observe applicationDidFinishLaunching to update state

mikeger:
- Initially the app is in the inactive state, so the UI is not getting loaded.


mikeger:
@typfel

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -157,6 +159,11 @@ extension AppStateController {
         recalculateAppState()
     }
     
+    func applicationDidFinishLaunching() {
```
typfel: Wouldn't this also fire when launching because of a push?

mikeger: True, but the state would be still inactive.

PR #1216, opened 2017-08-29
Fix/async account manager

johnxnguyen:
# What's in this PR?

This PR utilises the new `SyncEngine` API changes to observe file system access when attempting to create a `SessionManager`. Since the `SessionManager` instantiates an `AccountManager`, which in turn attempts to access the file system. However, It is possible that the file system is not yet unlocked. Therefore, we will only create a `SessionManager` object once we the file system is accessible.

The file system access functionality was introduce in the `SyncEngine` release `107.0.0`. Here we simply update the call to create the `SessionManager`.


mikeger:
🔥   Conflict

PR #1218, opened 2017-08-29
Log Out

mikeger:
- Updated the design for logging out
- Added confirmation
- Added deletion of the database
- Removed the feature flag for logout

# Dependencies

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/512


Wire-iOS/Sources/Helpers/syncengine/SessionManager+Shared.swift
```diff
@@ -26,4 +26,26 @@ extension SessionManager {
...
+        }
+        
+        guard let sharedContainerURL = Bundle.main.appGroupIdentifier.map(FileManager.sharedContainerDirectory) else {
+            preconditionFailure("Unable to get shared container URL")
+        }
+        
+        let selectedAccountID = selectedAccount.userIdentifier
+        
+        self.logoutCurrentSession(deleteCookie: true)
+        self.accountManager.remove(selectedAccount)
```
typfel: Is this safe to delete the account at this point? Maybe we should delete it inside SessionManager after it has torn down the user session.

mikeger: I was under impression that `SessionManager.logoutCurrentSession` is synchronously waiting for the teardown to be done: it is calling `ZMUserSession.closeAndDeleteCookie` which is calling the operation loop teardown.

typfel: You're right looks like it's blocking. I'd still argue removing the account and deleting the folder is something we should do inside SE.

typfel: Maybe add a `deleteAccount:` argument to `logoutCurrentSession()`

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -223,11 +219,36 @@ extension SettingsCellDescriptorFactory {
...
+        let logoutAction: ()->() = {
+            guard let selectedAccount = SessionManager.shared?.accountManager.selectedAccount else {
+                fatal("No session manager and selected account to log out from")
+            }
+            
+            SessionManager.shared?.logoutCurrentSession(deleteCookie: true)
+            SessionManager.shared?.delete(account: selectedAccount)
+            
+            if let otherAccount = SessionManager.shared?.accountManager.accounts.first {
+                SessionManager.shared?.accountManager.select(otherAccount)
```
typfel: Have you tried this scenario? Would it briefly show the registration before logging into the second account?

mikeger: It's semi-working, it shows login window with X button, when you press X you land in the other account, checking now what's the best way to fix it.

typfel: Miss read, thought you were calling `select()` on the session manager. So then it's working no? Or what behaviour do you want?

mikeger: I wanted to skip the login screen, but may be it's OK like that for the moment. Will check with Yorgos.

typfel: If you want to skip the login I think we would need to have two paths. 

1. If the user has as second account switch to it and *then* delete the first account. 
2. if the user has only one account logout and delete the account.

mikeger: Actually let's merge this one so I can show it to Yorgos :)

PR #1220, opened 2017-08-30
Fix/ephemeral picture ZIOS 8879

johnxnguyen:
## Problem
Ephemeral images can be liked, saved, copied, forwarded, edited and resent.

## Solution
For ephemeral images only:
- In fullscreen image view, remove all toolbar buttons except **delete**.
- In the image cell, remove all buttons except **expand**

## Note
The ephemeral image can expire while the fullscreen image view is open, however the image view will remain open. The design team have confirmed this is by design.

screen shot 2017-08-30 at 14 37 03

screen shot 2017-08-30 at 14 36 41


johnxnguyen:
Do we need a test for this? If so, I might need some help creating the snapshot test.

johnxnguyen:
hey @mikeger, I was not able to capture the actual image in the new `ConversationImagesViewController` snapshots, however the toolbar is correctly captured

PR #1224, opened 2017-08-31
Add command-line switch to inject storage container to test migrations

marcoconti83:
This PR adds the ability, when running on simulator, to inject a previously stored shared container. This can be useful when debugging migrations. 

To use this, first run an old version. Access the container and copy the content (**including hidden files**, or it will fail) to a folder. Then switch to the new version and set a command line argument:

```
--debug-data-to-install=<folder where you copied>
``` 

It replace the content of the shared container with the content of that folder. This is enough to copy over user defaults, cryptobox and database.

Depends on:
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/356


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -330,3 +332,26 @@ extension AppRootViewController : AppStateControllerDelegate {
     }
     
 }
+
+// MARK: - Debug
+extension AppRootViewController {
+    
+    /// Takes all files in the folder pointed at by `debugDataToInstall` and installs them
+    /// in the shared folder, erasing any other file in that folder.
+    fileprivate func installDebugDataIfNeeded() {
```
typfel: This method doesn't need to be in `AppRootViewController`. I'd move to it's own class or make an extension on the AppDelegate.

PR #1227, opened 2017-09-01
Fix crash when putting app in background while switching accounts

vytis:
The reason was that ConversationListViewController was leaking after tearing down UI and UserSession. It was listening for app lifecycle notifications and was doing things with user session after coming back from background.

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -189,9 +189,22 @@ class AppRootViewController : UIViewController {
         }
     }
     
+    
+    private func dismissModalsFromAllChildren(of viewController: UIViewController?) {
+        guard let viewController = viewController else { return }
+        for child in viewController.childViewControllers {
+            if child.presentedViewController != nil {
+                child.dismiss(animated: false, completion: nil)
+            }
+            dismissModalsFromAllChildren(of: child)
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -189,9 +189,22 @@ class AppRootViewController : UIViewController {
...
+            dismissModalsFromAllChildren(of: child)
+        }
+    }
+    
     func transition(to viewController : UIViewController, animated : Bool = true, completionHandler: (() -> Void)? = nil) {
         
-        visibleViewController?.dismiss(animated: false, completion: nil)
+        // If we have some modal view controllers presented in any of the (grand)children
+        // of this controller they stay in memory and leak on iOS 10.
+        dismissModalsFromAllChildren(of: visibleViewController)
```
typfel: So this is fixed in iOS11?

vytis: The same issue was in iOS10 also

vytis: Will have to check what's the status on ios11 though

vytis: @typfel same behaviour on iOS11, leaks without the fix

PR #1231, opened 2017-09-05
Fix/app rotation

johnxnguyen:
## Problem
The app was not able to rotate when viewing fullscreen images.

## Cause
The app was in fact not receiving any device rotation information (including `UIDeviceOrientationNotification`'s) from the system. The problem seemed to result in the recent changes to the root view controller instantiation with the introduction of the new `AppRootViewController` class. All window and root view controller objects are managed by this class, and this class was first created in the `AppDelegate`'s initializer. After much trial and error, we discovered that although the view controller hierarchy appeared to be properly constructed, the windows were not receiving any information regarding the device's orientation, and concluded that the windows were being created too early.

## Solution
Simply delaying the instantiation of `RootViewController` to the `application:WillFinishLaunchingWithOptions:` method solved the problem. Furthermore, the rotation handling of the `NotificationWindowRootViewController` needed to be updated to match exactly the `RootViewController`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.m
@@ -124,12 +124,22 @@
 
 - (BOOL)shouldAutorotate
 {
-    return YES;
+    // guard against a stack overflow
+    if (self != UIApplication.sharedApplication.wr_topmostViewController) {
+        return UIApplication.sharedApplication.wr_topmostViewController.supportedInterfaceOrientations;
+    } else {
+        return YES;
+    }
 }
 
 - (UIInterfaceOrientationMask)supportedInterfaceOrientations
 {
-    return [UIViewController wr_supportedInterfaceOrientations];
+    // guard against a stack overflow
+    if (self != UIApplication.sharedApplication.wr_topmostViewController) {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.m
```diff
@@ -120,16 +120,16 @@ - (void)transitionToLoggedInSession
     [self.voiceChannelController.view autoPinEdgesToSuperviewEdgesWithInsets:UIEdgeInsetsZero];
 }
 
-#pragma mark - Rotation handling (should match up with roort)
+#pragma mark - Rotation handling (should match up with root)
 
 - (BOOL)shouldAutorotate
 {
-    return UIApplication.sharedApplication.wr_topmostViewController.shouldAutorotate;
+    return YES;
```
mikeger: I think with this change the calling UI would start to rotate, if you receive the call while looking on the picture. The problem is the calling UI is not looking good on iPhone landscape. Please check this case.

johnxnguyen: ah ok, will do 👍

johnxnguyen:
@mikeger I've just notice that sometimes opening the app classy isn't finished loading before the UI is visible

mikeger:
@johnxnguyen classy issue is annoying, but let's fix the top issues first

PR #1234, opened 2017-09-05
Avoid asking notification permissions when testing in automation

vytis:
Since we moved code for handling this to SE there was no easy way to intercept the calls except from subclassing UIApplication.


mikeger:
That's a cool (and valid) hack!

PR #1236, opened 2017-09-05
Update drafts to new folder structure

typfel:
New folder structure is described here: https://github.com/wireapp/wire-ios-data-model/pull/358

NOTE: we don't attempt to migrate existing drafts since we haven't released
this feature in public yet.

depends on :
- [x]  https://github.com/wireapp/wire-ios-data-model/pull/358
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/517
- [x] https://github.com/wireapp/wire-ios-share-engine/pull/41


typfel:
Still need fix a thing in share engine. It creates a files in the `Caches` folder which will make the main app crash.

PR #1240, opened 2017-09-06
Fix: registration analytics events not fired correctly

vytis:
The original issue was that there were 2 identical events fired when registering with phone number, but after some refactoring the events were not fired at all.

Using new client registration notification to signal that registration is complete.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationEmailFlowViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationEmailFlowViewController.m
@@ -206,7 +206,7 @@
 
 #pragma mark - ZMAuthenticationObserver
 
-- (void)authenticationDidSucceed
+- (void)clientRegistrationDidSucceed
 {
     [self.analyticsTracker tagRegistrationSucceded];
     [self presentProfilePictureStep];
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationEmailFlowViewController.m
```diff
@@ -206,7 +206,7 @@ - (void)emailVerificationDidSucceed
 
 #pragma mark - ZMAuthenticationObserver
 
-- (void)authenticationDidSucceed
+- (void)clientRegistrationDidSucceed
```
typfel: So this will be called much later in the registration process now but that doesn't affect the registration flow?

vytis: It doesn't seem that it affects the flow. The problem is when registering with phone `authenticationDidSucceed` is not called at all, so I unified both flows to use client registration notifications.

vytis: Will investigate the reason `authenticationDidSucceed` is not fired, maybe this PR is not needed at all

vytis:
Made the fix for the underlying reason in SE: https://github.com/wireapp/wire-ios-sync-engine/pull/519

PR #1243, opened 2017-09-06
Show a skeleton of the app when while switching accounts

typfel:
NOTE: this currently showing the picture of the account you were switching from, I'll fix this in a separate PR since we need to change the SE API.


Wire-iOS/Sources/UserInterface/SkeletonViewController.swift
```diff
@@ -0,0 +1,287 @@
...
+            return button
+        }
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    func createConstraints() {
+    }
```
mikeger: I think constraints are escaped and now live in `init` :)

PR #1251, opened 2017-09-07
Add an option to delete ephemeral message

johnxnguyen:
## What's new?

We now allow ephemeral messages to be deleted prematurely by long pressing on the ephemeral message. This reveals the usual action menu, however the only item it will contain is `delete`.


mikeger:
@typfel @marcoconti83 how confident are we that this is supported by the SE?

typfel:
@johnxnguyen @mikeger I assume you tried?

johnxnguyen:
@typfel @mikeger yes I tested on the device for all types of messages

typfel:
@mikeger @johnxnguyen so should be fine then I guess.

PR #1254, opened 2017-09-07
Feature/single image view

johnxnguyen:
## What's new?
This PR contains some interactive/visual improvements for the Single Image View :

- Tapping the image hides/shows the top and bottom bars
- Zooming the image (pinch and double tap) hides the top and bottom bars

## Bonus 🎉 🎊 🎉 
A bug was fixed where swiping to dismiss from a point outside of the image (i.e the background) would make the image disappear.


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.h
```diff
@@ -30,13 +30,18 @@ NS_ASSUME_NONNULL_BEGIN
 - (nullable UIView *)backgroundScreenshotForController:(FullscreenImageViewController *)fullscreenController;
 @end
 
+@protocol FullscreenImageViewControllerDelegate <NSObject>
```
mikeger: I have some comments in regards of the naming:
1. I think the delegate's methods are not tight to `FullscreenImageViewController`, so we can also rename it to something like `MenuControllerType`.
2. I think it would be enough to have one property, like `menuVisible`, and build the toggle via swift protocol extension, in this way you could skip writing the toggle code for different menus.

PR #1255, opened 2017-09-08
store last opened conversation per account

johnxnguyen:
Currently all the settings are general and saved in user defaults. The last viewed conversation is also the part of those settings. We need to save the last viewed conversation per account.


johnxnguyen:
@mikeger do we want to remove `lastViewedConversation` from `Settings`? It's not being used anymore.

PR #1268, opened 2017-09-18
Update for new calling API

typfel:
### Why
The legacy V2 calling API has been removed in SE and and this PR updates the consumers of this API. Since the new API is swift only I had to port  some classes to Swift.

depends on:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/498
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/581


mikeger:
@typfel some major conflict here 💣

PR #1270, opened 2017-09-19
Inserted avatars into the Share View Controller

nicorsm:
In this PR I put the `ConversationAvatarView` inside the `ShareDestinationCell`. Since the two views are in different targets, I added the `avatarView()` function inside the `ShareDestination` protocol, so you can provide a generic `UIView` object to the cell.


WireExtensionComponents/Controllers/ShareDestinationCell.swift
```diff
@@ -19,22 +19,35 @@
 import Foundation
 import Cartography
 
-
 private let verifiedShieldImage = WireStyleKit.imageOfShieldverified()
 
 
 final class ShareDestinationCell<D: ShareDestination>: UITableViewCell, Reusable {
     let checmarkSize: CGFloat = 24
```
johnxnguyen: might be good to fix this typo :)

typfel:
Commit history doesn't look right. Let us know if you need help sorting it out.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -21,8 +21,13 @@ import WireSyncEngine
...
 
-extension ZMConversation: ShareDestination {}
-
+extension ZMConversation: ShareDestination {
+    public func avatarView() -> UIView? {
+        let avatarView = ConversationAvatarView()
+        avatarView.conversation = self
+        return avatarView
+    }
+}
```
johnxnguyen: I think we could make this a get only computed property

mikeger:
@nicorsm I think you diverged your branch from `master`, we usually work on the new stuff on `develop` and on `master` we keep the latest released version, we mostly follow this model: http://nvie.com/posts/a-successful-git-branching-model/

Please rebase your changes on develop to have the clean commit history.

PR #1280, opened 2017-09-24
Fix crash when switching the account with keyboard up

mikeger:
- Issue happens when the `SessionManager` tries to swap the acounts, and keyboard is dismissed. - The dismissal of the keyboard triggers the reload of the table view content when the `UserSession` is not active any more.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
```diff
@@ -317,7 +317,9 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     conversationCell.delegate = self.conversationCellDelegate;
     conversationCell.analyticsTracker = self.analyticsTracker;
     
-    [self configureConversationCell:conversationCell withMessage:message];
+    if (nil != [ZMUserSession sharedSession]) {
```
typfel: Let's add comment here why we need to check if we have an active user session otherwise it might get re-factored away in future.

PR #1287, opened 2017-09-25
Updated localizations for 2.43

mikeger:


typfel:
Should we also update the push strings or are there no changes there?

mikeger:
@typfel the push strings are not changed, we just added the title for team: https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Resources/Base.lproj/Push.stringsdict

PR #1291, opened 2017-09-25
Fix retain cycle in AudioTrackPlayer

vytis:
Was leaking the instance every time we switch between accounts


mikeger:
@vytis tests failed somehow (`AudioMessageCellTests`)

vytis:
Yeah, I added DoNotMerge until I fix this

PR #1293, opened 2017-09-25
Feature/chat heads calling

johnxnguyen:
## What's in this PR?

- chat heads now support calling notifications
- tapping on chat head switches account and presents call overlay

- The chat head label logic was refactored out to `ChatHeadsViewController` so that `ChatHeadView` was no longer dependent on a `ZMConversationMessage` object.
- `ChatHeadView` is not initialised with `title` and `content`.

## Note
Ideally, we should refactor the formatting logic up to the notification dispatcher layer (when the notification objects are created) so that we can initialise a `ChatHeadView` directly with the `UILocalNotification`s `title` and `alertBody` properties prefilled.

❗️ depends on: 
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/557


mikeger:
:hurtrealbad: Conflict

PR #1301, opened 2017-09-26
Prevents crash when message in FileTransferCell is nil

marcoconti83:


typfel:
In which scenario is the message of a cell nil?

marcoconti83:
@typfel I got this while switching accounts

PR #1302, opened 2017-09-26
Temporarily disable CBR

mikeger:
- It is not working properly with the current AVS.


marcoconti83:
@mikeger fixed my comment

PR #1315, opened 2017-09-28
Fix getting stuck when resuming an aborted login

typfel:
## Problem
Login can take several paths atm which makes it a bit complicated. The login would miss to register for initial sync in the `path 3` case and `path 5`.

Previously this was not an issue because then the login observation was:
```
    if (nil != [ZMUserSession sharedSession]) {
        self.initialSyncObserverToken = [ZMUserSession addInitialSyncCompletionObserver:self userSession:[ZMUserSession sharedSession]];
    } else {
        self.sessionCreationObserverToken = [[SessionManager shared] addSessionManagerObserver:self];
    }
```
But this had a problem in `path 4` where the we would start observing initial sync for the user session which was in the process of tearing down.

### path 1
1. launch app
2. login
3. register client
4. done

### path 2
1. launch app
2. login
3. fail to register client (because you have to many clients)
4 delete client
5. register client
6. done

### path 3
1. launch app
2. login
3. fail to register client (because you have too many clients)
4 kill the app
5. launch app 
6. app will try to open the user session you tried to active but
7. fail to register client (because you are missing credentials)
8 app will transition to login screen (but user session will still be around)
9. delete client
10. register client
11. done

### path 4
1. app is running with two accounts
2. logout from account 1
3. app will transition to account 2 (but account 2 is discovered to be logged out)
4 app will transition to login screen (but user session will still be around until the the transition is done)
5. register client
6. done

### path 5
1. launch app
2. login (client is already there)
3. done

## Solution

We now observe user session creation AND client registration and begin the initial sync on any of those two events.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationViewController.m
@@ -49,7 +49,7 @@
 
 #import "AnalyticsTracker+Registration.h"
 
-@interface RegistrationViewController (PostLoginAuthenticationObserver) <PostLoginAuthenticationObserver>
+@interface RegistrationViewController (UserSessionObserver) <SessionManagerObserver, PostLoginAuthenticationObserver>
 @end
 
 @interface RegistrationViewController () <UINavigationControllerDelegate, FormStepDelegate, ZMInitialSyncCompletionObserver>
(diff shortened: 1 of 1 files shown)


LucasDang:
I have the same problem like path 3.
My code now is on release/3.24.

PR #1318, opened 2017-10-04
[bugfix]  iOSZIOS-9012 Connection request screen disappears & reveals empty table view

billypchan:
Write the else case for if (self.connectionRequests.count != 0) in ConversationListViewModel.m


mikeger:
@billypchan don't mind the CLA assistant, we'll fix it today.

PR #1322, opened 2017-10-05
[fix]ZIOS-8958 Audio message is not playable after killing and reopening the app

billypchan:
Fix the no audio issue caused by AppDelegate.shared().mediaPlaybackManager?.audioTrackPlayer is nil when AUdioMessageView is created.

Move the audioTrackPlayer's assignment (from app delegate's shared property) from AudioMessageView to AudioMessageCell.

do nil check every time when AudioMessageCell want to asscees audioTrackPlayer and assign again if it is nil.


mikeger:
@billypchan please update styling and then we can merge it.

Issue #1153, opened 2017-07-26
System Sharing Extension and sharing of audio and location


EgbertW:
I encountered some limitations on the system sharing extension, getting in my way.

I wanted to share a location that I selected in the iOS Maps application. I can then select the Share button, but Wire does not show up as a target at all, while apps like Telegram and WhatsApp do. Seeing that I can share locations when I select the option within Wire it shouldn't be hard to fix this I suppose.

Additionaly, I had issues sharing a sound recording made in Telegram. I can select it in Telegram and share it, and Wire does show up as a target. But when I sent it this way, the showed up in Wire as a OGG-file, without Wire seeming to realize that OGG is indeed a valid sound format that it can play. It would be cool if Wire could be used to share recordings made elsewhere.

mikeger:
Hey @MadEgg ! I think the issue is that when we implemented the original share extension Wire used to have no support of location or audio recording sharing. Definitely something we should improve.

IpsmLorem:
The same way, a user cannot share a wallet ticket with wire. Wire doesn't show up in the sharing choice menu. The same thing happend with contact sharing too.

Wire would be nice to be able to share anything.

mikeger:
@IpsmLorem @EgbertW the issue you indicated is most likely fixed in the current version, can you confirm?

EgbertW:
I haven't encountered problems recently!

PR #1325, opened 2017-10-05
Share extension supporting sharing of locations and Wallet passes

nicorsm:
https://github.com/wireapp/wire-ios/issues/1153


typfel:
So how does it look when you share a location now? Does it use our own location message type?

nicorsm:
@typfel like this. It could be improved by using our custom location message type.
pass-example

mikeger:
@nicorsm @typfel I guess we can extract the necessary information to send it as a location message (lat, long) from the VCF file, but sounds like a bigger story, let's create the ticket for it and do it as a separate task.

PR #1327, opened 2017-10-05
Don't show unread dot for the current account in account switcher

typfel:
https://wearezeta.atlassian.net/browse/ZIOS-9011


mikeger:
@typfel should be possible to add the test to `AccountViewTests.swift` for the cases?

PR #1330, opened 2017-10-06
Hiding "No Conversations Yet" label when displaying ConversationListViewController

nicorsm:


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -168,6 +168,7 @@ - (void)viewDidLoad
     [self createViewConstraints];
     [self.listContentController.collectionView scrollRectToVisible:CGRectMake(0, 0, self.view.bounds.size.width, 1) animated:NO];
     
+    [self hideNoContactLabelAnimated:FALSE];
```
typfel: We normally use `YES` and `NO` in Objective-C instead of `TRUE` and `FALSE`

nicorsm: Right 🤦‍♂️

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -604,7 +605,7 @@ - (void)updateNoConversationVisibility;
     if (!self.hasConversations) {
         [self showNoContactLabel];
     } else {
-        [self hideNoContactLabel];
+        [self hideNoContactLabelAnimated:TRUE];
```
typfel: Sorry :) but this should then also be `YES`

PR #1335, opened 2017-10-09
Removed like tutorial

nicorsm:
I removed the "tap to like" tutorial that it was shown the first time a user taps on a message. I also removed the UI code.


Cartfile.resolved
```diff
@@ -11,9 +11,7 @@ github "wireapp/PureLayout" "v3.0.0"
 github "wireapp/TTTAttributedLabel" "2.0.0-swift3.0.2"
 github "wireapp/ZipArchive" "v2.0.8"
 github "wireapp/avs-ios-binaries" "3.6.113"
-github "wireapp/ios-snapshot-test-case" "2.1.4-swift3.1"
```
mikeger: Ok, something fishy here, we still need those frameworks I believe

nicorsm: Fixed, don't know why they were removed 🤔

PR #1336, opened 2017-10-09
[bugfix]ZIOS-9051-Opening-conversation-from-Contacts-UI-lead-to-an-empty-screen(Xcode9)

billypchan:
fix ConversationListViewController alpha = 0 issue.
check self.presentedViewController != nil before dismiss.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
@@ -327,7 +327,16 @@
 
     switch (state) {
         case ConversationListStateConversationList: {
-            [self.presentedViewController dismissViewControllerAnimated:YES completion:completion];
+            self.view.alpha = 1;
+            
+            if (self.presentedViewController != nil) {
+                [self.presentedViewController dismissViewControllerAnimated:YES completion:completion];
+            }
+            else {
+                if (completion) {
+                    completion();
+                }
+            }
         }
             break;
         case ConversationListStatePeoplePicker: {
--- a/Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
+++ b/Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
@@ -36,13 +36,13 @@
 {
     self = [super initWithNibName:nibNameOrNil bundle:nibBundleOrNil];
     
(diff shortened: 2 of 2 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -327,7 +327,16 @@ - (void)setState:(ConversationListState)state animated:(BOOL)animated completion
 
     switch (state) {
         case ConversationListStateConversationList: {
-            [self.presentedViewController dismissViewControllerAnimated:YES completion:completion];
+            self.view.alpha = 1;
```
typfel: Shouldn't this move into the `else` statement?

typfel:
You're changes here are good and make the this code more robust but as you said in the comments in https://wearezeta.atlassian.net/browse/ZIOS-9051

The root cause for this is that the dismiss is called twice. Shouldn't we then also remove the dismiss  in the `InviteContactsViewController` since it anyway is called by `selectConversation:focusOnView:animated:`

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
billypchan: Not sure, but I think it is logical to always make self.view.alpha = 1 when someone want to change the state to ConversationListStateConversationList (of ConversationListViewController).

billypchan:
@typfel Should we always dismiss the view from ZClientViewController?

Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.m
```diff
@@ -57,21 +57,19 @@ - (BOOL)sharingContactsRequired
...
 {
     if (user.isConnected) {
-        [self dismissViewControllerAnimated:YES completion:^{
             [[ZClientViewController sharedZClientViewController] selectConversation:user.user.oneToOneConversation
                                                                         focusOnView:YES
                                                                            animated:YES];
-        }];
     } else if (user.user.isPendingApprovalBySelfUser && ! user.user.isIgnored) {
-        [self dismissViewControllerAnimated:YES completion:^{
+//        [self dismissViewControllerAnimated:YES completion:^{
```
typfel: commented out code

PR #1343, opened 2017-10-10
Fix/local notifications

johnxnguyen:
## What's in this PR?
- Update `ChatHeadsViewController` and `ChatHeadView` to the new local notification changes in Sync Engine.

- [x] depends on: https://github.com/wireapp/wire-ios-sync-engine/pull/589
- [x] depends on: https://github.com/wireapp/wire-ios-sync-engine/pull/608


Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
```diff
@@ -22,21 +22,37 @@ import PureLayout
 
 class ChatHeadView: UIView {
 
-    private var userImageView: ContrastUserImageView!
+    private var userImageView: ContrastUserImageView?
```
typfel: Since you always create this view in `init` it's not really optional

johnxnguyen: but we only create it if the sender is not nil

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
```diff
@@ -153,15 +112,15 @@ class ChatHeadsViewController: UIViewController {
...
         
         // if call notification & in active account
-        if account.isActive && [ZMIncomingCallCategory, ZMMissedCallCategory].contains(note.category ?? "") {
+        if account.isActive && note.isCallingNotification {
             return false
         }
-
+        
         // if current conversation contains message & is visible
         if clientVC.currentConversation === conversation && clientVC.isConversationViewVisible {
```
typfel: Weren't you talking about comparing based on the UUID here so that you don't need to fetch the conversation which might not be saved yet?

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
```diff
@@ -64,65 +65,23 @@ class ChatHeadsViewController: UIViewController {
         }
         
         guard
-            let selfID = note.zm_selfUserUUID,
+            let selfID = note.selfUserID,
             let account = SessionManager.shared?.accountManager.account(with: selfID),
             let session = SessionManager.shared?.backgroundUserSessions[account.userIdentifier],
             let conversation = note.conversation(in: session.managedObjectContext),
```
typfel: Might not yet be saved?

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
typfel: Ah true missed that.

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
```diff
@@ -101,20 +117,25 @@ class ChatHeadView: UIView {
...
     
     private func createImageView() {
         userImageView = ContrastUserImageView(magicPrefix: "notifications")
-        userImageView.userSession = ZMUserSession.shared()
-        userImageView.isUserInteractionEnabled = false
-        userImageView.translatesAutoresizingMaskIntoConstraints = false
-        userImageView.user = self.sender
-        userImageView.accessibilityIdentifier = "ChatheadAvatarImage"
-        addSubview(userImageView)
+        userImageView!.userSession = SessionManager.shared?.backgroundUserSessions[userID]
```
typfel: Put the `userImageView` first into local variable only assign to `self.userImageView` then you don't need to force unwrap everywhere.

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
```diff
@@ -22,21 +22,37 @@ import PureLayout
...
+        NSFontAttributeName: FontSpec(.medium, .none).font!.withSize(14),
+        NSForegroundColorAttributeName: ColorScheme.default().color(withName: ColorSchemeColorChatHeadTitleText)
+    ]
+    private let titleMediumAttributes: [String: AnyObject] = [
+        NSFontAttributeName: FontSpec(.medium, .medium).font!.withSize(14),
+        NSForegroundColorAttributeName: ColorScheme.default().color(withName: ColorSchemeColorChatHeadTitleText)
+    ]
+    
+    private lazy var bodyFont: UIFont = {
+        let font = FontSpec(.medium, .regular).font!
```
vytis: Do we have some support for bigger font sizes in notifications? I think the app has some support through Classy magic?

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
johnxnguyen: in the end we need to have the conversation so that we can pass it to `userSession(:show:)` when we tap on the chathead

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
typfel: I think we should aim to reduce our reliance on Classy.

vytis: Totally agree, too much dynamic magic there!

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
johnxnguyen: but in this case, I would argue that we shouldn't show the chathead because we wouldn't have sufficient information

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
typfel: We could save after the tap and then fetch the conversation?

typfel: I mean otherwise wouldn't we sometimes miss chatheads if someone wrote in a new conversation?

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
johnxnguyen: what do you mean?

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
johnxnguyen: I haven't yet encountered this issue but I thinks it's better to be safer :)

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
```diff
@@ -64,67 +65,29 @@ class ChatHeadsViewController: UIViewController {
...
+            isEphemeral: note.isEphemeral
         )
         
-        chatHeadView!.onSelect = { conversation, account in
-            
+        chatHeadView!.onSelect = {
             SessionManager.shared?.withSession(for: account) { userSession in
-                SessionManager.shared?.userSession(userSession, show: conversation)
+                // save SYNC so we can fetch on UI
+                userSession.syncManagedObjectContext.saveOrRollback()
```
typfel: `syncManagedObjectContext` is private to the `ZMUserSession`. Also It might be threading violation to call save from another thread.

We might need to had a new public method on `ZMUserSession` along the lines of `synchronizeChanges(dispatch_block_t)block`, which switch to the sync mock and saves and then calls the block on the UI moc.

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
```diff
@@ -174,6 +135,19 @@ class ChatHeadsViewController: UIViewController {
         return clientVC.splitViewController.shouldDisplayNotification(from: account)
     }
     
+    /// If the given account is active, the title is trimmed to only include the
+    /// conversation name (i.e, trimming the possible team name). If no conversation
+    /// name is present then nil is returned.
+    ///
+    private func trimTitleIfNeeded(_ title: String, account: Account) -> String? {
+        var result = title
+        if let teamName = account.teamName, let range = title.range(of: "in \(teamName)"), account.isActive {
```
mikeger: As I understand it we want to remove the team reference if this team's account is active?

"in \(teamName)" is not localized, so most likely not going to work in other languages.

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadView.swift
```diff
@@ -131,34 +154,53 @@ class ChatHeadView: UIView {
...
+            }
         }
     }
     
+    private func attributedTitleText(_ title: String) -> NSAttributedString {
+        let attrText = NSMutableAttributedString(string: title, attributes: titleMediumAttributes)
+        
+        // if title contains "in [Team Name]", then we want "in" to be regular font
+        if let teamName = teamName {
+            if let teamRange = title.range(of: "in \(teamName)") {
```
mikeger: Same remark: the "in \(teamName)" can be localized.

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
johnxnguyen: Yes that's correct. But if it's not localized, wouldn't the "in" always remain the same?

mikeger: As I expect it to work, the message would be "In Wire: text", whereas in other language it would be "В Вайр:" so replace would not happen.

johnxnguyen: ok, currently the title is not constructed using localized strings, so I thought the "in" would remain constant. But I will change it so it is localized

mikeger: Good idea, we support 17 languages now, I think it would be fair to translate all the strings we are using in order to enable the translation for all those languages.

johnxnguyen:
@mikeger I've removed any reference to "in". Storing the the conversation name and team name in the `ZMLocalNotification`'s `userInfo` allows us to just use the conversation name if the account is active, and to style only these two strings with medium font for the title

PR #1346, opened 2017-10-11
Fixed sizes for share view controller on iPhone X

nicorsm:
Share View Controller was shown without margins on iPhone X.

### DIFF
--- a/WireExtensionComponents/Controllers/ShareViewController+Views.swift
+++ b/WireExtensionComponents/Controllers/ShareViewController+Views.swift
@@ -92,7 +92,10 @@ extension ShareViewController {
     internal func createConstraints() {
         constrain(self.view, self.blurView, self.containerView) { view, blurView, containerView in
             blurView.edges == view.edges
-            containerView.edges == view.edges
+            containerView.top == view.top + safeArea.top
+            containerView.bottom == view.bottom - safeArea.bottom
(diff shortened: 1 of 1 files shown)


WireExtensionComponents/Controllers/ShareViewController+Views.swift
```diff
@@ -162,4 +165,13 @@ extension ShareViewController {
             sendButton.bottom == -12 + view.bottom
         }
     }
+    
+    
+    var safeArea: UIEdgeInsets {
+        let size = UIScreen.main.bounds.size
+        if size.width == 375 && size.height == 812 {
+            return UIEdgeInsets(top: 44.0, left: 0, bottom: 34.0, right: 0)
```
nicorsm: I need to find a way to get these sizes programmatically, without hardcoding. The `UIScreen+SafeArea.swift` does not work because `UIApplication.shared` is not available in extensions.

typfel: Ah weird that they didn't provide a way to access it.

nicorsm: They provide it only via `UIApplication.shared.keyWindow`. There's a `safeAreaInsets` property, but when I call it from `self.view` of the `ShareViewController` it returns zero -.-"

PR #1347, opened 2017-10-11
Refactor NetworkActivityViewController and NetworkStatusViewController

typfel:
NetworkActivityViewController and NetworkStatusViewController have been replaced by a
new NetworkStatusViewController which does both tasks. The new controller is only
listen to the ZMNetworkState changes from SE

Offline bar has been removed from the registration flow. (Should we replace it with something?)

Observing Reachabillity and ZMNetworkState separately will lead to inconistency problems
were we are offline and syncing at the same.


Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+        offlineBarTimer = Timer.scheduledTimer(timeInterval: 2.0, target: self, selector: #selector(collapseOfflineBar), userInfo: nil, repeats: false)
+    }
+    
+    internal func collapseOfflineBar() {
+        if networkStatusView.state == .offlineExpanded {
+            update(state: .offlineCollapsed)
+        }
+    }
+    
+    fileprivate func enque(state: NetworkStatusViewState) {
```
mikeger: Typo: sounds the same, but I think `enqueue` is the correct spelling.

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+        
+        view.addSubview(networkStatusView)
+        
+        constrain(self.view, networkStatusView) { containerView, networkStatusView in
+            networkStatusView.left == containerView.left
+            networkStatusView.right == containerView.right
+            networkStatusView.top == containerView.top
+        }
+        
+        if let userSession = ZMUserSession.shared() {
```
mikeger: Should it crash if there is no user session?

typfel: Yes maybe that's better.

mikeger: Would break tests I guess 🗡

typfel: I'll try :)

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -0,0 +1,135 @@
...
+    internal func collapseOfflineBar() {
+        if networkStatusView.state == .offlineExpanded {
+            update(state: .offlineCollapsed)
+        }
+    }
+    
+    fileprivate func enqueue(state: NetworkStatusViewState) {
+        pendingState = state
+        NSObject.cancelPreviousPerformRequests(withTarget: self, selector: #selector(applyPendingState), object: nil)
+        perform(#selector(applyPendingState), with: nil, afterDelay: 1)
```
billypchan: Can I rewrite this with Timer.scheduledTimer()? I found that this function retains self if cancelPreviousPerformRequests did not called.

typfel: A timer will also retain the target, no?

typfel: Discussed with Bill in person and he'll replace this with a weak timer.

PR #1348, opened 2017-10-11
Only allow sounds to be played while app is active

typfel:
## Problem
Receiving a call in the background shows the media controls while playing the ringtone

## Solution
Don't play sounds while in the background. Ringtone will be played via the configured sound for the push notification / Callkit


Wire-iOS/Sources/Managers/SoundEventListener.swift
```diff
@@ -208,12 +210,16 @@ extension SoundEventListener {
             soundEventWatchDog.ignoreTime = 0.0
         }
     }
+    
+    func applicationDidEnterBackground() {
+        soundEventWatchDog.isMuted = true
```
mikeger: I think the isMuted is not send to false when the app is coming to the foregorund, or initialSyncCompleted is called then?

typfel: Was also thinking that but looking at https://github.com/wireapp/wire-ios-sync-engine/blob/3f65406522352953274443d6b2acc44ecbce6d5c/Source/UserSession/SyncStatus.swift#L162

it seems like it called after every sync

typfel:
Yes, so let's change it to observing the `ZMNetworkState` instead since that's really what it's doing now.

PR #1354, opened 2017-10-13
iOS 11, "minor" fixes

nicorsm:
In this PR:

- Fixed the "Invite more people" different behaviour between iOS 10 and 11
- Fixed the navigation bar size on iOS 10 (ZIOS-9063)
- Fixed conversation names inside navigation bar on iOS 11 (ZIOS-9074)
- Fixed a bug where switching between dark and light themes wasn't updating the status bar style on conversations until restart.
- Removed dumb animation on the conversation view while lightly swiping to conversation list and coming back
- Removed the blank space occurring with 3 logged accounts on Self Profile
- Updated bottom insets on keyboard for iPhone X (camera, location share, canvas)
- Added safe area to group creation and group info view controllers, media asset confirmation and Photogallery, participants view controller
- Fixed a bug on iOS 11 that was displaying an inset (taller as the screen size) on the connection requests view controller.
- updated UI tests


mikeger:
In regards of the new images `testThatItRendersTheConversationDisplayNameCorrectly_{98, 26.5}@2x.png` and `testThatItUpdatesTheTitleViewAndRendersTheVerifiedShieldCorrectly_{117, 26.5}@2x.png` & others: the image size is in the file name, so if your control is changing the size then record mode is creating the new image. You would have to delete the old image manually, i.e. the one with the name `testThatItUpdatesTheTitleViewAndRendersTheVerifiedShieldCorrectly_{<OLD SIZE>}@2x.png` where <OLD SIZE> is the size this control had before.

nicorsm:
@mikeger I'll check that in the next commit. Thanks! :)

Wire-iOS/Sources/UserInterface/Components/UINavigationBarContainer.swift
```diff
@@ -0,0 +1,67 @@
...
+            navigationBar.left == view.left
+            navigationBar.right == view.right
+            navigationBar.bottom == view.bottom
+        }
+    }
+    
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        guard let topMargin = topMargin, let navHeight = navHeight else { return }
+        let orientation = UIDevice.current.orientation
```
mikeger: I think that the device orientation is not good to check the interface orientation, since it includes 2 additional orientations (faceUp, faceDown) that are not convertible to the real UI orientation: https://developer.apple.com/documentation/uikit/uideviceorientation

mikeger: May be you can assign the UI orientation on this view from controller?

Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -45,7 +45,7 @@ final class ConversationImagesViewController: UIViewController {
     
     let collection: AssetCollectionWrapper
     
-    fileprivate let navigationBar = UINavigationBar()
+    fileprivate var navBarContainer: UINavigationBarContainer?
```
johnxnguyen: it looks like you're expecting this to always to be non nil, maybe you could implicitly force unwrap it here so that you don't have to explicitly force unwrap it every time you use it?

PR #1357, opened 2017-10-16
 Re-added the app icon

mikeger:
- During the recent app store submit we received the complaint from Apple that we must include the AppStore icon (1024px) to the bundle.


vytis:
Large icons looks pretty horribly scaled up...

billypchan:
The current icon has a plain white background but the Wire-83,5@2x for iPad Pro has a gradient background. Should the new icon follow the iPad Pro style?

mikeger:
Ok, I better only add the new icon, they are not all the same so re-generating the icons seems to bring more issues.

PR #1360, opened 2017-10-16
Fix/zios 9075 setting  username text field has wrong characters when using chinese cangjie input method

billypchan:
tested for both Chinese and English input with space cases.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/TailEditingTextField.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/TailEditingTextField.swift
@@ -46,11 +46,15 @@ class TailEditingTextField: UITextField {
         self.addTarget(self, action: #selector(TailEditingTextField.replaceNonBreakingSpacesWithNormalSpaces), for: UIControlEvents.editingDidEnd)
     }
     
-    func replaceNormalSpacesWithNonBreakingSpaces() {
-        self.text = self.text?.replacingOccurrences(of: type(of: self).normalSpace, with: type(of: self).nonBreakingSpace)
+    func replaceNormalSpacesWithNonBreakingSpaces(_ sender: Any) {
+        if let isContainsNormalSpace = (self.text?.contains(type(of: self).normalSpace)), isContainsNormalSpace {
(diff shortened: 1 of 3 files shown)


typfel:
Could explain why these changes solves the bug? Just reading the code, it looks like the behaviour is identical to before this changes. Shouldn't `replacingOccurrences` do nothing if there's no matches so the only difference is that we aren't re-setting the `self.text` property...

billypchan:
@typfel every time setting self.text property deselects the text field and chinese cangjie input method keyboard can not clean up the cangjie coding.

Wire-iOS/Sources/UserInterface/Settings/TailEditingTextField.swift
```diff
@@ -46,11 +46,15 @@ class TailEditingTextField: UITextField {
         self.addTarget(self, action: #selector(TailEditingTextField.replaceNonBreakingSpacesWithNormalSpaces), for: UIControlEvents.editingDidEnd)
     }
     
-    func replaceNormalSpacesWithNonBreakingSpaces() {
-        self.text = self.text?.replacingOccurrences(of: type(of: self).normalSpace, with: type(of: self).nonBreakingSpace)
+    func replaceNormalSpacesWithNonBreakingSpaces(_ sender: Any) {
+        if let isContainsNormalSpace = (self.text?.contains(type(of: self).normalSpace)), isContainsNormalSpace {
```
typfel: could be turned into a guard

Wire-iOS/Sources/UserInterface/Settings/TailEditingTextField.swift
```diff
@@ -46,11 +46,15 @@ class TailEditingTextField: UITextField {
...
+            self.text = self.text?.replacingOccurrences(of: type(of: self).normalSpace, with: type(of: self).nonBreakingSpace)
+        }
     }
     
     func replaceNonBreakingSpacesWithNormalSpaces() {
-        self.text = self.text?.replacingOccurrences(of: type(of: self).nonBreakingSpace, with: type(of: self).normalSpace)
+        if let isContainsNonBreakingSpace = (self.text?.contains(type(of: self).nonBreakingSpace)), isContainsNonBreakingSpace {
```
typfel: could be turned into a guard

PR #1368, opened 2017-10-20
resume Music playback after audio message completed playback.

billypchan:
Tested with iOS Music app and confirmed issue is fixed.
(Spotify can not be resumed)


typfel:
Let's sync up with the AVS team on these changes because they are controlling the audio category in our app.

mikeger:
Hey @z-dule , can you comment on this? Should we do it here for audio messages?

z-dule:
Please do not control the active state or the category outside of the mediamanager, doing so will break the purpose of the MediaManager.

mikeger:
@z-dule @typfel  should we (or are we already) tell mediamanger that we are about to play / finished playing the audio message?

z-dule:
@mikeger don't you already play the sound through the MediaManager?

mikeger:
As a matter of fact we do, through `MediaPlaybackManager`

mikeger:
@billypchan sorry for the confusion, @z-dule is going to take care of this case via AVS MediaManager.

PR #1370, opened 2017-10-23
Fix/zios 9130 improve ns string normalized filename to prevent app crash when user name can not convert to a posix standard file name

billypchan:
depends on: 
- [ ] https://github.com/wireapp/wire-ios-utilities/pull/29
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/385

fix: The video filename fit POSIX standard.
tested with username = "🇭🇰🇭🇰🇭🇰🇭🇰🇭🇰🇭🇰🇭🇰" case.


vytis:
Actually it would a good opportunity to move this functionality to https://github.com/wireapp/wire-ios-utilities. It would also be pretty easy to add some tests in there

billypchan:
@vytis I am thinking should I decouple ZMUser from the NSString extension? i.e. provide a string param "username" instead of access ZMUser singleton.

code:
extension NSString {
    static func filenameForSelfUser() -> NSString {
        return "\(ZMUser.selfUser().name!.normalizedFilename)-\(dateFormatter.string(from: Date()))" as NSString
    }

vytis:
It depends on where/how we use `filenameForSelfUser` method, maybe it should then be moved to an extension on `ZMUser` and the `normalizedFilename` could go to utilities then.

Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
```diff
@@ -29,13 +29,37 @@ extension NSString {
...
+    /// Convert to a POSIX "Fully portable filenames" (only allow A–Z a–z 0–9 . _ -)
+    /// Space will be converted to underscore first.
     var normalizedFilename: String {
         let ref = NSMutableString(string: self) as CFMutableString
         type(of: self).transforms.forEach { CFStringTransform(ref, nil, $0, false) }
-        return (ref as String).replacingOccurrences(of: " ", with: "-")
+        
+        let retString = (ref as String).replacingOccurrences(of: " ", with: "-")
+        
+        let characterSet = NSMutableCharacterSet() //create an empty mutable set
```
mikeger: 1. We can also use the Swift `CharacterSet`: https://developer.apple.com/documentation/foundation/characterset
2. Would be cool to move the creation of the character set to the static initializer in the category of `CharacterSet`, like 
```
extension CharacterSet {
    static func portableFilenameCharacters() -> CharacterSet
}
```

Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
```diff
@@ -29,13 +29,37 @@ extension NSString {
...
+
+        let unsafeChars = characterSet.inverted
+        let strippedString = retString.components(separatedBy: unsafeChars).joined(separator: "")
+        
+        return strippedString
     }
 
+    /// return a file name with length < 255 - 4(reserve for extension) - 37(reserve for WireDataModel UUID prefix for meta) characters
+    ///
+    /// - Returns: a string <= 214 characters
```
mikeger: Reminds me of DOS times somehow :hurtrealbad: 
```
┌─────────[X]┐
│MYFILE~1.TXT│
└────────────┘
```

Wire-iOS/Sources/UserInterface/Helpers/NSString+Filename.swift
billypchan: @vytis I would keep filenameForSelfUser in wire-iOS since it is depends on ZMUser.selfUser() (defined in "ZMUser+Additions.h" form wire-iOS), which is only used for UI.

mikeger: @billypchan you can do an extension on `ZMUser`, like `filenameForUser`, and then call it in UI on self user.

PR #1371, opened 2017-10-23
Fixed layout issues with lateral margins on Search Results

nicorsm:
Tested with iPhone 6 (iOS 11), iPhone 6 (iOS 10) and iPhone X. 

simulator screen shot - iphone 6 - 2017-10-23 at 15 32 54


typfel:
I think there's a regression in the add people screen because of this:
screen shot 2017-10-23 at 18 02 59

nicorsm:
😕 I'll take a look.

PR #1377, opened 2017-10-25
Fix to contents shared via Share Extension- links, images and passes

nicorsm:
In this PR:

- Fixed an issue that was causing the Share Extension to getting stuck in "Preparation..." while sending content from third party apps (Twitter, Safari, etc.) (`72253db`)
- Fixed an issue that was preventing images to be attached to the shared content while sending from third party apps (`20334ca`)
- Fixed an issue that was preventing Wallet passes to be attached to the shared content (`3d01fd2`)
- Added a control that appends the link to the shared content, if the link is not in the shared content or if the user accidentally deletes it. (`1ad8a1c`, `c7b0a8f`)


Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -68,21 +68,51 @@ class UnsentSendableBase {
...
         sharingSession.enqueue { [weak self] in
             guard let `self` = self else { return }
             let fetchPreview = !ExtensionSettings.shared.disableLinkPreviews
-            completion(self.conversation.appendTextMessage(self.text, fetchLinkPreview: fetchPreview))
+            let message = self.conversation.appendTextMessage(self.text, fetchLinkPreview: fetchPreview)
+            completion(message)
+        }
+    }
+    
+    func prepare(completion: @escaping () -> Void) {
```
typfel: I don't see this method being called anywhere? was it already defined somewhere else?

nicorsm: this is called when an URL Attachment is available (see line 78). I need it because the call to `fetchURL` is async.

typfel: Found it :)

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -68,21 +68,51 @@ class UnsentSendableBase {
...
+        precondition(needsPreparation, "Ensure this objects needs preparation, c.f. `needsPreparation`")
+        needsPreparation = false
+        
+        if let attachment = self.attachment, attachment.hasURL {
+            
+            self.attachment?.fetchURL(completion: { (url) in
+                
+                if let url = url?.absoluteString, !self.text.contains(url)  {
```
typfel: It's not so obvious what this does on a first glance, I would extract it out into a method called `appendURLToTextIfNotAlreadyPresent(_ URL : URL)`

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -68,21 +68,54 @@ class UnsentSendableBase {
...
+        needsPreparation = false
+        
+        if let attachment = self.attachment, attachment.hasURL {
+            
+            self.attachment?.fetchURL(completion: { (url) in
+                self.appendURLToTextIfNotAlreadyPresent(url)
+                completion()
+            })
+        } else {
+            completion()
```
typfel: If the attachment has no URL and the message is empty will then also be stuck in preparing?

nicorsm: I don't think so, but I'm not sure 100%. Before that, the message was stuck because the URL was recognised as `UnsentFileSendable` on `SendController:58`.

typfel: Ok

PR #1378, opened 2017-10-25
Workaround for failing to pre-fill password on iOS11

vytis:
The reason behind this is still unknown, maybe some bug in UIKit. It seems that if we switch the order of login and registration tabs (thus showing login first, not registration) all seems to be working fine. Another workaround was to simply activate the password field and then put focus to email field immediately.


johnxnguyen:
@vytis is there a noticeable glitch when you do this?

vytis:
There is a slight glitch, but barely visible. It seems that the key icon just flashes, nothing major at all.

PR #1379, opened 2017-10-25
Added associated domains

vytis:
This will enable us to have proper credentials autofill functionality


typfel:
We also don't need to update `Entitlements-Prod.entitlements`?

vytis:
Good catch! We definitely do

PR #1382, opened 2017-10-27
Added permission for writing to photo library

vytis:
Also one of the permissions text string was not present in the file.

### DIFF
--- a/Wire-iOS/Wire-Info.plist
+++ b/Wire-iOS/Wire-Info.plist
@@ -9,7 +9,7 @@
 	<key>NSCameraUsageDescription</key>
 	<string>In order to take and send pictures Wire needs access to your camera.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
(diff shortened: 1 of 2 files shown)


Wire-iOS/Wire-Info.plist
```diff
@@ -91,6 +91,8 @@
 	</dict>
 	<key>NSContactsUsageDescription</key>
 	<string>We’d like to access your Contacts so we can connect you with people you might know. Your data is safely stored with us and not shared with anyone else.</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>In order to store pictures you take Wire needs access to your  photo library.</string>
```
mikeger: Sometimes  you see them, but sometimes  not! Double space :)

vytis: I copy-pasted from the info.plist. I think it's some sort of non-breaking space. Still one character it seems

vytis: O yeah, was looking in a different place, nevermind

PR #1387, opened 2017-11-01
Improved flow for managing team

mikeger:
- Manage team item is only shown for admins and owners. Button is taking user to the login screen of the admin panel.
- Add item is showing now add team or add account, add team takes to the team creation flow, add user opens the log in screen.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
@@ -38,7 +38,7 @@ import Foundation
     func rootGroup() -> SettingsControllerGeneratorType & SettingsInternalGroupCellDescriptorType {
         var rootElements: [SettingsCellDescriptorType] = []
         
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -36,13 +36,14 @@ import Foundation
     }
     
     func rootGroup() -> SettingsControllerGeneratorType & SettingsInternalGroupCellDescriptorType {
-        var rootElements = [self.createTeamCell()]
+        var rootElements: [SettingsCellDescriptorType] = []
         
-        if SessionManager.shared?.accountManager.accounts.count < SessionManager.maxNumberAccounts {
-            rootElements.append(self.addAccountCell())
+        if ZMUser.selfUser().membership?.permissions.contains(.owner) ?? false || ZMUser.selfUser().membership?.permissions.contains(.admin) ?? false {
```
vytis: Maybe it would be good to move all permissions logic to one place? In this case we could have a flag `canManageTeam` which returns true when user is owner or admin. If new permissions are added it would be easier to update the logic and we could see all of it in one place.

mikeger:
@vytis updated

PR #1395, opened 2017-11-03
fix for filename with custom suffix would not exist 256-char limit

billypchan:


vytis:
Would be good if you could add a link to the Data Model pull request that this depends on

PR #1396, opened 2017-11-03
Crash on sharing images from Photo Library

nicorsm:
There was a bug when sharing images from the Photo Library. The extension was skipping the control to load pictures by using the local URL. Now, it tries to load the content from local URL first - if it fails, it loads the content directly.


mikeger:
@nicorsm should we merge it to `release/3.2` ?

nicorsm:
@mikeger I tested the changes with Photo, Safari, Twitter, Wallet and App Store, with images and/or links and everything works fine!

PR #1397, opened 2017-11-06
Fixed weak reference for SavableImage object in CollectionImageCell

nicorsm:
Images weren't saved while tapping on the "Save" button of the UIMenuController (via long press) due to the weak reference of the object - it was deallocated before completing the operations.

Related to ZIOS-9181.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionImageCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionImageCell.swift
@@ -161,8 +161,8 @@ final public class CollectionImageCell: CollectionCell {
         }
         
         saveableImage = SavableImage(data: imageData, orientation: orientation)
-        saveableImage?.saveToLibrary(withCompletion: {
-            self.saveableImage = nil
+        saveableImage?.saveToLibrary(withCompletion: { [weak self] in
+            self?.saveableImage = nil
         })
     }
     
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionImageCell.swift
```diff
@@ -153,13 +153,15 @@ final public class CollectionImageCell: CollectionCell {
...
+    
     func save(_ sender: AnyObject!) {
         guard let imageData = self.message?.imageMessageData?.imageData, let orientation = self.imageView.image?.imageOrientation else {
             return
         }
         
-        let savableImage = SavableImage(data: imageData, orientation: orientation)
-        savableImage.saveToLibrary()
+        savableImage = SavableImage(data: imageData, orientation: orientation)
+        savableImage?.saveToLibrary()
```
typfel: So this is an async operation? Couldn't this operation retain the reference until the save is complete?

Wire-iOS/Sources/UserInterface/Collections/CollectionImageCell.swift
```diff
@@ -153,13 +153,15 @@ final public class CollectionImageCell: CollectionCell {
         UIPasteboard.general.setMediaAsset(UIImage(data: imageData))
     }
     
+    var savableImage : SavableImage?
```
vytis: Should probably be "saveable"

Wire-iOS/Sources/UserInterface/Collections/CollectionImageCell.swift
nicorsm: Yes, the code inside `saveToLibrary()` does an async call to `PHPhotoLibrary.shared().performChanges()` with a completion block (`SavableImage.swift:40`)

PR #1398, opened 2017-11-06
Fix analyzer warnings

vytis:
We have not been running analyzer in a while, so there were quite a few errors. Will try to add a report in Danger about it so we would not forget to do it in the future.


Wire-iOS/Sources/UserInterface/ContactsUI/ContactsCell.m
```diff
@@ -100,7 +100,7 @@ - (void)setupSubviews
     [self.userNameContainerView addSubview:self.userSubtitleLabel];
     
     self.actionButton = [Button buttonWithStyleClass:@"dialogue-button-full"];
-    [self.actionButton setTitle:@"INVITE" forState:UIControlStateNormal];
+    [self.actionButton setTitle:NSLocalizedString(@"contacts_ui.action_button.invite", @"") forState:UIControlStateNormal];
```
mikeger: This is just a placeholder, we set the actual title in https://github.com/wireapp/wire-ios/blob/649e53c75c31ffdf30d9efcad28003b46b8d68fb/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.m#L456 , but should not hurt to silence the warning.

vytis: So maybe we can just not set it at all?

mikeger:
@vytis All good, but what's the new Danger error `summary file not found` ?

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsCell.m
mikeger: I'm not 100% sure, it could be that something breaks if it's empty, I think your solution is fine.

vytis:
Basically it takes info from custom `xcpretty` formatter. I tried to fix it few times but with iteration cycles of 20 mins I need to do it first on my local machine. That's why I added 'do not merge' for now.

WireExtensionComponents/Views/TokenField/TokenField.m
```diff
@@ -604,7 +604,7 @@ - (NSAttributedString *)stringForTokens:(NSArray *)tokens
 
 - (NSAttributedString *)collapsedString
 {
-    NSString *collapsedText = @" ...";
+    NSString *collapsedText = NSLocalizedString(@" ...", nil);
```
typfel: So you can't define any string without getting analyzer warning, or it's smart enough to check if it ends up in a label?

vytis: I think UIKit is annotated with which properties and parameters are user-facing, so it checks usage of those.

PR #1399, opened 2017-11-06
Fix/zios 9130 improve ns string normalized filename to prevent app crash when user name can not convert to a posix standard file name

billypchan:
apply new interface filenameForSelfUser(suffix:) to AudioRecordKeyboardViewController
clean up other Swift filse using filenameForSelfUser (return String instead ofNSString )


Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordKeyboardViewController.swift
```diff
@@ -194,11 +194,11 @@ import CocoaLumberjackSwift
             topContainer.left >= view.left + 16
             topContainer.top >= view.top + 16
             topContainer.right <= view.right - 16
-
+
```
vytis: To avoid these sort of changes there is a handy (and not well known) checkbox in Xcode "Text Editing" preferences:
screen shot 2017-11-06 at 14 31 52

mikeger:
@billypchan please bump the data model on this PR

PR #1405, opened 2017-11-07
Fix/zios 9199 prevent app start in slide mode with splitted layout

billypchan:
imp. viewWillTransition() for AppRootViewController


billypchan:
@mikeger It is part of the reasons, I am look into other parts of the code which may breaks the constraints in slide over/split mode.

billypchan:
continue working in https://github.com/wireapp/wire-ios/pull/1408

PR #1408, opened 2017-11-08
Fix/zios 9226 review login screen view size class handling

billypchan:
retested registration workflow with iPad mini 2.


Wire-iOS/Sources/UserInterface/Registration/ProfilePictureStepViewController.m
```diff
@@ -167,8 +167,8 @@ - (void)createConstraints
     
     [self.profilePictureImageView autoPinEdgesToSuperviewEdgesWithInsets:UIEdgeInsetsZero];
     [self.overlayView autoPinEdgesToSuperviewEdgesWithInsets:UIEdgeInsetsZero];
-    
-    if (self.traitCollection.userInterfaceIdiom == UIUserInterfaceIdiomPad) {
+
+    if ([SizeClass isIPadInFullScreenMode]) {
```
typfel: Let's aim to not have these kind of `if-constraints` in the new registration. It will break when the re-sizes the app, but let's not fix it now.

PR #1409, opened 2017-11-08
Fix/zios 9227 replace is ipad macro

billypchan:
replace all IS_IPAD macros and UIDevice currentDevice].userInterfaceIdiom == UIUserInterfaceIdiomPad with Swift 

tested the UI appearance of ParticipantsHeaderView, fixed the issue cancel button is not shown in slide over mode.


mikeger:
@billypchan All the code that is checking if it's an iPad is not entirely aligned with the guidelines, since according to Apple we should be good with the size class. Most of this code as you see is in Objective-C. It is because we decided not to port the old code to the new system. In the new swift code we try not to rely on the fact that we are working on the iPad. We should not rely on iPad / not-iPad logic looking forward, therefore I would prefer not to expose the `SizeClass` class to Swift/new code since someone can think that it is acceptable to use it.

mikeger:
In quite some places you are changing the logic from IS_IPAD to check also the full screen mode, have you tested all the cases changed?

billypchan:
@mikeger Thanks. I will rewrite it as a macro. This PR is focus on ParticipantsHeaderView and MagicConfig. More minor issues is WIP.

mikeger:
@billypchan I think it's OK for it to be a method, may be just port it to Obj-C. Thanks!

billypchan:
@mikeger I will separate the logics of "which kind of device" and "what kind of size class" into two parts in next PR.

PR #1410, opened 2017-11-08
Fix/multi acount sound

johnxnguyen:
## So what's it all about anyway?
Currently the wireapp only plays notification sounds for new messages & calls for the active account. The reason being, that only a single `SoundEventListener` object, which plays the sounds in response to new messages and call state changes, was being owned by `ZClientViewController` and thus tied only to the active account.

This PR adds sonic abilities for multiple accounts by creating a new `SoundEventListener` for each account. These listeners have been moved up to a higher layer and are now owned by `AppRootViewController`, which creates and destroys these objects in response to notifications posted by `SessionManager` (see the related PR below).

## Known Issue: first message for inactive account doesn't play sound
The `SoundEventListener` is created after the `ZMUserSession` is created, and so if a user session is created because of a remote notification (for a new message), then by the time the listener exists it is too late to detect the new message changes, resulting in no sound. 

- [x] depends on: https://github.com/wireapp/wire-ios-sync-engine/pull/626


mikeger:
@johnxnguyen I had an impression that we discussed it in the way that we would move the sound event listener to be the part of user session, but might be that I was thinking about it but haven't shared my thoughts with you :) current approach is looking good, but may be you can consider my idea, then:
1. Class `SoundEventListener` is moved to wire-ios-sync-engine.
2. Every instance of `ZMUserSession` has a property `soundEventListener` which is set to a new instance when `ZMUserSession` is created.
Then:
1. It is not necessary to observe the user session creation.
2. From UI perspective it just works.

Please think about it, let's discuss around it on Monday.

johnxnguyen:
@mikeger hey yes I did try this approach first but I found that moving ‘SoundEventListener’ to sync engine required a few other changes to remove the dependency of the UI, namely the keys for the sound names and also referencing AppDelegate. Wouldn’t we also need to move the sound files to SE too?

PR #1414, opened 2017-11-10
Do not create the preview of malformed PDF

mikeger:
# Issue

PDF with the layout size of 0 to 0 points would make the existing preview generator crash.

# Solution

Fail the preview generation in the graceful way.


mikeger:
Bonus: mac file preview cannot handle this file as well

screen shot 2017-11-13 at 09 36 17

PR #1415, opened 2017-11-13
Add option to disable sound for messages, ping and calls

typfel:
With this PR we support muting the sound separately for pings, messages and calls.

depends on:
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/632


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -93,14 +97,15 @@ class SettingsPropertyFactory {
             // Profile
         case .profileName:
             let getAction: GetAction = { [unowned self] (property: SettingsBlockProperty) -> SettingsPropertyValue in
-                return SettingsPropertyValue.string(value: self.selfUser.name)
+                return SettingsPropertyValue.string(value: self.selfUser?.name ?? "")
             }
             let setAction: SetAction = { [unowned self] (property: SettingsBlockProperty, value: SettingsPropertyValue) throws -> () in
                 switch(value) {
                 case .string(let stringValue):
+                    guard let selfUser = self.selfUser else { break }
```
mikeger: Should it assert when there is no self user?

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -79,7 +79,11 @@ class SettingsPropertyFactory {
         SettingsPropertyName.disableLinkPreviews        : UserDefaultDisableLinkPreviews,
     ]
     
-    init(userDefaults: UserDefaults, tracking: TrackingInterface?, mediaManager: AVSMediaManagerInterface?, userSession: ZMUserSessionInterface, selfUser: SettingsSelfUser) {
+    convenience init(userSession: ZMUserSessionInterface? = nil, selfUser: SettingsSelfUser? = nil) {
```
mikeger: This makes it easily breakable, I would vote to remove the default values, so user can create the instance, but only when explicitly passing the nil arguments.

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
typfel: Sure let's assert

PR #1416, opened 2017-11-13
Enable Callkit immediately after granted audio permissions

typfel:
## Problem
If the user has not granted us audio permission we will not enable Callkit, because the call wouldn't be established if answered outside of the app. The user would then need to restart app in order to have Callkit activated. This behaviour breaks automation test which can't restart the app in the middle of a test.

## Solution
Attempt to re-activate Callkit immediately after being granted audio permissions.

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
+++ b/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
@@ -24,15 +24,25 @@
 
 #import <AVFoundation/AVFoundation.h>
 
+NSString * const UserGrantedAudioPermissionsNotification = @"UserGrantedAudioPermissionsNotification";
+
 @implementation UIApplication (Permissions)
 
 + (void)wr_requestOrWarnAboutMicrophoneAccess:(void(^)(BOOL granted))grantedHandler
 {
+    BOOL audioPermissionNotDetermined = [AVCaptureDevice authorizationStatusForMediaType:AVMediaTypeAudio] == AVAuthorizationStatusNotDetermined;
+    
     [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
+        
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
```diff
@@ -24,15 +24,25 @@
...
+    BOOL audioPermissionNotDetermined = [AVCaptureDevice authorizationStatusForMediaType:AVMediaTypeAudio] == AVAuthorizationStatusNotDetermined;
+    
     [[AVAudioSession sharedInstance] requestRecordPermission:^(BOOL granted) {
+        
         dispatch_async(dispatch_get_main_queue(), ^{
             if (! granted) {
                 [self wr_warnAboutMicrophonePermission];
             }
+            
+            if (audioPermissionNotDetermined && granted) {
```
vytis: So what is "not determined" in this case? Before the permission was even asked? And'ing these two things sounds weird...

typfel: "not determined" is before the user has been asked yes. I can give it more obvious name like `willBeAskedForAudioPermissions`

vytis: so then maybe the flag should be inverted - `!audioPermissionNotDetermined`?

typfel: No, it should work like does now. I only want to send a notification after the first time the user grants us the audio permissions.

It's a bit confusing but the logic is: Send a notification if `we are granted audio permissions` *AND* `we have not asked for permissions previously`

vytis: Indeed confusing :D thanks for clarification!

PR #1417, opened 2017-11-14
Improved the touch area of close button in full screen image view

mikeger:
# Issue

The close button is hard to click from the first attempt.

# Solution

Increase the size and the tap area of the button.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionsView.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionsView.swift
@@ -89,7 +89,6 @@ import Cartography
         button.setIcon(.X, with: .tiny, for: .normal)
         button.frame = CGRect(x: 0, y: 0, width: 48, height: 32)
         button.accessibilityIdentifier = "close"
-        button.hitAreaPadding = CGSize(width: 16, height: 16)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionsView.swift
```diff
@@ -87,9 +87,10 @@ import Cartography
     public static func closeButton() -> IconButton {
         let button = IconButton.iconButtonDefault()
         button.setIcon(.X, with: .tiny, for: .normal)
-        button.frame = CGRect(x: 0, y: 0, width: 32, height: 20)
+        button.frame = CGRect(x: 0, y: 0, width: 48, height: 32)
         button.accessibilityIdentifier = "close"
-        button.imageEdgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: -16)
+        button.hitAreaPadding = CGSize(width: 16, height: 16)
```
typfel: icon buttons have a hit area padding of 20x20 by default.

typfel: So I wonder by the hit area wasn't big enough already? Usually it's because the button itself is inside another view which cancels the hit area padding.

PR #1419, opened 2017-11-14
Pull request template

nicorsm:
Proposal of a template for pull requests.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,10 @@
 
 _Briefly describe the issue you have solved or implemented with this pull request. If the PR contains multiple issues, use a bullet list._
 
+### Causes
+
+_Briefly describe the causes behind the issues. This could be helpful to understand the adopted solutions behind some nasty bugs or complex issues. (Optional)_
+
 ### Solutions
 
 _Briefly describe the solutions you have implemented for the issues explained above._
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,25 @@
+## What's new in this PR?
+
+### Issues
+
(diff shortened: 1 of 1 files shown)


johnxnguyen:
@nicorsm also I find sometimes it's useful after describing the issue, to describe the cause behind the issue. That usually leads nicely into the description of the solution. But of course, it's not always necessary, only for particularly nasty bugs that don't have an obvious reason.

nicorsm:
@johnxnguyen Sure!! It could be an optional field to add after "Issues".

.github/pull_request_template.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Solutions
+
+_Briefly describe the solutions you have implemented for the issues explained above._
+
+## Dependencies
+
+_If this issue depends from other issues available on GitHub, specify them here. (Optional)_
+
+This issue depends on:
```
vytis: I think this section could be more descriptive and say something like `Needs releases with:` because most of the time it's not really other issues/tickets that need to be solved, but code updated and merged.

vytis:
I think we can check the template by looking into several successful PRs and their descriptions (just few examples from recently):

https://github.com/wireapp/wire-ios/pull/1411
https://github.com/wireapp/wire-ios/pull/1410
https://github.com/wireapp/wire-ios/pull/1388
https://github.com/wireapp/wire-ios/pull/1349
https://github.com/wireapp/wire-ios-sync-engine/pull/628

Basically it boils down to "Issue", "Solution", "Dependencies" and maybe "Causes" or "Investigation". I think having less sections is better than more. This way we would avoid removing things and get into a habit of filling all sections.

nicorsm:
@vytis sure!

mikeger:
Please propagate to the SE (wire-ios-sync-engine). Let's try to live with it for some time, when it's good we can update all frameworks.

PR #1426, opened 2017-11-15
ZIOS-9241 Fix/i pad image cell margin

billypchan:
## What's new in this PR?
fix the issue of "when iPad in slide over mode, message cells have wide margins."

### Issues

In previous version there is no handling for iPad with compact horizontal size class margins.  

### Causes

Split mode is new feature of iOS11

### Solutions
Add left_margin_compact and right_margin_compact KVP for iPad compact mode.
Update layoutDirectionAwareLayoutMargins to handle iPad compact mode case
Imp. traitCollectionDidChange for ConversationCell (and its child ImageMessageCell)

## Notes

Further updates for other UI gadgets will be made for some other constant for iPad.


Wire-iOS/Resources/Magic/ipad.plist
```diff
@@ -98,8 +98,12 @@
 	<dict>
 		<key>left_margin</key>
 		<integer>96</integer>
+		<key>left_margin_compact</key>
```
typfel: These `plist` files are deprecated and we should not add more properties to them. I would take this opportunity to remove these content margin constant these files.

Wire-iOS/Resources/Magic/ipad.plist
```diff
@@ -98,8 +98,12 @@
...
+		<integer>50</integer>
```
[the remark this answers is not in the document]
typfel: The iPhone constants are defined in the `conversation.plist`

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/UIView+LayoutMargins.swift
```diff
@@ -25,6 +25,10 @@ extension UITableViewCell {
     class var layoutDirectionAwareLayoutMargins: UIEdgeInsets {
         var left = WAZUIMagic.cgFloat(forIdentifier: "content.left_margin")
         var right = WAZUIMagic.cgFloat(forIdentifier: "content.right_margin")
+        if UIDevice.current.userInterfaceIdiom == .pad && UIApplication.shared.keyWindow?.traitCollection.horizontalSizeClass == .compact {
```
typfel: Could we drop the `UIDevice.current.userInterfaceIdiom == .pad` and have this logic:

```
if horizontalSizeClass == .compact {
   "use small content insets" (56)
} else {
   "use large content insets" (96)
}
```

PR #1427, opened 2017-11-16
Fix Share Extension warnings

vytis:
### Issues

We had warnings about code not safe for extensions polluting the build log for a while.

### Causes

From the time that we had share extension with custom UI there was a framework `WireExtensionComponents`. It included a lot of UI elements and other things which imported a lot of frameworks (such as `WireSyncEngine`) that are not used in share extension. 

### Solutions

Isolated the code that is currently used in share extension and moved everything else into main application. Now `WireExtensionComponents` contains only things used from both UI and share extension. This turns out to be only ~10 files.

Also removed unneeded imports throughout the UI project and where needed replaced it with the headers of actual files in use.

## Notes

We still have 2 warnings left:
* HockeySDK. Unfortunately we cannot avoid this if we use Carthage for it. We should consider maybe integrate it with CocoaPods where you can select a variant that is safe for share extension.
* PINCache. It is a dependency of WireDataModel and it is not marked as safe for extensions. There is v3.0 brewing for a loooong time (first beta in Jul 23, 2016) which solves the issue. We could also consider moving to v3.0 without waiting until final release as it seems quite stable.


vytis:
No luck fixing the tests target build failure. Let's leave it for now.

vytis:
There seems to be an open Swift issue about it - https://bugs.swift.org/browse/SR-3381
In that case it is caused by `@objc protocol` defined in main target, but used in the test bundle. I tried looking at all such cases in our project and found several (`AudioRecordViewControllerDelegate` and  `VoiceChannelOverlayDelegate`), but removing those tests or following the steps to work around the issue doesn't solve it.

mikeger:
@vytis should we close this PR for the time being but keep the branch? Or you are planning to get back to it at some point?

vytis:
Yeah maybe we should close it, if I will come back to it I will have to do something different I think.

vytis:
Let's keep the branch for now though

PR #1430, opened 2017-11-20
Load MediaManager only after FlowManager

mikeger:
### Issues

The app takes a really long time to cold start (10s+).

### Causes

The MediaManger is waiting for some thread to spawn, but seems like it's happening too late.

The AVS libarary consists of several components, those are:
- FlowManager: the component for establishing the network media flows.
- MediaManager: the part responsible for audio routing on the device.
- wcall: the Calling3 implementation.

### Solutions

The entities must be initialized in certain expected order. The main requirement is that the MediaManager is only initialized after the FlowManager.

### DIFF
--- a/Wire-iOS/Sources/Helpers/sound/MediaManagerLoader.swift
+++ b/Wire-iOS/Sources/Helpers/sound/MediaManagerLoader.swift
@@ -95,10 +95,8 @@ extension MediaManagerState {
     
     override init() {
         super.init()
-        flowManagerObserver = NotificationCenter.default.addObserver(forName: FlowManager.AVSFlowManagerCreatedNotification, object: nil, queue: nil, using: { _ in
-            DispatchQueue.main.async { [weak self] in
-                self?.send(message: .flowManagerLoaded)
-            }
+        flowManagerObserver = NotificationCenter.default.addObserver(forName: FlowManager.AVSFlowManagerCreatedNotification, object: nil, queue: OperationQueue.main, using: { [weak self] _ in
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/Helpers/sound/MediaManagerLoader.swift
```diff
@@ -0,0 +1,108 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import CocoaLumberjackSwift
+import avs
+
+
+// The AVS libarary consists of several components, those are:
```
vytis: Typo - libarary

Wire-iOS/Sources/Helpers/sound/MediaManagerLoader.swift
```diff
@@ -0,0 +1,108 @@
...
+        
+        mediaManager.configureSounds()
+        mediaManager.observeSoundConfigurationChanges()
+        mediaManager.isMicrophoneMuted = false
+        mediaManager.isSpeakerEnabled = false
+    }
+    
+    override init() {
+        super.init()
+        flowManagerObserver = NotificationCenter.default.addObserver(forName: FlowManager.AVSFlowManagerCreatedNotification, object: nil, queue: nil, using: { _ in
```
vytis: Wouldn't it be the same by just adding a `queue` parameter instead of dispatching to main?

Issue #745, opened 2017-02-22
iPad UI


tschoof:
Hi, there are various UI glitches in the iPad version

1. Sometimes the "show more" Button in the input bar is stretched and misplaced, too close to the other buttons, and sometimes the share location button is too narrow
2. the input bar is sometimes empty, or the "show more" Button 
3.  in splitscreen as side app 
     a. The status bar is always visible but empty
4. In splitscreen as main app (on the left)
	 a. in the sharing location sheet, the X button is touching the iOS battery indicator
	 b. when launched into splitscreen (after force-quitting it), app isn't aware that it's only half-screen, so the contacts list is visible and chat window is cut off

That's what I could pin down in a short testing session, will update when I encounter new issues in use.

jenskasemets:
Hi @tschoof , thank you for your report.

Can you please clarify which iPad version you are were using for testing.

About first two comments, please can you send us also screenshots. 

I know it's extra work, but it would be always really helpful to get screenshots for visual bugs, then it's way easier to understand about which screen and problem is described.

tschoof:
It's an iPad mini, iOS 10.2.1 (current)

Will post screenshots, forgot that attaching files is supported here.

tschoof:
Oh well, of course it's an iPad mini 4.

tschoof:
Screenshots
![1](https://cloud.githubusercontent.com/assets/750370/23220081/113fae76-f921-11e6-8f8f-c1f239be7199.PNG)
![2 and 3a](https://cloud.githubusercontent.com/assets/750370/23220083/11559498-f921-11e6-9b10-aafa9e991e2f.PNG)
![4a](https://cloud.githubusercontent.com/assets/750370/23220082/1154318e-f921-11e6-9b2b-fe48f1bf1892.PNG)
![4b](https://cloud.githubusercontent.com/assets/750370/23220084/1173c508-f921-11e6-8e11-067b36809f4e.PNG)

tschoof:
And another one. When Wire was launched as overlayed app and is then accessed as only app, it doesn't quite get it and doesn't fill the space at all.

![img_0240](https://cloud.githubusercontent.com/assets/750370/23515480/8b1f6a5c-ff6a-11e6-8b7d-0dfe39cdc647.PNG)

billypchan:
@tschoof Thanks again for the detailed report. We had fixed a lot of issues in the latest release.

PR #1431, opened 2017-11-20
hide expandRowButton when it is not added to row one

billypchan:
### Issues

Conversation view - toolbar's expandRowButton is shown when trait collection changes to regular.

### Causes

expandRowButton is not included in first row but not removed form the view or set hidden

### Solutions

set expandRowButton hidden when multilineLayout == false


billypchan:
This PR should fix this issue:
https://github.com/wireapp/wire-ios/issues/745

PR #1432, opened 2017-11-20
Fix/team creation typo

billypchan:
fixed a copy and paste mistake, it generates wrong example for localization.


mikeger:
@billypchan this PR has an interesting commit history, most likely it was branched not from `develop` but from your original branch.

billypchan:
@mikeger I made this commit in a branch which deleted in remote. Next time I would create a new branch.

PR #1433, opened 2017-11-20
Fix/share extension connectivity

nicorsm:
### Issues

- In case of dropped or bad quality connection, Share Extension was stuck in "Preparing" or "Sending" screens
- The user wasn't informed about this kind of events and wasn't unable to get back to the share dialog

### Causes

This happens because the underlying layer (`SendableBatchObserver`) doesn't inform the share extension about timeouts or connection failures.

### Solutions

I've added:
- a timeout to the Send Controller, so if the message is stuck for 30 seconds, the user is pushed back to the Share dialog
- a label on the progress view ("No Internet Connection"), so the user is informed there's no internet connection (e.g. airplane mode enabled). This is based on Reachability, so it's not completely reliable.


Wire-iOS Share Extension/SendController.swift
```diff
@@ -102,15 +126,41 @@ class SendController {
             append(unsentSendables: unsentSendables, completion: completion)
         }
     }
-
+    
+    func tryToTimeout() {
+        if timedOut { return }
+        
+        cancelTimeout()
+        timeoutWorkItem = DispatchWorkItem { [weak self] in
```
mikeger: This can be assigned in `init` I believe.

Wire-iOS Share Extension/SendController.swift
```diff
@@ -71,19 +77,37 @@ class SendController {
...
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(SendController.networkStatusDidChange(_:)),
+                                               name: ShareExtensionNetworkObserver.statusChangeNotificationName,
+                                               object: nil)
     }
-
+    
+    @objc func networkStatusDidChange(_ notification: Notification) {
+        if let status = notification.object as? NetworkStatus, status.reachability() == .OK {
```
mikeger: Can you clarify the logic here? It should try to timeout only if the network is reachable?

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire Share Extension.xcscheme
```diff
@@ -79,8 +81,8 @@
       launchAutomaticallySubstyle = "2">
       <RemoteRunnable
          runnableDebuggingMode = "1"
-         BundleIdentifier = "com.apple.mobilesafari"
-         RemotePath = "/Applications/MobileSafari.app">
+         BundleIdentifier = "com.apple.mobileslideshow"
+         RemotePath = "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/Applications/MobileSlideShow.app">
```
mikeger: Seems like a redundant change, can you confirm?

Wire-iOS/Resources/Images/icon-no-appstore.xcassets/AppIcon-Debug.appiconset/Contents.json
```diff
@@ -133,6 +133,11 @@
       "idiom" : "ipad",
       "filename" : "83.5@2x.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ios-marketing",
+      "size" : "1024x1024",
+      "scale" : "1x"
```
mikeger: Also those files changes -- are they necessary?

nicorsm: I think this is a kind of placeholder for a new type of app icon in .xcassets. But yes, it's not necessary.

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire Share Extension.xcscheme
nicorsm: Yes, this is the default app launched together with Share Extension. It's useless.

Wire-iOS Share Extension/SendController.swift
nicorsm: Sure :)

Wire-iOS Share Extension/SendController.swift
nicorsm: Exactly

mikeger:
@nicorsm please update, also unfortunately contains the conflict now.

Wire-iOS Share Extension/SendController.swift
nicorsm: Just checked out. `cancel()` doesn't allow to reuse the same `DispatchWorkItem`, so I need to initialise again the variable.

PR #1436, opened 2017-11-21
Enable Local Authentication on Share Extension

nicorsm:
### Issues

If the user enabled Local Authentication (Passcode / Touch ID / Face ID) into the app, he was still able to see the list of contacts of the current user and to send contents on Wire via the Share Extension.

### Causes

This was made possible because Local Authentication wasn't implemented into the Share Extension.

### Solutions

- I added the same code from `AppLockViewController.swift` into the Share Extension. SE asks permission to the user only the first time he tries to choose a conversation in the current Sharing Session. This authentication is valid through the whole sharing session. That means, if the user dismiss the sharing dialog, he will be requested to re-authenticate.
- I refactored the code in order to avoid code duplication, by moving the authentication logic presented in `AppLockViewController.swift` to `AppLock.swift`.
- I moved the `SettingsPropertyName` enum to a proper file, in order to avoid useless dependencies with the Share Extension.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/48


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -262,6 +272,32 @@ class ShareExtensionViewController: SLComposeServiceViewController {
         
         pushConfigurationViewController(conversationSelectionViewController)
     }
+
+    /// @param callback confirmation; if the auth is not needed or is not possible on the current device called with '.none'
+    func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool?)->()) {
+        
+        guard #available(iOS 9.0, *), AppLock.isActive else {
```
mikeger: I know it's a copy-paste, but we don't support the iOS8 anyways, so may be the check it not necessary.

Wire-iOS/Sources/Components/Settings/SettingsPropertyName.swift
```diff
@@ -0,0 +1,90 @@
...
+ - PreferredFlashMode:     Flash mode for internal camera UI
+ - DarkMode:               Dark mode for conversation
+ - PriofileName:           User name
+ - SoundAlerts:            Sound alerts level
+ - DisableCrashAndAnalyticsSharing: Opt-Out analytics and Hockey
+ - DisableSendButton:      Opt-Out of new send button
+ - DisableLinkPreviews:    Disable link previews for links you send
+ - Disable(.*):            Disable some app features (debug)
+ */
+enum SettingsPropertyName: String, CustomStringConvertible {
```
mikeger: Have you added this one to both targets (app / share extension) or to ShareExtensionComponents?

nicorsm: Both targets. I could also move it to ShareExtensionComponents.

PR #1438, opened 2017-11-21
ZIOS-9273 update overlayWindow.frame when applicationWillEnterForeground instea…

billypchan:
This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1437
Fix the overlayWindow not cover the whole screen issue after device rotated.


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -56,7 +56,10 @@ class AppRootViewController : UIViewController {
 
     override public func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         mainWindow.frame.size = size
-        overlayWindow.frame.size = size
+
+        coordinator.animate(alongsideTransition: nil, completion: { _ in
+            self.overlayWindow.frame = UIScreen.main.bounds
```
typfel: Why is this necessary and won't behave badly on iPad when the app is in split screen mode?

billypchan:
@typfel @mikeger rewrite overlayWindow's frame update method with your comments: 

````
    func updateOverlayWindowFrame() {
        self.overlayWindow.frame = UIApplication.shared.keyWindow?.frame ?? UIScreen.main.bounds
    }
````

typfel:
@billypchan do we have any steps-to-reproduce for the issue that this fixes?

billypchan:
@typfel I recorded a video in ticket ZIOS-9273.

PR #1440, opened 2017-11-22
ZIOS-7752: No sound for media when phone on Silent

nicorsm:
### Issues

When playing a video in a conversation with iPhone in silent mode, the audio was silent too.

### Solutions

I chose to implement same behaviour of apps like Instagram: the sound is activated when pressing the volume buttons on iPhone. 

It's also possible to remove this control and activate the volume when the video starts, let's talk about it!


typfel:
We previously agreed with AVS that we will not directly change the audio category in the client, instead we will inform AVS via the media manager that we (the client) will play sound not controlled by AVS (sound cloud, video, audio clips). AVS will then take care of setting the correct audio category and change it back again after the playback ends.

Relevant APIs are: `AVSMediaDelegate` and `MediaPlaybackManager`

mikeger:
@typfel is this PR now redundant?

typfel:
@mikeger yes

PR #1441, opened 2017-11-23
Feature/text field ui components

billypchan:
### Issues

Team creation workflow needs a new textField written in Swift 

### Causes

The existing RegistrationTextField has to be refurbished with modern style.

### Solutions

A new AccessoryTextField class is created with these features:

1. Uses the font defined in FontSpec
2. When the text is validated, the ConfirmButton enable property will be updated
3. AccessoryTextFieldDelegate for string validation callback

## Notes

demo VC TeamNameStepViewController is updated for AccessoryTextField (tap the "Create team" button on landing screen to load it)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+            self.textFieldType = textFieldType
+        }
+    }
+
+    override init(frame: CGRect) {
+        let os = ProcessInfo().operatingSystemVersion
+
+        let leftInset: CGFloat = 16 + 8
+
+        if os.majorVersion < 11 {
```
typfel: Why not use the new `if #available(iOS 11, *) {` to check if we are on iOS 11?

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+        self.rightViewMode = .always;
+
+        self.font = AccessoryTextField.enteredTextFont
+        self.textColor = .textColor
+
+        autocorrectionType = .no
+        contentVerticalAlignment = .center
+        switch UIDevice.current.userInterfaceIdiom {
+        case .pad:
+            layer.cornerRadius = 4
```
typfel: Design is requesting us to have different design of this field for iPad?

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+
+        iconButton.isEnabled = false
+
+        return iconButton
+    }()
+
+    let textInsets: UIEdgeInsets = UIEdgeInsets(top: 0, left: 16, bottom: 0, right: 8)
+    let placeholderInsets: UIEdgeInsets
+
+    convenience init(frame: CGRect, textFieldType: TextFieldType? = .unknown) {
```
vytis: I guess if you have a default parameter then you don't need to optional here. As the users probably won't pass `nil` but instead omit the parameter altogether

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
billypchan: Yes, round corner is required on iPad

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+        self.init(frame: frame)
+
+        if let textFieldType = textFieldType {
+            self.textFieldType = textFieldType
+        }
+    }
+
+    override init(frame: CGRect) {
```
mikeger: I think that since we are using the autolayout everywhere, the parameter `frame` is redundant and can be emitted in both inits.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    static let enteredTextFont = FontSpec(.normal, .regular, .inputText).font!
+    static let placeholderFont = FontSpec(.small, .semibold).font!
+    static private let ConfirmButtonWidth: CGFloat = 32
+
+    var textFieldType: TextFieldType = .unknown
+
+    let confirmButton: IconButton = {
+        let iconButton = IconButton.iconButtonCircularLight()
+        iconButton.frame.size = CGSize(width: AccessoryTextField.ConfirmButtonWidth, height: AccessoryTextField.ConfirmButtonWidth)
```
vytis: Won't there be any potential problems trying to mix Autolayout and manual layouting? Maybe better to constrain the width/height?

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,187 @@
...
+            // Placeholder frame calculation is changed in iOS 11, therefore the TOP inset is not necessary
+            leftInset = 8
+        }
+
+        placeholderInsets = UIEdgeInsets(top: topInset, left: leftInset, bottom: 0, right: 16)
+
+        super.init(frame: frame)
+
+        self.rightView = self.confirmButton;
+        self.rightViewMode = .always;
```
mikeger: Semicolon ;;;;;;;)

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -135,9 +135,11 @@ final class LandingViewController: UIViewController {
         [createAccountButton, createTeamButton].forEach(containerView.addSubview)
 
         self.createConstraints()
+    }
 
-
-
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        UIApplication.shared.wr_setStatusBarHidden(true, with: .fade)
```
mikeger: Please try instead of forcing the status bar to be hidden implement `prefersStatusBarHidden` returning true on `LandingViewController` and call `UIApplication.wr_updateStatusBarForCurrentControllerAnimated` here.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/UIColor+TeamCreation.swift
```diff
@@ -24,6 +24,53 @@ extension UIColor {
...
     static let textColor = UIColor(red:0.20, green:0.22, blue:0.23, alpha:1.0)
     static let subtitleColor = UIColor(red:0.20, green:0.22, blue:0.23, alpha:0.56)
+    static let activeButtonColor = UIColor(hexString: "#2391D3")!
+    static let errorMessageColor = UIColor(hexString: "#FB0807")!
+    static let inactiveButtonColor = UIColor(hexString: "#33373A40")!
+    static let textfieldColor = UIColor.white
+    static let placeholderColor = UIColor(hexString: "#8D989F")!
+
+
+    public convenience init?(hexString: String) {
```
mikeger: Unfortunately as we are GPLv3 project we cannot take the code from the web to be included with our project, since it is going to violate the original license -- header of this file is suggesting it. 

Please remove the copied code and either reimplement it with the clean room approach.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
vytis: All this will be handled by the other controller that I am working on now, so I guess this can be removed so that we don't forget about it.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
mikeger: @vytis I am not 100% sure, but I think some dark corners of UIKit are not using the autolayout, for example the navigation bar. But may be here creating 2 constratins and laying the button out should also work.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
typfel: :/

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/UIColor+TeamCreation.swift
mikeger: forgot to finish: ...or use the colors with argb values :)

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
vytis: I checked the current `RegistrationTextField` and it seems we do use autolayout for accessory views there. For example [GuidanceDotView](https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/GuidanceDotView.m)

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/UIColor+TeamCreation.swift
billypchan: Rewrite with new interface with simpler logic.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
billypchan: Update IconButton, it will update its layer's round corner radius when its frame changes.

vytis:
@billypchan maybe you can add the validation as a separate PR? It seems quite a lot of code in one place already and there were just few minor things left before it was good to merge.

WireExtensionComponents/Views/Buttons/IconButton.m
```diff
@@ -221,6 +221,13 @@ - (void)setTitleColor:(UIColor *)color forState:(UIControlState)state
     }
 }
 
+- (void)setFrame:(CGRect)frame
```
typfel: Seems like this method can be removed since the corner radius is calculated based on the bounds and setting the frame will always set the bounds as well.

mikeger:
@billypchan please resolve the project file conflict, looks like we added some files simultaneously.

WireExtensionComponents/Views/Buttons/IconButton.m
billypchan: @typfel  This method is added because the parent view's layoutSubviews calls UITextField's _updateButtons, and _updateButtons calls setFrame but not setBound:

  * frame #0: 0x000000010b7f67ee WireExtensionComponents`-[IconButton setFrame:](self=0x00007fcd7d425e90, _cmd="setFrame:", frame=(origin = (x = 327, y = 12), size = (width = 32, height = 32))) at IconButton.m:228
    frame #1: 0x000000010842336e UIKit`-[UITextField _updateButtons] + 862
    frame #2: 0x00000001084268f9 UIKit`-[UITextField _layoutContent] + 92
    frame #3: 0x0000000108426c8f UIKit`-[UITextField layoutSubviews] + 274
    frame #4: 0x0000000104d051ee Wire`AccessoryTextField.layoutSubviews(self=0x00007fcd80051400) at AccessoryTextField.swift:117
    frame #5: 0x0000000104d05294 Wire`@objc AccessoryTextField.layoutSubviews() at AccessoryTextField.swift:0

typfel: Ok, how about moving the  `updateCircularCornerRadius ` to ` layoutSubviews` and remove both `setBounds:` and `setFrame:`

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -0,0 +1,230 @@
...
+    }()
+
+    let textInsets: UIEdgeInsets = UIEdgeInsets(top: 0, left: 16, bottom: 0, right: 8)
+    let placeholderInsets: UIEdgeInsets
+
+
+    /// init with textFieldType for keyboard style and validator type
+    ///
+    /// - Parameter textFieldType: the type for text field. If not set, default value .unknown is applied.
+    convenience init(textFieldType: TextFieldType?) {
```
vytis: In the previous comment about this I meant that the `textFieldType` can be not optional if we have default value:

```
convenience init(textFieldType: TextFieldType = .unknown)
```

PR #1442, opened 2017-11-23
ZIOS-9089: Do not allow sending the text messages longer than 8KB

nicorsm:
### Issues

Actually, you're able to send messages longer than 8KB.

### Solutions

- Added a control that shows the same alert that is shown on the web version. The alert is visible when tapping on the Send button, if the content of the input bar is greater than 8.000 characters.
- Added support to the `charactersRemaining` property of `SLComposeServiceViewController`. In this way, the remaining characters count is shown on the share extension.

## Attachments

simulator screen shot - iphone 7 - 2017-11-23 at 12 00 11
simulator screen shot - iphone 7 - 2017-11-23 at 12 01 05

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
@@ -1046,11 +1046,12 @@
 
 -(BOOL)checkMessageLength{
     
-    BOOL allowed = self.inputBar.textView.text.length <= 8000;
+    BOOL allowed = self.inputBar.textView.text.length <= (NSUInteger)SharedConstants.maximumMessageLength;
     
     if(!allowed) {
+        NSString *message = [NSString stringWithFormat:NSLocalizedString(@"conversation.input_bar.message_too_long.message", nil), SharedConstants.maximumMessageLength];
         UIAlertController *alert = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"conversation.input_bar.message_too_long.title", nil)
-                                                                       message:NSLocalizedString(@"conversation.input_bar.message_too_long.message", nil)
+                                                                       message:message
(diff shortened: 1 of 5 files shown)


vytis:
BTW does it show the count always or only when going towards the limit?

nicorsm:
@vytis Always, the label is black when the count is positive and red if negative. But indeed, since the property is visible only if `charactersRemaining` is non-nil, we could set the property only when reaching the limit (e.g. last 30 characters available).

nicorsm:
...also because you usually write 8.000 characters only via copy & paste 😄

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -1037,9 +1037,26 @@ @implementation ConversationInputBarViewController (Sending)
 
 - (void)sendButtonPressed:(id)sender
 {
-    [self.inputBar.textView autocorrectLastWord];
-    [self sendOrEditText:self.inputBar.textView.preparedText];
-    [self.inputBar.textView resetTypingAttributes];
+    if([self checkMessageLength]){
+        [self.inputBar.textView autocorrectLastWord];
+        [self sendOrEditText:self.inputBar.textView.preparedText];
```
johnxnguyen: Is the 8KB a strict limit? If so, we may exceed this limit in the case that the autocorrect and prepared text add more characters before sending.

nicorsm: It's not a strict limit by our side - I mean, you can send messages anyway. But yes, good point.

johnxnguyen: Also, another thing to consider is that emoji replacement happens after this point.

nicorsm:
@mikeger I'll take a look. Thanks for the feedback! :)

PR #1444, opened 2017-11-23
Digit code input field

mikeger:
## What's new in this PR?

The new multi-digit input field is designed to allow entering the 6-digit email code, as long as the other possible codes, for the new team creation flow.

Based on `UIControl`, the completely custom input is adopting `UITextInputTraits` to allow selecting the different type of keyboards.

Adopting `UIKeyInput` protocol is allowing to handle text input and character deletion.


Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,278 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import Cartography
+
+public protocol CharacterInputFieldDelegate: NSObjectProtocol {
+    func didChangeText(_ inputField: CharacterInputField, to: String)
```
vytis: I guess it would be good to have a callback when the input field got all the numbers in needs. Because we won't really be doing anything but checking `isFilled` and then doing something useful when it is true.

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,278 @@
...
+            return CGSize(width: 50, height: 56)
+        }
+        
+        override func didMoveToWindow() {
+            super.didMoveToWindow()
+            self.startCursorAnimationIfNeeded()
+        }
+        
+        private func startCursorAnimationIfNeeded() {
+            if !cursorView.isHidden && cursorView.layer.animation(forKey: "blinkAnimation") == nil {
```
vytis: Only so many times in life there is an opportunity to animate the blinking cursor!

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,278 @@
...
+                let animation = CAKeyframeAnimation(keyPath: "opacity")
+                animation.values = [1, 1, 0, 0]
+                animation.keyTimes = [0, 0.4, 0.7, 0.9]
+                animation.duration = 0.64
+                animation.autoreverses = true
+                animation.repeatCount =  .greatestFiniteMagnitude
```
vytis: Didn't have any useful comments to add for the PR except about this one horrible extra space!

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,278 @@
...
+        }
+        
+        self.text = valueToPaste
+        self.delegate?.didChangeText(self, to: storage)
+    }
+    
+    public override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        switch action {
+        case #selector(paste(_:)):
+            return UIPasteboard.general.string != nil
```
vytis: I guess we need to check here if the text we want to paste is valid. Or maybe even better signal with the delegate somehow that invalid info was pasted and we can show the error.

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
mikeger: Actually a copy-paste of our cursor animation from the input field. Not sure if it's displayed there any more.

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
billypchan: We need SwiftLint!

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,278 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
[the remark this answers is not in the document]
billypchan: https://github.com/billypchan/wire-swift-file-xctemplate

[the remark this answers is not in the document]
mikeger: I actually used it, not sure what went wrong, or I am a time traveller 🗡

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
mikeger: @vytis I've added the callback in the follow-up.

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -0,0 +1,297 @@
...
+            }
+            else {
+                characterView.character = .none
+            }
+        }
+    }
+    
+    fileprivate func notifyingDelegate(_ action: ()->()) {
+        let wasFilled = self.isFilled
+        let previosText = self.storage
```
vytis: Typo - should be `previousText`

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
vytis: Also need to think of what happens if we have few (or all) numbers already and then paste stuff over.

mikeger:
@vytis currently if you paste something it's going to replace the current content. I was thinking around it, but decided to leave it as it is and give it a try.

PR #1448, opened 2017-11-24
Team creation flow

vytis:
## What's new in this PR?

Functionality to create a team on mobile. This is the initial implementation that still has few flaws that are listed at the end.

## Architecture overview

The screens in the flow are very similar so to avoid creating multiple controllers that are almost exactly the same we have following components:

### `TeamCreationState`

It encodes all the steps in the flow and information that is gathered from the screens. It contains all the information needed to create a view controller.

It is the base of a simple state machine that transitions backwards and forwards through the states in order. It also knows what the relevant information for each screen is - titles, subtitles, buttons etc.

### `ViewDescriptor`

All components in the screens are encapsulated in a simple object that can create a view on demand and contains delegate logic etc. 

### `TeamCreationStepController`

A view controller that takes care of laying out all the labels, buttons etc. but has no knowledge of the flow. It takes `ViewDescriptor` instances for different elements, creates corresponding views and lays them out as needed.

### `TeamCreationFlowController`

 An object that is responsible for showing the right screen at the right time and handles all logic related to the server interaction. It is also responsible for transitioning through the `TeamCreationState`s when corresponding controller is shown.

Because in some screens there are API calls involved it is also a `RegistrationStatusDelegate`. When those calls are successful or fail the state is advanced or an error message is shown.

## Known issues

To be addressed in separate PRs:

- [ ] No tests
- [ ] Accepting terms and conditions
- [ ] Actually creating the team at the end and transitioning to logged-in state
- [ ] Making all text translatable
- [ ] Layout on iPad
- [ ] Text validation
- [ ] Open team creation flow from settings

Other small glitches:

- [ ] Webview "About Wire" has coloured background
- [ ] Webview should open same URL as currently from settings
- [ ] Autocapitalizing when entering email
- [ ] When opening activation code screen the keyboard is not open
- [ ] Wrong insets for text field placeholder
- [ ] Coming back from full name step should skip email verification step
- [ ] Coming back to landing screen shows keyboard for a bit
- [ ] Look through which textfield should have autocorrect enabled
- [ ] Use better error texts
- [ ] Clear error field after inputting new text
- [ ] Back button highlight color is wrong

### Note

This is pointed to merge to an integration branch (`feature/team-creation`) where we can gather all the other work until this is ready for showing to more people.


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -61,16 +61,10 @@ class AccessoryTextField : UITextField {
     let placeholderInsets: UIEdgeInsets
 
 
-    /// init with textFieldType for keyboard style and validator type
+    /// init with type for keyboard style and validator type
```
typfel: `type` should be `kind` now?

eike96:
- [ ] All font sizes are off
- [ ] line height of the description is off (look at “You’ve got mail” screen)
- [ ] webview "what is wire for teams" cannot be closed on iPhone X.  Close button is under status bar
- [ ] arrow in blue circle button is too small and has a wrong color
- [ ] transition between screens should only move right to left without  the shadow 
- [ ] Copy is not updated
- [ ] Some screens are missing the description, some have one even though they shouldnt
- [ ] hide cursor in code entry
- [ ] blue circle buttons inactive/active states are often wrong
- [ ] The Create Account and Log in screens should not be touched. Roll back to what we had
- [ ] first screen does not include interaction animation yet
- [ ] going back to “set email” screen and tapping blue circle button to proceed acts as if it proceeds but opens the same screen again. stuck.
- [ ] not all headlines are in exactly the same position
- [ ] when you tap “resend code” there should be a spinner until the mail has been sent

julesmummdry:
QA comments:

- [ ] App Crash when pressing the delete button for the 7th time on the verification code

- [ ] When coming back to the verification code and resend the code and type it in new it not automatically navigates to the next page...stuck on verification code page

mikeger:
@julesmummdry fix for Nr.1 here: https://github.com/wireapp/wire-ios/pull/1454

PR #1451, opened 2017-11-24
ZIOS-9293  Text validator written in Swift

billypchan:
### Issues

The ZMXXXValidator class is written in obj-c and it has an AutoreleasingUnsafeMutablePointer in the the parameter. It is not easy to use for Swift classes.

### Solutions

Create a new TextFieldValidator class for String checking. Use iOS build-in NSDataDetector for string validation.

## Notes

Further improvement such as invalid character checking will be added in next PR.


Wire-iOS Tests/AccessoryTextField/EmailAdresssValidatorTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        // WHEN & THEN
+
+        validEmailAddresses.forEach { email in
+            XCTAssert(email.isEmail, "failed for \(email)")
+        }
+    }
+
+    func testThatInvalidEmailsDoNotPassValidation(){
+        // GIVEN
+        let invalidEmailAddresses =
```
vytis: Since we use `NSDataDetector` under the hood these tests are sort of testing Apple's implementation. It's not bad, but since we can't control it might break with iOS updates. We had a similar issue in code that was using `NSLinguisticTagger` that we used for splitting names into fist name and last name. After updating to Xcode9 lots of examples that we were testing failed because the underlying implementation changed, for example https://github.com/wireapp/wire-ios-data-model/pull/373/files#diff-085d88cbcb7a4e9c347ff2c952551fcfL276

So in this case the test cases should be as short as possible to exercise our code without going too much deeper.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -20,14 +20,14 @@
 import Foundation
 import Cartography
 
-class AccessoryTextField : UITextField {
+protocol AccessoryTextFieldDelegate: class {
+    func validationErrorDidOccur(sender: UITextField, error: TextFieldValidationError?)
+    func validationSucceed(sender: UITextField)
```
vytis: More readable would be `validationSucceeded`

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/TextFieldValidator.swift
```diff
@@ -0,0 +1,107 @@
...
+    case name
+    case password
+    case unknown
+}
+
+enum TextFieldValidationError: Error {
+    case tooShort, tooLong, invalidEmail
+}
+
+protocol TextFieldValidatorDelegate: class {
```
mikeger: Question: why it's necessary to utilize a delegate pattern for the validator? Can't we potentially just receive the validation result as a return from the function call?

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -20,14 +20,14 @@
 import Foundation
 import Cartography
 
-class AccessoryTextField : UITextField {
+protocol AccessoryTextFieldDelegate: class {
+    func validationErrorDidOccur(sender: UITextField, error: TextFieldValidationError?)
```
vytis: Is there a reason to have the error optional? It seems logical to always produce an error if we fail validation.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Steps/TeamNameStepViewController.swift
```diff
@@ -158,3 +160,24 @@ class TeamNameStepViewController: UIViewController {
...
+    }
+
+    func validationErrorDidOccur(sender: UITextField, error: TextFieldValidationError?) {
+        if let error = error {
+            switch error {
+            case .tooLong:
+                errorLabel.text = "name.guidance.toolong".localized
+            case .tooShort:
+                errorLabel.text = "name.guidance.tooshort".localized
+            default:
```
vytis: Since there are not too many possible errors it's better to avoid using `default` case. If in the future we add more error cases the compiler will warn us that we are missing a description here and would not just fallback to "unknown error".

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/TextFieldValidator.swift
```diff
@@ -0,0 +1,107 @@
...
+            delegate?.validationSucceed()
+        }
+    }
+}
+
+// MARK:- Email validator
+
+extension String {
+    public var isEmail: Bool {
+        guard !self.hasPrefix("mailto:") else { return false }
```
vytis: I guess this is added because `mailto:some@email.com` is considered an email address by the NSDataDetector? Maybe add a short comment about it?

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/TextFieldValidator.swift
```diff
@@ -0,0 +1,107 @@
...
+
+        let dataDetector = try? NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue)
```
vytis: `guard let` would help to avoid optionals in all later lines.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/TextFieldValidator.swift
```diff
@@ -0,0 +1,107 @@
...
+        let range = NSRange(location: 0, length: self.characters.count)
+        let firstMatch = dataDetector?.firstMatch(in: self, options: NSRegularExpression.MatchingOptions.reportCompletion, range: range)
+
+        let numberOfMatches = dataDetector?.numberOfMatches(in: self, options: NSRegularExpression.MatchingOptions.reportCompletion, range: range)
+
+        if firstMatch?.range.location == NSNotFound { return false }
+        if firstMatch?.url?.scheme != "mailto" { return false }
+        if firstMatch?.url?.absoluteString.hasSuffix(self) == false { return false }
+        if numberOfMatches != 1 { return false }
```
vytis: Would be helpful to comment each of the lines above to say what case and why it passes. Might be tricky for another person to come back to this code in the future and understand the reasoning behind all of it.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/TextFieldValidator.swift
billypchan: Agree. I would refactor it just return a simple enum.

Wire-iOS Tests/AccessoryTextField/AccessoryTextFieldValidateionTests.swift
```diff
@@ -0,0 +1,240 @@
...
+        XCTAssertEqual(mockViewController.successCounter, 0, file: file, line: line)
+        XCTAssertFalse(sut.confirmButton.isEnabled, file: file, line: line)
+        XCTAssertEqual(expectedError, mockViewController.lastError, file: file, line: line)
+
+        /// check for localied error description
+        switch mockViewController.lastError {
+        case .tooShort(kind: let kind):
+            switch kind {
+            case .name:
+                XCTAssertEqual(mockViewController.lastError.localizedDescription, "At least 2 characters", file: file, line: line)
```
vytis: No need for checking the actual translations because they might change easily and then these tests will fail. It's enough to check that we get the expected kind of error.

typfel:
I'm a bit concerned that we are re-implementing the validation logic already existing in our data model layer. Now we will have two places where it will need to be updated if the business rules change.

Again what prevents us from using the validation methods exposed on `ZMUser`? Examples of how they can used from Swift can be found in `UnauthenticatedSession+Registration.swift` (wire-ios-sync-engine).

billypchan:
@typfel The requirement is different. The original ZMEmailAddressValidator accepts email address format in email header, e.g. "Meep \"_the_\" Møøp <Meep.Moop@ExAmple.com>" as a valid email address. The new created validator does not accept email with recipient's name.

typfel:
@billypchan @vytis if the requirement has changed shouldn't we then update the logic in the data model layer?

vytis:
It was my idea to do something better, but it seems I had a wrong impression of how it is supposed to be used (e.g. `ChangeEmailViewController`). We had a small tribe gathering together with @mikeger and @billypchan and agreed that it does not make much sense to have two implementations of the same thing. Which one is better for the uses cases we currently have in the app is not completely clear. However we are lagging behind schedule quite a bit and to save time lets merge this one and then maybe when we rewamp the registration flow we can think of how to simplify all of this.

PR #1456, opened 2017-11-27
Link previews not shown correctly and not clickable (iOS App)

nicorsm:
### Issues

Two main issues:
1. some shared links were not recognised as links, so it wasn't possible to open them via `openURL`.
2. images from Open Graph weren't displayed.

### Causes

1. in some cases, `permanentURL` was not resolved correctly - e.g. `http://` was not added to this link: designtoolsberlin.com/hackerguide . I (and @mikeger) added the `resolvedURL` parameter, so `ZMLinkPreview` can fallback to the correct resolved url (with `http://` and/or `www`). 
2. the image was downloaded correctly, but the container cell wasn't updated

### Solutions

Solutions implemented with this PR:
- added a delegate method to reload the cell when the image of the link is downloaded
- fixed tests

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-link-preview/pull/21
- [x] https://github.com/wireapp/wire-ios-data-model/pull/397


mikeger:
@nicorsm I have a question with regard to the images: the issue is only happening on the sender side, right? On the receiver side we would know that the link preview has the image data when we'll receive it.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/ArticleView.swift
```diff
@@ -27,6 +27,7 @@ import Classy
 @objc protocol ArticleViewDelegate: class {
     func articleViewWantsToOpenURL(_ articleView: ArticleView, url: URL)
     func articleViewDidLongPressView(_ articleView: ArticleView)
+    func articleViewAsksToRecalculateHeight()
```
mikeger: Please send the reference to `ArticleView` object in the delegate (as in the other callbacks).

nicorsm:
@mikeger yeah, it should happen only on sender side. Links are loading fine for me when receiving messages in the current release.

PR #1460, opened 2017-11-28
Completing the registration flow

vytis:
### Issues

Work started in #1448 continues here by completing the flow and enabling user to actually finish team registration

### Architecture changes

There was a need to add the last state `createTeam` that is the end of the flow and it doesn't have a corresponding controller. Before all the states had UI elements that were created from methods on `TeamCreationState` (`backButtonDescription`, `headline` etc.), but now we have to have an intermediate object that is optional. This made the code a bit more streamlined by grouping all UI elements on the screen in one class instead of scattering it in `switch` statements.

### Other changes

Wired up `registrationDelegate` method calls and various observers to know when we have logged in, synced and are ready to show UI.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/641


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -5746,7 +5770,7 @@
 				"$(SRCROOT)/Carthage/Build/iOS/WireProtos.framework",
 				"$(SRCROOT)/Carthage/Build/iOS/ProtocolBuffers.framework",
 				"$(SRCROOT)/Carthage/Build/iOS/Cartography.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/WireSyncEngine.framework",
+				"$(BUILT_PRODUCTS_DIR)/WireSyncEngine.framework",
```
mikeger: 🏎 Error!

PR #1461, opened 2017-11-28
Allow email code input field to get the responder automatically

mikeger:
### Issues

In the team creation flow only code view haven't received the responder state automatically (keyboard is not up when opening the email code screen).

### Causes

In every screen we call `mainView.becomeFirstResponder()`, in all other cases it's an input field, whereas on the email code screen the input is hidden inside the `UIView` wrapper.

### Solutions

Implement the wrapper that is forwarding the responder calls to the subview.

## Notes

In the past I was under an impression that `becomeFirstResponder` is a recursive call.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/VerificationCodeFieldDescription.swift
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/VerificationCodeFieldDescription.swift
@@ -24,7 +24,7 @@ final class VerificationCodeFieldDescription: NSObject, ValueSubmission {
     var constraints: [NSLayoutConstraint] = []
 }
 
-fileprivate final class RepsonderContainer: UIView {
+fileprivate final class ResponderContainer: UIView {
     private let responder: UIView
     
     init(responder: UIView) {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/VerificationCodeFieldDescription.swift
```diff
@@ -24,23 +24,48 @@ final class VerificationCodeFieldDescription: NSObject, ValueSubmission {
...
-        containerView.translatesAutoresizingMaskIntoConstraints = false
         let inputField = CharacterInputField(maxLength: 6, characterSet: .decimalDigits)
         inputField.keyboardType = .decimalPad
         inputField.translatesAutoresizingMaskIntoConstraints = false
         inputField.delegate = self
         inputField.accessibilityIdentifier = "VerificationCode"
         inputField.accessibilityLabel = "team.email_code.input_field.accessbility_label".localized
-        containerView.addSubview(inputField)
 
+        let containerView = RepsonderContainer(responder: inputField)
```
nicorsm: Typo, 'Repsonder' (if it's a quick change to do)

mikeger:
The responder chain is seems to be going upwards towards the UIViewController and UIApplication: https://developer.apple.com/library/content/documentation/General/Conceptual/Devpedia-CocoaApp/Responder.html
therefore forgetting to visit the subview of our UIView.

PR #1464, opened 2017-11-28
Few fixes to the flow

vytis:
### Issues

Solves several of the issues noticed:
* Tapping "what is wire" and "change email" buttons would do nothing
* After validating email and tapping back we should go back to email screen

### Causes

There were some weird issues happening because we would not keep a reference to `TeamCreationStepDescription`.

### Solutions

Rewamped the `TeamCreationStepController` so that it takes only one parameter and keeps a reference to the step description. This meant that all view descriptions inside a step need to be stored properties, not computed ones.

Another quick fix was to remove email verification controller when pushing a new one on top, so that when coming back we would skip it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ScreenDescriptions/SetEmailStepDescription.swift
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ScreenDescriptions/SetEmailStepDescription.swift
@@ -21,29 +21,19 @@ import Foundation
 final class SetEmailStepDescription: TeamCreationStepDescription {
 
     let controller: UIViewController
+    let backButton: BackButtonDescription?
+    let mainView: ViewDescriptor & ValueSubmission
+    let headline: String
+    let subtext: String?
+    let secondaryViews: [ViewDescriptor]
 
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ScreenDescriptions/SetEmailStepDescription.swift
```diff
@@ -21,29 +21,19 @@ import Foundation
...
-    }
-
-    var subtext: String? {
-        return nil
-    }
-
-    var secondaryViews: [ViewDescriptor] {
-        return []
+        backButton = BackButtonDescription()
+        mainView = TextFieldDescription(placeholder: "Email address", actionDescription: "Set e-mail", kind: .email)
```
billypchan: May be we should start mark down which strings need to be localized?

vytis: Yeah that is still one of the tasks to be done. Was waiting until we have the whole flow working so that we wouldn't have to do the same thing multiple times.

PR #1470, opened 2017-11-29
Fix/character input field fix for small screens and iPad

billypchan:
## What's new in this PR?

Fix the Code entering screen layout issue on iPad and small screen iPhones.
Decrease CharacterView's width on iPhone with 320 pt width.
Remove redundant constraints.


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
```diff
@@ -101,7 +101,7 @@ final class TeamCreationStepController: UIViewController {
     func updateKeyboardOffset(keyboardHeight: CGFloat) {
         self.keyboardOffset.constant = -(keyboardHeight + 10)
         UIView.performWithoutAnimation {
-            self.view.layoutIfNeeded()
+            self.view.setNeedsLayout()
```
mikeger: What is the intention here? To make view layout you need to do both: setNeedsLayout and immediately after layoutIfNeeded.

Wire-iOS/Sources/UserInterface/Components/CharacterInputField.swift
```diff
@@ -172,10 +177,12 @@ public class CharacterInputField: UIControl, UITextInputTraits {
     
     // MARK: - Overrides
     
-    init(maxLength: Int, characterSet: CharacterSet) {
+    init(maxLength: Int, characterSet: CharacterSet, width: CGFloat) {
```
mikeger: `width` is not descriptive. Also if we would like to change the height too then we'll need to introduce the new parameter. Please consider replacing with `charPlaceholderSize: CGSize`.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
billypchan: will undo the change

mikeger:
@billypchan I tried merging develop into this PR but failed merging `TeamCreationStepController.swift`. Please check if you can do it.

PR #1472, opened 2017-11-29
observe keyboard changes when view appears

johnxnguyen:
### Issues
The layout constraints on the team name page we're not updating to the keyboard offset when returning from the "what is teams for wire" webview is dismissed.

### Causes
This view controller registers to observe keyboard changes in `viewDidLoad` and removes itself as an observer in `viewWillDisappear`. Thus, when the view re-appears, the keyboard offset is initially zero, and it doesn't update when the keyboard eventually appears.

### Solutions
Move observer code to `viewWillAppear`.


mikeger:
@johnxnguyen I don't think the call is moved, seems like the same code. May be you forgot to stage the part?

johnxnguyen:
@mikeger oh yeah you're right! thanks!

PR #1473, opened 2017-11-29
Removed the cursor from character input field

mikeger:
## What's new in this PR?

Simplification of the character input view, as suggested by the design. As the field does not allow moving the cursor, showing it could be misleading.


vytis:
And that blink animation was moved somewhere else? I remember seeing the code in the original PR

mikeger:
@vytis it lives here https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Helpers/CAAnimation+Cursor.swift

PR #1481, opened 2017-11-30
Error localization

vytis:
### Issues

We were not showing text for errors returned from backend.

### Solutions

Localizations added in sync engine, the only change is update NSError API for Swift.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/645


mikeger:
I started the build for the new SE.

PR #1490, opened 2017-12-01
fix broken constraints

billypchan:
Remove unnecessary constraints for confirm button.


vytis:
Just a sidenote but it would be good if you could add some more descriptive description - it's not really needed to write what you did in the PR, because it's quite clear from the diff. However why this change was needed is not as clear - what layout issues does it fix? What was the constraint for? Why it is disabled initially? The PR templates has the main parts, but basically it should give enough context for a team member who didn't work on this code before enough information to review the pull request.

billypchan:
I will put the mainViewAlignVerticalCenter related change in another PR for iPad layout.

PR #1491, opened 2017-12-02
ZIOS-6963: If open graph image is < 480 pixel width or 160 pixel height it looks ugly scaled

nicorsm:
### Issues

When sharing a link with an image of small size declared on the open graph (`og:image`), the image on the `ArticleView` was ugly stretched.

### Causes

This was caused by the default behaviour of the view, which sets the `contentMode` of the image view by default at `.scaleAspectFill`.

### Solutions

When the image width is < 480px or height is < 160px, then `contentMode` is now set to `.center`.


mikeger:
@nicorsm should we have a screenshot test for this case?

nicorsm:
@mikeger why not? 🙂 I'll take a look at it.

mikeger:
@nicorsm some tests failed 😿

PR #1493, opened 2017-12-02
Fix the checkbox in the issue template

mat1th:
## What's new in this PR?
The .gittemplate did have a error.

### Issues
The checkbox did have wrong markdown syntax. This is fixed now.

### Causes
The syntax was `[ ] -`. This is not correct.

### Solutions
The syntax should be `* [ ]`.


vytis:
Thanks for spotting this!

PR #1494, opened 2017-12-04
fix constraint design for iPad and iOS 10

billypchan:
### Issues

Safe area is introduced on iOS 11 and  back buttons's vertical position is not correct on iOS <= 10 devices.

The constraints design does not match the design guideline.  

### Solutions

Set a different value of back buttons's top margin for ios 11/<=11.

Call setNeedsLayout instead of layoutIfNeeded to fix the backbutton move down visual flaw when iOS10 enter name screen shows.

Rewrite constraints for top and bottom spacing. Two spacer view are created and their heights are calculated with these rules:

|            | iPhone | iPad |
| ------- | ------- | ------- |
| keyboard is shown | topSpacer.height >= 20, bottomSpacer.height ==  24 + keyboardHeight | topSpacer.height == bottomSpacer.height + keyboardHeight |
| keyboard is hidden | topSpacer.height == bottomSpacer.height | topSpacer.height == bottomSpacer.height |

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
@@ -50,7 +50,6 @@ final class TeamCreationStepController: UIViewController {
     private var secondaryViews: [UIView] = []
     fileprivate var secondaryErrorView: UIView?
 
-    //    private var keyboardOffset: NSLayoutConstraint!
     private var mainViewWidthRegular: NSLayoutConstraint!
     private var mainViewWidthCompact: NSLayoutConstraint!
     private var topSpacerHeight: NSLayoutConstraint!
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
```diff
@@ -34,26 +33,28 @@ final class TeamCreationStepController: UIViewController {
...
     fileprivate var secondaryViewsStackView: UIStackView!
     fileprivate var errorViewContainer: UIView!
     private var mainViewContainer: UIView!
+    private var topSpacer: UIView!
+    private var bottomSpacer: UIView!
 
     private var backButton: UIView?
 
-    /// Text Field
+    /// mainView is a textField or CharacterInputField in team createion screens
```
vytis: Spelling - "createion"

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationStepController.swift
```diff
@@ -195,7 +218,11 @@ final class TeamCreationStepController: UIViewController {
         secondaryViewsStackView.spacing = 24
         secondaryViewsStackView.translatesAutoresizingMaskIntoConstraints = false
 
-        [backButton, headlineLabel, subtextLabel, mainViewContainer, errorViewContainer, secondaryViewsStackView].flatMap {$0}.forEach { self.view.addSubview($0) }
+        [backButton, headlineLabel, subtextLabel, mainViewContainer, errorViewContainer, secondaryViewsStackView, topSpacer, bottomSpacer].flatMap {$0}.forEach { self.view.addSubview($0) }
+
+        [topSpacer, bottomSpacer].forEach() { view in
+            self.view.sendSubview(toBack: view)
```
vytis: You can then add them to the beginning of the array few lines above.

PR #1503, opened 2017-12-05
zios-9040 fix crash when image screen rotates

billypchan:
### Issues

In NotificationWindowRootViewController, supportedInterfaceOrientations() is recursive called and it crashes when stack overflowed.

### Causes

1. A iOS internal class UIInputWindowController is the top view controller. When getting topViewController.supportedInterfaceOrientations, it calls NotificationWindowRootViewController.supportedInterfaceOrientations recursively. 

### Solutions

Check the topViewController's name. If it is "UIInputWindowController", return [UIViewController wr_supportedInterfaceOrientations].


typfel:
As a future task we should re-think this auto rotate logic since it seems easy to break and hard to understand.

PR #1507, opened 2017-12-06
Fix/zios 9334 collection button crash

billypchan:
### Issues

iPad with iOS9 crashes when presenting collection screen.

### Causes

In iOS9, when CollectionsViewController is loading, self.traitCollection.horizontalSizeClass is .unspecified instead of .regular, which causes supportedInterfaceOrientations and shouldAutorotate return invalid value(.portrait). 

### Solutions

Add missing orientation handling code.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
@@ -174,6 +174,8 @@ final public class CollectionsViewController: UIViewController {
 
     // MARK:- device orientation
 
+
+    /// Notice: for iPad with iOS9 in landscape mode, horizontalSizeClass is .unspecified (.regular in iOS11).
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Collections/CollectionsViewController.swift
```diff
@@ -170,22 +170,35 @@ final public class CollectionsViewController: UIViewController {
...
+        case .compact:
+            return false
+        default:
             return true
         }
-        else {
-            return false
+    }
+
+    override public var preferredInterfaceOrientationForPresentation: UIInterfaceOrientation {
```
typfel: Why is it necessary to implement `preferredInterfaceOrientationForPresentation`. The crash was caused by `supportedInterfaceOrientations` returning `.portrait` instead of `.all`?

billypchan:
@typfel removed preferredInterfaceOrientationForPresentation as your suggestion.

PR #1508, opened 2017-12-07
Add support for Availability

nicorsm:
## What's new in this PR?

I've introduced support for Availability on UI side:
- You can check and change your availability from self profile view and from the header of conversations list. 
- You can see others availability on their profile views, on the conversation list, on the Start UI, inside the participants view of a group conversation, and as a placeholder of the input bar in a 1:1 conversation.

Actually, I refactored the `ConversationTitleView` in order to have a shared behaviour between this view and the new one, `AvailabilityTitleView`, since they share most of the logic. Both classes derivate from `AbstractTitleView`, which has a `AbstractTitleViewTemplate` protocol that acts as Template Method (methods inside the protocol need to be implemented inside the subclasses in order to have a compliant class).

I made a common method that handles the composition of availability attributed strings (with/without icons, with/without arrow, etc.) inside `AvailabilityExtensions`. This extension will change and will be implemented as a class.

I also updated the `WireStyleKit` class (kudos to @eike96).

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-protos/pull/17
- [x] https://github.com/wireapp/wire-ios-data-model/pull/399
- [x] https://github.com/wireapp/wire-ios-message-strategy/pull/104

## Notes

**Code is subject to changes**: I need to deeply refactor the classes that makes use of this new feature, so classes and methods signatures will change.


Wire-iOS/Sources/UserInterface/Availability/AvailabilityExtensions.swift
```diff
@@ -0,0 +1,88 @@
...
+    var name: String {
+        switch self {
+        case .none:         return "availability.none".localized
+        case .available:    return "availability.available".localized
+        case .away:         return "availability.away".localized
+        case .busy:         return "availability.busy".localized
+        }
+    }
+    
+    var imageEnum: ZetaIconType? {
```
typfel: Since the name of the type is `ZetaIconType` I would name this var `iconType`

Wire-iOS/Sources/UserInterface/Availability/AvailabilityExtensions.swift
```diff
@@ -0,0 +1,88 @@
...
+
+import Foundation
+
+
+@objc public enum AvailabilityStyle: Int {
+    case profiles, lists, headers, placeholder
+}
+
+extension Availability {
+    var name: String {
```
typfel: Maybe be more specific and name it `localizedName`

Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -282,7 +284,12 @@ private struct InputBarConstants {
     }
 
     func updatePlaceholder() {
-        textView.placeholder = placeholderText(for: inputBarState)
+        
+        if let otherUser = otherUser, otherUser.isTeamMember {
```
typfel: I would prefer to keep this logic in the `InputBarViewController`. I propose instead of injecting `otherUser` we add a `otherUserAvailability : Availability` property which we can query here.

Wire-iOS/Sources/UserInterface/Availability/AbstractTitleView.swift
```diff
@@ -0,0 +1,119 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import UIKit
+import Cartography
+import Classy
+
+protocol AbstractTitleViewTemplate {
```
mikeger: Question: this protocol name is referring to the template, but it's neither having the generics nor it's a PAT. Is template referring to it being a protocol?

nicorsm: Yes. But this is totally removed from the last commits I've made that I'll push in the next minutes.

Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -0,0 +1,125 @@
...
+    private var user: ZMUser
+    private var style: AvailabilityTitleViewStyle
+    private var observerToken: Any?
+    
+    public init(user: ZMUser, style: AvailabilityTitleViewStyle) {
+        self.user = user
+        self.style = style
+        
+        var titleColor: UIColor?
+        var titleColorSelected: UIColor?
```
johnxnguyen: these don't need to be optional right? Since they are guaranteed to have a value after the if clause, you can declare these variables with `let`

Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -0,0 +1,125 @@
...
+        
+        var titleFont : UIFont?
+        if style == .header {
+            titleFont = FontSpec(.normal, .semibold).font
+        } else {
+            titleFont = FontSpec(.small, .semibold).font
+        }
+        
+        super.init(color: titleColor!, selectedColor: titleColorSelected!, font: titleFont!)
+
```
johnxnguyen: .... then you wouldn't need to force unwrap these.

Wire-iOS/Sources/UserInterface/Availability/TitleView.swift
```diff
@@ -0,0 +1,106 @@
...
+        createViews()
+    }
+    
+    private func createConstraints() {
+        constrain(self, titleButton) { view, button in
+            button.leading == view.leading
+            button.trailing == view.trailing
+            button.top == view.top
+            button.bottom == view.bottom
+        }
```
johnxnguyen: you can express this in one line with `button.edges == view.edges` 😜

Wire-iOS/Sources/UserInterface/Participants/ParticipantsListCell.m
```diff
@@ -74,8 +75,13 @@ - (void)addSubviews
 - (void)updateForUser:(ZMUser *)user inConversation:(ZMConversation *)conversation
 {
     self.userImageView.user = user;
-    self.nameLabel.text = [user.displayName uppercaseString];
     self.guestLabel.hidden = ![user isGuestInConversation:conversation];
+
+    if (ZMUser.selfUser.isTeamMember) {
```
johnxnguyen: this is a probably a bit nit-picky, but on line 52 in `ConversationListItemView+Update.swift` use check if `ZMUser.selfUser.hasTeam`. Although maybe very unlikely, `.hasTeam` could be nil even if the `.membership` is not nil, whereas `.isTeamMember` is equivalent to `nil != membership`. I think it would be a good idea to be completely consistent when determining if you should show this team only feature, otherwise it could happen that the feature shows in some places but not others.

Wire-iOS Tests/AvailabilityLabelTests.swift
```diff
@@ -0,0 +1,117 @@
...
+        let label = UILabel()
+        label.attributedText = attributedString
+        label.font = FontSpec(.medium, .regular).font
+        label.sizeToFit()
+        return label
+    }
+    
+    // MARK: - Placeholder labels
+
+    func testThatItRendersCorrectly_Placeholder_NoneAvailability() {
```
mikeger: From the screenshot it looks like the localization is missing for the status string.

Wire-iOS/Sources/Helpers/UIStackView+CustomSpacing.swift
```diff
@@ -0,0 +1,83 @@
...
+     
+     This is a approximation of the addCustomSpacing method only available since iOS 11. This method
+     has several constraints:
+     
+     - The stackview must be initialized with customSpacedArrangedSubviews
+     - spacing dosesn't update if views are hidden after this method is called
+     - custom spacing can't be smaller than 2x the minimum spacing
+     */
+    func wr_addCustomSpacing(_ customSpacing: CGFloat, after view: UIView) {
+        if let spacerIndex = subviews.index(of: view)?.advanced(by: 1) {
```
mikeger: I would consider using the `guard let` here to avoid the indentation.

Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
```diff
@@ -21,104 +21,48 @@ import UIKit
...
-
-        // The attachments contain images which break the centering of the text inside the button.
-        // If there is an attachment in the text we need to adjust the constraints accordingly.
-        let hasAttachment = configure(conversation, interactive: interactive)
-        frame = titleButton.bounds
-        createConstraints(hasAttachment)
+        
+        configure()
+        
+        self.tapHandler = { button in
```
mikeger: Seems like the default handler is empty, we can remove it then I guess?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -440,6 +449,24 @@ - (void)createTypingIndicatorView
...
+    }
+    
+    Availability connectedUserAvailability = self.conversation.connectedUser.availability;
+    
+    if (connectedUserAvailability == AvailabilityNone) {
+        self.inputBar.availabilityPlaceholder = nil;
+    } else {
+        self.inputBar.availabilityPlaceholder = [AvailabilityStringBuilder stringFor:self.conversation.connectedUser
+                                                                                with:AvailabilityLabelStylePlaceholder
+                                                                               color:_inputBar.placeholderColor];
```
mikeger: I think it is better for consistency reasons to use `self.inputBar` instead of the ivar.

Wire-iOS Tests/AvailabilityLabelTests.swift
nicorsm: This case shouldn't happen since the placeholder is the default one provided by the TextView in the case user doesn't have availability. Anyway, I've changed the test.

Wire-iOS/Sources/UserInterface/Participants/ParticipantsListCell.m
nicorsm: Changed check in `ConversationListItemView+Update.swift` to `.isTeamMember`.

Wire-iOS/Sources/Helpers/UIStackView+CustomSpacing.swift
```diff
@@ -0,0 +1,81 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+extension UIStackView {
```
mikeger: @nicorsm I was thinking about this implementation and I found the following:
1. Once initialized with `init(customSpacedArrangedSubviews:)`, the stack view would not be possible to use with other common stack view methods (for example, the distributed alignment would not make sense, adding another arranged subview would break the logic, etc.).
2. For any `UIStackView` that is not initialized with the `init(customSpacedArrangedSubviews:)`, calling `wr_addCustomSpacing(_:, after:)` would silently fail.

I think considering this it would be better to create the custom `UIView` subclass that would accommodate the  `UIStackView` and implement the 2 given methods (composition). In this way we can solve the 2 issues I underlined.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileDetailsViewController.m
```diff
@@ -149,7 +148,7 @@ - (void)createUserImageView
     self.userImageView = [[UserImageView alloc] initWithMagicPrefix:@"profile.user_image"];
     self.userImageView.userSession = [ZMUserSession sharedSession];
     self.userImageView.translatesAutoresizingMaskIntoConstraints = NO;
-    self.userImageView.size = UserImageViewSizeBig;
+    self.userImageView.size = IS_IPAD ? UserImageViewSizeGiant : UserImageViewSizeBig;
```
typfel: This will breakdown if you use the split screen mode on ipad

Issue #1125, opened 2017-07-11
Keyboard shortcuts


mergesort:
I’ve recently started using Wire on the iPad a lot more, and really wish that it had better keyboard shortcut support. Ideally there would be shortcuts for navigation, but even the ability to send a message with `⌘ + enter` would be a good enough start.

Thanks a lot!

mikeger:
Hey @mergesort , thanks for the request, this sounds perfectly valid, we'll prioritize the request. You can help us a lot if you would spend time defining the shortcuts, i.e. checking with other apps for the common key combinations to switch between conversations and other actions. Thanks a lot!

richpersaud:
Input focus should also be moved.  Currently, if you alt-tab or otherwise switch to Wire, a screen touch is needed to move input focus to the new message field.  It should be possible to type immediately after switching to the app.

mikeger:
@richpersaud can the focus be moved in Wire with tab? I have only experience with Android tablet where all the apps are allowing to be navigated with tab.

mikeger:
Thanks to @daehn we would have `⌘ + enter` sending in the 2.40 release.

daehn:
I think another shortcut which would be nice to have (and which we already have on desktop) is ↑ to start editing the last sent message.

mikeger:
Correction: it would go into 2.41, 2.40 was a hotfix.

PR #1511, opened 2017-12-07
Added support for up arrow key

vorobievalex:
## What's new in this PR?
Up arrow keyboard shortcut starts editing of last message.
(The same shortcut is already available on desktop app)

### Issues
Keyboard shortcuts support requested in #1125

### Solutions
Up Arrow key in InputBar initiates editLastMessage in ConversationContentViewController (via ConversationViewController).
If there is a lastEditableMessage in the conversation, wantsToPerformAction is sent to ConversationContentViewController for the message found.

## Dependencies
Needs releases with:
- [ ] https://github.com/wireapp/wire-ios-data-model/pull/400

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -214,6 +214,7 @@
 "conversation.connection_view.in_address_book" = "in Contacts";
 
 "conversation.input_bar.shortcut.send" = "Send Message";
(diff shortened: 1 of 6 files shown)


vytis:
I have tried running your fork on my machine and all seems to be working perfectly! Thanks again for the contribution.
Currently we are releasing an update, so we will wait until this is done to merge your changes in. In the meantime I am adding a "Do not merge" label, but this is only so we would not get confused on what gets out in the next release!

vytis:
Thanks again for implementing this 🍾

PR #1512, opened 2017-12-07
ZIOS-9337 iPad - conversation list is not visible after rotation

billypchan:
### Issues

- menu button in conversation view is not hidden after rotates to landscape mode
- conversation list is not visible after rotates to landscape mode

### Causes

In SplitViewController, updateLeftViewVisibility() is not called after updateForSize() is called.
in ConversationViewController, updateLeftNavigationBarItems() should be called when viewWillLayoutSubviews
### Solutions

Add the missing UI update code.


billypchan:
I will clean up UIViewController : + (UIInterfaceOrientationMask)wr_supportedInterfaceOrientations and refactor supportedInterfaceOrientations() and restart the PR.

PR #1516, opened 2017-12-08
Show alert when we suspect something is wrong and we want logs

typfel:
## What's new in this PR?

Show debug alert for internal users we might have a inconsistency in SE to prompt the user to send logs.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/403
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/652


Cartfile.resolved
```diff
@@ -20,14 +20,14 @@ github "wireapp/ono" "1.4.0"
...
-github "wireapp/wire-ios-data-model" "83.1.0"
+github "wireapp/wire-ios-data-model" "fix/remove-participants-logs"
 github "wireapp/wire-ios-images" "12.0.1"
 github "wireapp/wire-ios-link-preview" "7.0.0"
 github "wireapp/wire-ios-message-strategy" "69.0.0"
 github "wireapp/wire-ios-protos" "12.0.1"
 github "wireapp/wire-ios-request-strategy" "68.0.0"
 github "wireapp/wire-ios-share-engine" "55.0.0"
-github "wireapp/wire-ios-sync-engine" "134.1.2"
+github "wireapp/wire-ios-sync-engine" "fix/notify-when-we-detect-usual-state"
```
nicorsm: Reference to the local branch left in the code.

typfel: It's intentional. I want the tests to pass. I'll update when I release frameworks.

nicorsm: Right 👍

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
```diff
@@ -548,6 +549,13 @@ - (void)requestLoopNotification:(NSNotification *)notification;
     [DebugAlert showWithMessage:[NSString stringWithFormat:@"A request loop is going on at %@", path] sendLogs:YES];
 }
 
+#pragma mark - SE inconsistency notification
+
+- (void)potentialErrorNotification:(NSNotification *)notification;
+{
+    [DebugAlert showWithMessage:[NSString stringWithFormat:@"We detected a potential error, please send logs"] sendLogs:YES];
```
vytis: Maybe worth adding a more descriptive error, maybe using data from notification payload? Would be good for users to know what is happening there.

Issue #535, opened 2017-01-03
GIF: don't show search phrase


MadEgg:
When you search for GIF's using some search phrase, the recipient receives both the GIF and the text you were searching for. I don't like this behaviour: due to weird tagging of pictures I sometimes look for adapted search phrases that don't relate to the message I'm trying to convey. I'd prefer this info to stay at my own device.

Not really iOS-related I suppose, but since I'm using the iOS app I'm posting it here.

tschoof:
Definitely feels kinda broken, I often delete that message by hand. I shouldn't have to do that.

PR #1517, opened 2017-12-09
Prevented sending search text together with giphy gif

vorobievalex:
## What's new in this PR?
Search string used to find a gif at giphy.com will not be sent to the conversation.

### Issues
I agree with the author of https://github.com/wireapp/wire-ios/issues/535.
As a user, I do not expect a search text to be sent to a conversation. This can lead to some awkward situations if I use an inappropriate text to search.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -995,7 +995,6 @@
 "force.update.ok_button" = "Go to App Store";
 
 // Third Party
-"giphy.conversation.message" = "%@ · via giphy.com";
 "giphy.conversation.random_message" = "via giphy.com";
 "giphy.error.no_more_results" = "no more gifs";
 "giphy.error.no_result" = "no gif found";
--- a/Wire-iOS/Resources/ar.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/ar.lproj/Localizable.strings
@@ -995,7 +995,6 @@
 "force.update.ok_button" = "اذهب لـApp Store";
 
 // Third Party
-"giphy.conversation.message" = "%@ · عبر giphy.com";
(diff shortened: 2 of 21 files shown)


mikeger:
Hey @vorobievalex ! Thanks for the pull request, again an exceptional contribution. We are investigating right now the consequences in the light of the GIPHY API usage rules. Until it's cleared this one would unfortunately need to stay un-merged.

malicioustoker:
Any word on this?

mikeger:
Hey @malicioustoker , thanks for the reminder.

Small background: as we started to use the GIPHY service, we wanted to make the "translation" of text to GIF. Currently however I think we are using the GIPHY simply to find the emoticon or a meme relevant to the text.

# In the short run:

According to the [API rules of GIPHY](https://giphy.com/terms), we are prescribed to add the label "Powered by GIPHY" where the API is used. Therefore, generally, we cannot just remove the label from the message; what we should do is remove the label from message and add the label on the screen where the API is used.

In the current implementation, the GIPHY API is used only on the search GIF screen. So to merge this PR we only need to add one label on the GIF search screens. May be @eike96 can help us to design it properly.

# In the long run:

The other clients must adopt the same logic not to send the redundant text message together with GIF.

# In the ideal world:

The GIFs must be transferred as MP4/H.264 video messages to save the traffic. The flag to play on demand should be then added to the video messages.

vorobievalex:
@mikeger, this PR only removes search string from the message. The string "via giphy.com" is still attached to the gif.
Just like in case when no search is performed, when a trending gif is sent.

mikeger:
Spring cleanup: unfortunately I am going to close this PR for now, we are going to revisit the GIF integration and it would make sense to work on this part then. Thanks for the effort @vorobievalex !

TestingBytes:
Could this be re-opened? This string doesn't show up in any other application that uses Giphy.

PR #1521, opened 2017-12-11
add shortcut to cancel message editing

johnxnguyen:
## What's new in this PR?

I've added a minor extension to the  "edit last message" keyboard shortcut (thanks @vorobievalex ) so that it is aligned with the functionality found on desktop, namely that once we are editing a message, the `esc` will cancel editing. Furthermore, I've restricted which shortcuts are available depending on the input bar state, so that the `edit last message` shortcut is only available when the user is not already editing a message, and the `cancel` shortcut is only available when the user is currently editing a message.

### Attachments

simulator screen shot - ipad pro 9 7-inch - 2017-12-11 at 15 55 52

simulator screen shot - ipad pro 9 7-inch - 2017-12-11 at 15 56 10


johnxnguyen:
You're welcome @vytis !

PR #1531, opened 2017-12-15
ZIOS-9372 fix crash when fast tapping "REGISTRATION" and "LOG IN" tab buttons in Tabbar screen

billypchan:
### Issues

App crashes when user fast tapping "REGISTRATION" and "LOG IN" tab buttons in Tabbar screen

### Causes

UIViewController.transition() is called before earlier transition is done.

### Solutions

Add a flag to indicate the transition is in process. It is false in init, set to true when transition starts and set to false when transition completes. If the flag is true, transition is not called.

### DIFF
--- a/WireExtensionComponents/Views/TabBar/TabBarController.swift
+++ b/WireExtensionComponents/Views/TabBar/TabBarController.swift
@@ -58,7 +58,7 @@ open class TabBarController: UIViewController {
     fileprivate var presentedTabBarViewController: UIViewController?
     fileprivate var tabBar: TabBar?
     fileprivate var contentView: UIView!
-    fileprivate var isTransiting: Bool = false
+    fileprivate var isTransitioning: Bool = false
 
     required public init(viewControllers: [UIViewController]) {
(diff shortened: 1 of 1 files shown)


WireExtensionComponents/Views/TabBar/TabBarController.swift
```diff
@@ -1,71 +1,65 @@
...
-    
-    fileprivate var presentedTabBarViewController : UIViewController?
-    fileprivate var tabBar : TabBar?
-    fileprivate var contentView : UIView!
-    
+
+    fileprivate var presentedTabBarViewController: UIViewController?
+    fileprivate var tabBar: TabBar?
+    fileprivate var contentView: UIView!
+    fileprivate var isTransiting: Bool = false
```
typfel: typo: `isTransiting ` > `isTransitioning `

typfel:
It would be helpful for reviewers if you applied the code formatter in a separate commit in the future.

mikeger:
@billypchan please update the PR for review so we can merge it.

PR #1534, opened 2017-12-17
ZIOS-9347: Added support for switching between accounts in share extension

nicorsm:
### Issues

Currently, it's not possible to switch between accounts inside the Share Extension. You should open the container app and open the Self Profile view. It's pretty annoying!  😄

### Solutions

I've added the possibility to switch between accounts. The behavior is similar to the one already implemented for choosing conversations: if multiple accounts are logged into the app, the "Account" row appears before the "Conversation" row. Then you can select which account to switch. If the user is in a team, you'll also see the name of the team in the account selection view.

simulator screen shot - iphone x - 2017-12-17 at 12 20 17
simulator screen shot - iphone x - 2017-12-17 at 12 20 22

### DIFF
--- a/Wire-iOS Share Extension/AccountSelectionViewController.swift
+++ b/Wire-iOS Share Extension/AccountSelectionViewController.swift
@@ -40,7 +40,6 @@ class AccountSelectionViewController : UITableViewController {
         
         tableView.register(UITableViewCell.self, forCellReuseIdentifier: cellReuseIdentifier)
         
-        //preferredContentSize = UIScreen.main.bounds.size
         definesPresentationContext = true
     }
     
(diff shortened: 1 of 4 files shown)


typfel:
I don't think we should save the selected account when switching account in the share extension.

Wire-iOS Share Extension/AccountSelectionViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+    init(accounts: [Account], current: Account?) {
+        self.accounts = accounts
+        
+        super.init(style: .plain)
+        
+        self.current = current
+        
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: cellReuseIdentifier)
+        
+        //preferredContentSize = UIScreen.main.bounds.size
```
mikeger: Debug leftover.

mikeger:
@nicorsm please update the PR in order to get on the release train in time 🚅 🚋

PR #1538, opened 2017-12-18
Changed "Choose a status" label to "Set a status"

nicorsm:
## What's new in this PR?

Changed "Choose a status" label for Availability ActionSheet to "Set a status"

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
+++ b/Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
@@ -103,7 +103,7 @@ extension AvailabilityTitleView {
     
     var actionSheet: UIAlertController {
         get {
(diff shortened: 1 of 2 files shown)


nicorsm:
Should I remove the old translation entry for "Choose a status"?

mikeger:
@nicorsm if the string is not used it makes sense to remove it to save some work for the translators.

Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -103,7 +103,7 @@ extension AvailabilityTitleView {
     
     var actionSheet: UIAlertController {
         get {
-            let alert = UIAlertController(title: "availability.message.title".localized, message: nil, preferredStyle: .actionSheet)
```
typfel: So `availability.message.title` is unused now? We should delete it then.

nicorsm: Exactly

PR #1543, opened 2017-12-18
Fixed start UI camera picker using iOS integrated one (iPhone X)

mikeger:
### Issues

Users noticed that on iPhone X the custom picture picker camera UI is pretty broken. In other places of the app the custom UI seems to work fine.

### Solutions

I decided to replace the custom UI with iOS integrated one, since then we don't have the overhead of supporting it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
@@ -26,11 +26,49 @@
 
 #import "Wire-Swift.h"
 
-@interface ConversationListViewController (StartUI) <CameraViewControllerDelegate>
+@import MobileCoreServices;
+
+@interface ConversationListViewController (StartUI)
 
 @end
 
+typedef void (^ConversationCreatedBlock)(ZMConversation *);
+
 @implementation ConversationListViewController (StartUI)
+
+- (void)withConversationForUsers:(NSSet<ZMUser *>*)users callback:(ConversationCreatedBlock)onConversationCreated {
+    
+    [self dismissPeoplePickerWithCompletionBlock:^{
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
```diff
@@ -26,12 +26,50 @@
...
+            
+            ZMConversation __block *conversation = nil;
+            [[ZMUserSession sharedSession] enqueueChanges:^{
+                Team *team = ZMUser.selfUser.team;
+
+                conversation = [ZMConversation insertGroupConversationIntoUserSession:[ZMUserSession sharedSession]
+                                                                     withParticipants:users.allObjects
+                                                                               inTeam:team];
+            } completionHandler:^{
+                dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.3 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
```
typfel: What purpose does this `dispatch_after` have?

mikeger: Moved from the old code: https://github.com/wireapp/wire-ios/pull/1543/files#diff-13928cc74df935b3538fe3d30deb4fafL238

mikeger: Will try to remove it, let's see if it's necessary.

mikeger:
@typfel removed it, seems to be unnecessary.

typfel:
Can't you delete `CameraViewController`? This was the last usage of it?

mikeger:
@typfel I thought about it, but it is still used for picking user picture. There it works with iPhone X.

PR #1545, opened 2017-12-19
Fixed insets margin on Profile Devices Screen

nicorsm:
## What's new in this PR?

The devices Tableview was cut off by the Safe Area. I've changed the constraints for the TabBarController, allowing it to reach the bottom margin under the safe area, and I've moved the footer view of the profile in order to achieve the same appearance as before.


nicorsm:
Forgot to mention that this bug was relative to iPhone X 😅

PR #1552, opened 2017-12-21
add a guard for message nil check

billypchan:
### Issues

App crashes when ConversationCell is calling likeMessage()

### Causes

ConversationCell gets class member var message (type: id< ZMConversationMessage >), which may be nil at that time point.

### Solutions

Add a nil check guard in likeMessage()

## Notes

This PR is a simliar to https://github.com/wireapp/wire-ios/pull/1539 .

## Discussion
In ConversationCell, nullable keyword should be added  to 
````
@property (nonatomic, readonly) id<ZMConversationMessage>message;
````
to make sure Swift code must unwrap it before access. All other properties in ConversationCell should define Nullability also.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift
@@ -18,7 +18,6 @@
 
 import UIKit
 
-
 public extension ConversationCell {
 
     public func createLikeButton() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Like.swift [hunk 1]
```diff
@@ -34,28 +33,29 @@ public extension ConversationCell {
...
     }
-    
+
     @objc public func didDoubleTapMessage(_ sender: AnyObject!) {
         self.likeMessage(sender)
     }
-    
+
     @objc public func likeMessage(_ sender: AnyObject!) {
+        guard let _ = message else { return }
```
billypchan: only changes in this PR.

vytis: Just a small remark, but it would be best to avoid changing whitespace/formatting too much unless it's very annoying. The main reason is that sometimes it's useful to do `git blame` to get some context from the PR that introduced these changes. So when indentation is changed we see the last change only and not the PR that actually has any meaningful context.

PR #1563, opened 2017-12-28
fix: unwrap user name before access

billypchan:
### Issues

App crashes when access user name.

### Causes

user name is nil when access.

### Solutions

Unwrap user.name before access.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
+++ b/Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
@@ -27,7 +27,10 @@
         
         switch style {
             case .list: do {
-                title = user.name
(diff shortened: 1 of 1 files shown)


mikeger:
@billypchan one comment in regard of those fixes: I think the root cause of those issues is Obj-C portability. So next time someone who would write the code in Swift using, say, `name` property the issue would reappear, since the meta information of this property is not showing that it can eventually be nil. However, if we are going to change the optionality of a widely used property, like name, we would have to dig a lot of code around it. My proposal would be to try to understand the reason why the property is nil in some particular case, for example it is nil because the object is not anymore associated with the managed object context.

billypchan:
@mikeger we should improve the current crash logging message. e.g. when some essential property is nil, call fatalError() with some more info.

PR #1567, opened 2017-12-29
Fix/zios 9408 crash for unwrapped user name

billypchan:
### Issues

App crashes when sort a array of user by name if name is nil.

### Causes

user name is nil when access.

### Solutions

Unwrap user.name and give a default value for sorting.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
@@ -263,7 +263,15 @@ public class SearchResultsViewController : UIViewController {
         }
         
         contactsSection.contacts = contacts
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -263,7 +263,25 @@ public class SearchResultsViewController : UIViewController {
         }
         
         contactsSection.contacts = contacts
-        teamMemberAndContactsSection.contacts = Set(teamContacts + contacts).sorted { $0.name.compare($1.name) == .orderedAscending }
+
+        teamMemberAndContactsSection.contacts = Set(teamContacts + contacts).sorted {
+            var name0, name1: String
+            if let _ = $0.name {
```
vytis: This can be done much easier as 
```
let name0 = $0.name ?? ""
```

PR #1568, opened 2018-01-02
Mark conversation as read or unread

mikeger:
This feature is useful to mark some conversation as the one requiring attention when you already finished reading it or marking a non-important conversation as read.

Not every conversation can be marked as unread, since it must contain at least one message with the timestamp.

I also wanted to support marking the whole account's conversations as read, and marking all accounts conversations as read. This is useful for the app icon menu, where we can add the button to mark all conversations as read.

# Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/413
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/660


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -4083,6 +4156,8 @@
 		8F42C52F199244A700288E4D = {
 			isa = PBXGroup;
 			children = (
+				8751A6DC1FF6611D00804A58 /* WireDataModel.xcodeproj */,
```
vytis: External dependencies!

daehn:
@mikeger do the labels still apply?

mikeger:
@daehn needs the SE bump: https://github.com/wireapp/wire-ios-sync-engine/pull/660

PR #1571, opened 2018-01-03
prevent changing frame's origin after config is called.

billypchan:
### Issues

AvailabilityTitleView misplaced after status changes

### Causes

AvailabilityTitleView's frame is reset to zero after update.

### Solutions

reserve the frame's origin after update the content

## Notes

@nicorsm It is better not to call configure() for UI refresh since it will call createConstraints() every time.


mikeger:
Thanks @billypchan ! Merged directly to `release/3.5`.

PR #1574, opened 2018-01-04
Fix/zios 9404 camera keyboard collection view does not refreshed after toggling

billypchan:
### Issues

Some photos are not loaded in camera keyboard after refresh.

### Causes

RequestID was used as key to check the image is requested by the cell. However, requestID may be zeroed after cancelImageRequest is called.

### Solutions

Ref to apple's "Example app using Photos framework" sample, use PHAsset.localIdentifier as the key to check the image received in block is the image the cell requested.

https://developer.apple.com/library/content/samplecode/UsingPhotosFramework/Introduction/Intro.html#//apple_ref/doc/uid/TP40014575

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/AssetCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/AssetCell.swift
@@ -85,15 +85,15 @@ open class AssetCell: UICollectionViewCell, Reusable {
             let maxDimensionRetina = max(self.bounds.size.width, self.bounds.size.height) * (self.window ?? UIApplication.shared.keyWindow!).screen.scale
 
             representedAssetIdentifier = asset.localIdentifier
-            self.imageRequestTag = manager.requestImage(for: asset,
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/AssetCell.swift
```diff
@@ -26,8 +26,9 @@ open class AssetCell: UICollectionViewCell, Reusable {
     let imageView = UIImageView()
     let durationView = UILabel()
     
-    var imageRequestTag: PHImageRequestID = 0
-    
+    var imageRequestTag: PHImageRequestID = PHInvalidImageRequestID
+    var representedAssetIdentifier: String!
```
mikeger: Why this one is necessary? I think we never read the value of this property.

billypchan: @mikeger It is for manager.cancelImageRequest(self.imageRequestTag).

PR #1575, opened 2018-01-04
Camera View Controller minor layout issues with iPhone X

nicorsm:
## What's new in this PR?

Added safe area margins for top and bottom bar of Camera View Controller. Fixed lateral margins for icons.


nicorsm:
The insets were already fixed, I've adjusted a little _bump_ between icons of `ProfileSelfPictureViewController` and `CameraViewController`.

PR #1578, opened 2018-01-05
ZIOS-8954: pin ongoing conversations to the top of conversations list

nicorsm:
### Issues

On the desktop wrapper, when someone calls you, all the conversations with incoming calls are brought to the top. From here, you can accept or deny the call. In case you deny, you still can join the call later by clicking on the green "Join" button:

screen shot 2018-01-05 at 15 23 41

The current behavior of the iOS app is to bring the calling conversation to the top of the list - like any other conversation with new contents - but without pinning it to the top.

### Solutions

Since the UI behavior of the desktop wrapper is to "include" the incoming calls view inside the title view, I've replicated the same thing by inserting a generic `ConversationListPinnedItemView` object inside `ConversationListTopBar`. 
The `ConversationListPinnedItemView` contains a non-scrollable `UICollectionView` which acts similar to the one in the conversation list.
This collection view is populated with conversations having the `canJoinCall` flag set to `true`. All the other conversations with the flag set to `false` will go to the conversations list. 

_N.B.: the current implementation works fine, but it's not pretty elegant (those `NSPredicate`s are so ugly 😱). Proposals for improving it are welcome! 🙂_

simulator screen shot - iphone x - 2018-01-05 at 15 28 15


typfel:
One issue with pinning conversation by adding them to the top bar is that if enough calls are ongoing the entire conversation list will be blocked.

We previously considering pinning calls by keeping them at the top of the normal list which can be scrolled away.

nicorsm:
Good point @typfel. A solution could be pinning only a certain amount of conversations (e.g. 3 conv.). After that amount, the behavior could remain the same as now: completely scrollable list, with ongoing calls on top of the list.

nicorsm:
Needs approval from design.

nicorsm:
The correct behavior is the one currently implemented in `develop`. The Jira ticket was "resolved" but still in backlog. I'll leave the branch here for future developments. My fault, sorry 😞

typfel:
@nicorsm Oh this is not nice. We should take care in future not end up with false tickets in sprints.

nicorsm:
@typfel I know, I'll double check tickets in future. Lesson learned.

PR #1585, opened 2018-01-08
Let AVS handle the audio category when previewing audio messages

typfel:
### Issues

After playing an audio message preview AVS would be confused about in which state the audio session is.

### Causes

When playing an audio message preview we would manually change audio session  category to `.playAndRecord`. 

### Solutions

Generally AVS has the ownership of the audio session and we should therefore request any changes to to it by using the API they expose for telling them when we are about to play sounds on our own.

### Notes

- Removed `error` property from the `MediaPlayer` since it was never referenced.
- MediaPlayers without a `sourceMessage` won't get a media bar in the conversation.

### DIFF
--- a/Wire-iOS/Sources/Components/AudioPlayer/MediaPlayerController.swift
+++ b/Wire-iOS/Sources/Components/AudioPlayer/MediaPlayerController.swift
@@ -49,7 +49,7 @@ class MediaPlayerController: NSObject {
 extension MediaPlayerController: MediaPlayer {
 
     var title: String? {
-        return message.fileMessageData?.filename ?? ""
+        return message.fileMessageData?.filename
     }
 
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/Components/AudioPlayer/MediaPlayerController.swift
```diff
@@ -48,11 +48,11 @@ class MediaPlayerController: NSObject {
 
 extension MediaPlayerController: MediaPlayer {
 
-    var title: String {
+    var title: String? {
         return message.fileMessageData?.filename ?? ""
```
daehn: Seems like the nil-coalescing `??` should be removed if the result should be an optional instead of an empty string now.

PR #1586, opened 2018-01-09
Feature/zios 9426 update the animation style form gap animation to breath animation

billypchan:
## What's new in this PR?

Update UI design of network status bar animation form gap animation to breath animation

## Dependencies

Needs releases with:
(This PR should release with the PR of "ZIOS-9095 Change visual appearance of the network-status bar")


Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
```diff
@@ -0,0 +1,106 @@
...
+        super.layoutSubviews()
+
+        // restart animation
+        if animating {
+            startAnimation()
+        }
+    }
+
+    func applicationDidBecomeActive(_ sender: Any) {
+        if animating && !isAnimationRunning {
```
daehn: It looks to me like this will never enter the if clause here because `isAnimationRunning` is returning the stored value of `animating`.

Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
```diff
@@ -0,0 +1,106 @@
...
+    func applicationDidEnterBackground(_ sender: Any) {
+        if animating {
+            stopAnimation()
+        }
+    }
+
+    func startAnimation() {
+        let anim = CAKeyframeAnimation(keyPath: "opacity")
+        anim.values = [0.4, 1, 0.4]
+        anim.isRemovedOnCompletion = false
```
daehn: Not sure about this, but remembered that I read to be careful with setting `isRemovedOnCompletion = false` and `fillMode = kCAFillModeForwards` here https://www.objc.io/issues/12-animations/animations-explained/ as well as here https://twitter.com/andy_matuschak/status/464799423785336832.

> Alternatively, you can tell the animation to remain in its final state by setting its fillMode property to kCAFillModeForwards and prevent it from being automatically removed by setting removedOnCompletion to NO. However, it’s a good practice to keep the model and presentation layers in sync, so this approach should be used carefully.

However, I think this might not be an issue in our case as we are removing the animation explicitly in `stopAnimation`.

Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
billypchan: @daehn It is my mistake, isAnimationRunning should returns layer.animation(forKey: BreathLoadingAnimationKey) != nil

PR #1587, opened 2018-01-09
Show debug alert when registering a shake motion gesture

daehn:
### Issues

There was no functionality to send debug logs before being logged in.

### Solution

This PR adds showing the debug alert to send debug logs after shaking the device (only in case the developer menu is enabled).


typfel:
`DebugAlert` will do nothing on production?

daehn:
@typfel it will only show an alert in case `ENABLE_DEVELOPER_MENU` is set, which it is not in production.

PR #1588, opened 2018-01-09
ZIOS-9429: Update username for current user in the header of Conversation List

nicorsm:
### Issues

If you try to update your username on the settings, it won't get updated on the header of the Conversation List until you restart the app or switch to another account.

### Causes

Currently, the two versions of the header in Conversation List don't have a mechanism to update the username on changes.

### Solutions

I've updated the `UserChangeInfo` observer into the `AvailabilityTitleView` (Teams) class and I've added it into `ConversationListTopBar` class (Non-teams).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
+++ b/Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
@@ -100,8 +100,9 @@ import WireDataModel
 extension AvailabilityTitleView: ZMUserObserver {
     
     public func userDidChange(_ changeInfo: UserChangeInfo) {
-        guard changeInfo.availabilityChanged else { return }
-        
+        guard changeInfo.availabilityChanged || changeInfo.nameChanged,
+            let user = changeInfo.user as? ZMUser else { return }
+        self.user = user
         provideHapticFeedback()
         configure()
     }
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -100,8 +100,9 @@ import WireDataModel
 extension AvailabilityTitleView: ZMUserObserver {
     
     public func userDidChange(_ changeInfo: UserChangeInfo) {
-        guard changeInfo.availabilityChanged else { return }
-        
+        guard changeInfo.availabilityChanged || changeInfo.nameChanged,
+            let user = changeInfo.user as? ZMUser else { return }
+        self.user = user
```
typfel: Why are you setting the user here?

nicorsm: I'm using it inside the class, but in fact, I could pass it as a parameter to the `configure()` function without keeping the reference.

nicorsm: ouch I need it inside the `updateAccessibilityLabel()` method :(

PR #1558, opened 2017-12-27
[Fix] timer retain issue in ConversationCell and GiphySearchViewController

billypchan:
### Issues

In ConversationCell and GiphySearchViewController, Timers were created with  Timer.scheduledTimer(timeInterval: target:self  selector:  userInfo: , repeats: ) which cause self is still retained by the timer and can not be freed.

### Causes
Timer.scheduledTimer(timeInterval: target: selector:  userInfo: , repeats: ) makes a strong reference to target. 

### Solutions

Rewrite Timer.scheduledTimer related call with scheduledTimer(withTimeInterval interval: TimeInterval, repeats: Bool, block: @escaping (Timer) -> Swift.Void) and capture self weakly in the block. Create a scheduledTimer with block method for iOS9-.

## Notes

UIApplication (runDuration), NetworkStatusViewController and VoiceChannelViewController are still using Timer.scheduledTimer(timeInterval: target:  selector:  userInfo: , repeats: ). Will check they are necessary or not in next PR.

PR #1594, opened 2018-01-10
GiphySearchViewController: checking for indexPath.row is not over the array count

billypchan:
### Issues

App crashes when calling override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell

### Causes

Since the crash log does not shown the exact line causes the crash, possible reasons can be:

- collectionView.dequeueReusableCell does not return a cell

- indexPath.row is beyond searchResultsController.results's count

### Solutions

Add checking for two possible reasons mentioned above.

## Notes

It PR is a follow up of 
https://github.com/wireapp/wire-ios/pull/1558


mikeger:
I will close this one for the moment keeping the branch intact.

PR #1597, opened 2018-01-10
Notifications tracking

vytis:
### Issues

To have a better overview how we handle notifications we need to first add tracking for the processing.

### Solutions

More information about actual events and how it is done can be found in https://github.com/wireapp/wire-ios-sync-engine/pull/664

In this PR we only implement persistent attributes for analytics layer, so that we could actually save the intermediate values before dispatching event.

Also it improves casting of values sent to Mixpanel. At the moment all values sent are treated as strings and that makes viewing numeric data much more complicated on their web UI.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/418
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/664


cardassian-tailor:
can someone comment on this branch and give some details as to what notifications are being tracked, where the data is sent, if it can be disabled, and for what purposes the data is being collected?

vytis:
@cardassian-tailor More technical information and main code changes can be checked in another PR - https://github.com/wireapp/wire-ios-sync-engine/pull/664

To summarise what we are trying to do here is to figure out how many notifications sent to users are processed successfully in the app. We do not track or send to 3rd parties anything that is inside the notification. 

With these changes, on each app launch we will send 4 numbers to Mixpanel that will show how many times a certain part in notification processing code is reached. This will then enable us to have a better understanding which part of the code potentially has problems handling notifications.

There is an easy to turn off analytics in app settings and more information about this topic can be found in the privacy [whitepaper](https://wire-docs.wire.com/download/Wire+Privacy+Whitepaper.pdf).

cardassian-tailor:
Thank you @vytis

PR #1598, opened 2018-01-11
prevent collectionviewcell access out of bound data source 

billypchan:
### Issues

App crash when access an empty set (ZMConversation.voiceChannel?.participants).

### Causes

It may be a side effect of a not released VoiceChannelViewController, since onversation.voiceChannel?.participants should always have more than 1 member.
Another possible reason would be onversation.voiceChannel?.participant is cleaned in very short time when the UI is updating.

### Solutions

Check index is within indices of participants (NSOrderedSet type) before access.
change VoiceChannelViewController.collectionView for let to unowned let. The collectionView should not retained by VoiceChannelViewController.

## Notes
This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1560
We should study why the viewcontroller still calling cellForItemAt when numberOfItemsInSection returns 0


typfel:
Add bound check for a case we know should not happen is treating the symptoms without fixing the underlaying issue. I think we should dig deeper into `WireCallCenter` to find out how we can end up in this situation.

billypchan:
@typfel I will add an integration test to reproduce the crash and prove the fixe solves the issue.

daehn:
IIRC there was a similar crash in a different place at some point where the issue was a queue switch or performing a runloop tick in-between updating the datasource and the view. Can't remember what exactly it was but it lead to the datasource and view update not being in sync. I would also propose not to work around this bug but instead to try to maybe simplify the logic around it.

mikeger:
I will close this one for the moment keeping the branch intact.

PR #1600, opened 2018-01-12
ZIOS-8936: Wrongly positioned or missing avatar background in skeleton when switching accounts

nicorsm:
### Issues

1) While switching between accounts, the blur effect of `SkeletonViewController` is different than the one in `BackgroundViewController`.
2) While switching from personal to team account, the `SkeletonViewController` background is grey.

### Causes

1) The `UIVisualEffectView` of `SkeletonViewController` wasn't scaled by 1.4x.
2) The personal avatar was saved inside the `Account` object only on personal accounts, not team accounts. See https://github.com/wireapp/wire-ios-data-model/pull/419 for further information.

### Solutions

1) I've unified the blurring logic in both views.
2) I've added a `teamImageData` field inside the `Account` object.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/419
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/665


Wire-iOS/Sources/UserInterface/SkeletonViewController.swift
```diff
@@ -254,9 +254,12 @@ class SkeletonViewController: UIViewController {
         super.init(nibName: nil, bundle: nil)
         
         backgroundImageView.contentMode = .scaleAspectFill
-        backgroundImageView.transform = CGAffineTransform(scaleX: 1.4, y: 1.4)
         
-        if let imageData = account.imageData, let image = UIImage(data: imageData) {
+        let factor = BackgroundViewController.backgroundScaleFactor
+        backgroundImageView.transform = CGAffineTransform(scaleX: factor, y: factor)
+        blurEffectView.transform = CGAffineTransform(scaleX: factor, y: factor)
```
typfel: So we are essentially scaling the image two times? Once on the imageView and then again on the effectView.

nicorsm: That's the behavior already used in `BackgroundViewController`. I've checked out that you can achieve the same effect by using an image with a size of 40px and scaling only the `UIImageView`. Changes in the next commit 🔜.

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -125,8 +125,9 @@ final public class BackgroundViewController: UIViewController {
         
         self.updateStatusBarBlurStyle()
         
-        self.blurView.transform = CGAffineTransform(scaleX: 1.4, y: 1.4)
-        self.imageView.transform = CGAffineTransform(scaleX: 1.4, y: 1.4)
+        let factor = BackgroundViewController.backgroundScaleFactor
+        //self.blurView.transform = CGAffineTransform(scaleX: factor, y: factor)
```
typfel: Commented out code

typfel:
@nicorsm will you update this PR?

nicorsm:
@typfel Yes, I'm updating dependencies locally right now.

PR #1602, opened 2018-01-15
ZIOS-9434: Audio Message Fixes

johnxnguyen:
### Issues

This PR addresses a few bugs relating to the audio messages UX:

1. In the audio message keyboard, the recording timer label is not visible.
2. Whilst previewing a recording, tapping the *redo* icon returns to the recording view, however, the previous recording will play until it finishes.
3. If the max recording limit is reached (25min), the user is presented with an alert view with two actions *cancel* and *send*. Meanwhile, the audio recording is lost. Cancelling dismisses the alert, but the user can not preview nor send the recording. Tapping *send* does nothing.

### Causes

1. The font color of the timer label is exactly the same as the background colour of it's superview.
2. Although the `AVAudioPlayer` within the `AudioEffectsPickerViewController` is stopped in `viewWillDisappear(:)`, this is not called when the player is currently playing.
3. When recording is stopped manually (before 25mins), some post-recording code is executed to ensure the file exists and if so, `fileURL` is set, which will be used when picking an effect. When the recording is stopped automatically (at 25mins), this functionality is not executed, so `fileURL` is nil.

### Solutions

1. Change the font color
2. Stop the audio player when `AudioEffectsPickerViewController` is removed from the view heirarchy.
3. Ensure the post-recording code is executed. Also, change the max duration alert to contain only a single action *ok*, which will simply dismiss the alert. This allows the user to preview, apply affects and send or redo as desired (See screenshots, but note the time limit has been reduced to 5 sec for debugging purposes).

### Attachments

simulator screen shot - iphone 8 - 2018-01-10 at 10 09 01

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
@@ -176,10 +176,10 @@ public final class AudioRecorder: NSObject, AudioRecorderType {
     
     @discardableResult public func stopRecording() -> Bool {
         audioRecorder?.stop()
-        return postRecordingCleanUp()
+        return postRecordingProcessing()
     }
     
-    fileprivate func postRecordingCleanUp() -> Bool {
+    fileprivate func postRecordingProcessing() -> Bool {
         recordLevelCallBack?(0)
         removeDisplayLink()
         guard let filePath = audioRecorder?.url.path , fm.fileExists(atPath: filePath) else { return false }
@@ -303,7 +303,7 @@ extension AudioRecorder: AVAudioRecorderDelegate {
         
         // in the case that the recording finished due to the maxRecordingDuration
         // reached, we should still clean up afterwards
-        if recordedToMaxDuration { _ = postRecordingCleanUp() }
+        if recordedToMaxDuration { _ = postRecordingProcessing() }
         
         self.recordEndedCallback?(recordedToMaxDuration)
     }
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
```diff
@@ -176,6 +176,10 @@ public final class AudioRecorder: NSObject, AudioRecorderType {
     
     @discardableResult public func stopRecording() -> Bool {
         audioRecorder?.stop()
+        return postRecordingCleanUp()
+    }
+    
+    fileprivate func postRecordingCleanUp() -> Bool {
```
typfel: Maybe `cleanUp` is a bit misleading since it actually does something important for actually sending the audio message (assigning `FileURL`).

How about something like: `postRecordingProcessing`.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
```diff
@@ -289,12 +289,12 @@ public final class AudioRecorder: NSObject, AudioRecorderType {
...
-        let recordedToMaxDuration: Bool
+        
+        var recordedToMaxDuration = false
+        
         if let maxRecordingDuration = self.maxRecordingDuration {
-            recordedToMaxDuration = currentDuration >= maxRecordingDuration
-        }
-        else {
-            recordedToMaxDuration = false
+            let duration = AVURLAsset(url: recorder.url).duration.seconds
```
typfel: So when hitting this issue with the bluetooth headset will it still end the recording but not show the max-duration alert?

johnxnguyen: ahh, so I had the BT fix in here but decided to separate it into a separate PR, but it seems I forgot to take this out. But yes it is as you describe, calculating the duration this way will correctly pass the max-duration check, but we will still see an incorrect current duration while recording.

PR #1603, opened 2018-01-15
Add invite members step to team on boarding flow

daehn:
## What's new in this PR?

**Add an invite team members step after creating a new team during the boarding flow.**

This screen differs from the other team flow screens as it has an expanding list in which newly invited emails are inserted to. As the list should push content towards the top when inserting a new email the `TeamMemberInviteViewController` uses an `UpsideDownTableView` instead of a regular one.  

## Dependencies

- [x] `wire-ios-sync-engine` after the branch `feature/team-invitation` has been merged.

## Notes

* The initial top and bottom content insets of the `UpsideDownTableView` are not yet calculated correctly for each screen size.

* This PR also includes the tracking call for the event `"team.sent_invite"` [(ZIOS-9479)](https://wearezeta.atlassian.net/browse/ZIOS-9479).

* This PR also fixes a bug in which the `TextFieldDescription` validation was not run when the return key was pressed.


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/TextFieldDescription.swift
```diff
@@ -26,12 +26,14 @@ final class TextFieldDescription: NSObject, ValueSubmission {
     var valueValidated: ValueValidated?
     var acceptsInput: Bool = true
     var validationError: TextFieldValidator.ValidationError
+    let uppercasePlaceholder: Bool
```
mikeger: Just to check, may be it's possible to set the placeholder string upper-cased? Any particular reason to have this logic in the control?

daehn: Before it was always set to be uppercased in the `AccessoryTextField` object itself so there was no control over it from the descriptor.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/TextFieldDescription.swift
```diff
@@ -26,12 +26,14 @@ final class TextFieldDescription: NSObject, ValueSubmission {
...
 
     fileprivate var currentValue: String = ""
 
-    init(placeholder: String, actionDescription: String, kind: AccessoryTextField.Kind) {
+    init(placeholder: String, actionDescription: String, kind: AccessoryTextField.Kind, uppercasePlaceholder: Bool = true) {
```
mikeger: Kind can be potentially passed as the generic parameter, then you don't have to cast for `create()`.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Team Invites/TeamMemberInviteHeaderView.swift
```diff
@@ -0,0 +1,70 @@
...
+        [titleLabel, subtitleLabel, bottomSpacerView].forEach(stackView.addArrangedSubview)
+        titleLabel.textAlignment = .center
+        subtitleLabel.textAlignment = .center
+        titleLabel.textColor = UIColor.Team.textColor
+        subtitleLabel.textColor = UIColor.Team.subtitleColor
+        addSubview(stackView)
+    }
+    
+    func updateHeadlineLabelFont(forWidth width: CGFloat) {
+        titleLabel.font = width > 320 ? TeamCreationStepController.headlineFont : TeamCreationStepController.headlineSmallFont
```
mikeger: I think it's more consistent to use iOS integrated mechanism to shrink the font size, like
```
titleLabel.numberOfLines = 1
titleLabel.minimumFontSize = 8
titleLabel.adjustsFontSizeToFitWidth = true
```

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/ViewDescriptions/TextFieldDescription.swift
daehn: I think this is related to the protocol `ViewDescriptor` returning a `UIView` (and not having an associated type). I didn't wan't to change too much here as this is a part of the existing team flow and was already there.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Team Invites/TeamMemberInviteHeaderView.swift
daehn: Yeah, that's correct. This behaves the same way (and was taken from) the other headers in the team creation flow do to be consistent (they shrink when the width is lower than 320).

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Team Invites/TeamMemberInviteViewController.swift
```diff
@@ -0,0 +1,168 @@
...
+        super.viewDidLoad()
+        setupViews()
+        createConstraints()
+    }
+    
+    private func setupViews() {
+        tableView.backgroundColor = .clear
+        tableView.allowsSelection = false
+        tableView.separatorStyle = .none
+        tableView.rowHeight = 56
```
mikeger: `tableView` setup is here and then again on lines 58-60: should it be in one block? May be would be cool to create the separate method for it.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Team Invites/UIEdgeInsets+Adjusting.swift
```diff
@@ -0,0 +1,34 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+extension UIEdgeInsets {
```
mikeger: Should this file/class go to `Source/User Interface/Helpers` ?

Issue #243, opened 2016-09-24
BUG: The app is each time asking for permission 


mat1th:
# Short description 
The app is asking for permission each time i will send a picture.

# Behavior: 
I'm using the app without it giving permission to my camera. And each time I will send a picture it alerts: 'Wire needs access to the Camera'.  Thats very frustrating. In the attachment there is a photo of the alert. (The alert is also not on top so I can't click on Open in Wire settings) 

# Expected behavior 
I will expect that the app will ask it once. And when I have said no to the permission it only asks when I hit the camera button. (see the orange annotation in the photo) 

![Camera button](https://cloud.githubusercontent.com/assets/6486389/18807548/6d4c3d72-824a-11e6-99e3-3cb3d65f11fd.jpg)
![The wrong alert ](https://cloud.githubusercontent.com/assets/6486389/18807549/6d78ab0a-824a-11e6-937c-e30dd6326979.jpg)

Thanks for your nice app!

KoalaBoomka:
Hi, @mat1th!
Thanks for the report. What model of the phone and iOS version do you use?

mat1th:
I'm using a iPhone 5s with iOS 10 buy I had the same issue with iOS 9. I've the latest app version. So I use version 2.16

KoalaBoomka:
@mat1th, thanks for the additional information. We really appreciate your bug report and will prioritize the issue.

mat1th:
Hello,
Is there already any information about the progress of the bug? 
Thanks!

KoalaBoomka:
Hello, @mat1th! 
Thanks for following up the issue. 
We managed to reproduce it, scheduled the ticket and will post here as we have any update on the progress.

mat1th:
@KoalaBoomka In the app store version 2.29 the bug still exists. When is the ticked scheduled? its now more than 4 months ago.. Thanks.

KoalaBoomka:
@mat1th, sorry, we haven't made any progress on this issue.

mat1th:
Okay, Thanks for the update. Its still annoying. I'll hope you will fix the bug soon.

PR #1610, opened 2018-01-16
ZIOS-7285: New camera keyboard permissions handler

nicorsm:
### Issues

There's a problem in the current implementation of `CameraKeyboardViewController`. In case the user doesn't allow to access their camera or photo library, the relative parts of the view remain empty, with a black camera view on the left side, and a grey collection view on the right side. The "open Camera Roll" button is also visible, even if we don't grant access to the library*.
Furthermore, you're continually prompted to give Wire permissions only for the camera when opening the camera keyboard. This is boring if you decide to definitely don't give permissions to the app.

_* which is ok from a technical side, because iOS 11 users still have the possibility to open the camera roll system dialog, but it's not clear to the user from a security point of view - "Why you still can access to my photos, even if I haven't authorized you?"._

### Solutions

I've implemented a new overlay view in case you don't have rights to access the camera and/or photo library, which replaces the respective section of the camera keyboard view. The "Settings" button inside this view automatically opens the Settings screen of the app.

testinitialstatelayoutsizecompact_cameraaccessgranted_ 320 216 2x


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -230,6 +230,12 @@
 "camera_access.denied.instruction" = "";
 "camera_access.denied.open_settings" = "Enable it in Wire Settings";
 
+// Camera and photos access, keyboard
+"keyboard_photos_access.denied.keyboard.settings" = "Settings";
+"keyboard_photos_access.denied.keyboard.camera_and_photos" = "Wire needs access to your\ncamera and photos.";
```
johnxnguyen: I think the newline char could confuse some translators, they might not know where to put it or they could possibly leave it out altogether. I would suggest either to put a comment here for them, or consider to leave the newline out of the string and adjust the `UILabel` so it splits over two lines. What do you think?

mat1th:
This fixes this issue https://github.com/wireapp/wire-ios/issues/243 :tada:

Wire-iOS/Resources/Base.lproj/Localizable.strings
nicorsm: There are other strings with this issue, like `"Start a conversation or\ncreate a group."`. I think that the reason why we put these line breaks is that we need to split the string into two lines with an equal amount of text for each line. For example, without a line break, the text could be "Wire needs access to your camera and (\n) photos." on bigger screens. Furthermore, Crowdin automatically checks for missing symbols like \n or %@. I'll add a comment anyway.

johnxnguyen: Cool, I wasn't sure but good to know!

nicorsm:
Tests are working fine locally. I'll merge the PR.

PR #1613, opened 2018-01-17
ZIOS-9487: Authentication flag is now stored inside ShareExtensionViewController

nicorsm:
### Issues

There's a problem with authentication inside Share Extension. If you have Local Authentication enabled and you switch between multiple accounts in SE, the authentication is requested another time.

### Causes

This is caused by the reset of the `SharingSession` object, which actually stores the `isLocalAuthenticationGranted` flag.

### Solutions

I've stripped the `isLocalAuthenticationGranted` from `SharingSession` and now I'm handling the flag inside `ShareExtensionViewController` (`localAuthenticationStatus` variable), so you're asked to authenticate only once. I've created a new enum, `LocalAuthenticationStatus`, just to make things more clear.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/51


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -356,27 +362,31 @@ class ShareExtensionViewController: SLComposeServiceViewController {
         pushConfigurationViewController(accountSelectionViewController)
     }
 
-    /// @param callback confirmation; if the auth is not needed or is not possible on the current device called with '.none'
-    func requireLocalAuthenticationIfNeeded(with callback: @escaping (Bool?)->()) {
+    /// @param callback confirmation; called when authentication evaluation is completed.
+    func requireLocalAuthenticationIfNeeded(with callback: @escaping ()->()) {
```
mikeger: Would it not still be helpful to return the status to the callback block?

nicorsm: The only two places where I've used it are in `presentChooseAccount()` and `presentChooseConversation()`. But yes, I could use the callback result there so I don't need to access to `self.localAuthenticationStatus`.

PR #1621, opened 2018-01-18
Move InviteTeamMemberCell into a separate section without a header

typfel:
### Issues

The invite team member cell was hiding the guests if a team owner was only connected to guests.

### Causes

The logic for showing hiding the invite team member cell was only considering team members

### Solutions

Create a new section for the invite team member cell which is only visible when the team owner doesn't have any team members.

## Notes

The invite team member cell is now also only showing upp in the "list mode" in the search UI.


vytis:
Test failure should be fixed by #1623, we probably need to merge that one into release as well.

PR #1622, opened 2018-01-18
Fix dotColor & fonts of tokenfield not being present

daehn:
### Issue

When creating a group conversation from a 1-on-1 and being asked to select more users to add the doe in the search field was cut off at the bottom.

### Cause

The size image view showing the dot is depending on the font size, which is not yet set when we are adding the user participating in the 1-on-1 conversation.

### Solution

Don't use Classy for this view but set the fonts manually. Some classes had to be moved from `WireExtensionComponents` into the main target in order to use the dynamic font functions, but said classes were not used in any extensions.


WireExtensionComponents/Controllers/ShareViewController+Views.swift
```diff
@@ -40,6 +40,9 @@ extension ShareViewController {
             self.shareablePreviewWrapper = shareablePreviewWrapper
         }
 
+        let scheme = FontScheme(contentSizeCategory: .medium)
+        self.tokenField.font = scheme.font(for: .init(.normal, .regular))
```
daehn: This will currently break dynamic type in the main app, I suggest moving all of the token textfield related views  from WireExtensionComponents into the main app as they aren't used in any extension anyways.

typfel:
It will still break if the dynamic font is changed but we'll try to make that work?

daehn:
The font size will change, but the token alignment might still break at the moment, yes.

PR #1627, opened 2018-01-22
Remove force unwrapping of selfUser

daehn:
## What's new in this PR?

As we allow sending logs without being logged in now we can no longer assume a self user will be present.


vytis:
Also on the logs front, maybe it's possible to disable shake gesture when user is logged in? It is quite easy to do it while normally using the app.

daehn:
@vytis yeah we could do that if that's preferred!

PR #1631, opened 2018-01-22
Use correct TOS link for team accounts

daehn:
## What's new in this PR?

* Use correct TOS link for team accounts.


daehn:
Closed in favour of https://github.com/wireapp/wire-ios/pull/1632

PR #1634, opened 2018-01-22
remove guest label if the user is a bot

billypchan:
## What's new in this PR?

Remove Guest label if the user is a serviceUser

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/422


Wire-iOS/Sources/UserInterface/Participants/ParticipantsListCell.m
```diff
@@ -77,6 +77,10 @@ - (void)updateForUser:(ZMUser *)user inConversation:(ZMConversation *)conversati
     self.userImageView.user = user;
     self.guestLabel.hidden = ![user isGuestInConversation:conversation];
 
+    if ([user isServiceUser]) {
+        self.guestLabel.hidden = true;
+    }
```
daehn: I think some Swift leaked into this Objective-C file, though I assume it'll still work using`YES` might be better.

PR #1637, opened 2018-01-23
Show informational alert if password is needed after phone verification

daehn:
## What's new in this PR?

Users with an associated email address have to enter their password after trying to login via phone number. To make this flow less confusing an alert controller will now be presented in that case.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -932,6 +932,9 @@
 "registration.signin.email_button.title" = "Email";
 "registration.signin.phone_button.title" = "Phone";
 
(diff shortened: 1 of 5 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -932,6 +932,9 @@
 "registration.signin.email_button.title" = "Email";
 "registration.signin.phone_button.title" = "Phone";
 
+"registration.signin.alert.password_needed.title" = "Password needed";
+"registration.signin.alert.password_needed.message" = "Please enter your Password in order to log in.";
```
vytis: I guess needs to be looked at by @infotexture

daehn:
Closed in favour of https://github.com/wireapp/wire-ios/pull/1638

PR #1642, opened 2018-01-23
Remove deprecated calls to .characters

beurdouche:
## What's new in this PR?
Restore build on current IOS/XCode

### Issues

The application wasn't building anymore in recent versions of XCode

### Causes

`.characters` seems to be deprecated

### Solutions

Simple removal of all `.characters.`

## Dependencies

None

Needs releases with:

- [ ] GitHub link to the pull request


vytis:
Thanks for useful contribution! We are still using Xcode 9.0.1 because migrating to 9.2 is almost full day of work and we were busy on other things. 
Maybe you would like to look into other internal frameworks we use and see how it works with newest Xcode? Or even trying to port them to Swift4? I am @vytis_wire on Wire, let me know if you have any questions!

PR #1647, opened 2018-01-24
Update zoom after setting an image instead of in viewDidLayoutSubviews

daehn:
## What's new in this PR?

The initial "double tap to zoom" gesture was broken on some devices (e.g. iPhone 8), only adjusting the zoom rect after setting an image instead of in each `viewDidLayoutSubviews` fixes this issue.


typfel:
test

PR #1649, opened 2018-01-25
iPhone X margins

nicorsm:
## What's new in this PR?

Fixed some misalignments on iPhone X:

- bottom margin on Search Results View (shared by Start UI and Add Participants View Controllers)
- bottom margin on Share View Controller


nicorsm:
@mikeger we're already using `KeyboardAvoidingViewController` with `ShareViewController`, but it's used to manage the bottom offset of the keyboard between Share View Controller and the Keyboard Avoiding controller. 
Here, I'm using the observer to adjust the bottom margin of the safe area, which should be = 0 when the keyboard is open. Since I need to handle this case, is there a way to avoid adding a second observer to do the same thing on a child view controller?

nicorsm:
No problem 🙂

mikeger:
@nicorsm can be merged now?

nicorsm:
@mikeger still need to fix one little thing with Start UI.

PR #1653, opened 2018-01-25
Services: Fixed background color of the top bar in the startUI

mikeger:
screen shot 2018-01-25 at 17 32 12


mikeger:
The tests are passing locally.

PR #1654, opened 2018-01-25
Feature/zios 9442 i want to distinquish integrations from team members

billypchan:
## What's new in this PR?

Added collection view section for services in ParticipantsViewController. The service users will be shown in second section of the collection view. A section header view will be displayed if service section exists.

# TODO
- [x] adjust ParticipantsListCell top margin
- [x] test on iPad


Wire-iOS/Sources/UserInterface/Participants/ParticipantsCollectionHeaderView.swift
```diff
@@ -0,0 +1,58 @@
...
+    public var title: String = "" {
+        didSet {
+            titleLabel.text = title.localized.uppercased()
+            titleLabel.sizeToFit()
+        }
+    }
+
+    let titleLabel: UILabel = {
+        let label = UILabel()
+        label.textColor = UIColor.wr_color(from: "rgb(141, 152, 159)")
```
mikeger: My opinion is that using this function from Classy is not consistent with our existing codebase. Please 
Please make sure the color mentioned does not exist already in ColorScheme. 
If it does, then please use `ColorScheme.default.color(withName:)`
If it does not exist then please create the color from integers or the HEX value.

Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
```diff
@@ -0,0 +1,122 @@
...
+        return headerView
+    }
+}
+
+extension ParticipantsViewController {
+    // MARK: - collectionview layout configuration
+    func configCollectionViewLayout() {
+
+        switch DeviceScreenSize.screenSizeOfThisDevice {
+        case .iPhone4_7Inch, .iPhone5_5Inch, .iPhone5_8Inch, .iPhoneBiggerThan5_8Inch:
```
mikeger: Suggestion: May be it would be better to check the horizontal size class?

billypchan: This logic is replace of magic config. The original logic was having a bigger cell for screen size bigger than iPhone 4 inch. All iPhones have same horizontal size class in portrait mode.

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -3,7 +3,7 @@
    LastUpgradeVersion = "0800"
    version = "1.3">
    <BuildAction
-      parallelizeBuildables = "YES"
+      parallelizeBuildables = "NO"
```
vytis: Make sure to revert this before merging!

Wire-iOS/Sources/Helpers/syncengine/Conversation+SortedParticipants.swift
```diff
@@ -16,15 +16,26 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
+public enum UserType: Int {
+    case user = 0
```
vytis: I think there is no need for this enum to be `Int` because simple ones are `Hashable` and can be used as dictionary keys.

Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
billypchan: @mikeger I will chnage the code to determine the cell size by the ParticipantsViewController's view's width, since iPad UI has different width in split mode.

Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
```diff
@@ -0,0 +1,124 @@
...
+
+extension ParticipantsViewController: UICollectionViewDataSource {
+
+    public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        guard let userType = UserType(rawValue:section),
+            let array = groupedParticipants[userType] as? [ZMUser]
+            else { return 0 }
+
+        return array.count
+    }
```
billypchan: Use enum UserType's vlaue for collectionView sections' data source access.

Wire-iOS/Sources/Components/DeviceNativeBoundsSize.swift
```diff
@@ -0,0 +1,97 @@
...
+}
+
+extension CGSize: ExpressibleByStringLiteral {
+    public init(stringLiteral value: String) {
+        let size = CGSizeFromString(value)
+        self.init(width: size.width, height: size.height)
+    }
+
+    public init(extendedGraphemeClusterLiteral value: String) {
+        let size = CGSizeFromString(value)
```
daehn: It might be possible to just call `init(stringLiteral` from `init(extendedGraphemeClusterLiteral` and `init(unicodeScalarLiteral`.

PR #1659, opened 2018-01-26
Update avatar image views for service users

daehn:
## What's new in this PR?

Update `UserImageView`, `ConversationAvatarView` and `AvatarImageView` to support service users. A new case `AvatarImageViewShapeRoundedRelative` has been added to the `AvatarImageViewShape` to support the necessary relative corner radius.


WireExtensionComponents/Views/AvatarImageView.m
```diff
@@ -83,6 +89,9 @@ - (void)updateCornerRadius
         case AvatarImageViewShapeRounded:
```
vytis: Maybe this is then not needed anymore, we just need to check if it produces the right corners in conversation view

daehn: You're right, should't be necessary any more.

daehn: @vytis I removed the hardcoded value case and adjusted the divider so it produces the correct values.

PR #1663, opened 2018-01-30
ZIOS-9538: Single bot conversations shows up in people search not as that 1:1 conversation but as a bot service

nicorsm:
### Issues

After creating a conversation with a service, that conversation was shown in the search of Start UI with the same layout and the same behavior of a service (the same of the "Services" tab).

### Causes

This was caused by:
- behavior: `searchResultsViewController(_ searchResultsViewController:, didTapOnConversation:)` was not checking the case with service users.
- UI: while deciding which UI elements to show, the conversation was treated as a service user - and not as a search result

### Solutions

Now I'm checking in both places if the conversation has a service user.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+SearchResults.swift
+++ b/Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+SearchResults.swift
@@ -50,7 +50,8 @@ extension StartUIViewController: SearchResultsViewControllerDelegate {
     }
     
     public func searchResultsViewController(_ searchResultsViewController: SearchResultsViewController, didTapOnConversation conversation: ZMConversation) {
-        if conversation.conversationType == .group {
+        if conversation.conversationType == .group
+            || (conversation.conversationType == .oneOnOne && conversation.includesServiceUser) {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+SearchResults.swift
```diff
@@ -50,7 +50,8 @@ extension StartUIViewController: SearchResultsViewControllerDelegate {
     }
     
     public func searchResultsViewController(_ searchResultsViewController: SearchResultsViewController, didTapOnConversation conversation: ZMConversation) {
-        if conversation.conversationType == .group {
+        if conversation.conversationType == .group
+            || (conversation.conversationType == .oneOnOne && conversation.includesServiceUser) {
```
daehn: This change breaks opening a conversation from the top people section [ZIOS-9629](https://wearezeta.atlassian.net/browse/ZIOS-9629).

daehn: @nicorsm I am not 100% sure what this change was for, as the appearance of the service user conversation shouldn't be affected by this.

Issue #1582, opened 2018-01-06
Spacing issue on iPhone X


malicioustoker:
This issue occurs when you load a place within Wire that launches the in app browser. As you can see at the top, the text gets all squished up with the top sensor bar, the time, and the battery. ![GitHawk Upload by malicioustoker](https://i.imgur.com/Hger6x4.jpg)

You can’t see it in the photo, but on the iPhone X itself, where it says “Wire-Support” part of that gets cut off by the “notch”

mikeger:
Thanks for the report, definitely something to fix on our side, @nicorsm something we can easily replace with `SFSafariViewController` ?

nicorsm:
@mikeger I'll take a look.

nicorsm:
Fix has been merged into `develop`.

PR #1664, opened 2018-01-30
ZIOS-9546: Fixed margins for Browser View Controller on iPhone X

nicorsm:
## What's new in this PR?

- Fixed margins for Browser View Controller on iPhone X
- Removed usage of `useWithStatusBar`, since now we're calculating the space occupied by the status bar in `UIScreen.safeArea` (even on iOS versions before 11.0).

## Note

Also referenced in https://github.com/wireapp/wire-ios/issues/1582

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserBarView.m
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserBarView.m
@@ -27,7 +27,6 @@
 
 @property (nonatomic) CAShapeLayer *progressLayer;
 @property (nonatomic) BOOL initialConstraintsCreated;
-@property (nonatomic) BOOL useWithStatusBar;
 
 @end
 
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserBarView.m
```diff
@@ -74,7 +66,7 @@ - (void)updateConstraints
     if(! self.initialConstraintsCreated) {
         self.initialConstraintsCreated = YES;
 
-        CGFloat offset = self.useWithStatusBar ? 10 : 0;
+        CGFloat offset = UIScreen.safeArea.top / 2.0;
```
mikeger: Not sure that I am following it correctly, what is the reason for taking the half of the inset?

nicorsm: Because the size of the top bar includes the size of the status bar, and the items inside the top bar are centered within the entire view: https://github.com/wireapp/wire-ios/pull/1664/files/b5ff720484345f858ba696ab687065f1aba71961#diff-3071398d212a91f049170c9afe445abdR92

PR #1665, opened 2018-01-30
ZIOS-9545: Crash while adding a bot and selecting a conversation for the second time

nicorsm:
### Issues

Crash occurring after choosing a service in `ShareViewController` for the second time.

App was crashing at `ShareViewController.swift:173` due to the forced cast:
https://github.com/wireapp/wire-ios/blob/e8a810c3af6a04a1327b3e505bf69b0310117835/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift#L173

### Causes

This was caused by the `representedObject` property of `Token`, which was set to `weak`. 
https://github.com/wireapp/wire-ios/blob/db824447f9f19a3363f8813b55e593c19c7a22ea/Wire-iOS/Sources/Components/TokenField/Token.h#L28

For this reason, `representedObject` was deallocated just after https://github.com/wireapp/wire-ios/blob/db824447f9f19a3363f8813b55e593c19c7a22ea/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift#L146
in this way, not letting us make any more a comparison between selected and unselected items in the list.

### Solutions

I've marked the `representedObject` property as `strong`.

### DIFF
--- a/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift
+++ b/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift
@@ -145,6 +145,10 @@ public class ShareViewController<D: ShareDestination, S: Shareable>: UIViewContr
         
         self.tokenField.addToken(forTitle: destination.displayName, representedObject: destination)
         
+        if self.tokenField.tokens.first?.representedObject == nil {
+            print("STOPP")
+        }
+        
         self.selectedDestinations.insert(destination)
         
         if !self.allowsMultipleSelection {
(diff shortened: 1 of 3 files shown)


vytis:
There might be a good reason why this was `weak`, maybe it introduces retain cycles otherwise? Did you check if things that are stored in `representedObject` actually gets released after closing the controller?

nicorsm:
@vytis I need to take a look. Actually, `representedObject` (which is an `id`) can be `ZMSearchUser`, `ZMUser`, `MessageDraft`, `ZMConversation` and `ServiceConversation` (which can contain a `ZMConversation` object). The last one is an enum and it's the only case that gets deallocated after creating the token object.

nicorsm:
I guess it shouldn't introduce retain cycles since there's no type of object stored into `representedObject` which has a reference to `Token`... right?

vytis:
You are right, probably it shouldn't be, as the token is only stored inside the `tokenField`

nicorsm:
Btw tests are working fine locally.

PR #1674, opened 2018-02-01
Unified the nav bars in Start UI, AddParticipants

mikeger:
### Issues

The service detail controller is designed to be shown as the part of the navigation sequence. Therefore during the original implementation of services I had to put in the ugly solution of
1. Wrapping Start UI/Add Participants VC into the navigation controller.
2. Showing / hiding the nav bar depending on the screen being looked on.

It was semantically inconsistent and simply not future-proof.

### Solutions

It was noticed that the start UI view controller is showing it's title in the way that is very similar to the navigation view controller's navigation bar. Therefore the proposed solution is to replace the title line of Start UI / Add people with the navigation bar. The benefits of this solution:

1. Style of the bar is consistent between the screens.
2. No need to hide / show the bar.

## Additional changes

The `ConversationListViewController` has a not that nice state variable that is showing which controller is shown now. We used the delegate call to dismiss the start UI and reset the state. It also seems not consistent with the regular iOS modal / navigation presentation logic. The state variable is up to be removed in the future.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
@@ -20,6 +20,19 @@ import Foundation
 import Cartography
 import Classy
 
+
+class AddParticipantsNavigationController: UINavigationController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.navigationBar.tintColor = .black
+        self.navigationBar.setBackgroundImage(UIImage(), for:.default)
+        self.navigationBar.shadowImage = UIImage()
+        self.navigationBar.isTranslucent = true
+        self.navigationBar.titleTextAttributes = [NSForegroundColorAttributeName: UIColor.black,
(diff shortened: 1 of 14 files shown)


Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
```diff
@@ -20,6 +20,19 @@ import Foundation
...
 
+
+class AddParticipantsNavigationController: UINavigationController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.navigationBar.tintColor = .black
+        self.navigationBar.setBackgroundImage(UIImage(), for:.default)
+        self.navigationBar.shadowImage = UIImage()
+        self.navigationBar.isTranslucent = true
+        self.navigationBar.titleTextAttributes = [NSForegroundColorAttributeName: UIColor.black,
```
nicorsm: Could this override the light/dark theme settings? As far as I remember, the title color with the dark theme is white.

mikeger: True, forgot to test on the dark theme!

PR #1677, opened 2018-02-02
Add snapshot tests for share preview views

daehn:
## What's new in this PR?

Add missing snapshot tests to ensure we notice in case the share previews break again (see also #1675 and #1676).


nicorsm:
Snapshot tests for Share View Controller are coming too! 🚀

Wire-iOS Tests/MessagePreviewViewTests.swift
```diff
@@ -0,0 +1,73 @@
...
+            resolvedURLString: "https://www.example.com/article/1",
+            offset: 0
+        )
+
+        article.title = "You won't believe what happened next!"
+        let textMessageData = MockTextMessageData()
+        textMessageData.linkPreview = article
+        textMessageData.imageDataIdentifier = "image-id-unsplash_matterhorn.jpg"
+        textMessageData.imageData = UIImageJPEGRepresentation(image(inTestBundleNamed: "unsplash_matterhorn.jpg"), 0.9)
+        textMessageData.hasImageData = true
```
vytis: The snapshot for this seem to be missing the image, is that expected?

daehn: Yes (kind of), the other tests only testing article views are waiting a couple of fractions of a second until it becomes available, but this didn't work out in that specific case.

daehn: @vytis fixed with 14215bc

vytis: Hmm maybe there is some sort of loop we can do to wait until it appears? Seems a bit weird to have test that will fail when it will actually render it correctly:)

daehn: It's not like that anymore though, I rerecorded the correct reference image containing an image.

PR #1678, opened 2018-02-02
Added tests for Share View Controller

nicorsm:
## What's new in this PR?

Added some tests for Share View Controller. Work in progress.


Wire-iOS Tests/ShareViewControllerTests.swift
```diff
@@ -0,0 +1,118 @@
...
+    }
+    
+    func testThatItRendersCorrectlyShareViewController_LocationMessage() {
+        let location = LocationData.locationData(withLatitude: 43.94, longitude: 12.46, name: "Stranger Place", zoomLevel: 5)
+        groupConversation.appendMessage(with: location)
+        makeTestForShareViewController()
+    }
+    
+    func testThatItRendersCorrectlyShareViewController_Photos() {
+        /// TODO fix image loading
```
vytis: Not sure if it's the same problem, but @daehn fixed it somehow on https://github.com/wireapp/wire-ios/pull/1677

daehn: I only fixed it for link previews, and even there it was more related to test execution timing, as when running a single test the image would always fail to load. For images you can have a look at how we do it in the tests for image message cells, but I think this approach is not suitable here as we just set the image directly on the cell there (which we can't do here as we don't have access to the cell).

vytis: Maybe it would be possible to introduce a similar approach as we do in sync engine - have a dispatch group everywhere so we could actually wait for these things to finish?

daehn: Yeah, but we'd have to change the image loading logic to use a dispatch group and make it injectable (which would be nice).

vytis: Definitely out of scope for this PR, just something to think about if we keep having problems like this.

nicorsm: Yes, the main point here is to test the general behavior of Share View Controller: the test that the table view is visible, that the bottom bar is visible or not in the two possible cases, that the top view is displayed correctly. I think that most of the test cases for cell layouts are covered with the tests made by @daehn.

nicorsm:
Tests are working fine locally.

vytis:
I re-run the tests, it seems they crashed somewhere in the middle.

mikeger:
Weird, failed again, I am going to restart the tests.

vytis:
Still crashing, maybe worth investigating more, doesn't seem like a random issue...

mikeger:
@nicorsm what's the status of this one?

nicorsm:
@mikeger I'll take a look later, I'm on the test failures for https://github.com/wireapp/wire-ios-share-engine/pull/52

nicorsm:
Update: I'm experiencing a crash on `V2Asset`. Something different than what we see from CircleCI (a generic failure on audio calls), but I'll investigate.

nicorsm:
`testThatItRendersCorrectlyShareViewController_Photos()` is causing a crash with the cache in `UserConnectionViewTests`.

nicorsm:
I think this time Carthage has crashed...

```
carthage(678,0x7fff9fd003c0) malloc: *** error for object 0x10bc2dc68: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
./setup.sh: line 46:   678 Abort trap: 6           carthage bootstrap --platform ios
Exited with code 134
```

PR #1682, opened 2018-02-05
Fix layout of the participants when only one bot is added

mikeger:
### Issues

When there are no regular users in the group conversation, the collection view used to show the second section slightly scrolled down from the top of the collection view.

### Causes

The collection view that is showing the conversation participants is designed to show 2 sections: users and bots. When the first section is empty, the collection view still thinks it's in the view, even without the cells. The collection view layout is configured in the way that both sections are having the inset, so the second section is off from the top on the height of the first section insets.

### Solutions

There is a way to supply the inset per section dynamically via the collection view flow layout delegate.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
+++ b/Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
@@ -20,12 +20,15 @@ import Foundation
 
 extension ParticipantsViewController: UICollectionViewDataSource {
 
+    func participants(of type: UserType) -> [ZMUser] {
+        return groupedParticipants[type] as? [ZMUser] ?? []
+    }
+    
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        guard let userType = UserType(rawValue:section),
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Participants/ParticipantsViewController+CollectionView.swift
```diff
@@ -94,6 +102,34 @@ extension ParticipantsViewController: UICollectionViewDelegate {
     }
 }
 
+extension ParticipantsViewController: UICollectionViewDelegateFlowLayout {
+    public func collectionView(_ collectionView: UICollectionView,
+                               layout collectionViewLayout: UICollectionViewLayout,
+                               insetForSectionAt section: Int) -> UIEdgeInsets {
+        let section = UserType(rawValue: section)!
+        
+        let hasFirstSection = ((groupedParticipants[UserType.user] as? Array<ZMUser>)?.count ?? 0) > 0
```
vytis: there is a handy `.isEmpty()` method for this!

mikeger: Only simplifying it a little, since `groupedParticipants` is missing the type information. Will try to make it more readable.

PR #1691, opened 2018-02-06
Fix/profile header view tests

billypchan:
### Issues

ProfileHeaderViewTests failed to compile after ProfileHeaderView's init medthod updated.

### Solutions

Rewrite ProfileHeaderViewTests and create a flag in ProfileHeaderView to prevent traitCollectionDidChange (called by verifyInAllPhoneWidths) changes headerStyle during the snapshot tests.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileHeaderView.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileHeaderView.swift
@@ -77,14 +77,14 @@ final class ProfileHeaderView: UIView {
 
         let detailViewMargin = horizontalMargin + 32
 
-        constrain(self, detailView) { (view: LayoutProxy, detailView: LayoutProxy) -> Void in
+        constrain(self, detailView) { (view: LayoutProxy, detailView: LayoutProxy) -> () in
             detailView.top == view.top + topMargin
             detailView.leading == view.leading + detailViewMargin
             detailView.trailing == view.trailing - detailViewMargin
             detailView.bottom == view.bottom - 12
         }
 
-        constrain(self, dismissButton, verifiedImageView, detailView.titleLabel) { (view: LayoutProxy, dismiss: LayoutProxy, verified: LayoutProxy, title: LayoutProxy) -> Void in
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileHeaderView.swift
```diff
@@ -76,15 +76,15 @@ final class ProfileHeaderView: UIView {
         let horizontalMargin = WAZUIMagic.cgFloat(forIdentifier: "profile_temp.content_left_margin")
 
         let detailViewMargin = horizontalMargin + 32
-        
-        constrain(self, detailView) { (view: LayoutProxy, detailView: LayoutProxy) -> () in
+
+        constrain(self, detailView) { (view: LayoutProxy, detailView: LayoutProxy) -> Void in
```
mikeger: I think that doing the style changes on other's code is not too useful.

mikeger:
@billypchan can you explain a little bit more why the tests where broken with the trait change update?

billypchan:
@mikeger the verifyInAllPhoneWidths method form ZMSnapshotTestCase calls traitCollectionDidChange when it test for each Phone size. The flag disableHeaderStyleUpdate is added to ProfileHeaderView in this PR to prevent the initial headStyle value set for each snapshot test being updated.

typfel:
@billypchan so why is it breaking when the `traitCollectionDidChange` is called in the tests, because you wan't the the no button scenario which would then always get re-configured?

Maybe we should change the test to run on all device sizes, then we would actually tests the trait collection code as well.

billypchan:
@mikeger @typfel I added a guard statement in traitCollectionDidChange to prevent the snapshot test triggers traitCollectionDidChange. The flag disableHeaderStyleUpdate is removed.

mikeger:
I don't like the current solution, since introducing the flag for the test makes the tests slightly less valuable. Also this makes the test implementation entangled with the code. However let's merge this for the moment and create the ticket to improve it later.

mikeger:
My last comment is related to the original implementation.

billypchan:
I tried verifyInAllDeviceSizes method in ZMSnapshotTestCase, but it only changes the testing view's size to match iPad's screen size. The better way to cover all possible cases of ProfileHeaderView would be run this test suite in both iPhone and iPad simulator.

PR #1692, opened 2018-02-06
Remove separator line from collections view

typfel:
## What's new in this PR?
Remove the separator line above the search field in the collections view controller.

## Notes
The design has changed so that the navigation bar and the content view has different colours making the separator line unnecessary.


Wire-iOS/Sources/UserInterface/Components/Controllers/DefaultNavigationBar.swift
```diff
@@ -41,10 +41,9 @@ class DefaultNavigationBar : UINavigationBar {
                                NSForegroundColorAttributeName: ColorScheme.default().color(withName: ColorSchemeColorTextForeground)]
         setTitleVerticalPositionAdjustment(-2.0, for: .default)
         
-        let separatorPixel = UIImage.singlePixelImage(with: ColorScheme.default().color(withName: ColorSchemeColorSeparator))
+        let separatorPixel = UIImage.singlePixelImage(with: UIColor.clear)
```
mikeger: May be the whole thing with `retinaSeparatorPixel` is not necessary any more, i.e. can be removed all together?

PR #1696, opened 2018-02-07
Profile: Fix missing close button

mikeger:
### Issues

The close button is missing from user profile when entering it on iPhone.

### Causes

[Some refactoring](https://github.com/wireapp/wire-ios/pull/1684) was done on the header of participants view in order to adopt it better for the traits change. During the refactoring we broke the tests, so [another fixes](https://github.com/wireapp/wire-ios/pull/1691) have to be done to address this. During those fixes, we missed to check the functionality on the iPhone: the call to `updateHeaderStyle` is necessary in order to make sure the initial close button style is configured.

### Solutions

Missing call is added, so the view is initially correctly configured.


billypchan:
Thanks @mikeger! I should write a new set for test for the new icon update feature for both iPad test targets.

PR #1697, opened 2018-02-07
Fix self profile retain cycle

typfel:
### Issues

The app was crashing in the `AvailabilityTitleView` due to  an exception: `An NSManagedObject must have a valid NSEntityDescription`

### Causes

There was a retain cycle in the `ProfileView` which contains a `AvailabilityTitleView`. The steps leading up to a crash:

1. Launch app into  account 1
2. Switch to account 2 (this will leak a `AvailabilityTitleView`)
3. App gets a memory warning which will shutdown the user session for account 1
4. Open app and the `AvailabilityTitleView` will try access no longer existing user object and crash.

### Solutions

Fix retain cycle.


mikeger:
👍 real fix!

PR #1701, opened 2018-02-07
ZIOS-9393: Added loggedOut error handling in Share Extension

nicorsm:
### Issues

Share Extension was crashing because the `reachability` property of `ZMTransportSession` was not torn down correctly. 

### Solutions

I'm currently calling `self.reachability.tearDown()` inside `ZMTransportSession` (see dedicated PR in `wire-ios-transport`). I've also added the error handling of `SharingSession.InitializationError.loggedOut` with an alert view that informs the user about the error. This error was already thrown but not caught by `ShareExtensionViewController`. 

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/52
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/72


vytis:
We should not tear down reachability inside transport session because in multiple accounts we have a transport session for each account, but they all share a single reachability object. So if we tear down reachability when we tear down one of the transport sessions it will stop working for the other one(s).
Basically who created the reachability should be responsible for tearing it down. In the share extension case it is created here: https://github.com/wireapp/wire-ios-share-engine/blob/develop/Sources/SharingSession.swift#L267

nicorsm:
Totally missed that instance of Reachability! Thanks @vytis!

mikeger:
@nicorsm is this one ready to be reviewed?

nicorsm:
@mikeger yes!

PR #1704, opened 2018-02-08
Cartography 3

mikeger:
Cartography 3 offers quite some nice updates, most and foremost is the possibility of creating constraints not only between views but also with layout guides. This allows constrainting to safe areas of view controllers.


Cartfile
```diff
@@ -1,7 +1,7 @@
 github "wireapp/wire-ios-ziphy" ~> 7.0
 github "wireapp/Classy" "v0.2.6-swift3.0.2"
 github "wireapp/CocoaLumberjack" ~> 3.3
-github "wireapp/Cartography" ~> 1.1
+github "wireapp/Cartography" ~> 3.0
```
vytis: We have been sleeping quite long it seems!

PR #1707, opened 2018-02-09
Fixed multiple issues with conversation creation controller

mikeger:
- Deselect the cell when the controller is presented.
- Fix the unnecessary keyboard animation or show the keyboard initially.
- Fix navigation buttons selection color.
- Allow ClearBackgroundNavigation to work with non-transparent controllers.

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIColor+Hex.swift
+++ b/Wire-iOS/Sources/Helpers/UIColor+Hex.swift
@@ -35,11 +35,8 @@ extension UIColor {
     
     public var alpha: CGFloat {
         get {
-            var red: CGFloat = 0
-            var green: CGFloat = 0
-            var blue: CGFloat = 0
             var alpha: CGFloat = 0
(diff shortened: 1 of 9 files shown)


Wire-iOS/Sources/Helpers/UIColor+Hex.swift
```diff
@@ -32,4 +32,19 @@ extension UIColor {
...
         self.init(red: r, green: g, blue: b, alpha: alpha)
     }
+    
+    public var alpha: CGFloat {
+        get {
+            var red: CGFloat = 0
+            var green: CGFloat = 0
+            var blue: CGFloat = 0
+            var alpha: CGFloat = 0
+            guard self.getRed(&red, green: &green, blue: &blue, alpha: &alpha) else {
```
daehn: Should be possible to pass `nil` to everything except alpha here (as they are not used and optional).

PR #1708, opened 2018-02-09
New conversation system messages

vytis:
### Issues

Updated conversation creation flow needs to have improved system messages that would have conversation name.

### Solutions

Added another type of cell that displays the conversation name only and updated participants cell to have shorter text. They are displayed together so there were some changes needed for padding between system messages.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/432


vytis:
After discussing with @typfel I will re-do the functionality to avoid having two system messages. Closing this and related PR.

PR #1721, opened 2018-02-13
Ignore dark mode for linear conv creation flow

mikeger:
### Issues

The linear conversation creation flow can be opened from both start UI and the conversation details. The `AddParticipantsViewController` is designed to respect the user's dark theme mode, since it's shown from the conversation details. 

Therefore, if user has the dark mode selected then when opening the linear conv creation flow from start UI, the conv name screen is white and add participants screen in dark.

### Solutions

For the moment the linear conv creation flow would always be white, does not matter from where it is presented. This is achieved by passing in the color scheme variant.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
@@ -57,6 +57,7 @@ public class AddParticipantsViewController: UIViewController {
         case create(ConversationCreationValues)
     }
     
+    fileprivate let variant: ColorSchemeVariant
     fileprivate let searchResultsViewController : SearchResultsViewController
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/AddContacts/AddParticipantsViewController.swift
```diff
@@ -97,7 +98,9 @@ public class AddParticipantsViewController: UIViewController {
         }
     }
     
-    public init(context: Context) {
+    public init(context: Context, variant: ColorSchemeVariant? = nil) {
+        self.variant = variant ?? ColorScheme.default().variant
```
nicorsm: Just a little thing: here you could pass `ColorScheme.default().variant` as the default parameter of the constructor, instead of `nil`.

PR #1730, opened 2018-02-14
Temporarily revert the audio recording API

mikeger:
### Issues

Some issue connected to the iOS client and AVS makes the audio messages recording impossible. Until the proper solution is not found, we'll revert the audio messages recording API and use the old method.

Original PR: https://github.com/wireapp/wire-ios/pull/1693


mikeger:
AVS fixed the API, no need to revert.

PR #1736, opened 2018-02-16
Do not generate the preview on huge PDFs

mikeger:
### Issues

User can send the huge PDF, that would make the client crash.

### Causes

PDF preview generator is not taking in account the document size.

### Solutions

Check the PDF viewport size before rendering.


WireExtensionComponents/Utilities/FilePreviewGenerator.swift
```diff
@@ -262,7 +262,11 @@ func AspectFitRectInRect(_ fit: CGRect, into: CGRect) -> CGRect
         contextRef.setAllowsAntialiasing(true)
         let cropBox = pageRef.getBoxRect(CGPDFBox.cropBox)
         
-        guard cropBox.size.width != 0, cropBox.size.height != 0 else {
+        guard cropBox.size.width != 0,
+              cropBox.size.width < 16384,
```
vytis: Maybe that's a bit extensive and something lower would work also? Or it's still pretty fast on old devices?

mikeger: 10k width is working fine (blink of an eye), also trying to be future-proof here :)

PR #1737, opened 2018-02-16
Fix top people not opening conversation 

daehn:
## What's new in this PR?

* Fix top people not opening the associated 1-1 conversation anymore.


typfel:
I don't think you should be able to select them, we should open the conversation.

daehn:
@typfel yeah, wrong wording on my side, they were selectable and now they open :)

PR #1740, opened 2018-02-16
Only use default transitions for non-clear opaque backgrounds

daehn:
## What's new in this PR?

* Only use default transitions when the background is neither clear nor translucent.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
@@ -85,11 +85,19 @@ class ClearBackgroundNavigationController: UINavigationController {
         }
     }
     
-    func navigationController(_ navigationController: UINavigationController, didShow viewController: UIViewController, animated: Bool) {
-        let clearBackround = viewController.view.backgroundColor == .clear
-        let opaqueBackground = fabs(viewController.view.backgroundColor?.alpha ?? 1.0 - 1.0) < CGFloat.ulpOfOne
-        useDefaultPopGesture = !clearBackround && opaqueBackground
+    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
+        if let avoiding = viewController as? KeyboardAvoidingViewController {
+            updateGesture(for: avoiding.viewController)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
```diff
@@ -86,7 +86,9 @@ class ClearBackgroundNavigationController: UINavigationController {
     }
     
     func navigationController(_ navigationController: UINavigationController, didShow viewController: UIViewController, animated: Bool) {
-        useDefaultPopGesture = fabs(viewController.view.backgroundColor?.alpha ?? 1.0 - 1.0) < CGFloat.ulpOfOne
+        let clearBackround = viewController.view.backgroundColor == .clear
+        let opaqueBackground = fabs(viewController.view.backgroundColor?.alpha ?? 1.0 - 1.0) < CGFloat.ulpOfOne
+        useDefaultPopGesture = !clearBackround && opaqueBackground
```
mikeger: I am not fully following, if background is .clear, should it have also alpha == 0, so then color.alpha would be != 1?

daehn: If the alpha is clear, or the background has an alpha value lower than 1 we should use the default gesture. Before it was only checking whether or not the alpha was lower than 1, which might not be the case for clear background colours.

daehn: `opaqueBackground` should be named `translucentBackground` here though

mikeger: So there can be a clear color that would not match "alpha < 1" ?

daehn: You're right, I think I misread this, I thought we were checking the view's alpha, but we are already checking the color's alpha! Good point, will have a look again.

daehn: @mikeger fixed now!

PR #1743, opened 2018-02-16
Fix to Address Book correlation label color, adjusted margins

nicorsm:
### Issues

1) `AddressBookCorrelationFormatter` in `SearchResultCell` was receiving a fixed white color, so the label was almost not visible on white backgrounds
2) Consequently, that label was overlapped by the right checkmark view

### Solutions

1) I'm passing the color from the current scheme
2) I'm now calculating the right margin of the label within all the possible cases, also with both checkmark view and guest label visible.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/Common/Cells/SearchResultCell.m
+++ b/Wire-iOS/Sources/UserInterface/StartUI/Common/Cells/SearchResultCell.m
@@ -31,6 +31,7 @@
 @property (nonatomic, strong) UIView *gesturesView;
 @property (nonatomic, strong) BadgeUserImageView *badgeUserImageView;
 @property (nonatomic, strong) ConversationAvatarView *conversationImageView;
+@property (nonatomic, strong) NSMutableDictionary *correlationFormatters;
 @property (nonatomic, strong) UILabel *nameLabel;
 @property (nonatomic, strong) UIView *avatarContainer;
 @property (nonatomic, strong) IconButton *instantConnectButton;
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/StartUI/Common/Cells/SearchResultCell.m
```diff
@@ -77,7 +77,9 @@ + (AddressBookCorrelationFormatter *)correlationFormatter
     static dispatch_once_t onceToken;
     static AddressBookCorrelationFormatter *formatter = nil;
     dispatch_once(&onceToken, ^{
-        formatter = [[AddressBookCorrelationFormatter alloc] initWithLightFont:self.lightFont boldFont:self.boldFont color:UIColor.whiteColor];
+        formatter = [[AddressBookCorrelationFormatter alloc] initWithLightFont:self.lightFont
+                                                                      boldFont:self.boldFont
+                                                                         color:[[ColorScheme defaultColorScheme] colorWithName:ColorSchemeColorTextDimmed]];
```
daehn: In some cases the cell we use a different color theme than the default though (e.g. when used in StartUI).

mikeger: @daehn is right, please use `colorSchemeVariant` to find which variant is used on current screen.

nicorsm: `correlationFormatter` is a static method, so I need to provide variant as a static method like
```
+ (ColorSchemeVariant)defaultVariant {
    return ColorSchemeVariantDark;
}
```
and then, replacing occurrencies in the class or assigning like `self.colorSchemeVariant = self.class.defaultVariant;`

mikeger: You cannot really use it as static: the color can be white from start UI and dark from conv details. This needs to be refactored to work :(

nicorsm: A more flexible approach could be introducing a property setter for `self.colorSchemeVariant` that updates at the same time the AB Correlation Formatter singleton, which its `color` property should be changed into a variable.

mikeger: This would not work if two cells of different style would be displayed on one screen (like on iPad). I think you can create the dictionary that contains the key-values as following: [ColorSchemeVariant: AddressBookCorrelationFormatter]. Then one can just pick the correct formatter with `correlationFormatters[self.colorSchemeVariant]`.

Wire-iOS/Sources/UserInterface/StartUI/Common/Cells/SearchResultCell.m
```diff
@@ -72,15 +73,21 @@ + (UIFont *)boldFont
...
 }
 
-+ (AddressBookCorrelationFormatter *)correlationFormatter
-{
-    static dispatch_once_t onceToken;
-    static AddressBookCorrelationFormatter *formatter = nil;
-    dispatch_once(&onceToken, ^{
-        formatter = [[AddressBookCorrelationFormatter alloc] initWithLightFont:self.lightFont boldFont:self.boldFont color:UIColor.whiteColor];
-    });
+-(void)setColorSchemeVariant:(ColorSchemeVariant)colorSchemeVariant{
```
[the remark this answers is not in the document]
nicorsm: But how you would handle the case with two formatters if we only have one variable static? I was thinking about making `correlationFormatters` static, so each instance of `SearchResultCell` could pick one of the two formatters available in the dictionary (light or dark).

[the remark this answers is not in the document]
mikeger: You can make the dictionary static. Then it would have several static formatters for each case inside. For example like that:
```swift
static private var formatters: [ColorSchemeVariant: AddressBookCorrelationFormatter] = [:]

public static func addressBookFormatter(for variant: ColorSchemeVariant) -> AddressBookCorrelationFormatter {
    if let existing = self.formatters[variant] {
        return existing
    }

    /// ... creation code
    /// then store to `formatters`
   return new
}
```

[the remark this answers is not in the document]
nicorsm: Perfect, that's the solution I was thinking about! 🙂

PR #1751, opened 2018-02-21
Fix/down integration

johnxnguyen:
## What's new in this PR?

This PR contains **Down** markdown framework integration, which replaces the old **Marklight** framework. 

### Issues

The old framework has some problems:
- Parsing conflicts in links and with our emoji substitution.
- Too rigid to change and extend
- Slow performance for large messages

### Causes

Some of these result from the extensive modifications needed to suit our requirements and needs, but the main issue stems from the use of regular expressions and the need to constantly scan the input text after each change in order to update the visual feedback in the input bar.

### Solutions

**Down** is a Swift wrapper for the **CMark** C library, which uses an Abstract Syntax Tree instead of regular expressions. This makes parsing **extremely fast**. Furthermore, it has been found to be much more reliable and more capable than the old framework.

The major change in switching frameworks is how the user styles text in the input bar. Previously the user would interact with the Markdown toolbar, which would insert the appropriate syntax directly in the text view. **Down** differs from **Marklight** in that it parses a `String` containing markdown syntax and returns an `NSAttributedString` with the syntax removed. In order to see the text styling as it is being written, another input mechanism is needed. Tapping on the markdown icons in the toolbar now toggle the text styles which are then applied as the user types. As a result, no markdown syntax is visible in the input bar. When the message is sent, the styled input text is then converted into a plain string containing the necessary markdown syntax, which is then parsed by the receiver of the message to display the styled text once more.

Needs releases with:

- [x] https://github.com/wireapp/Down/pull/1

## Notes

Still to do:

- [x] Update snapshot tests
- [x] Finalize styling with design
- [x] Last combthrough for refactoring & improvements
- [x] Security review


Wire-iOS/Sources/UserInterface/MarkdownBarView.swift
```diff
@@ -94,64 +92,60 @@ public final class MarkdownBarView: UIView {
...
-            case .italic: buttonsToHighlight.append(italicButton)
-            case .bold: buttonsToHighlight.append(boldButton)
-            case .numberList: buttonsToHighlight.append(numberListButton)
-            case .bulletList: buttonsToHighlight.append(bulletListButton)
-            case .code: buttonsToHighlight.append(codeButton)
-            default: break
-            }
+    // MARK: - Conversions
+    
+    private func buttonFor(_ markdown: Markdown) -> IconButton? {
```
typfel: Could be named `button(for markdown: Markdown)` to be consistent with the other functions in this class.

Wire-iOS/Sources/UserInterface/MarkdownBarView.swift
```diff
@@ -94,64 +92,60 @@ public final class MarkdownBarView: UIView {
...
+        case boldButton:        return .bold
+        case italicButton:      return .italic
+        case codeButton:        return .code
+        case numberListButton:  return .oList
+        case bulletListButton:  return .uList
+        default:                return nil
+        }
+    }
+    
+    public func updateIcons(forMarkdown markdown: Markdown) {
```
typfel: Could be named `updateIcons(for markdown : Markdown)` for consistency with other functions in this class.

Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
```diff
@@ -0,0 +1,570 @@
...
+        return self.parser.parse(attributedString: self.attributedText)
+    }
+    
+    /// Set when newline is entered, used for auto list item creation.
+    private var newlineFlag = false
+    
+    /// The current attributes to be applied when typing.
+    fileprivate var currentAttributes: [String : Any] = [:]
+
+    /// The currently active markdown. This determines which attribtues
```
typfel: typo: attribtues

PR #1760, opened 2018-02-23
ZIOS-9649: Fix for TailEditingTextField width on iOS 10

nicorsm:
### Issues

The width of the input field inside `TailEditingTextField` (used in `SettingsCell`) was not updated after typing something inside it, only on iOS 10. So, for example, if you write a short display name inside the Account settings, you go back and return on that screen, the width of the text field doesn't get updated until you press the spacebar.

### Solutions

I'm setting the `text` property of the text field every time. This is causing the text field to update its frame.


billypchan:
@nicorsm Some input method, e.g. Chinese Cangjie input method replace its characters component to a signal character after space is pressed. This fix caused the character components are inputted in the textfield.

PR #1771, opened 2018-02-28
Tweak colors for darkmode

typfel:
### Issues

Colors were off group details when in dark mode.

### Solutions

Fix colors


typfel:
@mikeger oh :) yes they are

PR #1783, opened 2018-02-28
Remove guests when toggle is switched off during group creation

daehn:
## What's new in this PR?

* Remove guests when toggle is switched off during group creation.


Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -166,16 +166,27 @@ final class ConversationCreationController: UIViewController {
...
         
         bottomViewContainer.isHidden = nil == ZMUser.selfUser().team
     }
+    
+    private func updateAllowGuests(_ allow: Bool) {
+        // Remove guests if the toggle was switched off.
+        let participants: Set<ZMUser> = {
+            guard let current = values?.participants else { return [] }
+            guard !allow else { return current }
+            return Set(Array(current).filter { $0.team != ZMUser.selfUser().team || (!$0.isServiceUser) })
```
mikeger: We do the same in the init of `ConversationCreationValues`: https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift#L31

PR #1784, opened 2018-02-28
Enable swipe back gesture in group details

typfel:
### Issues

Swipe back gesture doesn't work in the group details view controller.

### Causes

Setting a custom `leftBarButtonItem` disables the swipe back gesture.

### Solutions

Change the appearance of the back button by modifying the `backIndicatorImage` instead and providing a `backBarButtonItem` without a title.

### Notes

- Unify the appearance of the navigation bar title.


Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
```diff
@@ -79,6 +79,20 @@ class ClearBackgroundNavigationController: UINavigationController {
...
+    override func setViewControllers(_ viewControllers: [UIViewController], animated: Bool) {
+        viewControllers.forEach { viewController in
+            viewController.navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
+        }
+        
+        super.setViewControllers(viewControllers, animated: animated)
+    }
+    
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        viewController.navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
```
mikeger: As I understood you are creating the back item to suppress the default pop gesture? This could generate the issue that is hard to trace, namely: someone creates the new plain controller, creates the `backButtonItem`, pushes it, and the item is not visible.

typfel: The back button icon is always visible, that code just hides the back button title.. This way get the back button behaviour we want without having do anything in our view controllers.

But you right if would be a hard to find bug if you try to something custom. How about if I only  set `backBarButtonItem` if it's not already set.

PR #1790, opened 2018-03-01
Add accessibility identifiers, adjust remove participant confirmation sheet

daehn:
## What's new in this PR?

* Use standard components for remove participant sheet.
* Add accessibility identifiers.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
+++ b/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
@@ -134,7 +134,8 @@
     [super viewDidLoad];
     
     self.colorSchemeController = [[ColorSchemeController alloc] init];
-
+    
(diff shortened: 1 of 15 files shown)


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
```diff
@@ -134,7 +134,8 @@ - (void)viewDidLoad
     [super viewDidLoad];
     
     self.colorSchemeController = [[ColorSchemeController alloc] init];
-
+    
+    [[UIView appearanceWhenContainedInInstancesOfClasses:@[UIAlertController.class]] setTintColor:[ColorScheme.defaultColorScheme colorWithName:ColorSchemeColorTextForeground variant:ColorSchemeVariantLight]];
```
daehn: Not sure where the best place to put this is.

mikeger: May be can be put into an extension on UIAlertController in the separate file? And called from here

PR #1798, opened 2018-03-01
Fix guest icon being shown for users in non-team group

typfel:
### Issues

When team user is added to non-team group he would see all participants as guests.

### Causes

Guest icon logic was not taking the conversation into account.

### Solutions

Use the `isGuest(_ in: ZMConversation)` function to determine if a group participant is a guest.


mikeger:
@typfel conflict 🙉

PR #1802, opened 2018-03-02
Show badges in group participants avatars

typfel:
## What's new in this PR?

Show badges in group participants avatars


typfel:
@mikeger not yet, I'll add one.

PR #1805, opened 2018-03-02
All titles in settings navigation controllers are now uppercase

nicorsm:
## What's new in this PR?

All navigation titles in Settings view controllers are now uppercase. The only leftover is the default Edit/Done button in Devices list provided by `UITableView`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/SettingsTableViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/SettingsTableViewController.swift
@@ -151,7 +151,7 @@ class SettingsTableViewController: SettingsBaseTableViewController {
     required init(group: SettingsInternalGroupCellDescriptorType) {
         self.group = group
         super.init(style: group.style == .plain ? .plain : .grouped)
-        self.title = group.title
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Settings/SettingsTableViewController.swift
```diff
@@ -151,7 +151,7 @@ class SettingsTableViewController: SettingsBaseTableViewController {
     required init(group: SettingsInternalGroupCellDescriptorType) {
         self.group = group
         super.init(style: group.style == .plain ? .plain : .grouped)
-        self.title = group.title
+        self.title = group.title.uppercased()
```
typfel: maybe localised uppercase?

nicorsm: Titles should be already provided with the proper localization, I'll check.

nicorsm: Confirmed, titles are already localised.

PR #1806, opened 2018-03-02
Bump AVS to 4.2.

jspittka:
Please update to latest AVS 4.2. branch. No API changes. All initial tests are positive. Release notes are here: https://github.com/wearezeta/avs/blob/release-4.2/ChangeLog.txt

### DIFF
--- a/avs-versions
+++ b/avs-versions
@@ -17,4 +17,4 @@
 # along with this program. If not, see http://www.gnu.org/licenses/.
 #
 
-export APPSTORE_AVS_VERSION=4.1.20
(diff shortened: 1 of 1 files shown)


mikeger:
@jspittka this would go into the next release, I would roll back.

jspittka:
Depends on schedule. If there is time for a round of testing it makes more sense to include it. The roaming feature is fairly straight forward. What needs more attention (basically make lots of calls) is the updated audio buffer handling for the new iOS audio system. Otherwise, I agree.

PR #1809, opened 2018-03-04
Fix typo in German translation.

fisx:
This PR fixes a simple typo, and is best understood by looking at the commit itself.

thanks!  (-:

### DIFF
--- a/Wire-iOS/Resources/de.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/de.lproj/Localizable.strings
@@ -401,7 +401,7 @@
 
 "content.file.too_big" = "Du kannst Dateien bis zu %@ senden";
 // Someone pinged
-"content.ping.text" = "%1$@ hast gepingt";
+"content.ping.text" = "%1$@ hat gepingt";
 
 // Current user pinged
 "content.ping.text.you" = "Du";


mikeger:
Hey @fisx ! Thanks for the PR. Wire is relying heavily on the external contribution in regards of the translations and localization. Therefore we have created the advanced system that is helping the translators to keep the project up to speed with the most recent changes in the English version of localization. This system is working in the following way:

- As the new feature being developed, the strings are added and removed from the English version of the translation files.
- On every commit the English version of localizations is being uploaded to Crowdin.
- The registered translators are getting notified about the updated string in their languages and updating the thanslations on Crowdin.
- When the release is upcoming, the translations are downloaded from Crowdin and replaced the translations in the files in git.

As you can see, we only transfer the translations in 2 directions: English from git to Crowdin, others from Crowdin to git. Therefore, if I am going to approve your PR then when we are going to download the resources next time your changes are going to be replaced with the Crowdin version of the translation.

So I would have to close this PR without merging. Please consider registering on Crowdin and updating the translation there.

# Note on the issue itself

The issue with German localization is the following: we are using the same string for "Du hast..." and "<Er[Username]> hat...". Therefore if the change is going to be accepted, the other way of using this string would break. The good fix would be to decouple the two strings and use dedicated string for each case. If you feel like digging this, I can help and advise which parts of code can be changed to fix it. Otherwise we also have this issue to be fixed on our radar.

fisx:
ok, i'll leave this to you then, that's probably overall more efficient.  thanks for the exhaustive explanation!

cheers..

mikeger:
@fisx the issue would be fixed in the release 3.9: https://github.com/wireapp/wire-ios/pull/1821

PR #1812, opened 2018-03-05
Updated color, fixed accessibility label

mikeger:
## Issue

The colors of the guest labels updated, also made sure to use the same icons everywhere.

Also fixed the accessibility label on the cell.


typfel:
This will also change colour on the user profile screen which is not correct I think.

Wire-iOS/Sources/UserInterface/Components/Views/GuestIndicator.swift
```diff
@@ -29,11 +29,20 @@ public class GuestIndicator: UIImageView, Themeable {
     }
     
     func applyColorScheme(_ colorSchemeVariant: ColorSchemeVariant) {
-        image = UIImage(for: .guest, iconSize: .tiny, color: UIColor.wr_color(fromColorScheme: ColorSchemeColorIconNormal, variant: colorSchemeVariant))
+        image = UIImage(for: .guest, iconSize: .tiny, color: iconColor)
+    }
+    
+    var iconColor: UIColor {
```
typfel: I know the current colouring system is a bit clunky I still don't think we should abandon it until we have replacement.

daehn:
I think some snapshots might need to be updated after this change.

Wire-iOS/Sources/UserInterface/Components/Views/GuestIndicator.swift
mikeger: I agree, but Eike meant that this icon has this special color, which is not the case for all the other icons. I think we need to develop some framework to follow with colors, then it would be possible to implement it in code.

mikeger:
@typfel I am checking about the participants details screen, indeed it does not look optimal.

mikeger:
@typfel updated

PR #1640, opened 2018-01-23
Feature/zios 9095 change visual appearance of the network status bar

billypchan:
## What's new in this PR?

Change the visual appearance of the network-status bars.

The network status bars will from now on only appear in the list UI and the conversation view

The network status bars will sit on top of the navigation bars and push down the navigation bars and the content below.

The new network status bar and loading bar has rounded corners, new color, and updated size and margins. The "No network connection" toast UI stays on the screen until network resumes. When the toast or loading bar toggles, the height of its embedded navigation bar animates at the same time.

| Idiom/Horizontal size class/Orientantion | ConversationRootViewController shows NetworkStatusView? | ConversationListViewController shows NetworkStatusView? |
| :---: | :---: | :---: |
| iPad/regular/any | Yes | No |
| any/compact/any | Yes | Yes |

TODO list to finish this PR:
- [x] adjust animation timing
- [x] add status bar to conversation list screen
- [x] ~~add status bar to splash screen~~
- [x] management logic to prevent status bar showing on both conversation list and conversation view on iPad
- [x] fix conversation screen top left icons unnecessary animation
- [x] response the size class changing on iPad, refresh network status bar visibility on conversation screen
- [x] For iPad full screen landscape mode, the network status bar should be should on conversation view
- [x] Unit test for horizontal size class and NetworkStatusViewController state
- [x] final design confirmation

This PR is the final step. It removes network status bars from NotificationWindowRootViewController and embed it in ConversationListViewController and ConversationRootViewController

PR #1813, opened 2018-03-05
mock device property injection for test

billypchan:
### Issues

Currently, to test the UI updating when device orientation or size class changes, we have to try it manually on simulator or real device. When we want to test different behavior on different idioms, we have to rerun the same tests to cover the idiom logic related code on different simulators. 

### Causes

For example, in LandingViewController we get device idiom from UIDevice.current.userInterfaceIdiom and current size class from self.traitCollection.horizontalSizeClass. Mocking size class change events and userInterfaceIdiom is not possible.

### Solutions

Create UIIdiomSizeClassOrientationProtocol and injects UIIdiomSizeClassOrientation as a param in the init method. For the unit test, injects MockIdiomSizeClassOrientation instead. Then we can mock the devices properties update events and write tests to check UI updates as we want.

## Notes

This PR is part of https://github.com/wireapp/wire-ios/pull/1640. A new test for NetworkStatusViewController size class changing will be written.


vytis:
I think there is a super easy way to override trait collections by calling `setOverrideTraitCollection(_:forChildViewController:)`. So basically you just need to override the `userInterfaceIdiom`? 

The approach here on injecting dependencies is very convoluted I would say. The easiest way to do it is to create a protocol that exposes the method you need and use that instead of concrete type. There are plenty of examples in other places how we do it, just one that I found quickly:
https://github.com/wireapp/wire-ios/blob/26591d081c987d678ba0cf555a273bd96b15aa8e/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/PhotoPermissionsController.swift#L28

billypchan:
@vytis I get your point, I will refactor the code to match the pattern.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -32,6 +32,8 @@ final class LandingViewController: UIViewController {
 
     private let tracker = AnalyticsTracker(context: AnalyticsContextRegistrationEmail)
 
+    fileprivate var uIIdiomSizeClassOrientationProtocol: UIIdiomSizeClassOrientationProtocol.Type
```
daehn: I would suggest either capitalising both letter at the beginning or none.

Wire-iOS/Sources/Components/UserInterfaceIdiom.swift
```diff
@@ -0,0 +1,27 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol UserInterfaceIdiomProtocol {
+    static var idiom: UIUserInterfaceIdiom {get}
```
vytis: In this case the simplest approach is to create a protocol that `UIDevice` would conform without any changes. This way you can pass `UIDevice` where `UserInterfaceIdiomProtocol` is expected or your mock implementation that returns the `userInterfaceIdiom` you want.

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -144,6 +146,19 @@ final class LandingViewController: UIViewController {
         return button
     }()
 
+    /// init method for injecting mock UserInterfaceIdiomProtocol
+    ///
+    /// - Parameter userInterfaceIdiom: a struct Type compliant with UserInterfaceIdiomProtocol. Default Type is UserInterfaceIdiom.self. Provide this param for testing only.
+    init(userInterfaceIdiom: UserInterfaceIdiomProtocol.Type = UserInterfaceIdiom.self) {
```
vytis: In here you could the just do `UserInterfaceIdiomProtocol = UIDevice.current`

Wire-iOS Tests/LandingViewControllerTests.swift
```diff
@@ -0,0 +1,84 @@
...
+    var mockUserInterfaceIdiom: MockUserInterfaceIdiom! = MockUserInterfaceIdiom()
+
+    override func setUp() {
+        super.setUp()
+        sut = LandingViewController(idiom: mockUserInterfaceIdiom)
+        mockParentViewControler.addChildViewController(sut)
+    }
+    
+    override func tearDown() {
+        super.tearDown()
```
vytis: tearDown should be called at the end of the method

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -144,6 +146,19 @@ final class LandingViewController: UIViewController {
         return button
     }()
 
+    /// init method for injecting mock UserInterfaceIdiomProtocol
+    ///
+    /// - Parameter userInterfaceIdiom: a struct Type compliant with UserInterfaceIdiomProtocol. Default Type is UserInterfaceIdiom.self. Provide this param for testing only.
```
vytis: The comment is outdated and I think a bit verbose - it's probably enough to just say that it's used for testing

PR #1821, opened 2018-03-06
Fix: German "You/Someone pinged" conjugation handling

billypchan:
### Issues

German localization of "You/Someone pinged" are not translated correctly

### Causes

German localization of "You/Someone pinged" have different conjugations. (Du hast/ Man hat gepingt) is not handled in this case. 

### Solutions

Apply the solution from https://github.com/wireapp/wire-ios/pull/943

## Notes

A unit test for PingCell.pingMessage() will be added in next PR.


Wire-iOS/Resources/de.lproj/Localizable.strings
```diff
@@ -402,7 +402,8 @@
 
 "content.file.too_big" = "Du kannst Dateien bis zu %@ senden";
 // Someone pinged
-"content.ping.text" = "%1$@ hast gepingt";
+"content.ping.text" = "%1$@ hat gepingt";
+"content.ping.text-you" = "%1$@ hast gepingt";
```
typfel: Why keep the `%1$@`? won't it always be `du`?

Wire-iOS/Resources/de.lproj/Localizable.strings
```diff
@@ -402,7 +402,8 @@
...
 
 // Current user pinged
 "content.ping.text.you" = "Du";
```
typfel: This can then be removed and we avoid problematic concatenation of strings.

Wire-iOS/Resources/de.lproj/Localizable.strings
daehn: AFAIK the pov system might need it?

billypchan: @daehn I checked and It is not necessary.

PR #1825, opened 2018-03-06
Implement iOS App Store ratings and review dialog

alexaubry:
## What's new in this PR?

After a successful audio call, the call quality survey will be presented to the user. If they rate the call 4 or 5 stars, the app will ask them for an App Store review.

The quality survey will be shown every 10 days, after the user disconnects from a successful call that ended without errors.

## Notes

Testing the call quality survey requires analytics to be enabled.

## To-Do

- [x] Fix copyright headers
- [x] Move constraint setup to `createConstraints`
- [x] Create enum for colors
- [x] Filter out short calls
- [x] Explicitly mention the mute duration in days
- [x] Create constants for system animation curves


Wire-iOS/Sources/UserInterface/Components/Controllers/CallQualityController/CallQualityController.swift
```diff
@@ -105,22 +113,114 @@ class CallQualityViewController : UIViewController {
...
+            bottomAnchor = view.bottomAnchor
+        }
+
+        iphone_bottomConstraint = contentView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -8)
+        ipad_centerYConstraint = contentView.centerYAnchor.constraint(equalTo: view.centerYAnchor)
+        ipad_centerXConstraint = contentView.centerXAnchor.constraint(equalTo: view.centerXAnchor)
+
+
+        updateLayout(for: traitCollection)
+
```
nicorsm: I'd suggest moving the setup of constraints to a separate method called `createConstraints()`, like we're doing in the other view controllers, just to have things in order.

Wire-iOS/Sources/UserInterface/Components/Controllers/CallQualityController/CallQualityController.swift
```diff
@@ -139,26 +239,29 @@ class CallQualityView : UIStackView {
...
         scoreButton.tag = buttonScore
         scoreButton.circular = true
         scoreButton.setTitle(String(buttonScore), for: .normal)
+        scoreButton.titleLabel?.font = UIFont.monospacedDigitSystemFont(ofSize: 18, weight: UIFontWeightRegular)
         scoreButton.setTitleColor(UIColor.cas_color(withHex: "#272A2C"), for: .normal)
         scoreButton.setTitleColor(.white, for: .highlighted)
         scoreButton.setTitleColor(.white, for: .selected)
         scoreButton.addTarget(self, action: #selector(onClick), for: .primaryActionTriggered)
-        scoreButton.setBackgroundImageColor(.white, for: UIControlState.normal)
+        scoreButton.setBackgroundImageColor(UIColor.cas_color(withHex: "#F8F8F8"), for: UIControlState.normal)
```
nicorsm: Maybe it's worth moving all the colors to a proper enum like we're doing in https://github.com/wireapp/wire-ios/blob/7c82e8a5f9e1469235b34f46491719fc9f1afffa/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/UIColor%2BTeamCreation.swift, so we can avoid hardcoded hex colors in the code.

Wire-iOS/Sources/Analytics/CallQualityScoreProvider.swift
```diff
@@ -64,3 +63,33 @@ final class CallQualityScoreProvider: NSObject, AnalyticsType {
         return nextProvider?.persistedAttributes(for: event)
     }
 }
+
+// MARK: - Survey Mute Filter
+
+let UserDefaultLastCallSurveyDate = "LastCallSurveyDate"
+
+// Show the survey only once every 10 days (10 * 24 * 3600)
+let CallSurveyMuteInterval: TimeInterval = 864_000
```
vytis: Just a nitpick, but maybe it makes sense to write the equation in code. Comments tend to get outdated and noone will bother to calculate if this is still 10 days

Wire-iOS/Sources/UserInterface/Components/Controllers/CallQualityController/CallQualityPresentationTransition.swift
```diff
@@ -0,0 +1,67 @@
...
+            contentView.transform = CGAffineTransform(scaleX: 0, y: 0)
+            
+        default:
+            contentView.transform = CGAffineTransform(translationX: 0, y: containerView.frame.height)
+        }
+        
+        // Animate Presentation
+        
+        let duration = transitionDuration(using: transitionContext)
+        let curve = UIViewAnimationOptions(rawValue: 7 << 16)
```
vytis: I think there should be some constants here that you could use

Wire-iOS/Sources/UserInterface/Overlay/ActiveVoiceChannelViewController.swift
```diff
@@ -176,31 +177,79 @@ extension ActiveVoiceChannelViewController : WireCallCenterCallStateObserver {
...
+        }
+
+        if case let .terminating(reason) = callState {
+            
+            guard answeredCalls.contains(conversation.remoteIdentifier!) else {
+                return
+            }
+            
+            // Only show the survey if the called finished without errors
+            guard reason == .normal || reason == .stillOngoing else {
```
vytis: Just a minor detail, but maybe it's worth checking for call length as well, sometimes ir happens that call is hung up immediately for some reason. It is considered a successful call, but nothing useful actually happened. Maybe @jspittka can chip in here.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -245,7 +245,15 @@ import Foundation
         developerCellDescriptors.append(shareCryptobox)
         let reloadUIButton = SettingsButtonCellDescriptor(title: "Reload user interface", isDestructive: false, selectAction: SettingsCellDescriptorFactory.reloadUserInterface)
         developerCellDescriptors.append(reloadUIButton)
-        
+
+        if (UseAnalytics.boolValue || AutomationHelper.sharedHelper.useAnalytics) &&
+            !TrackingManager.shared.disableCrashAndAnalyticsSharing {
+
+            let resetSurveyMuteButton = SettingsButtonCellDescriptor(title: "Show call quality survey", isDestructive: false, selectAction: SettingsCellDescriptorFactory.resetCallQualitySurveyMuteFilter)
```
vytis: Title needs to be localized

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -385,6 +393,18 @@ import Foundation
         
         rootViewController.reload()
     }
+
+    private static func resetCallQualitySurveyMuteFilter(_ type: SettingsCellDescriptorType) {
+        guard let controller = UIApplication.shared.wr_topmostController(onlyFullScreen: false) else { return }
+
+        CallQualityScoreProvider.resetSurveyMuteFilter()
+
+        let alert = UIAlertController(title: "Mute Filter Removed",
```
vytis: Same here

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
alexaubry: Do we need to localize it even if it's in the internal developer menu?

vytis: Ah i didn't realize it is in developer's menu. You are right, we do not localize things in dev menu, so ignore my comments :)

Wire-iOS/Sources/UserInterface/Components/Controllers/CallQualityController/CallQualityController.swift
alexaubry: Good point, implemented this in 80287ce03f0610df4b774419b36d5fdb9ac63156.

malicioustoker:
After every 10 days I’m going to be asked to rate the app if the quality of the audio is good?? Thats going to get annoying....

alexaubry:
@malicioustoker Not exactly. After every 10 days, you’ll be asked to rate the quality of the call. But since we’re using the [native review controller](https://developer.apple.com/app-store/ratings-and-reviews/), the system will make sure the App Store reviews don’t get annoying.

If you give an App Store review after rating the call, we will not ask you for another review before at least 365 days.

If you don’t give a review, the system will only ask for a review two more times within a 365 days period, when it thinks it’s appropriate (not necessarily the second and third time you rate a call).

PR #1826, opened 2018-03-06
ZIOS-9720: Login issues - Registering with existing phone number

nicorsm:
### Issues

App was going into an infinite loop while signing up with an existing phone number that also had email assigned in the account.

### Causes

This was caused by an ineffective call to `popToRootViewController` on the wrong navigation controller.

### Solution

We're calling the pop function on the correct navigation controller. After that, an alert is shown on screen requiring a password.


Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -114,7 +114,7 @@
          </CommandLineArgument>
          <CommandLineArgument
             argument = "-ZMBackendEnvironmentType staging"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
mikeger: Debug: Please revert.

PR #1827, opened 2018-03-06
Added safe area margins in signup screens

nicorsm:
## What's new in this PR?

Added iPhone X safe area margins to view controllers in the registration flow.


mikeger:
@nicorsm can we pin to `safeBottomAnchor` instead of using the safe area?

nicorsm:
Let's try! 🙂

PR #1828, opened 2018-03-06
mock device orientation injection for test

billypchan:
### Issues

This PR is related to https://github.com/wireapp/wire-ios/pull/1813.
Currently, to test the logic when device orientation changes, we have to try it manually on the simulator or device. 

### Causes

Take CameraCell as an example, It reads UIDevice.current.orientation and calculate for new AVCaptureVideoOrientation value. it is not possible to mock the device orientation change event.

### Solutions

Create DeviceOrientationProtocol and injects UIDevice.current as a param in the init method. For the unit test, injects MockDeviceOrientation instead. Then we can mock the device's orientation update events and write tests to check UI updates as we want. (CameraControllerProtocol and MockCameraController are created, for injects a mock CameraController for simulator)

## Notes

This PR is part of https://github.com/wireapp/wire-ios/pull/1640. A new test for NetworkStatusViewController size class changing will be written.


Wire-iOS/Sources/Components/DeviceOrientation.swift
```diff
@@ -0,0 +1,27 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol DeviceOrientationProtocol {
```
vytis: Maybe it's better to unify this with the protocol in #1813. It makes sense to mock all device properties and pass them down to views. For example if we would start doing something different in CameraCell on iPad we would need to have same property in two mock protocols. 
So I propose making something like MockDevice and using it everywhere.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraCell.swift
```diff
@@ -36,14 +36,17 @@ open class CameraCell: UICollectionViewCell, Reusable {
...
+
     deinit {
         NotificationCenter.default.removeObserver(self)
     }
-    
+
     override init(frame: CGRect) {
         self.cameraController = CameraController()
-        
+        self.deviceOrientation = UIDevice.current
```
vytis: It's a bit inconsistent how the dependency is injected - there is a default parameter in the other initializer but then just setting a property in here. Maybe it's better to add the parameter here and have only one init method? Or it doesn't work because we need it from objc?

Wire-iOS/Sources/Components/DeviceOrientation.swift
billypchan: I will merge this protocol with UserInterfaceIdiomProtocol. Thanks!

Wire-iOS/Sources/Components/DeviceProtocol.swift
```diff
@@ -18,10 +18,11 @@
 
 import Foundation
 
-protocol UserInterfaceIdiomProtocol {
+protocol DeviceProtocol {
```
vytis: So now we have two identical protocols - MockDevice and DeviceProtocol. You don't need to have one in tests one in code and can simply use this in tests as well.

billypchan: @vytis MockDevice is a class conforms DeviceProtocol, which mimics UIDevice class. It would be clearer to copy the UIDevice's original design.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraCell.swift
```diff
@@ -268,3 +276,14 @@ open class CameraCell: UICollectionViewCell, Reusable {
         Settings.shared().preferredCamera = cameraController.currentCamera
     }
 }
+
+extension CameraCell {
+    /// init method with a param for injecting mock device
+    ///
+    /// - Parameters:
+    ///   - device: Provide this param for testing only
+    convenience init(device: DeviceProtocol) {
```
billypchan: In the future this for-test-only init method would be better to put in another module, to reduce the release code size.

Wire-iOS/Sources/Components/DeviceProtocol.swift
vytis: Oh yeah, somehow I misread the code. All is good here 👍

PR #1833, opened 2018-03-07
ZIOS-9682, 9724: Login issues - Infinite loading when logging in via registration tab with phone number

nicorsm:
### Issues

There was an infinite loop while attempting to register for the second time with a phone number already associated with a user. 

### Causes

This was partially fixed by https://github.com/wireapp/wire-ios/pull/1826, but it was still pushing back the user to the sign in view controller after verifying the code and setting up the email address.

### Solutions

We've decided to remove these steps and guide the user to do the login via the classic login section of the sign in view controller. 

## Notes

This PR also covers ZIOS-9724 - _Login issues: Logging in with phone on a second device_ with this commit: https://github.com/wireapp/wire-ios/pull/1833/commits/5a4ba1ba791ef808c46adf04ea816e6cc4326228 .

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios/pull/1826
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/705


Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/PhoneNumberViewController.h
```diff
@@ -31,6 +31,7 @@
 
 @property (nonatomic, readonly) RegistrationTextField *phoneNumberField;
 @property (nonatomic, readonly) Country *country;
+@property (nonatomic) NSString *phoneNumber;
```
typfel: copy?

nicorsm: No, this is a property for setting correctly the string of a phone number inside the text field, letting us split it between international prefix and local phone number.

nicorsm: (See the setter in the .m file)

Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/PhoneNumberViewController.m
```diff
@@ -186,34 +186,19 @@ - (void)setEditable:(BOOL)editable
...
-
-- (BOOL)textField:(UITextField *)textField shouldChangeCharactersInRange:(NSRange)range replacementString:(NSString *)string
+-(void)setPhoneNumber:(NSString *)phoneNumber
 {
-    NSString *newString = [textField.text stringByReplacingCharactersInRange:range withString:string];
-    NSString *phoneNumber = [NSString phoneNumberStringWithE164:self.country.e164 number:newString];
-    
-    NSError *error = nil;
-    [ZMUser validatePhoneNumber:&phoneNumber error:&error];
+    _phoneNumber = phoneNumber;
```
typfel: we should do `[phoneNumber copy]` here then.

Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/PhoneNumberViewController.m
```diff
@@ -249,6 +234,37 @@ - (BOOL)textField:(UITextField *)textField shouldPasteCharactersInRange:(NSRange
...
+{
+    NSString *newString = [textField.text stringByReplacingCharactersInRange:range withString:string];
+    NSString *phoneNumber = [NSString phoneNumberStringWithE164:self.country.e164 number:newString];
+    
+    NSError *error = nil;
+    [ZMUser validatePhoneNumber:&phoneNumber error:&error];
+    
+    if (error.code == ZMObjectValidationErrorCodeStringTooLong ||
+        error.code == ZMObjectValidationErrorCodePhoneNumberContainsInvalidCharacters) {
+        return NO;
```
mikeger: Should the code go through in case of other (potential?) error?

nicorsm: The other possible validation errors are `ZMObjectValidationErrorCodeStringTooShort` and `ZMObjectValidationErrorCodeEmailAddressIsInvalid`, two errors that shouldn't appear while inserting phone numbers.

mikeger: Sure, but just in case we would add some additional error or change the logic in error generation the case would go through, I think it's more correct to break on any error.

PR #1839, opened 2018-03-08
Use new SE API to check if the conversation access can be managed

mikeger:
- Cleaner implementation, tested and implemented in one place.


daehn:
@mikeger there is a similar check in `ParticipantsCellViewModel`.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsCellViewModel.swift
```diff
@@ -99,10 +99,9 @@ struct ParticipantsCellViewModel {
     }
     
     var showInviteButton: Bool {
-        guard case .started = message.actionType else { return false }
-        return true == message.conversation?.allowGuests
-            && nil != message.conversation?.team
-            && ZMUser.selfUser().isTeamMember
+        guard case .started = message.actionType,
+                let conversation = message.conversation else { return false }
+        return conversation.canManageAccess
```
mikeger: @daehn please confirm that I am holding it right.

mikeger: @daehn bump

daehn: Yes, looks correct to me!

PR #1841, opened 2018-03-08
Add UI test cases

michelleXue:
### Issues

Add UI test cases to test GUIs of users' daily usage

### Causes

No existing UI test cases

### Solutions

Add 34 UI test cases with 40% code coverage

## Dependencies

no dependencies

Needs releases with:

- [x] GitHub link to the pull request

## Notes

GUI Test cases can test the app from users aspect

### DIFF
--- a/Wire-UI-Tests/ConversationsTest.swift
+++ b/Wire-UI-Tests/ConversationsTest.swift
@@ -1,0 +1,450 @@
+//
+//  ConversationsTest.swift
+//  Wire-iOS
+//
+//  Created by Xue Qin on 1/18/18.
+//  Copyright © 2018 Zeta Project Germany GmbH. All rights reserved.
+//
+
+import XCTest
+
+class ConversationsTest: Wire_UI_Tests {
+    
+    func testAddConversation() {
+        app.buttons["bottomBarPlusButton"].tap()
+        sleep(5)
+        if app.alerts["“Wire” Would Like to Access Your Contacts"].exists {
+            app.alerts["“Wire” Would Like to Access Your Contacts"].buttons["OK"].tap()
+        }
+        sleep(1)
+        app.collectionViews.staticTexts["Michelle"].tap()
+        sleep(2)
+        app.buttons["OPEN"].tap()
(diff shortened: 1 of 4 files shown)


mikeger:
Hey @michelleXue ! Thanks for the remarkable work. We had no automated xcode-based UI tests so far, however internally we already utilize the UI tests using Selenium (Appium) framework. This being said, I think that having the second set of tests is not something that can hurt us.

I would like to add some comments on how your tests can be improved, I hope that they would be helpful.

1. Your tests are seem to be based on the fact that the app is signing in with one particular user that has a particular email and password. The problem that can happen here is for example that someone or even stranger on the internet would find the login credentials to this users (they are now plain text in your PR) and can log in to it breaking the testing routine. I think this is an issue that is not that easy to fix with your current setup. 

I have no good solution except than injecting the https://github.com/wireapp/wire-ios-mocktransport which is in the nutshell the fake Wire backend written in Swift. If you would like to continue on this path please contact me so I can explain more in detail how this can be done.

2. Generally, when the test is created it's necessary to define the criteria that would indicate the success of the test (could be formulated like https://en.wikipedia.org/wiki/Given-When-Then ). For example: 
*Given:* I am on the conversation view
*When:* I post the message
*Then:* I see the message appearing in the conversation view
Having this `then` part reassures that the aim of the test, i.e. to verify the functionality, is achieved.

I think in some of the tests that you have created this `then` part is not being completely verified. May be it makes sense to add this given-when-then labels and having it define the structure of your tests.

mikeger:
@michelleXue as we are unable to merge your contribution in the current state I would suggest to cancel the existing pull request and recreate it once you are up to changing the tests in the way that it was suggested. Sorry for the inconvenience.

michelleXue:
Hi Mike,
Actually, we are currently using Wire app, but just need to confirm the GUI quality. So we end up with writing the UI test cases ourselves. We are glad to hear you already have the UI tests using Selenium framework. Is there any chance that you can share these test cases with me? Or do you know where I can find it?

Thank you so much!

PR #1844, opened 2018-03-12
ZIOS-9722: Login issues - iPad fixes for registration

nicorsm:
### Issues

The behavior of logging in from the registration flow with an already registered user on iPad was broken. It was going forward to the next steps (e.g. picture choosing) and, after that, was pushing back the user to the login screen.

### Solutions

I've changed the registration flow in case of an already registered user: after the Terms of Service screen, now the user is pushed back to the login view with username and password prefilled.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/708

## Notes

Still WIP, needs cleaning and refactoring. Comments are welcome!


vytis:
So is there anything left to do here? Maybe time to remove "do not merge"?

PR #1849, opened 2018-03-12
Feature/network status bar delegates

billypchan:
## What's new in this PR?

Originally NetworkStatusViewController belongs to NotificationWindowRootViewController. In the new design it belongs to ConversationRootViewController and ConversationListViewController.

In the new design, NetworkStatusView's visibility depends on device's properties:

| Idiom/Horizontal size class/Orientantion | ConversationRootViewController shows NetworkStatusView? | ConversationListViewController shows NetworkStatusView? |
| :---: | :---: | :---: |
| iPad/regular/landscape | No | Yes |
| iPad/regular/portrait | Yes | No |
| any/compact/any | Yes | Yes |

## Notes

This PR is part of https://github.com/wireapp/wire-ios/pull/1640.


Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
```diff
@@ -0,0 +1,221 @@
...
+import XCTest
+@testable import Wire
+
+final class MockConversationRootViewController: UIViewController, NetworkStatusBarDelegate {
+    var isViewDidAppear: Bool = true
+    
+    var networkStatusViewController: NetworkStatusViewController!
+    
+    var shouldShowNetworkStatusUIInIPadRegularLandscape: Bool {
+        get {
```
vytis: No need for `get {}`, only the line with `return` will work fine

Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
```diff
@@ -0,0 +1,221 @@
...
+    var mockConversationList: MockConversationListViewController!
+    
+    override func setUp() {
+        super.setUp()
+        mockDevice = MockDevice()
+        
+        mockConversationList = MockConversationListViewController()
+        sutList = NetworkStatusViewController(device: mockDevice)
+        mockConversationList.networkStatusViewController = sutList
+        mockConversationList.addChildViewController(sutList)
```
vytis: There are some more things you need to do to properly add a child view controller - https://developer.apple.com/library/content/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html#//apple_ref/doc/uid/TP40007457-CH11-SW13

Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
```diff
@@ -0,0 +1,221 @@
...
+                                                    horizontalSizeClass: UIUserInterfaceSizeClass,
+                                                    orientation: UIDeviceOrientation,
+                                                    listState: NetworkStatusViewState,
+                                                    rootState: NetworkStatusViewState,
+                                                    file: StaticString = #file, line: UInt = #line) {
+        // GIVEN & WHEN
+        setUpSut(userInterfaceIdiom: userInterfaceIdiom, horizontalSizeClass: horizontalSizeClass, orientation: orientation)
+
+        // THEN
+        checkResult(listState: listState, rootState: rootState)
```
vytis: To have test failures showed in correct place you need to pass `file` and `line` parameters to this function from outside. Otherwise all the failures will be shown in this line and it makes it hard to see what went wrong. Easy to check what happens by making a test fail artificially.

Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
billypchan: We have UIViewController.addToSelf(_ viewController:) helper for this.
https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Helpers/UIViewController%2BaddViewController.swift

Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
```diff
@@ -0,0 +1,213 @@
...
+    /// check for networkStatusView state is updated after device properties are changed
+    ///
+    /// - Parameters:
+    ///   - userInterfaceIdiom: updated idiom
+    ///   - horizontalSizeClass: updated size class
+    ///   - orientation: updated orientation
+    ///   - listState: expected networkStatusView state in conversation list
+    ///   - rootState: expected networkStatusView state in conversation root
+    ///   - file: optional, for XCTAssert logging error source
+    ///   - line: optional, for XCTAssert logging error source
```
vytis: I think it's a bit too much to add comments to a fileprivate function inside a test file. It's pretty obvious what it does and exact meaning of parameters can be figured out easily from how it is used

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,35 +19,82 @@
...
-    
+
+    /// default init method with a parameter for injecting mock device
+    ///
+    /// - Parameter device: Provide this param for testing only
+    init(device: DeviceProtocol = UIDevice.current) {
+        super.init(nibName: nil, bundle: nil)
+
+        self.device = device
+        NetworkStatusViewController.selfInstances.append(self)
```
vytis: Could you explain a bit more why you need to store all created instances in an array? As I understand there can be more than one bar when on iPad(?) and NetworkStatusViewController.shared returns only the one that is currently shown?

billypchan: @vytis Yes, we will have two NetworkStatusViewController instance. The array is for static func notifyWhenOffline() to call all instance to update(). 

(And thanks it reminds me that clean up is needed in deinit())

vytis: So maybe it would be better to let them handle inside if they show the contents or not depending on some flag/object. Because it's not very good practice to keep accumulating (it stores strong references) stuff in static variables. For example on logout/change accounts we basically reload the UI and set up everything again, so you need to be extra careful that this also gets cleaned up. And same (possibly) applies to tests.

billypchan: It reminders me that this class has timer retain issue not yet fixed (related story:  https://github.com/wireapp/wire-ios/pull/1558) 

I agree that a static array may be problematic. Simply replace the array use two static variable is easier to handle.

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,51 +19,142 @@
...
+    var shouldShowNetworkStatusUIInIPadRegularLandscape: Bool {get}
+
+    /// if return false, NetworkStatusViewController will not disapper in iPad regular mode portrait orientation.
+    var shouldShowNetworkStatusUIInIPadRegularPortrait: Bool {get}
+}
+
+extension NetworkStatusViewController {
+    // static pointer for global access
+    static weak var selfInConversationRootView: NetworkStatusViewController?
+    static weak var selfInConversationListView: NetworkStatusViewController?
```
billypchan: @vytis I write the array into two weak static var to simplify the pattern.

vytis:
We talked with @billypchan in person - I think it's a bit problematic that we have these variables and directly say to multiple `NetworkStatusViewController` what to do. Also if we add one more indicator somewhere then we would need to add a third variable and manage it. Would be much better to decouple this and make the `NetworkStatusViewController` show/hide itself depending on the way it is setup. So for example one controller could be shown on compact mode on ipad, but the other one on all devices in landscape orientations (just a random combination). We could trigger all of this with a notification for example. @mikeger what do you think? Or the current approach is simpler?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -7130,8 +7145,10 @@
 				1651F9C01D36789C00A9FAE8 /* Message+FormattingTests.swift in Sources */,
 				87BEB0E21C734DB60094BFE9 /* MockMessage.swift in Sources */,
 				162917CC1F8CC176000F3E92 /* NetworkStatusViewTests.swift in Sources */,
+				EF2128EE2056D55400C1673B /* NetworkStatusViewControllerTests.swift in Sources */,
 				EE75B774203C6F5300855099 /* MarkdownTextStorageTests.swift in Sources */,
-				87D5E4431C3E79DA00EB8289 /* UserClient+Fingerprint.swift in Sources */,
+
+87D5E4431C3E79DA00EB8289 /* UserClient+Fingerprint.swift in Sources */,
```
mikeger: Weird indentation: consequence of merge?

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,51 +19,142 @@
 import Foundation
 import Cartography
 
+typealias NetworkStatusBarDelegate = NetworkStatusViewControllerDelegate & NetworkStatusViewDelegate
+
+protocol NetworkStatusViewControllerDelegate: class {
+    /// if return false, NetworkStatusViewController will not disapper in iPad regular mode landscape orientation.
+    var shouldShowNetworkStatusUIInIPadRegularLandscape: Bool {get}
```
mikeger: I don't have a good feeling about the current naming, I think it's too long right now. Moreover, I think it's more flexible to create the delegate method like `func shouldShow(networkStatus viewController: NetworkStatusViewController, in mode: Mode)` since we might change the list of modes potentially.

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,51 +19,142 @@
...
+
+    /// if return false, NetworkStatusViewController will not disapper in iPad regular mode portrait orientation.
+    var shouldShowNetworkStatusUIInIPadRegularPortrait: Bool {get}
+}
+
+extension NetworkStatusViewController {
+    // static pointer for global access
+    static weak var selfInConversationRootView: NetworkStatusViewController?
```
mikeger: How about this solution: 
- Remove global references.
- In `notifyWhenOffline` send the notification that all instances would listen to.
- In the listener check current state and show accordingly.

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,51 +19,142 @@
 import Foundation
 import Cartography
 
+typealias NetworkStatusBarDelegate = NetworkStatusViewControllerDelegate & NetworkStatusViewDelegate
```
mikeger: I think I am missing some details, which classes are implementing this delegate? Or those are the part of the other PR?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -7130,9 +7145,10 @@
 				1651F9C01D36789C00A9FAE8 /* Message+FormattingTests.swift in Sources */,
 				87BEB0E21C734DB60094BFE9 /* MockMessage.swift in Sources */,
 				162917CC1F8CC176000F3E92 /* NetworkStatusViewTests.swift in Sources */,
+				EF2128EE2056D55400C1673B /* NetworkStatusViewControllerTests.swift in Sources */,
 				EE75B774203C6F5300855099 /* MarkdownTextStorageTests.swift in Sources */,
 				87D5E4431C3E79DA00EB8289 /* UserClient+Fingerprint.swift in Sources */,
-				BF7ED2DC1DF1D1CF003A4397 /* IncomingConnectionViewTests.swift in Sources */,
+                BF7ED2DC1DF1D1CF003A4397 /* IncomingConnectionViewTests.swift in Sources */,
```
mikeger: Still appears off, please check your editor. I think other lines have 4 tabs.
screen shot 2018-03-14 at 10 57 40

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
billypchan: It will added in next PR.

Wire-iOS/Sources/UserInterface/Overlay/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -19,51 +19,101 @@
...
             update(state: viewState(from: userSession.networkState))
             networkStatusObserverToken = ZMNetworkAvailabilityChangeNotification.addNetworkAvailabilityObserver(self, userSession: userSession)
         }
         
         networkStatusView.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(tappedOnNetworkStatusBar)))
     }
-    
-    public func notifyWhenOffline() -> Bool {
+
+    func chnageStateFormOfflineCollapsedToOfflineExpanded() {
```
mikeger: Typo: `chnage`

PR #1852, opened 2018-03-13
Fix share controller tint, remove empty screen for group details

daehn:
## What's new in this PR?

* Fix share controller tint.
* Remove empty screen for group details.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation Options/TintCorrectedActivityController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation Options/TintCorrectedActivityController.swift
@@ -1,0 +1,35 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+// Subclass intended to work around https://stackoverflow.com/questions/25795065/ios-8-uiactivityviewcontroller-and-uialertcontroller-button-text-color-uses-wind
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation Options/TintCorrectedActivityController.swift
```diff
@@ -0,0 +1,35 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+// Subclass intended to work around https://stackoverflow.com/questions/25795065/ios-8-uiactivityviewcontroller-and-uialertcontroller-button-text-color-uses-wind
+final class TintCorrectedActivityViewController: UIActivityViewController {
```
typfel: This can't be solved by UIApperance?

typfel: Ah read the stackoverflow link :)

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
```diff
@@ -131,6 +131,7 @@ - (instancetype)init
         [[SectionHeader appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setColorSchemeVariant:ColorSchemeVariantDark];
         [[GroupConversationCell appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setColorSchemeVariant:ColorSchemeVariantDark];
         [[GroupConversationCell appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setContentBackgroundColor:UIColor.clearColor];
+        [[UIView appearanceWhenContainedInInstancesOfClasses:@[UIAlertController.class]] setTintColor:[ColorScheme.defaultColorScheme colorWithName:ColorSchemeColorTextForeground variant:ColorSchemeVariantLight]];
```
typfel: Is this redundant then?

Wire-iOS/Sources/UserInterface/Conversation Options/TintCorrectedActivityController.swift
daehn: No, I tried but it get's overwritten by the UIWindows tintColor unfortunately (see the linked SO thread).

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
daehn: This is for UIAlertController and was there before, I only moved it to the other appearance calls from further down in the same file.

Issue #1485, opened 2017-11-30
Fail to paste image from clipboard into text field


AndreaVEH:
## Checklist (please check before submitting)

- [x]  I checked the existing issues, and the issue was not reported before.

## The problem

After copying an image elsewhere, I cannot paste the image into text field. There is no response after I tap the paste of edit menus. In Notes app, I can normally paste the image whether it is from local clipboard or universal clipboard. But in Wire, it always fails. No response when tapping paste option of edit menus from local clipboard. And when I want to invoke edit menus to paste from universal clipboard, Wire will not show edit menus, instead it will automatically show a panel displaying "Pasting from X's device" and fade away like nothing has happened. After that failure, I can invoke edit menus, but tapping paste option will have no response like pasting from local clipboard. 

But when if I cut the image into the local clipboard, then I can normally paste it into text field and send it to friends. 

And copying and pasting images works in other third-party apps. 

## Environment

* Wire version that exhibits the issue: **Version 3.2(2921)**
* Last Wire version that did not exhibit the issue (if applicable): **Version 3.1**
* iOS version used to run Wire: **iOS 11.1.2**
* iOS device type: **iPhone SE**
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): **Wi-Fi**
* Exact date, time and timezone (only for the crash issues): 

## Details

1. Copy an image elsewhere into clipboard(local clipboard when on the same iOS device or universal clipboard when on the other iOS or macOS device)
2. Open Wire then touch and hold or double-tap in a text field to invoke edit menus
3. From local clipboard paste option will appear and tap it, but nothing will happen. 
4. From universal clipboard paste option will not appear at first, instead a panel showing "Pasting from XXX" will appear and fade away automatically. After that, edit menus can be invoked, but nothing will happen after tapping paste option.
5. Cut an image from other app on the same device and can be able to paste into text field through tapping paste option of edit menus.

AndreaVEH:
The problem still exists in iOS 11.2. And I find the problem only appears when copy item is an image. There is no problem if copying text via local clipboard or universal clipboard.

AndreaVEH:
The problem still appears on Version 3.3(2934).

mikeger:
Hey @AndreaVEH ! Thanks for reporting the issue, we'll look into it.

AndreaVEH:
So does Version 3.4 (2941).

AndreaVEH:
So does Version 3.5 (2945).

[...]

PR #1854, opened 2018-03-13
ZIOS-9738: Cannot copy images from the clipboard

nicorsm:
### Issues

As reported here: https://github.com/wireapp/wire-ios/issues/1485, there are some cases where the images are not copied correctly from the clipboard.

### Causes

The main cause is that the images copied from Universal Clipboard are not properly recognized as images: while debugging, `[self containsPasteboardTypes:UIPasteboardTypeListImage]` returns `false` while `self.hasImages` returns `true`. While copying images from other sources like the local clipboard. This method is available only from iOS 10.
Furthermore, calling `pasteboard.image` from `canPerformAction:withSender:` is causing the system to display immediately the "Pasting from <Device>" popup in case of pasting from Universal Clipboard (and it's useless since APIs provide the `hasImages` method).

### Solutions

I've implemented three new methods in the `UIPasteboard` category for checking if the clipboard contains images, strings and urls.


vytis:
In my experience these sort of compatibility extensions are quite hard to get rid of afterwards. A nice approach that solves this is to make a separate file with a category that provides the compatibility methods ( `wr_hasImages` and friends). Inside those methods use `if (@available(iOS 10, *))` check and call `hasImages` method. This way when we want to drop iOS 9 we can simply remove the category and replace all calls to `wr_hasImages` with `hasImages`.

PR #1856, opened 2018-03-13
ZIOS-9491: Fix audio recording button reachability

alexaubry:
### Issues

This makes the long press gesture to record audio messages in the conversation view controller more accessible and reliable.

### Causes

In some cases, the long press gesture recognizer would not recognize gestures. Furthermore, the minimum duration for the long press was not specified.

### Solutions

This issue was fixed by using a tap gesture recognizer instead of a button event, which avoids conflicts between the two. A shorter minimum delay before presentation was specified.

Conflicts between gesture recognizers were also reduced by disabling the pan-to-close gesture of the split view when recording audio.

An issue where the overlay would appear green at presentation was also fixed.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
@@ -67,7 +67,6 @@ extension ConversationInputBarViewController {
                 audioRecordViewController.setRecordingState(.recording, animated: false)
                 audioRecordViewController.beginRecording()
                 self.inputBar.buttonContainer.isHidden = true
-                (self.parent as? ConversationViewController)?.lockInterfaceInCurrentConversation()
             }
         case .changed:
             if let audioRecordViewController = self.audioRecordViewController {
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -57,9 +63,11 @@ extension ConversationInputBarViewController {
         case .began:
             self.createAudioRecord()
             if let audioRecordViewController = self.audioRecordViewController , showAudioRecordViewControllerIfGrantedAccess() {
+                audioRecordViewController.setOverlayState(.expanded(0), animated: true)
                 audioRecordViewController.setRecordingState(.recording, animated: false)
                 audioRecordViewController.beginRecording()
                 self.inputBar.buttonContainer.isHidden = true
+                (self.parent as? ConversationViewController)?.lockInterfaceInCurrentConversation()
```
mikeger: May be the same functionality can be achieved by implementing the delegate methods of the gesture recognizers and defining which gesture recognizer should fail first? As I understand the problem, the horizontal pan recognizer is conflicting with the long tap. In theory you can implement the long press gesture recognizer delegate and say that those two recognizers can function simultaneously?

mikeger: Or using this one  https://developer.apple.com/documentation/uikit/uigesturerecognizerdelegate/1624229-gesturerecognizer

PR #1860, opened 2018-03-14
[ZIOS-9417] Implement 3D Touch actions for image preview

alexaubry:
### Issues

In the current implementation, previewing an image message with 3D Touch displays an image controller with the navigation bar and an action buttons row. These buttons could not be tapped and the user could not quickly interact with the image (like/share/... from the preview).

### Solutions

This PR introduces system preview action items on the image controller. This allows the user to swipe up when previewing and quickly interact with the image from there.

#### Before

peek_before

#### After

peek_after_image

peek_after_actions

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
@@ -67,6 +67,12 @@ final class ConversationImagesViewController: UIViewController {
             self.createNavigationTitle()
         }
     }
+
+    public var isPreviewing: Bool = false {
+        didSet {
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -325,7 +339,9 @@ final class ConversationImagesViewController: UIViewController {
     }
     
     @objc public func saveCurrent(_ sender: UIButton!) {
```
nicorsm: Is it possible to make `sender` optional (`_ sender: UIButton? = nil`), in order to make a clean call to the function without passing the `nil` parameter?

PR #1864, opened 2018-03-15
Added the possibility to show the global top overlay

mikeger:
This should be useful if we would like to display the global overlay that should push the whole app content down.


typfel:
Is there a design for this? How would it work on iPad?

mikeger:
So in this PR there is no "it", what is implemented is the possibility to set the view to sit on top of the logged in controller. On iPad with current implementation it would push the app content (both left and right panels) down.

typfel:
Exactly, that was my concern. If that's not how design wants it we have a problem. Why not develop the entire feature in a feature branch?

mikeger:
@typfel yeah I wanted to save my code until someone would pick up this stuff, I guess it can stay on the branch and be merged later

PR #1866, opened 2018-03-15
ZIOS-8754: Support VoiceOver in login flow

alexaubry:
### Issues

Logging into the app cannot be performed with VoiceOver. 

### Solutions

We ensure that accessibility labels and traits are set and localized correctly. We also make sure that all data is accessible and understandable for users of assistive technologies (images, transitions)...

### DIFF
--- a/WireExtensionComponents/Views/TabBar/TabBarController.swift
+++ b/WireExtensionComponents/Views/TabBar/TabBarController.swift
@@ -37,6 +37,9 @@ public extension UIViewController {
     }
 
     @objc public func takeFirstResponder() {
+        if UIAccessibilityIsVoiceOverRunning() {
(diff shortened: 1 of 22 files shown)


WireExtensionComponents/Views/TabBar/TabBarController.swift
```diff
@@ -37,6 +37,9 @@ public extension UIViewController {
     }
 
     @objc public func takeFirstResponder() {
+        if UIAccessibilityIsVoiceOverRunning() {
+            return
+        }
```
mikeger: Is this code necessary? I think the purpose of this overriding method is to suppress the default implementation of the `takeFirstResponder`.

PR #1867, opened 2018-03-15
Add expiration time to user cell subtitle, fix avatar accent color

daehn:
## What's new in this PR?

* Add expiration time to user cell subtitle.
* Fix wireless avatar accent color.


Wire-iOS/Sources/Helpers/syncengine/ZMUser+ExpirationTimeFormatting.swift
```diff
@@ -36,7 +36,8 @@ final class WirelessExpirationTimeFormatter {
         return formatter
     }()
     
-    func string(for user: ZMUser) -> String? {
+    func string(for user: ZMBareUser) -> String? {
+        guard let user = user as? ZMUser else { return nil }
         return string(for: user.expiresAfter)
```
daehn: Might be worth moving the `expiresAfter` as well as `isWirelessUser` into ZMBareUser.

mikeger: True, we get the search users only form the search, and temporary users are not (hopefully) included there, so my thinking is we can implement it in `SearchUser` as always returning `false`.

mikeger:
@daehn I think UserCellTests failed

PR #1869, opened 2018-03-15
Make ToggleView themeable

typfel:
### Issues

The toggle in the group creation flow was white in dark mode

### Causes

Colors are hard coded

### Solutions

Implement the Themeable protocol for the `ToggleView`


daehn:
Test might have to be adjusted I think.

typfel:
Don't think there's any tests for this controller 😱

PR #1873, opened 2018-03-16
Move conversation options to native action sheets

daehn:
## What's new in this PR?

With the redesigned group participant view we started using native UIAlertController action sheet to show conversation options (like leave / mute / archive). This PR converts most appearances of the custom `ActionSheetController` to use native controls as well (the only remaining appearance so far is the conversation degraded dialog).


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -878,57 +877,49 @@ - (void)conversationDidChange:(ConversationChangeInfo *)note
...
+                    [self presentViewController:participantsController animated:YES completion:nil];
+                }
+                break;
+        }
+    }];
+
+    [self presentViewController:controller animated:YES completion:nil];
 }
 
 - (void)degradedConversationDismissed:(id)sender
```
typfel: I guess this is never called anymore after deleting line: https://github.com/wireapp/wire-ios/pull/1873/files#diff-d0adf7620aab7082451935f800968beaL923

daehn:
@mikeger @typfel I will address the presentation bug I found while working on this in a separate PR.

PR #1876, opened 2018-03-16
Remove magic

typfel:
## What's new in this PR?

Magic was a library for configuring views using yaml files which in theory could be changed by designers. We don't use it actively anymore and should therefore remove it completely.

## Notes

Some fonts changed slightly because was were configuring fonts in a non-standard way. 

## Todo

- [x] design review


Wire-iOS/Sources/UserInterface/Helpers/UIFont+FontSpec.swift
```diff
@@ -0,0 +1,76 @@
...
+import Foundation
+
+// Objective-C compatiblity layer for the Swift only FontSpec
+@objc
+extension UIFont {
+    
+    // MARK: - Small
+    
+    class var smallLightFont: UIFont {
+        return FontSpec.init(.small, .light).font!
```
daehn: I think the `.init` can be dropped in all these cases.

mikeger:
@typfel small conflict and we are ready to merge 🎊

PR #1884, opened 2018-03-19
ZIOS-9751 fix/crash when calling cellForRow

billypchan:
### Issues

App crashes sometimes with calling cellForRow after scrollToRow is called.

### Causes

UITableView.cellForRow returns nil if cell is not visible or index path is out of range. Delay for 0.1 sec does not promise that the cell is visible after scrolling without animation.

### Solutions

Unwrap cell before calling completion closure.
Call scrollToRow inside a UIView.animate's closure with duration = 0.0
Call completion closure in UIView.animate's completion closure, instead of hard code delay.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
@@ -222,10 +222,13 @@ extension ConversationContentViewController {
     
     func scroll(toIndex indexToShow: Int, completion: ((ConversationCell)->())? = .none) {
         let cellIndexPath = IndexPath(row: indexToShow, section: 0)
-        self.tableView.scrollToRow(at: cellIndexPath, at: .middle, animated: false)
-        
-        delay(0.1) {
-            completion?(self.tableView.cellForRow(at: cellIndexPath) as! ConversationCell)
-        }
+
+        UIView.animate(withDuration: 0.0, animations: {
+            self.tableView.scrollToRow(at: cellIndexPath, at: .middle, animated: false)
+        }, completion: { (didComplete) in
+            if let cell = self.tableView.cellForRow(at: cellIndexPath) {
+                completion?(cell as! ConversationCell)
+            }
+        })
     }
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -222,10 +222,13 @@ extension ConversationContentViewController {
     
     func scroll(toIndex indexToShow: Int, completion: ((ConversationCell)->())? = .none) {
         let cellIndexPath = IndexPath(row: indexToShow, section: 0)
-        self.tableView.scrollToRow(at: cellIndexPath, at: .middle, animated: false)
-        
-        delay(0.1) {
-            completion?(self.tableView.cellForRow(at: cellIndexPath) as! ConversationCell)
-        }
+
+        UIView.animate(withDuration: 0.0, animations: {
```
daehn: @billypchan I think `cellForRowAtIndexPath:` should never return nil in case there is a cell for this index path, even though if it's not visible at the moment (because in that case it is the responsibility of the method to dequeue and configure a cell). To me it looks like the root of the issue is that when the block is executed the data source might have changed, effectively invalidating the index path that is being used, so we can no longer guarantee that there is still a cell at this point, but I am pretty certain there is one when the method is called initially.

I think wrapping this in an `animate`  block does the same as performing a run-loop tick. I am still not sure why there was a delay in the first place or why performing a run-loop tick is necessary here, but maybe you can elaborate.

billypchan: @daehn In Apple's doc, it said "An object representing a cell of the table, or nil if the cell is not visible or indexPath is out of range."
https://developer.apple.com/documentation/uikit/uitableview/1614983-cellforrowatindexpath?language=objc

I would agree that put the scrollToRow inside the closure is not necessary since we don't need animation for scrolling. The crash occurs about one to two times a day and I can not reproduce it as I did a few days ago. I will keep the unwrap part and remove the animation code.

PR #1886, opened 2018-03-20
ZIOS-7949/ZIOS-9729: Fix call survey presentation

alexaubry:
### Issues

After merging the automation fix in #1851, the call survey no longer appeared after a call for all users.

### Causes

The criteria used to detect if the app was running in automation were not correct and would return `false` for every case.

### Solutions

We now only check if analytics are disabled since automation tests will be included. This also fixes layout issues on bigger screens.


mikeger:
@alexaubry after the data model bump the tests target cannot compile, I will fix it on dev so you can check your branch.

PR #1887, opened 2018-03-20
Updated "de|it|ja|pt-BR|es-ES|ru|uk|tr|nl|da|fr|zh-CN|sl|ar|et|fi|pl|…

billypchan:
Updated strings for release 3.9.


Wire-iOS/Resources/de.lproj/Localizable.strings
```diff
@@ -314,6 +318,9 @@
 "content.system.conversation.team.member-leave" = "%@ wurde aus dem Team entfernt.";
 "content.system.conversation.guest.joined" = "%@ ist beigetreten";
 
+"content.system.conversation.invite.title" = "People outside your team can join this conversation.";
```
mikeger: @daehn would be cool to translate this stuff as well.

billypchan: I will start another localization update when German translation reaches 100%.

PR #1893, opened 2018-03-21
ZIOS-9778: Fix crash on video player dismissal

alexaubry:
### Issues

On iOS 9 and 10, dismissing a video player caused the app to crash.

### Causes

This crash was caused because KVO observers of the player controller were not removed at the time of `deinit`. The crash was not observed on iOS 11 because observers are removed automatically from this version.

### Solutions

This introduces a protocol used to synchronously remove the observers before nil-ing out the player controller at dismissal time.

### DIFF
--- a/Wire-iOS/Sources/Managers/MediaPlaybackManager.h
+++ b/Wire-iOS/Sources/Managers/MediaPlaybackManager.h
@@ -31,11 +31,11 @@ FOUNDATION_EXPORT NSString *const MediaPlaybackManagerPlayerStateChangedNotifica
 /// An object that observes changes in the media playback manager.
 @protocol MediaPlaybackManagerChangeObserver
 
-/// A media player did become active.
-- (void)mediaPlaybackManager:(MediaPlaybackManager *)mediaPlaybackManager didStartMediaPlayer:(id<MediaPlayer>)mediaPlayer;
+/// The title of the active media player changed.
+- (void)activeMediaPlayerTitleDidChange;
 
-/// The media player did become inactive.
-- (void)mediaPlaybackManager:(MediaPlaybackManager *)mediaPlaybackManager didRemoveMediaPlayer:(id<MediaPlayer>)mediaPlayer;
+/// The state of the active media player changes.
+- (void)activeMediaPlayerStateDidChange;
 
 @end
 
--- a/Wire-iOS/Sources/Managers/MediaPlaybackManager.h
+++ b/Wire-iOS/Sources/Managers/MediaPlaybackManager.h
@@ -24,14 +24,27 @@
 
 
 @class AudioTrackPlayer;
+@class MediaPlaybackManager;
 
 FOUNDATION_EXPORT NSString *const MediaPlaybackManagerPlayerStateChangedNotification;
+
+/// An object that observes changes in the media playback manager.
+@protocol MediaPlaybackManagerChangeObserver
+
+/// A media player did become active.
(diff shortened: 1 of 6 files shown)


billypchan:
@alexaubry I had fixed same issue before( https://github.com/wireapp/wire-ios/pull/1529), but the code seems overwritten later. It is good to see another solution.

Wire-iOS/Sources/UserInterface/Conversation/Content/MessagePresenter.m
```diff
@@ -43,7 +43,7 @@ - (BOOL)prefersStatusBarHidden
 
 - (void)viewWillDisappear:(BOOL)animated {
     [super viewWillDisappear:animated];
-    
+    [self.wr_playerController tearDown];
```
mikeger: I think the necessity of calling `tearDown` on `MediaPlayerController` makes it's API less intuitive. Also now the user of this class must think when the media player should become unnecessary and can be torn down. It would be way easier to just rely on dealloc (deinit) to understand when the player can be torn down.

Wire-iOS/Sources/Managers/MediaPlaybackManager.h
```diff
@@ -24,14 +24,27 @@
...
 @class AudioTrackPlayer;
+@class MediaPlaybackManager;
 
 FOUNDATION_EXPORT NSString *const MediaPlaybackManagerPlayerStateChangedNotification;
 
+/// An object that observes changes in the media playback manager.
+@protocol MediaPlaybackManagerChangeObserver
+
+/// A media player did become active.
+- (void)mediaPlaybackManager:(MediaPlaybackManager *)mediaPlaybackManager didStartMediaPlayer:(id<MediaPlayer>)mediaPlayer;
```
mikeger: As I understood, the root of the problem is that the KVO observer is still alive when the object is already deallocated?

I think the aim of KVO on `MediaBarViewController` is to detect the changes in the media player, precisely in `activeMediaPlayer.state` and `activeMediaPlayer.title`. So it is attached to both, the `activeMediaPlayer`, and to the `MediaPlayerController`. This makes it problematic to deallocate the observers before the objects. As I understand it, both the `activeMediaPlayer` object instance and the player's state can change. So my proposal can be the following:

1. Remove KVO in `MediaBarViewController`.
2. Add KVO (if not added already) in `MediaPlayerController`.
3. Observe the changes in both player instance and instance state via the delegate.

-> What we gain: we hide the KVO from `MediaPlayerController` user, so we can control better the lifetime of the observers and objects.

Wire-iOS/Sources/Managers/MediaPlaybackManager.m
```diff
@@ -156,4 +164,32 @@ - (void)mediaPlayer:(id<MediaPlayer>)mediaPlayer didChangeToState:(MediaPlayerSt
...
+    [((NSObject *)mediaPlayer) removeObserver:self.titleObserver forKeyPath:@"title"];
+    self.titleObserver = nil;
+}
+
+- (void)activeMediaPlayerTitleChanged:(NSDictionary *)change
+{
+    [self.changeObserver activeMediaPlayerTitleDidChange];
+}
+
+- (void)activeMediaPlayerStateChanged:(NSDictionary *)change
```
typfel: Is this method ever called, there's no KVO on this or am I missing something?

alexaubry: Oh right I forgot to remove it. We're not using KVO for the state since we already have `mediaPlayer:didChangeToState:`.

PR #1894, opened 2018-03-21
Feature/zios 9780 network status bar ui adjustment

billypchan:
## What's new in this PR?

This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1890.

UI changes in this PR:
- network bar needs bottom margins 8pt but only in list ui
- network bar needs top margins 8pt for all non iPhone X devices
- collapsed bar should be 4pt high instead of 2pt
- pulse animation of collapsed bar should be a bit slower and should animate from 64% opacity to 100% opacity (change animation curve)

code clean up: 
- remove legacy offlineCollapsed state and related methods


Wire-iOS Tests/NetworkStatusView/NetworkStatusViewControllerTests.swift
```diff
@@ -155,42 +159,6 @@ final class NetworkStatusViewControllerTests: XCTestCase {
                                        rootState: .offlineExpanded)
     }
 
-    func testThatNotifyWhenOfflineShowsOneNetworkStatusViewOnIPad() {
```
mikeger: Please remove the corresponding PNGs from the repo. (also `testThatNotifyWhenOfflineShowsBothNetworkStatusViewOnIPhone`).

billypchan: The snapshot tests are moved to NetworkStatusViewControllerSnapshotTests. Checked and confirmed that no old PNGs are still there.

PR #1902, opened 2018-03-22
Allow mark read unread only on internal/dev versions

mikeger:
The feature needs the cross-platform support, so must be hidden from the external people for the moment.


Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Actions.m
```diff
@@ -103,10 +104,10 @@ - (NSOrderedSet *)availableActionsForConversation
     }
 
     NSArray *unreadMessages = self.unreadMessages;
-    if (unreadMessages.count > 0) {
+    if (DeveloperMenuState.developerMenuEnabled && unreadMessages.count > 0) {
```
mikeger: @daehn BTW I think this file is not used any more after the menu refactroing, can you confirm?

daehn: Yeah shouldn't be needed anymore. Good catch!

daehn: Thi logic moved to `ConversationActions.swift`.

PR #1906, opened 2018-03-22
Added tracking of the added_team_name property

nicorsm:
### Issues

Title says it all :)


Wire-iOS/Sources/Analytics/Registration/AnalyticsTracker+Registration.swift
```diff
@@ -40,6 +40,10 @@ extension AnalyticsTracker {
         self.tagEvent("team.verified")
     }
     
+    func tagTeamCreationAddedTeamName() {
+        self.tagEvent("team.added_team_name")
```
mikeger: I think currently the events must be whitelisted in `AnalyticsMixpanelProvider`

nicorsm: Right 👍

PR #1909, opened 2018-03-22
Fix misaligned account selector outline

typfel:
### Issues

account outline

### Causes

Size changed from 46pt to 44pt since it now lives in the navigation bar

### Solutions

Update constant so size adds up to 44pt


mikeger:
@typfel no screenshot tests?

typfel:
@mikeger maybe :) I'll check

PR #1912, opened 2018-03-22
Declassify [2]

typfel:
## What's new in this PR?

Remove more usage of classy


typfel:
So the snapshots broke because of the way we export fonts to classy. The change comes from the fact that iOS switches between two fonts (Pro Text / Display) when creating a system font bigger than 20pt. That doesn't happen for the classy fonts.

daehn:
@typfel to me it looks like the difference between those fonts is mostly their kerning? I don't think it's a big deal I this case though, but maybe worth checking with design.

PR #1914, opened 2018-03-23
Fix: [iPad]Portrait split mode - conversation list view not dismiss after another app added to screen

billypchan:
## What's new in this PR?

This PR is follow up and bug fix of https://github.com/wireapp/wire-ios/pull/1903

### Issues

After the previous fix, swipe gesture to open list screen does not work.

### Causes

openPercentage is reset when viewWillLayoutSubview is called.

### Solutions

Instead of reset openPercentage after size change, only update leftViewOffsetConstraint.constant and rightViewOffsetConstraint.constant.

### Notes

a test will be written to cover "swipe to open list view" functionalities.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
@@ -21,6 +21,7 @@
 
 
 #import "SplitViewController.h"
+#import "SplitViewController+internal.h"
 #import "CrossfadeTransition.h"
 #import "SwizzleTransition.h"
 #import "VerticalTransition.h"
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -263,7 +261,7 @@ - (void)updateForSize:(CGSize)size
     self.futureTraitCollection = nil;
 
     // update right view constraits after size changes
-    [self resetOpenPercentage];
+    [self updateRightAndLeftEdgoConstants: self.openPercentage];
 }
```
billypchan: bug fix for swipe gesture not work issue

PR #1915, opened 2018-03-23
ZIOS-9786: Fix share screen visuals

nicorsm:
### Issues

Inside the `ShareDestinationCell` of `ShareViewController`, the shield icon was displayed after the selection icon, instead of before.

abbc3fbf-c846-471e-ade4-4093d1c42cf0

### Causes

This is caused by the current implementation of the shield icon, which is considered the `accessoryView` of the cell.

### Solutions

I'm using a custom `UIImageView`. I've also added tests and adjusted the top inset of the view.


Wire-iOS/Resources/Images/icon-no-appstore.xcassets/AppIcon-Debug.appiconset/Contents.json
```diff
@@ -133,6 +133,11 @@
       "idiom" : "ipad",
       "filename" : "83.5@2x.png",
       "scale" : "2x"
+    },
+    {
+      "idiom" : "ios-marketing",
```
typfel: These changes doesn't seem relevant for the PR.

PR #1918, opened 2018-03-26
Removal of classy [3]

typfel:
## What's new in this PR?

More removal of Classy usage in views


mikeger:
@typfel conflict 🔥

typfel:
@mikeger the broken snapshots were either minimal font differences or invisible pixel changes.

PR #1919, opened 2018-03-26
Fix - List UI turns black during transition to conversation view

billypchan:
### Issues

list ui turns black during the transition to conversation view

### Causes

The leftView was set to hidden immediately in the line:
self.leftView.hidden = self.openPercentage == 0;

### Solutions

Do not call updateLeftViewVisibility every time in updateForSize, which is called by viewWillLayoutSubviews.

for iPad, call updateLeftViewVisibility after orientation changes, to hide the left view when rotates to portrait mode, and shows the left view when rotates to landscape mode.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
@@ -174,11 +174,6 @@ NSString *SplitLayoutObservableDidChangeToLayoutSizeNotification = @"SplitLayout
 
 @implementation SplitViewController
 
-- (void)dealloc
-{
-    [[NSNotificationCenter defaultCenter] removeObserver:self];
-}
-
 - (void)viewDidLoad 
 {
     [super viewDidLoad];
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
@@ -197,8 +197,6 @@ NSString *SplitLayoutObservableDidChangeToLayoutSizeNotification = @"SplitLayout
     self.horizontalPanner = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(onHorizontalPan:)];
     self.horizontalPanner.delegate = self;
     [self.view addGestureRecognizer:self.horizontalPanner];
-
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(updateLayoutSizeAndLeftViewVisibility) name:UIApplicationDidChangeStatusBarFrameNotification object:nil];
 }
 
 - (void)viewWillLayoutSubviews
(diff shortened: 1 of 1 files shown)


billypchan:
@mikeger I rewrite this PR without affecting hide property after the right view is revealed.

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -174,6 +174,11 @@ @interface SplitViewController () <UIGestureRecognizerDelegate>
 
 @implementation SplitViewController
 
+- (void)dealloc
+{
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
```
mikeger: This is not necessary since iOS 9.

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -197,6 +202,8 @@ - (void)viewDidLoad
     self.horizontalPanner = [[UIPanGestureRecognizer alloc] initWithTarget:self action:@selector(onHorizontalPan:)];
     self.horizontalPanner.delegate = self;
     [self.view addGestureRecognizer:self.horizontalPanner];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(updateLayoutSizeAndLeftViewVisibility) name:UIApplicationDidChangeStatusBarFrameNotification object:nil];
```
mikeger: May be it's sufficient to observe the trait collection change?

billypchan: This is for UI orientation change, in the viewWillTransitionToSize method, it is no clue that the new orientation is landscape or portrait.

mikeger: If you check `updateLayoutSizeForTraitCollection:size:` method that you are calling, it's checking the `IS_IPAD_PORTRAIT_LAYOUT` macro, so it does not depend on the status bar frame.
I think I am missing the logic here to understand why updating from `viewWillTransitionToSize:withTransitionCoordinator:` is not enough and why the additional observation is needed. I expect that if the view size is observed correctly there is no need for the hacks to observe the status bar frame, which seems irrelevant to this controller.

billypchan: Actually IS_IPAD_PORTRAIT_LAYOUT needs the current statusBarOrientation:

#define IS_IPAD_PORTRAIT_LAYOUT (IS_IPAD_FULLSCREEN && UIInterfaceOrientationIsPortrait([UIApplication sharedApplication].statusBarOrientation))

UIContentContainer protocol lacks a method "viewDidTransitionToSize", we just know the new size in viewWillTransitionToSize:withTransitionCoordinator:, but not the statusBarOrientation.

mikeger: @billypchan have you tried the transition coordinator approach that we discussed the other day?

billypchan: @mikeger I rewrote and tested. Thanks!

PR #1920, opened 2018-03-26
Authentication flow changes for backup

vytis:
### Issues

We need to add a screen right after login that would let users choose to import their backup. This slightly duplicates the functionality of the screen that shows that you will not have any history, we have decided to remove it from the flow altogether.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/719


mikeger:
@vytis can this be merged now?

PR #1921, opened 2018-03-26
ZIOS-9797: Fix audio player crash when making an outgoing call

alexaubry:
### Issues

When making an audio call as an audio message is playing, the app crashes.

### Causes

This happens because the observer for `title` is removed twice.

The workaround that caused this issue was added because the observer was removed from inside the `deinit` of the object. As this causes the object to be removed from the `KeyValueObserver` before we can remove it, we needed to explicitly call `removeObserver`.

### Solutions

We fix this by tearing down the observer outside of `deinit`.

### DIFF
--- a/WireExtensionComponents/Utilities/KeyValueObserver.m
+++ b/WireExtensionComponents/Utilities/KeyValueObserver.m
@@ -22,7 +22,7 @@
 
 
 @interface KeyValueObserver ()
-@property (nonatomic, retain) id observedObject;
+@property (nonatomic, weak) id observedObject;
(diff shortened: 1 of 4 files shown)


WireExtensionComponents/Utilities/KeyValueObserver.m
```diff
@@ -22,7 +22,7 @@
 
 
 @interface KeyValueObserver ()
-@property (nonatomic, weak) id observedObject;
+@property (nonatomic, retain) id observedObject;
```
typfel: I think now you will have retain cycle and the `KeyValueObserver` will never deallocate.

PR #1923, opened 2018-03-26
Fix not setting textColor for centerStatusLabel

typfel:
### Issues

"Video not available" message was not visible in the call UI

### Causes

text color was not set

### Solutions

Set text color


daehn:
@typfel no snapshots that need to be updated?

typfel:
@daehn yep

PR #1925, opened 2018-03-26
ZIOS-9771: enable opt-in and out analytics data

nicorsm:
## What's new in this PR?

Mixpanel was not tracking the opt-in and opt-out actions because:
- event was ignored in `Analytics+Settings`
- `settings.changed_value` event was not whitelisted


mikeger:
@nicorsm have you checked with analytics people if it's fine to track the settings changes? Also I think we expect in Mixpanel to have `settings.opted_out_tracking` event, not the settings change. Since all platforms are sending the events to one logical Mixpanel events store, it's important to send the exact events as other platforms do.

nicorsm:
I'm closing this PR since the correct event was still there and whitelisted. Seems to already work fine from our side, need to investigate why mixpanel is not tracking the opt-in event.

PR #1928, opened 2018-03-27
Feature/zios 9800 add a delay before network status bar shows sync bar

billypchan:
### Issues

Sync bar shows every time when the app starts. We want to have a show it after some delay when the network state changes to loading.

### Causes

In viewDidLoad and viewWillAppear, the network state applies immediately, not delayed. Other state change are delayed before showing on the UI. 

### Solutions

Enqueue the initial network state with a delay. The later network state change will override the init state.

## Notes

This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1894.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusView.swift
@@ -151,11 +151,7 @@ class NetworkStatusView: UIView {
 
     private lazy var topMargin = UIScreen.hasNotch ? CGFloat(0) : CGFloat.NetworkStatusBar.topMargin
 
-    public weak var delegate: NetworkStatusViewDelegate! {
-        didSet {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusView.swift
```diff
@@ -139,11 +139,7 @@ class NetworkStatusView: UIView {
 
     private lazy var topMargin = UIScreen.hasNotch ? CGFloat(0) : CGFloat.NetworkStatusBar.topMargin
 
-    public weak var delegate: NetworkStatusViewDelegate! {
-        didSet {
-            createConstraints()
-        }
-    }
+    public weak var delegate: NetworkStatusViewDelegate!
```
mikeger: May be it's worth having the weak var as optional (?), I think this is the default behavior for the delegate properties.

PR #1929, opened 2018-03-27
Back up UI

mikeger:
- Basic screen that gives the possibility to save the back up for the user.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -842,8 +842,8 @@
 // History backup
 
 "self.settings.conversations.title" = "Conversations";
-"self.settings.history_backup.title" = "Histroy backup";
-"self.settings.history_backup.description" = "Back up your conversation history to a safe place. If you lose your phone or switch to a new one, your conversation will be safe. You can restore your conversation histroy when you log in. You can only restore history from a backup of the same platform.\nThe back up is not protected by Wire end-to-end encryption.";
+"self.settings.history_backup.title" = "History backup";
+"self.settings.history_backup.description" = "Back up your conversation history to a safe place. If you lose your phone or switch to a new one, your conversation will be safe. You can restore your conversation history when you log in. You can only restore history from a backup of the same platform.\nThe back up is not protected by Wire end-to-end encryption.";
 "self.settings.history_backup.action" = "Back Up Now";
 "self.settings.history_backup.error.title" = "Error";
 
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -839,6 +839,14 @@
 // Invite from Settings
 "self.settings.invite_friends.title" = "Invite people";
 
+// History backup
+
+"self.settings.conversations.title" = "Conversations";
+"self.settings.history_backup.title" = "Histroy backup";
+"self.settings.history_backup.description" = "Back up your conversation history to a safe place. If you lose your phone or switch to a new one, your conversation will be safe. You can restore your conversation histroy when you log in. You can only restore history from a backup of the same platform.\nThe back up is not protected by Wire end-to-end encryption.";
+"self.settings.history_backup.action" = "Back Up Now";
+"self.settings.history_backup.error.title" = "Error";
+
 // +++++++++++++++++++++++++++++++++++++++++++++++++++++
 // System status
 "system_status_bar.no_internet.title" = "No Internet";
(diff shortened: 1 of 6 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -839,6 +839,14 @@
 // Invite from Settings
 "self.settings.invite_friends.title" = "Invite people";
 
+// History backup
+
+"self.settings.conversations.title" = "Conversations";
+"self.settings.history_backup.title" = "Histroy backup";
```
typfel: typo: `Histroy` > `History`

Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+        tableView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(tableView)
+        tableView.delegate = self
+        tableView.dataSource = self
+        
+        // this is necessary to remove the placeholder cells
+        tableView.tableFooterView = UIView()
+        cells = [BackupStatusCell.self, BackupActionCell.self]
+        
+        Set(cells.map { HashableType<UITableViewCell>($0) }).forEach {
```
vytis: Is it worth adding all this machinery for just registering two kinds of cells?

vytis: or maybe it will be/is used somewhere else as well?

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -0,0 +1,51 @@
...
+            return UIView.userInterfaceLayoutDirection(for: .unspecified)
+        }
+    }
+    
+    @discardableResult func fitInSuperview(with insets: UIEdgeInsets = .zero) -> [NSLayoutConstraint] {
+        guard let superview = self.superview else {
+            fatal("Not in view hierarchy: self.superview = nil")
+        }
+        
+        let (leadingInset, trailingInset) = self.layoutDirection == .leftToRight ?
```
typfel: Are you sure you want flip the insets when you are using leading and trailing?

mikeger: I think in this case yes, since the insets in `UIEdgeInsets` are left and right, if I would not swap them then on RTL left would be on the right. I guess from the developer perspective the name `left` should be considered as `leading`.

Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
mikeger: In this case yes, it's possible to just enumerate cells since they are always distinct. My idea is that it can be reused in the future.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
typfel: I though the constant was attached to the leading/trailing constraint. So that if you have leading inset 64 and trailing inset 16 it would on a RTL setup move the 64 inset to the right side.

mikeger: Okay, may be I don't need to flip it, makes it confusing.

typfel: Ah or you mean that since the insets named`left` and `right` they should counteract the leading and trailing behaviour? But then you also use `left` and `right` constraints instead of `leading` and `trailing`?

Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
vytis: But there is no problem to just do `.forEach` with the `cells` array? Current approach prevents us from registering the same type multiple times, but since `reuseIdentifier` is a static var it won't make a difference I think.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
mikeger: I think I would refactor it not to use the edge insets, it's not useful to have left and right naming.

Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
mikeger: I agree that it does not hurt to re-register the registered cell, it should still work. However, I still think it's better only to register the cell once, since there is no information on how the table view implementation works.

vytis: Sure it's better to be safe and make sure we register only once, but in this case IMHO it's a bit too much since the array is static and there is no risk to have something weird happening. Anyway I'm good to keep the code here also, not a blocker 👯‍♂️

mikeger:
@vytis @typfel updated!

mikeger:
@vytis I would add the tests when we would know how it should look like. Otherwise every UI change would break them.

PR #1935, opened 2018-03-28
ZIOS-8802: Fix music pausing when removing someone from a group

alexaubry:
### Issues

When removing someone from a group and music was playing, the music was paused for a few seconds.

### Causes

The alert sound played before confirming the removal of the participant required the music to be paused. 

### Solutions

This was fixed by allowing the alert sound to be mixed with other currently playing sounds.


mikeger:
Easy... if you know where to look :)

PR #1938, opened 2018-03-28
Fix/zios 9807 giphy search view controller crash

billypchan:
### Issues

App crashes when collectionView accesses an out-of-bound array item. The app may crash when the user fast typing in the search field and scroll collection view.

### Causes

The collection view's data source (searchResultsController.results) is wiped when a new search starts.  Data source is empty for a short time before results arrive. If the collection calls cellForItemAt() or didSelectItemAt() when the data source is empty, the app crashes.

### Solutions

ZiphySearchResultsController provides a delegate method didCleanResults() to let the UI knows refesh the UI and data source is wiped.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-ziphy/pull/12


billypchan:
@mikeger both ziphy and wire-ios are merged with develop.

PR #1944, opened 2018-03-29
Fix: List UI turns black during transition to conversation view

billypchan:
## What's new in this PR?

This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1919.

### Issues

list UI turns black during the transition to conversation view if the user chooses another conversation quickly (<0.55 sec).

### Causes

If second animation starts before the first animation ends, the first animation's completion closure hides left view before second animation finishes.

### Solutions

Check leftview is in the middle of the animation before hiding the leftview.

### Note

setLeftViewControllerRevealed() is translated to Swift for easier debugging.


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+leftView.swift
```diff
@@ -0,0 +1,59 @@
...
+                    completion()
+                }
+
+                if self.layoutSize != .regularLandscape {
+                    self.leftViewController?.endAppearanceTransition()
+                    self.rightViewController?.endAppearanceTransition()
+                }
+                if self.openPercentage == 0 &&
+                   self.layoutSize != .regularLandscape &&
+                   self.leftView.layer.presentation()?.frame == self.leftView.frame {
```
billypchan: Bug fixed by added checking for animation is finished or not.

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+leftView.swift
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        let completionBlock: () -> Void = {
+            if let completion = completion {
+                completion()
+            }
+
+            if self.openPercentage == 0 &&
+                self.layoutSize != .regularLandscape &&
+                (self.leftView.layer.presentation()?.frame == self.leftView.frame || (self.leftView.layer.presentation()?.frame == nil && !animated)) {
```
billypchan: Bug fixed by added checking for animation is finished or not.

PR #1945, opened 2018-03-29
ZIOS-9812: Update and unify tab bar design

alexaubry:
### Issues

Two tab bar designs coexisted inside the app: the "line" (start UI), and the "arrow" (everywhere else).

### Solutions

The `TabBar` object was updated to use the new line design. It can be seen in:

- Participant details view
- Registration/login flow
- Add participant view
- Start UI

The tab bar was also refactored to use stack views, which reduces code duplication with the `RadioButtonsView` class.

The contacts/Start UI search controllers also use the tab bar now, instead of the custom buttons that were previously used.


Wire-iOS/Sources/UserInterface/Components/TabBar/Tab.swift
```diff
@@ -0,0 +1,80 @@
...
+        addSubview(selectionLineView)
+        constrain(self, selectionLineView) { selfView, selectionLineView in
+            selectionLineView.leading == selfView.leading + 16
+            selectionLineView.trailing == selfView.trailing - 16
+            selectionLineView.height == 1
+            selectionLineView.bottom == selfView.bottom
+        }
+
+        titleLabel?.font = FontSpec(.small, .semibold).font
+        isSelected = false
```
mikeger: Is the UIButton selected per default?

alexaubry: I don't think it is actually. I just copied this init method from `SearchGroupButton`which used to set that to false.

PR #1949, opened 2018-03-29
ZIOS-9804: Added back button in the email verification view

nicorsm:
## What's new in this PR?

Added back button on the email verification view, so now users can go back to the previous view when they provide a wrong email address. Added tests.


vytis:
I think this is the step where we keep trying to login with this email every 5 mins. We need to stop this process somehow, but probably there is a method to reset the authentication state

Wire-iOS/WireBridgingHeader.h
```diff
@@ -96,6 +96,7 @@
 #import "ConfirmAssetViewController.h"
 #import "ProfileSelfPictureViewController.h"
 #import "AddEmailPasswordViewController.h"
+#import "EmailVerificationStepViewController.h"
```
mikeger: You sure this is necessary?

nicorsm: I need it just for test purposes, will move to the proper tests bridging header.

vytis:
I think stopping the timer should be possible by resetting the authentication status somehow - clearing credentials or firing a notification. You should check if there is a situation already where we quit the authentication flow an see what is done there.

nicorsm:
@vytis just found that we're already calling `[[UnauthenticatedSession sharedSession] cancelWaitForEmailVerification]` in `viewWillDisappear`. Need to check the appearing/disappearing of the back button since it's not working on the registration flow.

PR #1950, opened 2018-03-29
Fix crash: when presenting video editor on iPad

billypchan:
### Issues

App crashes when showing video editor on iPad.

### Causes

"Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'On iPad, Wire.StatusBarVideoEditorController must be presented via UIPopoverController'"

### Solutions

Present videoEditor in popover style on iPad. 

###Note 
This PR fixed a run time warning for dismiss spinner in non-main thread.


mikeger:
@billypchan please check if you may be changed some GIT config, since it does not recognize your commits as yours any more.

PR #1952, opened 2018-03-29
Added checkmark view, other fixes

mikeger:
- Do not show alert when not necessary.
- Erase leftover backups.


mikeger:
Yeah it‘s all autogenerated :)

Sent from a Mobile Device

PR #1955, opened 2018-04-03
ZIOS-9808: Fix visual issues on landing page

alexaubry:
### Issues

The cancel button on the landing screen was too big and mispositioned; which caused a visual glitch in the transition from the self profile view controller.

### Causes

The cancel button was sized and positioned manually. The close button in the self profile is managed by a navigation bar, and thus has a different position on every device (adapted by the `UINavigationBar` depending on the size of the screen).

Because the two buttons do not have the same size, and a slightly different position, this caused the animation between the two to be "off".

### Solutions

We now use a standard navigation bar to size and position the cancel button on the landing page.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
@@ -144,7 +144,7 @@ final class LandingViewController: UIViewController {
         let button = IconButton()
         button.translatesAutoresizingMaskIntoConstraints = false
         button.accessibilityIdentifier = "CancelButton"
-        button.setIcon(.cancel, with: .small, for: .normal)
+        button.setIcon(.cancel, with: .tiny, for: .normal)
         button.addTarget(self, action: #selector(LandingViewController.cancelButtonTapped(_:)), for: .touchUpInside)
 
         return button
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -168,31 +165,31 @@ final class LandingViewController: UIViewController {
...
 
-        self.view.backgroundColor = UIColor.Team.background
-        cancelButton.setIconColor(.black, for: .normal)
-        cancelButton.setIconColor(UIColor.black.withAlphaComponent(0.6), for: .highlighted)
-
         [headerContainerView, buttonStackView, loginHintsLabel, loginButton].forEach(view.addSubview)
 
         [logoView, headline].forEach(headlineStackView.addArrangedSubview)
-        [headlineStackView, cancelButton].forEach(headerContainerView.addSubview)
+        [headlineStackView].forEach(headerContainerView.addSubview)
```
daehn: I guess the `forEach` can be removed now.

alexaubry: Removed it, thanks for pointing out.

PR #1957, opened 2018-04-03
Back up to tmp on simulator

mikeger:
Store the back up on simulator to the `tmp` folder to allow automation to fetch it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
@@ -190,11 +190,6 @@ extension BackupViewController: UITableViewDataSource, UITableViewDelegate {
                 #if arch(i386) || arch(x86_64)
                     let tmpURL = URL(fileURLWithPath: "/var/tmp/").appendingPathComponent(url.lastPathComponent)
                     try! FileManager.default.moveItem(at: url, to: tmpURL)
-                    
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Settings/BackupViewController.swift
```diff
@@ -187,11 +187,21 @@ extension BackupViewController: UITableViewDataSource, UITableViewDelegate {
                 self.present(alert, animated: true)
                 BackupEvent.exportFailed.track()
             case .success(let url):
-                let activityController = UIActivityViewController(activityItems: [url], applicationActivities: nil)
-                activityController.completionWithItemsHandler = { _, _, _, _ in
-                    SessionManager.clearPreviousBackups()
-                }
-                self.present(activityController, animated: true)
+                #if arch(i386) || arch(x86_64)
+                    let tmpURL = URL(fileURLWithPath: "/var/tmp/").appendingPathComponent(url.lastPathComponent)
```
mykola-mokhnach: how `url.lastPathComponent` is constructed? If we cannot get value easily in autpmated script then it would be better to hardcode it

mikeger: Name would be like `Wire-Backup_YYYYMMDD.ios_wbu`.

PR #1959, opened 2018-04-03
Feature/zios 9829 network status bar animation finetuning

billypchan:
## What's new in this PR?

Rewrite the network status bar animation for better appearance.

UI animation properties changed:
1. sync bar animates the size to same as network status bar
2. sync bar animates the alpha value from 0 to 1
3. sync bar animates the corner radius to same as network status bar
4. sync bar top margin should match the network status bar top margin.
5. network status bar's text label should have a fade in/out animation when it appears/dismisses.

The state variables of classes OfflineBar and BreathLoadingBar are replaced with NetworkStatusViewState, which matches 3 states of NetworkStatusViewController.


Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
```diff
@@ -40,6 +43,44 @@ class BreathLoadingBar: UIView {
         }
     }
 
+    private var _state: NetworkStatusViewState = .online
+
+    var state: NetworkStatusViewState {
+        set {
+            update(state: newValue, animated: false)
```
vytis: No need to have another private variable for these things, you can simply use `didSet{}` in there to call `update` method

Wire-iOS/Sources/UserInterface/Components/NetworkStatus/OfflineBar.swift
```diff
@@ -0,0 +1,98 @@
...
+        constrain(self, offlineLabel) { containerView, offlineLabel in
+            offlineLabel.center == containerView.center
+            offlineLabel.left >= containerView.leftMargin
+            offlineLabel.right <= containerView.rightMargin
+
+            heightConstraint = containerView.height == 0
+        }
+    }
+
+    private func updateViews(animated: Bool = true) {
```
vytis: The code for `OfflineBar` and `BreathLoadingBar` seems quite similar - maybe there is a chance to unify the views together by adding something like a `BarKind` enum?

Wire-iOS/Sources/UserInterface/Components/NetworkStatus/OfflineBar.swift
```diff
@@ -0,0 +1,98 @@
...
+
+class OfflineBar: UIView {
+
+    private let offlineLabel: UILabel
+    private var heightConstraint: NSLayoutConstraint?
+    private var _state: NetworkStatusViewState = .online
+
+    var state: NetworkStatusViewState {
+        set {
+            update(state: newValue, animated: false)
```
vytis: Same issue here

Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
billypchan: It seems like a legacy objc style setter with a private variable with a prefixed underscore. I will make it more swifty.

Wire-iOS/Sources/UserInterface/Components/NetworkStatus/OfflineBar.swift
billypchan: Actually they are quite different. Only common attribute is they have a state.

PR #1965, opened 2018-04-04
ZIOS-9791: Fix transparent navbar in fullscreen image

alexaubry:
### Issues

The navigation bar in the fullscreen image view did not cover the content under the status bar, so the images would appear blended with the status bar.

### Solutions

We are now using a child view controller instead of a view as the `UINavigationBarContainer`. This simplifies the positioning of the navigation bar, as we can use the `safeTopAnchor` to constrain it.

Thanks to this change, we now only need to pin the navbar container to the top/left/right of the view. Safe area/status bar changes will be handled automatically.

Furthermore, the Auto Layout constraints inside the container were changed so that:

- the navbar container is always pinned to the top of the screen
- the navigation bar is pinned to the top of the safe area (to the bottom of the status bar on < iOS 11)
- the navbar container extends to the bottom of the navigation bar


typfel:
nice 👍

Wire-iOS/Sources/UserInterface/Conversation/ConversationRootViewController.swift
```diff
@@ -74,30 +74,33 @@ import Cartography
         navbar.barTintColor = UIColor.wr_color(fromColorScheme: ColorSchemeColorBarBackground)
         navbar.tintColor = UIColor.wr_color(fromColorScheme: ColorSchemeColorTextForeground)
 
-        self.customNavBar = UINavigationBarContainer(navbar)
+        let navBarContainer = UINavigationBarContainer(navbar)
```
billypchan: It is confusing that have a local variable has the same name as a class variable.

PR #1967, opened 2018-04-04
Restore: Do not remove the checkmark after the animation

mikeger:
When user restores the backup due to the timing issue when the check mark is removed the view controller is not dismissed yet.

The proposed solution is to keep the check mark and just dismiss the whole controller.

### DIFF
--- a/WireExtensionComponents/Views/Loading/UIViewController+LoadingView.h
+++ b/WireExtensionComponents/Views/Loading/UIViewController+LoadingView.h
@@ -29,6 +29,6 @@
 @property (strong, nonatomic) ProgressSpinner *activityIndicator;
 @property (assign, nonatomic) BOOL showLoadingView;
 
-- (void)indicateLoadingSuccessWithCompletion:(dispatch_block_t)completion;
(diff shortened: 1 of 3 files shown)


WireExtensionComponents/Views/Loading/UIViewController+LoadingView.h
```diff
@@ -29,6 +29,6 @@
 @property (strong, nonatomic) ProgressSpinner *activityIndicator;
 @property (assign, nonatomic) BOOL showLoadingView;
 
-- (void)indicateLoadingSuccessWithCompletion:(dispatch_block_t)completion;
+- (void)indicateLoadingSuccessWithCompletion:(dispatch_block_t)completion removeCheckmark:(BOOL)removeCheckmark;
```
billypchan: Is it better to move completion argument to the last, then Swift can use it with trailing closure?

mikeger: You are right, it is better that way.

PR #1968, opened 2018-04-05
Fix: List UI turns black during transition to conversation view

billypchan:
## What's new in this PR?

This PR is a follow up of https://github.com/wireapp/wire-ios/pull/1919.

### Issues

list UI turns black during the transition to conversation view if the user chooses another conversation quickly (<0.55 sec).

### Causes

If second animation starts before the first animation ends, the first animation's completion closure hides left view before second animation finishes.

### Solutions

Check leftview is in the middle of the animation before hiding the leftview.

### Note

setLeftViewControllerRevealed() is translated to Swift for easier debugging.


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -532,53 +531,19 @@ - (BOOL)transitionFromViewController:(UIViewController *)fromViewController
...
 {
-    if (animated) {
-        [self.view layoutIfNeeded];
-    }
-    
-    self.leftView.hidden = NO;
-    
     _leftViewControllerRevealed = leftViewControllerRevealed;
-    [self resetOpenPercentage];
+    [self updateLeftViewControllerWithAnimated:animated completion:completion];
```
mikeger: I think the name of the method is not reflecting what it is doing, also has this annoying `With` from automatic conversion from Swift. May be it is better to call it `updateLeftViewControllerVisibilityAnimated:completion:` ?

billypchan:
@makingthematrix Thanks!

PR #1971, opened 2018-04-05
Fix: When app language set to Chinese simplified or Italian, "Why verify  conversations?" is not clickable 

billypchan:
### Issues

In device fingerprint screen, no link for the text "Why verify conversations?" for Chinese simplified, Italian and some other languages.

### Causes

In some of the translations, "profile.devices.fingerprint_message.link" is not a substring of "profile.devices.fingerprint_message". Part of "profile.devices.fingerprint_message" is not mark as a link in these languages.

### Solutions

Rewrite the logic to combine two strings, instead of finding "profile.devices.fingerprint_message.link" in "profile.devices.fingerprint_message".

## Notes

"general.space_between_words" is added for localize space for languages which do not have spaces.

### DIFF
--- a/Wire-iOS/Resources/zh-Hans.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/zh-Hans.lproj/Localizable.strings
@@ -17,6 +17,7 @@
 //
 
 
+// General strings
 "general.ok" = "好";
 "general.next" = "继续";
 "general.cancel" = "取消";
@@ -24,6 +25,9 @@
 "general.open_settings" = "开启Wire设定";
 "general.edit" = "修改";
 "general.done" = "完成";
+"general.confirm" = "好";
+// Language like Chinese does not use space to sperate words or sentences.
+"general.space_between_words" = "";
 
 "general.guest-room-name" = "访客聊天室";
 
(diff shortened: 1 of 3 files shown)


typfel:
Why wasn't the link string a substring of the whole string in Chinese and Italian?

mikeger:
@typfel I guess it's the translator error or hard for the translator to realize that the strings have to be exactly the same.

billypchan:
@typfel The translators can not see the relation of these two strings, they translate them in a different way:

Base:
profile.devices.fingerprint_message" = "Wire gives every device a unique fingerprint. Compare them with %@ and verify your conversation. Why verify conversations?";
"profile.devices.fingerprint_message.link" = "Why verify conversations?";

Italian:
"profile.devices.fingerprint_message" = "Wire dà un'impronta unica a ogni dispositivo. Confrontale con %@ e verifica la tua conversazione. Perché verificare la conversazione?";
"profile.devices.fingerprint_message.link" = "Perché verificare le conversazioni?";

Wire-iOS/Resources/zh-Hans.lproj/Localizable.strings
```diff
@@ -566,7 +570,7 @@
 // Device list
 "profile.devices.title" = "设备";
 "profile.devices.fingerprint_message_unencrypted" = "%@使用旧版本的Wire，无法显示设备。";
-"profile.devices.fingerprint_message" = "Wire 给每一台设备一组唯一的指纹，并比较它与%@来验证您的对话。为什么要验证？";
+"profile.devices.fingerprint_message.title" = "Wire 给每一台设备一组唯一的指纹，并比较它与%@来验证您的对话。";
```
mikeger: Don't forget to update it on Crowdin, otherwise this one would be overwritten.

typfel:
Don't forget to sync this with crowdin

PR #1972, opened 2018-04-05
ZIOS-9422: Send boolean type values to Mixpanel as boolean, not string

nicorsm:
### Issues

Booleans were sent to Mixpanel as hardcoded strings ("true" / "false")

### Solutions

Now I'm passing primitive types (BOOL wrapped in NSNumber or Swift booleans) to Mixpanel.


nicorsm:
@mikeger from our side, I've verified that we're sending these events. Don't know if now they're correctly handled on Mixpanel (I don't have access).

PR #1976, opened 2018-04-05
Fix: Change 'anruf' to 'Anruf'

billypchan:
### Issues

"Bill calling" is translated to "Bill anruf" in German. The correct translation should be "Bill Anruf"

### Causes

"Anruf" is lowercase in the code.

### Solutions

Keep the original Uppercased localization. Handle the casing in Crowdin.


mikeger:
I think this should break the screenshot tests.

billypchan:
@mikeger  I would keep English string same as before for the snapshot tests.

PR #1980, opened 2018-04-06
ZIOS-9795: Disable call quality overlay if UseAnalytics is set to 0

nicorsm:
### Issues

Call quality survey overlay was shown when `UseAnalytics` flag was set to zero.

### Solutions

I'm now checking if `UseAnalytics` is active before showing the survey overlay.


mikeger:
Let's wait for @alexaubry to review this.

PR #1982, opened 2018-04-06
Notify UI that sign in is successful

mikeger:
This portion of code used to show the "You logged in on the new device" screen, which was moved in the flow.


mikeger:
diff with the old code:
screen shot 2018-04-06 at 12 24 37

mikeger:
The bug is that the sign in on iPad does not show the conv. list.

PR #1983, opened 2018-04-06
Feature/zios 9856 network status bar breathing animation fine tune

billypchan:
## What's new in this PR?

Fine-tune the animation with a slower and smoother appearance:
1. Opacity values should be 40% and 100%
2. use a sine curve instead of easeInOutQuart
3. fix network status bar missing in conversation view issue.
4. refactor ConversationRootViewController
5. Fix: sync bar's accent color on conversation list is not correct issue. (When the user switches the account, the accent color is not updated.)


Wire-iOS/Sources/UserInterface/Conversation/ConversationRootViewController.swift
```diff
@@ -66,30 +75,22 @@ import Cartography
...
+        networkStatusViewController.createConstraintsInContainer(bottomView: navBarContainer.view,
+                                                                 containerView: self.view, 
+                                                                 topMargin: UIScreen.safeArea.top)
  
         constrain(navBarContainer.view, view, contentView, conversationViewController.view) {
             navBarContainer, view, contentView, conversationViewControllerView in
 
             navBarContainer.left == view.left
             navBarContainer.right == view.right
-            navBarContainer.top == view.top
```
billypchan: The constraint for navBarContainer.top is created in networkStatusViewController.createConstraintsInContainer() ready.

Wire-iOS/Sources/UserInterface/Components/Views/BreathLoadingBar.swift
```diff
@@ -101,6 +84,27 @@ class BreathLoadingBar: UIView {
...
+        case .online:
+            heightConstraint?.constant = 0
+            alpha = 0
+            layer.cornerRadius = 0
+        case .onlineSynchronizing:
+            heightConstraint?.constant = CGFloat.SyncBar.height
+            alpha = 1
+            layer.cornerRadius = CGFloat.SyncBar.cornerRadius
+
+            backgroundColor = UIColor.accent()
```
billypchan: update background color before it shows.

PR #1984, opened 2018-04-06
ZIOS-9838: Start UI title turns black

nicorsm:
### Issues

1. Open the Start UI
2. Tap on "Create group"
3. Go back
4. The title inside the navigation bar is black

### Solutions

Setting the tint color and the translucency of the navigation bar inside the `viewDidLoad()` seems to solve the problem

## Notes

The title color seems to be updated only after completing the transition animation between the two view controllers.


Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController.m
```diff
@@ -146,6 +146,9 @@ - (void)viewDidLoad
     self.quickActionsBar = [[StartUIInviteActionBar alloc] init];
     [self.quickActionsBar.inviteButton addTarget:self action:@selector(inviteMoreButtonTapped:) forControlEvents:UIControlEventTouchUpInside];
 
+    self.navigationController.navigationBar.barTintColor = [UIColor clearColor];
+    [self.navigationController.navigationBar setTranslucent:YES];
```
typfel: I think you can move the move the two lines at https://github.com/wireapp/wire-ios/pull/1984/files#diff-b59a5ec836885e4540638817487d237bR170 here as well.

PR #1986, opened 2018-04-09
Add request password UI on backup export

mikeger:
- Includes the copy update and visual fixes for the back up screen.
- Also includes the tests.


Wire-iOS/Sources/UserInterface/Settings/BackupPasswordViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        navigationItem.rightBarButtonItem?.tintColor = .accent()
+        navigationItem.rightBarButtonItem?.isEnabled = false
+    }
+    
+    fileprivate func updateState(with text: String) {
+        password = Password(text)
+        navigationItem.rightBarButtonItem?.isEnabled = nil != password
+    }
+    
+    dynamic fileprivate func cancel() {
```
daehn: We can already start by doing `@objc private func` to be ready for Swift 4.

Wire-iOS/Sources/UserInterface/Settings/BackupPasswordViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+     
+        if let _ = string.rangeOfCharacter(from: CharacterSet.newlines) {
+            if let _ = password {
+                completeWithCurrentResult()
+            }
+            return false
+        }
+        
+        let newString = ((textField.text ?? "") as NSString).replacingCharacters(in: range, with: string)
```
daehn: I guess we should also think about whitespaces (I assume a password only consisting of spaces is not allowed?).

mikeger: No, only whitespaces password is not good.

Wire-iOS/Sources/UserInterface/Settings/BackupPasswordViewController.swift
```diff
@@ -24,15 +24,15 @@ struct Password {
     let value: String
     
     init?(_ value: String) {
-        guard value.count >= Password.minimumCharacters else { return nil }
+        guard value.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).count >= Password.minimumCharacters else { return nil }
```
daehn: Maybe we should also forbid the user to enter whitespaces altogether

mikeger: I was advised to make it work exactly as the other password field in registration, there entering spaces is not forbidden.

PR #1999, opened 2018-04-12
Fix/link alert

johnxnguyen:
### Issues

Tapping on **any** link (including detected addresses/phone numbers etc) displays an open link alert. 

### Causes

The open link alert is shown when a user taps a link that hides its underlying URL. Unfortunately, data detected links will always qualify to show the open link alert.

### Solutions

Use the markdown identifier attribute to determine which links were created from markdown, and show the alert for all such links.


johnxnguyen:
@mikeger FYI I've updated this PR

PR #2003, opened 2018-04-12
ZIOS-9100: Adapt UI to fetch new log files

nicorsm:
### Issues

Logs are now saved on `current.log` and `previous.log` files, inside the Caches directory. We need to send both files via UI.

### Solutions

I've adapted `DebugAlert` and `SettingsTechnicalReportViewController` classes to support the sharing of the two files using the `MFMailComposeViewController` component or , in case the user has no mail accounts in Mail.app, using `UIActivityViewController`. 

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-system/pull/20


Wire-iOS/Sources/AppDelegate.m
```diff
@@ -127,6 +127,8 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     }];
     self.launchOptions = launchOptions;
     
+    [ZMSLog switchCurrentLogToPrevious];
```
mikeger: Would it make sense to do it immediately when this method is called? Otherwise some logs would fall to the prev file (just a guess).

Wire-iOS/Sources/AppDelegate.m
```diff
@@ -113,6 +113,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 {
     ZMLogInfo(@"application:didFinishLaunchingWithOptions START %@ (applicationState = %ld)", launchOptions, (long)application.applicationState);
     
+    [ZMSLog switchCurrentLogToPrevious];
```
mikeger: Previous `ZMLogInfo` line would go to the previous log I believe.

mikeger: I see no easy way to overcome this, since logging is async.

PR #2010, opened 2018-04-16
ZIOS-9870: iPad - Impossible to register with email from settings

nicorsm:
### Issues

If the user is logged in with an account and wants to add an unverified account, after inserting data, is redirected to the landing page instead of the Terms of Use page and Verify email view.

### Causes

The app state switches from `unauthenticated` with `ZMUserSession` error 21 (`ZMUserSessionAddAccountRequested`) to `unauthenticated` with `nil` error. Since the default value for `addingNewAccount` is `true`, the user is redirected to the Landing page.

### Solutions

I was thinking about setting the flag to `false`, so in this case we display the landing page only if the `ZMUserSessionAddAccountRequested` error is emitted (and not by default). @vytis what do you think?

## Notes

There's also a little fix I'm working on, about disabling the loop of checking the email verification when dismissing the Terms of Use view.


nicorsm:
The fix is not working in case of a clean install. I'll work on that.

PR #2014, opened 2018-04-17
Fix: ZIOS-9873 prevent configure ConversationCell if message is nil

billypchan:
### Issues

App crashes when trying config a message cell with nil message.

### Causes

ConversationCell.message is wiped before config. 

### Solutions

Check message is nil before configuration.


billypchan:
@mikeger Thanks! I updated the PR.

billypchan:
This method is private, only called inside the class.

typfel:
@billypchan release branch is already created, maybe rebase on top of `release/3.11`

PR #2016, opened 2018-04-18
Fix/zios 9794 setting navigation bar covers table

billypchan:
### Issues

On iOS11, the verify email/phone screen's table top line is clipped

### Causes
This fix is not necessary on iOS11
tableView.contentInset = UIEdgeInsets(top: -32, left: 0, bottom: 0, right: 0)

### Solutions

Remove the contentInset custom value for iOS11.

## Notes

Snapshot tests are created and rewrite force unwrap of ZMUserSession.shared().


Wire-iOS Tests/ConfirmPhoneViewControllerTests.swift
```diff
@@ -0,0 +1,40 @@
...
+import XCTest
+@testable import Wire
+
+final class ConfirmPhoneViewControllerTests: ZMSnapshotTestCase {
+    
+    var sut: ConfirmPhoneViewController!
+    
+    override func setUp() {
+        super.setUp()
+        sut = ConfirmPhoneViewController(newNumber: "012345678901", delegate: nil)
```
vytis: Would be good to maybe add it to navigation controller so that it would be similar setup as in real life. No need to commit it, but ideally the screenshot would show the bug in action, the view is shifted before the fix and as expected after the fix

Wire-iOS/Sources/UserInterface/Settings/ChangingDetails/ConfirmEmailViewController.swift
```diff
@@ -46,7 +46,10 @@ final class ConfirmEmailViewController: SettingsBaseTableViewController {
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        observer = UserChangeInfo.add(observer: self, for: ZMUser.selfUser(), userSession: ZMUserSession.shared()!)
+
+        if let userSession = ZMUserSession.shared() {
```
vytis: If `userSession` session is nil then we are in real trouble, basically UI should not even be shown before it is correctly set up and this is why we can safely force-unwrap it everywhere

Wire-iOS/Sources/UserInterface/Settings/SettingsTableViewController.swift
```diff
@@ -79,6 +79,11 @@ class SettingsBaseTableViewController: UIViewController {
         view.addSubview(footerContainer)
         footerContainer.addSubview(footerSeparator)
         footerSeparator.inverse = true
+
+        if #available(iOS 11.0, *) {
+        } else {
+            tableView.contentInset = UIEdgeInsets(top: -32, left: 0, bottom: 0, right: 0)
```
typfel: So this is now always applied for all settings screens I assume you've verified this doesn't cause any problems?

billypchan: @typfel I check on iOS10 device and confirmed it was a hack to prevent the tableview content covered by navigation bar.

PR #2018, opened 2018-04-18
Fix/zios 7149 Taking a picture should save it in the iOS Camera Roll

billypchan:
### Issues

Photo shot in full screen mode is not store in camera roll

### Causes

Shot in camera keyboard and shot in full screen mode are written in two classes, they have different work flow to pass an image to conversation.

### Solutions

clean up the logic and write a function for save the image, call it before image message is sent

## Notes

It is better to unify the workflow in ConversationInputBarViewController+Camera.swift and ConversationInputBarViewController+Files.m.


billypchan:
Close this PR as the issue shoved in https://github.com/wireapp/wire-ios/pull/2020 .

PR #2025, opened 2018-04-20
Correctly display add/remove participant error

mikeger:
### Issues

The error message was not displayed to the user. Moreover, there was no special case for offline action.

### Causes

The controller we tried to present over was not the topmost one.
There was no copy for the error message.

### Solutions

Select the correct controller to present the alert over. Use the existing offline copy.


Podfile.lock
```diff
@@ -53,4 +53,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 5f2dad8ae96f93f0bdb37167e3672155aff83bf1
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.4.0
```
typfel: Oh this been annoying me since forever but does Jenkins coup with this?

mikeger: I'll revert for the time being, let's check with 3.12

PR #2027, opened 2018-04-20
Adjust unknown message body copy

infotexture:
The outdated client version is not necessarily the cause, so we need to soften this message accordingly.

In addition to revising the copy of the error message, the “[Why is this happening?](https://github.com/wireapp/wire-ios/blob/fix/error-copy/Wire-iOS/Resources/Base.lproj/Localizable.strings#L394)” link should be removed.


mikeger:
@infotexture I will remove the `Why is this happening?` part in this PR.

PR #2030, opened 2018-04-23
Fix: do not set ColorScheme.default().variant in snapshot tests

billypchan:
### Issues

After https://github.com/wireapp/wire-ios/pull/2021 is merged, some snapshot tests fails color difference.

### Causes

Change ColorScheme.default().variant may cause side effects to icon button's border color.

### Solutions

Inject the color variant to ImageMessageCell, instead of changing ColorScheme.default().variant.


Wire-iOS Tests/ImageMessageCellTests.swift
```diff
@@ -26,19 +26,20 @@ class ImageMessageCellTests: ZMSnapshotTestCase {
 
     override func setUp() {
         super.setUp()
+        UIView.setAnimationsEnabled(false)
```
daehn: This is already done in the super class `ZMSnapshotTestCase` itself (same for the `UIView.setAnimationsEnabled(true)` in `tearDown`).

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ImageMessageCell.h
```diff
@@ -20,10 +20,12 @@
 #import "ConversationCell.h"
 @class SavableImage;
 @class FLAnimatedImageView;
+@import WireExtensionComponents;
 
 @interface ImageMessageCell : ConversationCell
 
 @property (nonatomic, readonly) FLAnimatedImageView *fullImageView;
 @property (nonatomic, readonly) SavableImage *savableImage;
+@property (nonatomic) ColorScheme *colorScheme;
```
daehn: It seems like only the variant and not the scheme itself is used, maybe it is sufficient to make this a `ColorSchemeVariant` property.

PR #2031, opened 2018-04-23
ZIOS-9889: Audio message is discarded after Wire call

nicorsm:
### Issues

Audio recordings started before incoming or outgoing calls were deleted due to an unhandled case.

### Causes

This is because the keyboard got dismissed on `viewWillDisappear()`, but the continuation of the flow (adding effects on the registration and sending it) was not handled.

### Solutions

I'm now handling the cases where the keyboard should reappear after the call termination by monitoring changes on `WireCallCenterCallStateObserver` inside `ConversationInputBarViewController+Audio.swift`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
@@ -188,20 +188,16 @@ extension ConversationInputBarViewController : WireCallCenterCallStateObserver {
     
     public func callCenterDidChange(callState: CallState, conversation: ZMConversation, caller: ZMUser, timestamp: Date?) {
         
-        let isRecording = self.audioRecordKeyboardViewController?.recorder.state == .recording
+        let isRecording = self.audioRecordKeyboardViewController?.isRecording
         
-        switch callState {
+        switch (callState, isRecording, self.wasRecordingBeforeCall) {
             
-        case .incoming(_, let shouldRing, _) where shouldRing && isRecording && !self.wasRecordingBeforeCall:
-            fallthrough
-        case .outgoing where isRecording && !self.wasRecordingBeforeCall:
-            self.wasRecordingBeforeCall = true
-            
-        case .incoming(_, let shouldRing, _) where !shouldRing && self.wasRecordingBeforeCall:
-            fallthrough
-        case .terminating where self.wasRecordingBeforeCall:
-            displayRecordKeyboard()
-            
+        case (.incoming(_, true, _), true, false),      // receiving incoming call while recording an audio
+             (.outgoing, true, false):                  // making an outgoing call while recording an audio
+            self.wasRecordingBeforeCall = true          // -> remember that we were recording an audio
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -175,3 +182,34 @@ extension ConversationInputBarViewController: AudioRecordViewControllerDelegate
     }
     
 }
+
+
+extension ConversationInputBarViewController : WireCallCenterCallStateObserver {
+    
+    public func callCenterDidChange(callState: CallState, conversation: ZMConversation, caller: ZMUser, timestamp: Date?) {
+        
+        let isRecording = self.audioRecordKeyboardViewController?.recorder.state == .recording
```
vytis: Could use the `isRecording`property defined above

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -175,3 +182,34 @@ extension ConversationInputBarViewController: AudioRecordViewControllerDelegate
...
+        
+        switch callState {
+            
+        case .incoming(_, let shouldRing, _) where shouldRing && isRecording && !self.wasRecordingBeforeCall:
+            fallthrough
```
vytis: Maybe would be possible to simplify the logic a bit by switching on a tuple `(callState, isRecording)` instead. That many variables and `fallthrough` makes it pretty hard to see what's going on.

daehn: Also I think it is possible to do `case .incoming(_, true, _)` instead of `case .incoming(_, let foo, _) where foo`.

nicorsm: I was thinking the same as @daehn, the syntax becomes less verbose but it's more clear what happens - anyway, I've added some comments for each instruction.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -175,3 +182,30 @@ extension ConversationInputBarViewController: AudioRecordViewControllerDelegate
...
+    public func callCenterDidChange(callState: CallState, conversation: ZMConversation, caller: ZMUser, timestamp: Date?) {
+        
+        let isRecording = self.audioRecordKeyboardViewController?.isRecording
+        
+        switch (callState, isRecording, self.wasRecordingBeforeCall) {
+            
+        case (.incoming(_, true, _), true, false),      // receiving incoming call while recording an audio
+             (.outgoing, true, false):                  // making an outgoing call while recording an audio
+            self.wasRecordingBeforeCall = true          // -> remember that we were recording an audio
+        case (.incoming(_, false, _), _, true),         // refusing an incoming call
```
vytis: 👍 for comments!

PR #2046, opened 2018-04-25
Recommend strong password for backup file

infotexture:
Revise backup description to include strong password recommendation.

@mikeger @daehn I also updated the current German translation in Crowdin to reflect the revised version of the string, but merging this may mark the modified string as untranslated, so I'll push an extra commit with the revised translation to save time.


mikeger:
@infotexture feel free to update the German copy directly in the file in this PR, this would save time and we would update it properly in crowdin as the second step.

mikeger:
@infotexture I would also change the base branch, so it would go directly into release.

mikeger:
Oops, this one failed, let me fix the branch.

PR #2050, opened 2018-04-26
Bump AVS to 4.3.16

mikeger:
With the new message recording API we are asking AVS to prepare the audio session for the recording, AVS then is setting the session category and waiting for the confirmation from the operating system. Unfortunately on the Simulator the confirmation never arrives, so AVS is never telling us that it is ready to start the recording.

AVS fixed the issue from their side in the way that on the simulator they are immediately calling the callback without waiting for the iOS confirmation.

### DIFF
--- a/avs-versions
+++ b/avs-versions
@@ -17,4 +17,4 @@
 # along with this program. If not, see http://www.gnu.org/licenses/.
 #
 
-export APPSTORE_AVS_VERSION=4.2.28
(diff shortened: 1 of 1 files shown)


mikeger:
@typfel I noticed that this also moves us from AVS 4.2 to 4.3, should be safe? I tested the basic calling, seems OK.

PR #2055, opened 2018-04-26
Add new calling UI

daehn:
## What's new in this PR?

* Add call actions view for new call UI.


Wire-iOS/Sources/UserInterface/Calling/CallActionsView/IconLabelButton+CallActionsView.swift
```diff
@@ -0,0 +1,163 @@
...
+    var backgroundColorNormal: UIColor {
+        switch self {
+        case .audio: return UIColor.wr_color(fromColorScheme: ColorSchemeColorGraphite, variant: .light).withAlphaComponent(0.08)
+        case .video: return UIColor.white.withAlphaComponent(0.24)
+        }
+    }
+    
+    var backgroundColorSelected: UIColor {
+        switch self {
+        case .audio: return UIColor.wr_color(fromColorScheme: ColorSchemeColorGraphite, variant: .light)
```
typfel: I guess the plan is to update this later when we work on supporting dark mode?

daehn: Checking with design right now, I was under the impression that there was no real dark mode but only the difference between audio / video call. But maybe the audio call UI will be dark in dark mode – in which case it might make more sense to make this a simple `dark` / `light` enum and let the controlling component decide.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/IconButton+CallActions.swift
```diff
@@ -0,0 +1,63 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+extension IconButton {
+    
+    static let width: CGFloat = 64
```
mikeger: Can be private?

mikeger: Ah, see it now, used in another file. May be worth renaming then, since now I think it is too generic.

Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/ThemeableContainer.swift
```diff
@@ -0,0 +1,39 @@
...
+            
+            for subview in subviews {
+                if var themeableView = subview as? Themeable {
+                    themeableView.colorSchemeVariant = colorSchemeVariant
+                }
+            }
+        }
+    }
+    
+    func applyColorScheme(_ colorSchemeVariant: ColorSchemeVariant) {
```
mikeger: This one under construction?

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
```diff
@@ -0,0 +1,254 @@
...
+    var variant: ColorSchemeVariant
+    
+    init(properties: CallProperties, variant: ColorSchemeVariant = ColorScheme.default().variant) {
+        self.variant = variant
+        self.properties = properties
+        statusView = CallStatusView(configuration: properties.configuration(variant: variant))
+        super.init(nibName: nil, bundle: nil)
+    }
+    
+    @available(*, unavailable)
```
mikeger: I am not sure if this is necessary.

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
```diff
@@ -0,0 +1,254 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    private func setupViews() {
+        [stackView, bitrateLabel].forEach(addSubview)
+        stackView.distribution = .fill
+        stackView.alignment = .center
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        bitrateLabel.translatesAutoresizingMaskIntoConstraints = false
```
mikeger: Is it necessary for the arranged views?

Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/ThemeableContainer.swift
typfel: No, the container doesn't need to do anything in the method. It's required by the `Themeable` protocol though.

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
daehn: `bitrateLabel` is not inside the stackView

Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/ThemeableContainer.swift
mikeger: Ah ok

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
daehn: It just makes sure this initialiser isn't exposed and can't be called from anywhere.

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
mikeger: True, sorry.

Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
mikeger: True, just this one is never called directly anyway, but good to have the compile time declaration for holistic reasons.

PR #2059, opened 2018-04-27
Fix/After double tapping GIF in full screen image viewer, the image moves to down right corner

billypchan:
### Issues

Steps to recreate:
1. open a small size GIF(GIF's width < view's width) in the image viewer
2. double tap it
Actual result:
the gif animates to bottom right corner
Expected result:
the gif should be zoomed in 

### Causes

Calling scrollView.zoom(to rect:  animated: ) for a FLAniamtedImageView already in max zoom scale cause this issue

### Solutions

Do not zoom in the image if it is in max zoom scale and it is smaller than the view.


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController+zoom.swift
```diff
@@ -57,6 +57,33 @@ extension FullscreenImageViewController {
...
+
+        UIMenuController.shared.isMenuVisible = false
+        let scaleDiff: CGFloat = scrollView.zoomScale - scrollView.minimumZoomScale
+        // image view in minimum zoom scale, zoom in to a 50 x 50 rect
+        if scaleDiff < kZoomScaleDelta {
+            // image is smaller than screen bound and zoom sclae is max(1), do not zoom in
+            if (image.size.width < self.view.bounds.width &&
+                image.size.height < self.view.bounds.height &&
+                scrollView.zoomScale == 1) == false {
+                let point: CGPoint? = doubleTapper?.location(in: doubleTapper?.view)
```
vytis: Having `?? .zero` at the end would make the line below a bit clearer. Also there is no need for type annotation.

Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController+internal.h
```diff
@@ -23,7 +23,7 @@ static CGFloat const kZoomScaleDelta = 0.0003;
 @property (nonatomic) CGFloat lastZoomScale;
 @property (nullable, nonatomic, readwrite) UIImageView *imageView;
 
-- (void)handleDoubleTap:(nonnull UITapGestureRecognizer *)doubleTapper;
```
vytis: It seems here `doubleTapper` was marked as `nonnull`, but in Swift version it's optional. Was that intentional?

billypchan: I agree that mark it as non-optional is more clear as it is meaningless it doubleTapper is nil. Thanks!

PR #2065, opened 2018-04-30
Fix/zios 9959 fix crash in wire data model.message.is normal

billypchan:
### Issues

App crashes when calling WireDataModel.Message.isNormal.

### Causes

Lack of nil check in ZMConversationMessageWindow.isPreviousSenderSameForMessage. It may pass a nil argument value to Message.isNormal.

### Solutions

Rewrite ZMConversationMessageWindow.isPreviousSenderSameForMessage in Swift and unwrap message argument.

### DIFF
--- a/Wire-iOS Tests/Mocks/MockConversation.h
+++ b/Wire-iOS Tests/Mocks/MockConversation.h
@@ -24,12 +24,13 @@
 
 @interface MockConversation : NSObject<Mockable>    
 
-@property (nonatomic, copy) NSString *displayName;
-@property (nonatomic) ZMUser *connectedUser;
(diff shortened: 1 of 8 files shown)


vytis:
You also seem to have tried to fix this crash with https://github.com/wireapp/wire-ios/pull/2014 and there was is a nil check there already so this method should not get called with `message = nil`. 

It seems the changes were not shipped with 3.11 version though, when I checkout code from the `3.11` tag the changes are not there: https://github.com/wireapp/wire-ios/blob/appstore/3.11/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m#L330-L333

PR #2068, opened 2018-05-02
Add call status view

daehn:
## What's new in this PR?

* Add call status view + tests.


Wire-iOS/Sources/UserInterface/Calling/CallStatusView.swift
```diff
@@ -0,0 +1,130 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+import WireExtensionComponents
+
+final class CallStatusView: UIView {
```
typfel: If you implement the `Themable` protocol you get can snapshot testing of themes for free.

daehn: Ah nice, didn't know about that. This shouldn't be dependent on the global theme though as video calls always should use the dark variant.

PR #2070, opened 2018-05-02
ZIOS-9942: Show warning when switching accounts

nicorsm:
## What's new in this PR?

I've added an alert (action sheet) which will be shown every time the user is in an ongoing call and tries to:
- add an account
- switch between different accounts

The method is implemented as static in `SelfProfileViewController` and has a completion block that is called when the user decides to switch account. It covers both cases - ongoing or not ongoing call.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/753


Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -199,3 +201,23 @@ extension SelfProfileViewController: UserImageViewDelegate {
...
+extension SelfProfileViewController: SessionManagerSwitchingDelegate {
+    
+    func confirmSwitchingAccount(completion: @escaping (Bool) -> Void) {
+        
+        guard let controller = UIApplication.shared.wr_topmostController(onlyFullScreen: false) else {
+            completion(false)
+            return
+        }
+        
+        let alert = UIAlertController(title: "self.settings.add_account.switch_accounts.title", message: nil, preferredStyle: .actionSheet)
```
nicorsm: Maybe the title should be `.localized` 🙃

mikeger:
@nicorsm what's the status in this one? Should be merged? Some conflicts appeared meanwhile.

nicorsm:
@mikeger It has this dependency that should be reviewed first: https://github.com/wireapp/wire-ios-sync-engine/pull/753

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -542,6 +543,32 @@ extension AppRootViewController: LandingViewControllerDelegate {
     }
 }
 
+// MARK: - Ask user if they want want switch account if there's an ongoing call
+
+extension AppRootViewController: SessionManagerSwitchingDelegate {
+    
+    func confirmSwitchingAccount(completion: @escaping (Bool) -> Void) {
+        
+        guard let session = ZMUserSession.shared(), let callCenter = session.callCenter, !callCenter.activeCallConversations(in: session).isEmpty else { return completion(true) }
```
daehn: Maybe we can extract this into a computed property on `ZMUserSession` like `isCallOngoing` or something like that, as we will need this in other places as well (warning when placing a call when there is one already ongoing).

PR #2072, opened 2018-05-02
Wrap call status view in view controller

daehn:
## What's new in this PR?

* Wrap status view in view controller that starts and stops a call duration timer.


Wire-iOS Tests/CallStatusViewTests.swift
```diff
@@ -25,8 +25,9 @@ final class CallStatusViewTests: ZMSnapshotTestCase {
 
     override func setUp() {
         super.setUp()
+        recordMode = true
```
typfel: Should be deleted?

PR #2075, opened 2018-05-03
Add first implementation of CallInfoController

daehn:
## What's new in this PR?

* Add first implementation of CallInfoController.


Wire-iOS/Sources/UserInterface/Calling/CallViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+
+// The ouput actions a `CallInfoViewController` can perform.
+enum CallAction {
```
typfel: Couldn't we lose either `CallActionsViewAction` or `CallAction`?

daehn: Yeah I guess we could lose `CallActionsViewAction` and only use `CallAction`

daehn: @typfel updated!

typfel:
Nice! Let's merge it?

PR #2077, opened 2018-05-03
Initial version: hide the calling overlay

mikeger:
## What's new in this PR?

Initial version to allow dismissing the calling overlay and showing it again. Function works similarly to the regular iOS calling UI.

Pending changes: (current state can be merged).

- [ ] Interactive dismissal.
- [ ] Fix the status label: must be the same as on the call overlay.
- [ ] Verify what happens with the second call.


nicorsm:
I'll also give a try on the simulator.

PR #2078, opened 2018-05-03
Create utility method to enable/disable hockey

marcoconti83:
## What's new in this PR?

Refactored to control the turning on and off of tracking with a single method. Also cleaned up the initialization part of hockey and added comments on why we do certain things in initialization.

## Notes

There's still one XXX left, waiting for comment from someone that knows the code a bit better

⚠️ This is not tested on a real device. It should be tested before merging ⚠️


WireExtensionComponents/Utilities/Hockey.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Hockey.swift
+//  WireExtensionComponents
+//
+//  Created by Marco Conti on 03.05.18.
+//  Copyright © 2018 Zeta Project Germany GmbH. All rights reserved.
```
mikeger: Default header 🙈

mikeger:
This can be reviewed / merged now.

PR #2079, opened 2018-05-04
fix: disable testThatItMigratesTheOptOutFromLocalytics

billypchan:
### Issues

testThatItMigratesTheOptOutFromLocalytics always fail on the testing server and local machine.

### Solutions

Disable it until it is fixed.


billypchan:
Lo

PR #2080, opened 2018-05-04
Fix call participants list themeing

daehn:
## What's new in this PR?

* Fix call participants list themeing + add more tests.


typfel:
Something is going on with the accessory views in the participants list. Some have the `>` some don't.

daehn:
@typfel I will have a look!

PR #2081, opened 2018-05-04
Working calling UI

typfel:
## What's new in this PR?

This PR adds another view controller which binds the calling components with the voice channel.

Many things are still missing or not working as intended.

working:
- [x] accepting / declining calls
- [x] muting

not working
- [ ] showing participant list
- [ ] toggling video

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallViewController.swift
@@ -1,0 +1,189 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallViewController.swift
```diff
@@ -0,0 +1,189 @@
...
+}
+
+extension CallViewController: CallInfoViewControllerDelegate {
+    
+    func infoViewController(_ viewController: CallInfoViewController, perform action: CallAction) {
+        guard let userSession = ZMUserSession.shared() else { return }
+        
+        switch action {
+        case .acceptCall:
+            conversation?.joinCall()
```
typfel: Would like to clean up up these helper methods. Kinda confusing that we call some on `ZMConversation` and some on `VoiceChannel`.

Wire-iOS/Sources/UserInterface/Calling/CallViewController.swift
```diff
@@ -0,0 +1,189 @@
...
+
+struct CallInfoConfiguration  {
+    
+    let voiceChannel: VoiceChannel
+    
+}
+
+extension CallInfoConfiguration: CallInfoViewControllerInput {
+    
+    var accessoryType: CallInfoViewControllerAccessoryType {
```
typfel: I think we need to change `CallInfoViewControllerAccessoryType` since there are cases were show no avatar and no participant list (outgoing group call, camera turned off during incoming video call).

daehn: Yeah, I also wasn't too happy with this in general. What's your suggestion, adding a new case `.none` or changing this in general?

PR #2087, opened 2018-05-07
Extract call accessory, fix alignment when no accessory is shown

daehn:
## What's new in this PR?

* Extract call accessory component into separate view controller.
* Fix alignment when no accessory is shown.


Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoViewController.swift
```diff
@@ -98,35 +97,31 @@ final class CallInfoViewController: UIViewController, CallActionsViewDelegate {
...
+    // MARK: - Actions + Delegates
 
     func callActionsView(_ callActionsView: CallActionsView, perform action: CallAction) {
         Calling.log.debug("\(action) button tapped")
         delegate?.infoViewController(self, perform: action)
     }
+    
+    func callParticipantsViewControllerDidSelectShowMore(viewController: CallParticipantsViewController) {
+        Calling.log.debug("Show more participants tapped")
+        // TODO: Do we need this or should we directly push the new controller?
```
typfel: Can't it be considered internal state of the `CallInfoViewController`

daehn: Yeah, I think so. We can just create + push the detail vc in here.

daehn: Ah no, I think that `CallViewController` should be responsible for that, as that one will be the one presented inside a `UINavigationController`.

typfel: Ok

PR #2089, opened 2018-05-07
Remove Localytics

billypchan:
Remove Localytics and related tests


typfel:
@mikeger should we merge this?

PR #2091, opened 2018-05-07
Feature/video grid zios 9921

typfel:
## What's new in this PR?

Add support for showing a grid of video streams when in a video call.


Wire-iOS/Sources/UserInterface/Overlay/VoiceChannelViewController.swift
```diff
@@ -69,7 +69,7 @@ class VoiceChannelViewController: UIViewController {
         voiceChannelView.constantBitRate = conversation.voiceChannel?.isConstantBitRateAudioActive ?? false
         
         callStateObserverToken = conversation.voiceChannel?.addCallStateObserver(self)
-        receivedVideoObserverToken = conversation.voiceChannel?.addReceivedVideoObserver(self)
+//        receivedVideoObserverToken = conversation.voiceChannel?.addReceivedVideoObserver(self)
```
daehn: Is still still needed? Otherwise maybe worth adding a `// TODO` so Danger picks it up later.

typfel: That observer is gone. But I guess we'll delete this whole file at some point.

PR #2094, opened 2018-05-08
feature: zios-9970 [Tablet] Gallery view is opened full screen instead of being opened inside a popover

billypchan:
## What's new in this PR?

Display camera roll image picker with popover style on iPad regular mode.


mikeger:
@billypchan conflict on the project file 🔥

Wire-iOS/Sources/UserInterface/Camera/CameraBottomToolsViewController.m
```diff
@@ -187,20 +187,11 @@ - (IBAction)closeCamera:(id)sender
 
 - (IBAction)openPhotoLibrary:(id)sender
 {
-    UIImagePickerController *picker = [[UIImagePickerController alloc] init];
+    UIImagePickerController *picker = [UIImagePickerController popoverForIPadRegularWithSourceRect:CGRectInset(self.libraryButton.bounds, 4, 4) sourceView:self.libraryButton presentViewController:self sourceType:UIImagePickerControllerSourceTypePhotoLibrary];
```
mikeger: I think that this method's signature is too long, may be it is better to create the immutable configuration object that would hold all those parameters? Like `ImagePickerPresentationContext`, or something similar?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/UIImagePickerController+iPadPopover.swift
```diff
@@ -0,0 +1,43 @@
...
+
+extension UIImagePickerController {
+    @objc class func popoverForIPadRegular(sourceRect: CGRect,
+                                           sourceView: UIView,
+                                           presentViewController: UIViewController,
+                                           sourceType: UIImagePickerControllerSourceType) -> UIImagePickerController {
+        let picker = UIImagePickerController()
+        picker.sourceType = sourceType
+
+        if presentViewController.isIPadRegular(device: UIDevice.current) {
```
mikeger: I think that the method `isIPadRegular` is not relevant to `UIViewController`, would be significantly better to have it on something relevant to it, for example on `UIDevice`.

billypchan: @mikeger I will rewrite it like: 

if presentViewController.traitCollection.horizontalSizeClass.isIPadRegular(device: UIDevice.current)

mikeger: Ah, I see now that it checks the trait collection, then I guess it is fine on UIViewController.

PR #2095, opened 2018-05-08
Add logic to show / hide calling information overlay

daehn:
## What's new in this PR?

* Add timer logic to show / hide overlay.
* Remove unused code.
* Cleanup folder structure.


Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/CallParticipantsViewController.swift
```diff
@@ -69,6 +70,11 @@ class CallParticipantsViewController: UIViewController, UICollectionViewDelegate
         updateAppearance()
     }
     
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        topConstraint?.constant = navigationController?.navigationBar.frame.maxY ?? 0
```
typfel: This is to avoid showing the list under the navigation bar? Isn't there some safe inset for this?

daehn: This is a workaround to avoid having to make the navigation bar non-clear and changing its color depending on the state (video / audio call and the current theme).

PR #2097, opened 2018-05-08
Flood conversation with messages (debug)

mikeger:
### Issues

The option to append 10.000 text messages to the conversation. The messages are not scheduled for sending.

### Causes

This is necessary for testing the performance and the backup-restore scenarios.


vytis:
I am just not sure if it saves every time we append the message. Maybe it would be faster to do it on the sync thread inside a grouped block?

mikeger:
Good point, I will check in the profiler what is the best way to schedule the insertions.

vytis:
@mikeger What do we need to merge this one?

mikeger:
I need some time to make sure this works in the best possible way in regards of performance.

mikeger:
@vytis I've optimized the appending of the messages, the single potential problem left is when I save the context, the change notification is really huge and takes a long time to process, haven't figured out the way around it.

PR #2101, opened 2018-05-09
Fix incorrect grid layout when third person leaves

typfel:
### Issues

When the third person leaves a video call the grid would show two tall vertical cells.

### Solutions

We need re-arrange the views when one of the two stacksviews gets empty.


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -67,10 +67,26 @@ class GridView: UIStackView {
...
     }
     
-    func updateVisibleStacksViews() {
+    private func reinsert(view: UIView) {
+        remove(view: view)
+        append(view: view)
+    }
+    
+    private func rearrangeViews() {
+        if lowerHorizontalStackerView.arrangedSubviews.isEmpty && upperHorizontalStackerView.arrangedSubviews.count > 1, let view = upperHorizontalStackerView.arrangedSubviews.last {
```
mikeger: May be if you are using `,` it can be used between all the logical elements, replacing the `&&`. Also would be good to make it more narrow :)

PR #2103, opened 2018-05-09
Add floating self video preview

daehn:
## What's new in this PR?

Integrate the floating `PinnableThumbnailViewController` created by @alexaubry into the current state of the new call UI.


Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -23,21 +23,35 @@ struct VideoConfiguration {
...
+            case .connected(videoState: let state) where state.isSending: return user.remoteIdentifier
+            default: return nil
             }
-            
-            return nil
         }
         
-        return [ZMUser.selfUser().remoteIdentifier] + otherParticipants
+        // TODO: Do we need this? Right now the participants array is empty for 1-1 calls.
+        if otherParticipants.isEmpty, voiceChannel.conversation?.conversationType == .oneOnOne,
```
typfel: Let's fix it in SE so that `VoiceChannel.particiants` have a sensible value also for 1-1 calls

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -23,21 +23,35 @@ struct VideoConfiguration {
...
+            let otherUser = voiceChannel.conversation?.firstActiveParticipantOtherThanSelf()?.remoteIdentifier {
```
typfel: for 1-1 is better to use `connectedUser` instead of `firstActiveParticipantOtherThanSelf()` it works better in the tests.

Wire-iOS/Sources/UserInterface/Calling/Thumbnail/PinnableThumbnailViewController.swift
```diff
@@ -0,0 +1,267 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+@objc class PinnableThumbnailViewController: UIViewController {
```
mikeger: This implementation does not really care what kind of view `thumbnailView` is, would be potentially cool to make it more generic and not coupled with calling at all. Just the controller that can contain _any_ view and add the pinning behavior for it.

daehn: That's possible though, it's not coupled with calling right now, is it?

daehn: You can set any `UIView` as content view of the thumbnail using `setThumbnailContentView`. This class doesn't know anything about calling.

mikeger: Ah right, khm, sorry ⚡️

mikeger: I thought `ContinuousCornersView` is the video renderer for some reason.

typfel:
@daehn so let's merge this?

daehn:
> You're own videoState will actually not be .sending until after the call is established but let's change this in SE.

Yeah, that's why right now I check if the call is not established yet but is an outgoing video call and insert the self identifier for the preview manually.

PR #2104, opened 2018-05-09
Fix/zios 9982 TOS screen iPad size class update handling and snapshot

billypchan:
## What's new in this PR?

This PR is a follow-up of https://github.com/wireapp/wire-ios/pull/2096.

Add size class update handling for iPad in TOS screen. Add snapshot tests for code coverage.


Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/Steps/TermsOfUseStepViewController.h
```diff
@@ -27,6 +27,7 @@
 @interface TermsOfUseStepViewController : RegistrationStepViewController
 @property (nonatomic, readonly) ZMIncompleteRegistrationUser *unregisteredUser;
 
+- (instancetype)initWithDevice:(id)device;
```
mikeger: Can it be defined as `id<DeviceProtocol>`? In Obj-C you can forward declare the protocol using `@protocol DeviceProtocol;`.

billypchan: @mikeger I had tried defined it as id<DeviceProtocol>. It requires define DeviceProtocol with @objc keyword and needs to rewrite some methods to fulfil objective-c compiler requirement. I would prefer do it in another way such as rewrite TermsOfUseStepViewController completely in Swift.

mikeger: Then may be let's not create the initializer that require device, since it is not used during the initialization. Would be enough to have it as the property and set it on the instance?

mikeger:
Conflict, also please check my last comment. Thanks!

PR #2107, opened 2018-05-11
ZIOS-9939: Show camera icon in the user list

nicorsm:
## What's new in this PR?

Added a video icon inside `UserCell`, so in this way, we can display the users that have camera enabled during a call by using the `sendsVideo` property of the `CallParticipantsCellConfiguration` enum. The icon is added to the `UIStackView` of the cell and shown when needed.


nicorsm:
@mikeger `UserCell` has a corresponding test class, I'll try to write a test.

daehn:
@nicorsm it seems like some tests are still failing.

PR #2108, opened 2018-05-11
fix: App stops at launch screen after tried to login with an unverified account

billypchan:
### Issues

After trying to login with an unverified account and restart the app, the app stops at launch screen

### Causes

In AppRootViewController, when (SessionManager.shared?.accountManager.accounts.count == 0), Landing screen will be shown, but in the above SessionManager.shared?.accountManager.accounts.count is 1 and neither RegistrationViewController or LandingViewController will be shown.

### Solutions

In AppStateController, assign authenticationError if account is nil, which happens when the app launchs with no account.

### Note
- [x] tests for AppStateController of different use cases.


typfel:
@mikeger @billypchan so what about including this into the RC?

billypchan:
@typfel I will create some tests for this PR first to make sure it does not affect other use cases.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -128,8 +128,8 @@ extension AppStateController : SessionManagerDelegate {
         let selectedAccount = SessionManager.shared?.accountManager.selectedAccount
 
         // We only care about the error if it concerns the selected account, or the loading account.
-        // ignore the error when account is nil (selectedAccount and loadingAccount are nil also) when app launchs
-        if account != nil && (selectedAccount == account || loadingAccount == account) {
+        if account == nil ||
+           (account != nil && (selectedAccount == account || loadingAccount == account)) {
```
mikeger: The account can be either nil or non-nil. Let's say A = (account == nil), B = (selectedAccount == account || loadingAccount == account), then:
- if A true: true || (false && B) = true
- if A is false: false || (true && B) = B

-> account != nil check is not necessary in the second part.

PR #2109, opened 2018-05-11
Fix/accessibility bold text

johnxnguyen:
### Issues
When the accessibility setting "bold text" is enabled, all lightweight fonts are not rendered bold. The light weight font is the standard font in the app, so to adhere to accessibility guidelines we should make sure they are rendered bold.

### Causes
The light weight fonts (ultra-light, thin, light) do not render as bold (at least not noticeably).

### Solutions
We can check if the bold text setting is enabled, and if so, use the regular font weight whenever a light weight is being used.

### Attachments

Bold text disabled:

img_6267

img_6268

Bold text enabled:

img_6270

img_6269

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -19,6 +19,11 @@ DEPENDENCIES:
   - RBBAnimation (from `https://github.com/wireapp/RBBAnimation`, commit `7dd8d9a3cf610be5f7c1e4459692d555d70704c7`)
   - SCSiriWaveformView (from `https://github.com/wireapp/SCSiriWaveformView`, tag `v1.0.3`)
 
+SPEC REPOS:
+  https://github.com/cocoapods/specs.git:
+    - GoogleAPIClient
+    - GTMSessionFetcher
+
 EXTERNAL SOURCES:
   ARCollectionViewMasonryLayout:
     :git: https://github.com/wireapp/ARCollectionViewMasonryLayout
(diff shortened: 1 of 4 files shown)


billypchan:
@johnxnguyen is it possible not to set the toolbar icons to bold style?

Podfile.lock
```diff
@@ -44,10 +49,10 @@ CHECKOUT OPTIONS:
...
-  GTMSessionFetcher: 5fa5b80fd20e439ef5f545fb2cb3ca6c6714caa2
+  GTMSessionFetcher: ebaa1f79a5366922c1735f1566901f50beba23b7
   RBBAnimation: 74263069442b91350aa4264089c1c05c0395503e
   SCSiriWaveformView: b8b2756af42addcd31900c3945f24edafefb5395
 
-PODFILE CHECKSUM: 5f2dad8ae96f93f0bdb37167e3672155aff83bf1
+PODFILE CHECKSUM: 7fd8d2c0e2667df6f65cde0e829ab67e41259431
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.5.2
```
mikeger: Better not to include that stuff, otherwise we need to update cocoapods, which is not bad, just annoying to do.

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -6177,21 +6106,6 @@
...
-			inputPaths = (
-			);
-			name = "[CP] Copy Pods Resources";
-			outputPaths = (
-			);
-			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "\"${SRCROOT}/Pods/Target Support Files/Pods-WireExtensionComponents/Pods-WireExtensionComponents-resources.sh\"\n";
-			showEnvVarsInLog = 0;
-		};
```
mikeger: Also this stuff seems to be not relevant.

johnxnguyen: ah yeah, I couldn't compile without updating my cocoapods, but I'll take this stuff out :)

johnxnguyen:
@billypchan it seems that bold icons are a result of the rendering mode `UIImageRenderingModeAlwaysTemplate` for the `UIButton` (see this [link](http://averagepro.com/2014/03/17/a-couple-of-ios-7-1-ui-development-surprises-not-the-good-kind/)). I suppose a workaround would be to use the original rendering mode and tint the icons manually.

PR #2110, opened 2018-05-11
 Add overlay showing mute state to self video preview thumbnail

daehn:
## What's new in this PR?

* Add overlay showing mute state to self video preview thumbnail.


vytis:
Interesting that code coverage says almost nothing in the diff is covered by tests...

PR #2112, opened 2018-05-14
ZIOS-9934: Disable camera if phone is held against ear

nicorsm:
### Issues

Camera was not paused on video calls if the phone was held against ear.

### Solutions

I've added a mechanism similar to the one implemented in `AudioMessageView`. The `proximityStateDidChange` method is attached to the Proximity manager and it's called every time the proximity sensor changes his state. Inside that method, I'm starting or pausing the video.


nicorsm:
@daehn tested right now with your changes, it works really well! Good job!

PR #2117, opened 2018-05-14
Feature/Show Data Usage dialogue when log in to a new device

billypchan:
## What's new in this PR?

Create a dialogue for the Crash and Usage reports when the user land on the List UI, before the Notifications dialogue.

## Notes
Test cases:
- [ ] When the user switching account, the dialogue should not be shown

### Attachments

simulator screen shot - iphone x - 2018-05-14 at 17 32 02

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
@@ -29,11 +29,11 @@ extension ConversationListViewController {
         let alertController = UIAlertController(title: "conversation_list.date_usage_permission_alert.title".localized, message: "conversation_list.date_usage_permission_alert.message".localized, preferredStyle: .alert)
 
         alertController.addAction(UIAlertAction(title: "general.accept".localized, style: .default, handler: { (_) in
-//            handler(false)
-            ///TODO: update setting
+            TrackingManager.shared.disableCrashAndAnalyticsSharing = false
         }))
 
         alertController.addAction(UIAlertAction(title: "general.skip".localized, style: .cancel, handler: { (_) in
+            ///TODO: save the timestamp and do not ask again in the future?
         }))
 
 
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
@@ -33,12 +33,9 @@ extension ConversationListViewController {
         }))
 
         alertController.addAction(UIAlertAction(title: "general.skip".localized, style: .cancel, handler: { (_) in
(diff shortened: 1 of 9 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -0,0 +1,42 @@
...
+
+        guard TrackingManager.shared.disableCrashAndAnalyticsSharing else { return }
+
+        let alertController = UIAlertController(title: "conversation_list.date_usage_permission_alert.title".localized, message: "conversation_list.date_usage_permission_alert.message".localized, preferredStyle: .alert)
+
+        alertController.addAction(UIAlertAction(title: "general.accept".localized, style: .default, handler: { (_) in
+            TrackingManager.shared.disableCrashAndAnalyticsSharing = false
+        }))
+
+        alertController.addAction(UIAlertAction(title: "general.skip".localized, style: .cancel, handler: { (_) in
```
daehn: Handler can be set to `nil` in this case.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ConversationListViewController {
+    func showDataUsagePermissionDialogIfNeeded() {
+        guard !isComingFromRegistration,
```
daehn: So the alert should not be shown when coming from registration?

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.h
```diff
@@ -40,6 +40,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic) BOOL isComingFromRegistration;
 
+@property (nonatomic) BOOL needToShowDataUsagePermissionDialog;
```
daehn: Maybe both the setter and getter of this property can be overridden and just get / set the storage in `self.conversationListViewController`.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
```diff
@@ -261,7 +261,7 @@ - (void)setupConversationListViewController
 {
     self.conversationListViewController = [[ConversationListViewController alloc] init];
     self.conversationListViewController.isComingFromRegistration = self.isComingFromRegistration;
-    self.conversationListViewController.needToShowDataUsagePermissionDialog = self.needToShowDataUsagePermissionDialog;
+    self.conversationListViewController.needToShowDataUsagePermissionDialog = false;
```
daehn: I think it's better to use `NO` in Objective-C.

PR #2127, opened 2018-05-16
Ignore all system windows when looking for the top most view controller

typfel:
### Issues

App would get into an infinite loop when asking the `NotificationWindowRootViewController` if it `shouldAutorotate` and crash when long pressing on a message.

### Causes

Long pressing on message shows a context menu, which is displayed in a an window above all other windows. When this is window is displayed the `NotificationWindowRootViewController` would get asked if it `shouldAutorotate` which in turn looks for the top most controller in the front most window. This would after change from yesterday always pick an internal class from apple which displays the context menu.

### Solutions

Ignore all window not created by us when looking for the top most view controller.


typfel:
Caused by: https://github.com/wireapp/wire-ios/pull/2122

PR #2128, opened 2018-05-16
Feature/zios 9990 as a person registering to wire account i want to be asked whether i want to send usage and crash reports to wire when i create any type of account

billypchan:
## What's new in this PR?

Display the dialogue for the Crash and Usage reports at the moment that the user lands on the List UI.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
@@ -21,10 +21,10 @@ import Foundation
 extension ConversationListViewController {
     func showDataUsagePermissionDialogIfNeeded() {
         guard needToShowDataUsagePermissionDialog,
-              ZMUser.selfUser().handle != nil,
               !AutomationHelper.sharedHelper.skipFirstLoginAlerts else { return }
 
-        guard TrackingManager.shared.disableCrashAndAnalyticsSharing else { return }
+        guard isComingFromRegistration ||
+              (!isComingFromRegistration && ZMUser.selfUser().handle != nil && TrackingManager.shared.disableCrashAndAnalyticsSharing) else { return }
 
         let alertController = UIAlertController(title: "conversation_list.date_usage_permission_alert.title".localized, message: "conversation_list.date_usage_permission_alert.message".localized, preferredStyle: .alert)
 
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -21,18 +21,20 @@ import Foundation
 extension ConversationListViewController {
     func showDataUsagePermissionDialogIfNeeded() {
         guard needToShowDataUsagePermissionDialog,
-              ZMUser.selfUser().handle != nil,
               !AutomationHelper.sharedHelper.skipFirstLoginAlerts else { return }
 
-        guard TrackingManager.shared.disableCrashAndAnalyticsSharing else { return }
+        guard isComingFromRegistration ||
+              (!isComingFromRegistration && ZMUser.selfUser().handle != nil && TrackingManager.shared.disableCrashAndAnalyticsSharing) else { return }
```
typfel: You can remove `!isComingFromRegistration` inside the `()` since it can only be reached if `isComingFromRegistration = false`

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -21,18 +21,20 @@ import Foundation
 extension ConversationListViewController {
     func showDataUsagePermissionDialogIfNeeded() {
         guard needToShowDataUsagePermissionDialog,
-              ZMUser.selfUser().handle != nil,
               !AutomationHelper.sharedHelper.skipFirstLoginAlerts else { return }
 
-        guard TrackingManager.shared.disableCrashAndAnalyticsSharing else { return }
+        guard isComingFromRegistration ||
+              (ZMUser.selfUser().handle != nil && TrackingManager.shared.disableCrashAndAnalyticsSharing) else { return }
```
typfel: Why is it relevant to check if  `ZMUser.selfUser().handle != nil` in order to decide if we should ask for data usage permission? Are we waiting for the slow sync to finish?

billypchan: Agree, ZMUser.selfUser().handle is not related to data usage permission setting.

typfel: So we can remove it?

billypchan:
@typfel This alert appears when the user logs in to a new device. needToShowDataUsagePermissionDialog flag is set to true by AppRootViewController when appState changes form .unauthenticated to .authenticated.

mikeger:
@billypchan just for the future, I think that having the shorter and easier to read branch names is helpful, with the 2 line branch name it's super hard to get what it is about.

PR #2129, opened 2018-05-16
ZIOS-10000: Support denied camera access (blurred self preview)

alexaubry:
### Issues

When receiving a video call, the background is black if the user denied camera access or didn't grant it yet.

### Solutions

We display the blurred profile picture of the user and a description label if required.


daehn:
@alexaubry as far as I understood the spec we should blur the self video feed and not the profile picture.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -124,6 +125,20 @@ extension CallInfoConfiguration: CallInfoViewControllerInput {
...
         return ColorScheme.default().variant
     }
+
+    var videoPlaceholderState: CallVideoPlaceholderState {
+
+        guard voiceChannel.isVideoCall else {
+            return .hidden
+        }
+
+        if case .stopped = voiceChannel.videoState {
```
daehn: This should also depend on the call state or not? E.g. if the call is established we should never display the blurred avatar, as the other users feeds might be visible? If no feed at all is visible than the audio UI will / should be shown.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -106,11 +107,27 @@ final class CallViewController: UIViewController {
     }
     
     fileprivate func updateConfiguration() {
+        refreshAuthorizationState()
```
daehn: Not sure if this is actually a big performance hit, but `updateConfiguration` is called very often, is there a way to put this somewhere else (e.g. `viewWillAppear`) to avoid querying the authorisation status so frequently? Of course `callInfoConfiguration.preferedVideoCallState ` should be updated every time the user either changes it in settings or as a response of us asking for permissions.

Podfile.lock
```diff
@@ -44,10 +49,10 @@ CHECKOUT OPTIONS:
...
-  GTMSessionFetcher: 5fa5b80fd20e439ef5f545fb2cb3ca6c6714caa2
+  GTMSessionFetcher: ebaa1f79a5366922c1735f1566901f50beba23b7
   RBBAnimation: 74263069442b91350aa4264089c1c05c0395503e
   SCSiriWaveformView: b8b2756af42addcd31900c3945f24edafefb5395
 
-PODFILE CHECKSUM: 5f2dad8ae96f93f0bdb37167e3672155aff83bf1
+PODFILE CHECKSUM: 7fd8d2c0e2667df6f65cde0e829ab67e41259431
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.5.2
```
mikeger: It's nice to update the CocoaPods, but this can break our CI and also everyone needs to update the pods on their ends after this one would be merged. TL;DR: better to create the separate PR for it.

alexaubry: I hadn't noticed I committed it. Reverted this in 0dd555af.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
```diff
@@ -195,5 +199,10 @@ final class CallActionsView: UIView {
         default: fatalError("Unexpected Button: \(button)")
         }
     }
+
+    override var intrinsicContentSize: CGSize {
```
typfel: Why do we need to define `intrinsicContentSize` for the `CallActionsVIew` now?

alexaubry: I have seen cases where the actions view would take the entire height of the info stack view and it would compress the accessory. That seemed to fix the issue. Do you think it's necessary?

alexaubry: When we hide the participants list to display the "Wire doesn't have access to the camera" label, the stack view completely hides the accessory view controller and this results in this if we don't define the intrinsic content size of the actions button:

screen shot 2018-05-22 at 14 26 58

typfel: Oh ok I haven't seen that maybe we need to do it then, but this seems unrelated to task. If it's not required by your other changes I would prefer to have this layout fix as separate PR.

typfel: Ah didn't see you second message. So it is dependent on your changes.

alexaubry: Yeah, I needed to add the intrinsic content size so that the camera access denied message I added in this PR is displayed. Otherwise we can't see this label:

testgroupvideodeniedvideopermissions_ 375 667 2x

daehn: I think it would be good to put the label into the accessory view

daehn: So adding the label to the accessory view, centering it in there and then the accessory view should decide depending on it's state if it should be hiding or showing it .

daehn: Because right now there is already a state in which nothing is displayed, but the accessory view is still not hidden, it just internally hides it's subviews.

daehn: So basically adding a new enum case to `CallInfoViewControllerAccessoryType` an making the configuration return this when it's appropriate should work. In general as this label is contained in the center area which is managed by the accessory controller it makes more sense to move all ownership and control over the label into that component I think,

daehn: Ah sorry, I just saw that the label is already managed by the accessory view with your changes 😅.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallAccessoryViewController.swift
```diff
@@ -56,16 +57,26 @@ final class CallAccessoryViewController: UIViewController, CallParticipantsViewC
...
+        videoPlaceholderStatusLabel.text = "video_call.camera_access.denied".localized
     }
     
     private func createConstraints() {
         participantsViewController.view.translatesAutoresizingMaskIntoConstraints = false
         avatarView.translatesAutoresizingMaskIntoConstraints = false
+        videoPlaceholderStatusLabel.translatesAutoresizingMaskIntoConstraints = false
         participantsViewController.view.fitInSuperview()
         avatarView.fitInSuperview()
+        videoPlaceholderStatusLabel.fitInSuperview()
```
daehn: This might prevent the accessory view from growing vertically if I'm not mistaken. Maybe the aforementioned workaround is not needed when the label is pinned to the center instead?

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
alexaubry: @typfel @daehn I fixed this accessory label issue in my latest commit.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
```diff
@@ -176,6 +178,8 @@ final class CallActionsView: UIView {
         }()
         topStackView.spacing = isCompact ? verticalStackView.spacing : 32
         bottomStackView.spacing = isCompact ? verticalStackView.spacing : 32
+
+        invalidateIntrinsicContentSize()
```
typfel: This can also be deleted?

alexaubry: Oh sorry I forgot to remove it.

PR #2131, opened 2018-05-16
ZIOS-9994: Username not visible in calling overlay during calls with users with a profile picture

nicorsm:
### Issues

Username and calling duration labels were not visible during a call with a contact that has a profile picture.

### Causes

This was caused by the internal `UIImageView` stretching priority settings, automatically set to 750 instead of the expected value (249).

### Solutions

I'm applying the new stretching priority to both views, the `userImageView` container and the inner `imageView`.


daehn:
@nicorsm nice catch!

PR #2133, opened 2018-05-16
Fix terminating state when hanging up

typfel:
### Issues

When hanging up a call the call UI would switch to the incoming call layout for the action buttons.

### Causes

When leaving a call SE will treat it as having the state `.incoming(video: false, shouldRing: false, degraded: false)`.

### Solutions

Display incoming calls with `shouldRing = false` as terminating


typfel:
@daehn  It was fixed for the status labels but not the action buttons

mikeger:
@typfel no tests affected?

typfel:
@mikeger let's see :)

PR #2139, opened 2018-05-17
Update AVS to 4.4..22

typfel:
## What's new in this PR?

AVS has been updated to 4.4.22 where we control when the video capture is started or stopped.


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -76,18 +75,33 @@ class VideoGridViewController: UIViewController {
...
     
     func updateState() {
-        updateFloatingVideo(with: configuration.floatingVideoStream)
+        let selfStreamId = ZMUser.selfUser().remoteIdentifier!
+        let selfInGrid = configuration.videoStreams.contains { $0.stream == selfStreamId }
+        let selfInFloatingOverlay = nil != configuration.floatingVideoStream
+        let isShowingSelf = selfInGrid || selfInFloatingOverlay
+        
+        // Create self preview if there is none but we should show it
+        if isShowingSelf && nil == selfPreviewView {
```
mikeger: May be `if let _ = ` is more swifty than comparing with nil?

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -76,18 +75,33 @@ class VideoGridViewController: UIViewController {
...
+            selfPreviewView?.translatesAutoresizingMaskIntoConstraints = false
+        }
+        
         updateVideoGrid(with: configuration.videoStreams)
+        updateFloatingVideo(with: configuration.floatingVideoStream)
+        
+        // Clear self preview we we shouldn't show it anymore
+        if !isShowingSelf, let _ = selfPreviewView {
+            selfPreviewView = nil
+            precondition(selfPreviewView == nil)
```
mikeger: Not sure about the logic here, you just set it to nil, so may be no reason to check?

typfel: It's a check that nothing else still holds a reference to it like it being in the view hierarchy.

mikeger: Well this would not work I think, since you just nullified that same reference one line before, it cannot be not nil? Or I am missing something? I.e. even if something is still holding to that object, our reference was nullified.

typfel: It's a property :) but it's hidden without syntax highlighting

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
typfel: Not sure about this  `== nil`  is more obvious I think. But anyway maybe we can discuss this at Friday afternoon 🍻

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
typfel: Ah wasn't thinking straight. You're right the check Is pointless, I'll remove it.

PR #2154, opened 2018-05-18
FeatureI want to decide if i want to receive news and offers when i join a team

billypchan:
## What's new in this PR?

Show the data usage dialog when a user first time login to Wire, i.e. the user had set the user name in the login workflow.

Add checks to prevent data usage dialog appear more then once.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -21,10 +21,11 @@ import Foundation
 extension ConversationListViewController {
     func showDataUsagePermissionDialogIfNeeded() {
         guard !AutomationHelper.sharedHelper.skipFirstLoginAlerts else { return }
-
+        guard !dataUsagePermissionDialogDisplayed else { return }
         guard needToShowDataUsagePermissionDialog else { return }
 
         guard isComingFromRegistration ||
+              isComingFromSetUsername ||
```
mikeger: Should we also check if user has the team, i.e. `ZMUser.selfUser().isTeamUser` ? It can be that the old Wire user is signing now in, the one that has no team, but also no user name: should the takeover be also displayed for such user?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController.m
```diff
@@ -114,6 +114,11 @@ @interface ConversationListViewController ()
 
 @property (nonatomic) CGFloat contentControllerBottomInset;
 
+/// for data usage dialog
+@property (nonatomic) BOOL isViewDidAppear;
```
mikeger: @johnxnguyen what would be the correct way to formulate it in English? `viewAppearedBefore`, `viewUsedToAppear` ? The logic is: viewDidAppear was called.

PR #2165, opened 2018-05-23
Update strings for GDPR compliance release

billypchan:
## What's new in this PR?

Update two sentences in English and localization in other languages.


mikeger:
@billypchan you need to add all the new strings to all languages, for example "news_offers.consent.title" does not exist in Arabic, so the user would see localized string title instead of the text in English.

mikeger:
Ah, I see it now. Sorry.

PR #2169, opened 2018-05-23
Disable video for calls in large groups

typfel:
### Issues

Design want the video button to be disabled in large group but still show an alert when it's tapped informing the user why video calling is not possible.

### Solutions

Since a disabled button is not responding to touches we place a ghost button on top of the video button which is only enabled when the video button is disabled.


typfel:
@mikeger no screenshots affected this time.

PR #2172, opened 2018-05-23
ZIOS-10000: Disable video button on incoming call without permission

alexaubry:
### Issues

When the user denied camera permissions and receives an incoming call, the "video" button is displayed, and when they tap it it shows an alert to take them to the settings app.

It was requested by design that we disabled this button when the permissions are denied, and that we show the speaker icon instead of the flip button, since audio will be the only possible calling mode.

### Solutions

Instead of determining the video permissions when the user accepts the call, we determine them when setting up the UI so that we can show the correct buttons.

## Notes

Please test this branch on a physical device before approving.


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -76,7 +76,7 @@ extension CallInfoConfiguration: CallInfoViewControllerInput {
         case .outgoing, .incoming(video: false, shouldRing: _, degraded: _):
             return false
         default:
-            return true
+            return (permissions.canAcceptVideoCalls || permissions.isPendingVideoPermissionRequest) && (permissions.canAcceptAudioCalls || permissions.isPendingAudioPermissionRequest)
```
typfel: Why is it important to check we that we have audio permissions here?

alexaubry: We only request audio permissions once the user taps the "Accept call" button, so I thought we should disabled the toggle call mode in the case the user didn't grand audio permission yet. Not sure if this is necessary though.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -97,13 +98,13 @@ final class CallViewController: UIViewController {
 
     @objc private func resumeVideoIfNeeded() {
         guard voiceChannel.isVideoCall, voiceChannel.videoState.isPaused else { return }
-        voiceChannel.videoState = .started
+        voiceChannel.videoState = permissions.videoStateIfAllowed(.started)
```
typfel: The app will restart if the user changes any permission in settings this is not necessary I think.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
typfel: I see, but we don't need audio permission to show your own video preview I assume? If that's the case I think it's not necessary.

typfel:
Conflicts 💣

alexaubry:
@typfel Could you please verify the conflict resolution commit  aa47bb5acadf69c33392cc945f4417857a5aa79b?

typfel:
@alexaubry looks good.

typfel:
@alexaubry tests are failing and maybe we should a test case in `CallInfoConfigurationtTests`

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -76,12 +76,19 @@ extension CallInfoConfiguration: CallInfoViewControllerInput {
         case .outgoing, .incoming(video: false, shouldRing: _, degraded: _):
             return false
         default:
-            if voiceChannel.videoState == .stopped {
+            // When the call is incoming, we should check if the user has permission and if they're eligible for video calls
+            if case .incoming = voiceChannel.state {
                 let canAcceptMediaCall = !permissions.isVideoDisabledForever && !permissions.isAudioDisabledForever
                 return (voiceChannel.conversation?.isEligibleForVideoCalls ?? false) && canAcceptMediaCall
```
typfel: We should not prevent prevent non team users from accepting incoming video calls in group conversations.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
```diff
@@ -172,6 +172,12 @@ final class CallActionsView: UIView {
         acceptCallButton.isHidden = !input.canAccept
         firstBottomRowSpacer.isHidden = input.canAccept || isCompact
         secondBottomRowSpacer.isHidden = isCompact
+
+        if !input.permissions.canAcceptVideoCalls {
```
typfel: In what scenario is this necessary? Isn't the media state already `notSendingVideo` here?

alexaubry: This `if` handles the case where the user did not accept or deny the permissions yet (in this case the media state is `sendingVideo`).

typfel: I see. How about moving this logic to the `CallInfoConfiguration.mediaState`and have it return `.notSendingVideo` for this scenario then it would be testable in a logic test.

neongreen:
congratulations on the 10000 get! :tada:

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
alexaubry: @typfel I've updated `mediaState` to handle this case, and added tests for these cases.

Wire-iOS Tests/Mocks/MockCallInfoViewControllerInput.swift
```diff
@@ -28,10 +29,35 @@ struct MockCallInfoViewControllerInput: CallInfoViewControllerInput {
...
-    var mediaState: MediaState
+    var preferredMediaState: MediaState
     var state: CallStatusViewState
     var isConstantBitRate: Bool
     var title: String
     var isVideoCall: Bool
     var variant: ColorSchemeVariant
+
+    var mediaState: MediaState {
+        return permissions.mediaStateIfAllowed(preferredMediaState)
```
typfel: What's the reason for running the `mediaState` throw this function only in tests?

PR #2173, opened 2018-05-23
Fix the constraints for the top bar

mikeger:
### Issues

The minimized call bar stopped appearing.

### Cause

The changes with the iOS status bar style are clashed with the changes from the call bar minimization:
 the new constraints logic was not taking the top overlay container in account.

### Fix

Take the top overlay in account when creating the constraints.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
@@ -100,11 +100,10 @@ extension ZClientViewController {
         topOverlayContainer.translatesAutoresizingMaskIntoConstraints = false
         view.addSubview(topOverlayContainer)
 
-        contentTopRegularConstraint = splitViewController.view.topAnchor.constraint(equalTo: safeTopAnchor)
-        contentTopCompactConstraint = splitViewController.view.topAnchor.constraint(equalTo: view.topAnchor)
+        contentTopRegularConstraint = topOverlayContainer.topAnchor.constraint(equalTo: safeTopAnchor)
+        contentTopCompactConstraint = topOverlayContainer.topAnchor.constraint(equalTo: view.topAnchor)
         
         NSLayoutConstraint.activate([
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -100,11 +100,10 @@ extension ZClientViewController {
...
-        contentTopRegularConstraint = splitViewController.view.topAnchor.constraint(equalTo: safeTopAnchor)
-        contentTopCompactConstraint = splitViewController.view.topAnchor.constraint(equalTo: view.topAnchor)
+        contentTopRegularConstraint = topOverlayContainer.topAnchor.constraint(equalTo: safeTopAnchor)
+        contentTopCompactConstraint = topOverlayContainer.topAnchor.constraint(equalTo: view.topAnchor)
         
         NSLayoutConstraint.activate([
-            topOverlayContainer.topAnchor.constraint(equalTo: view.topAnchor),
             topOverlayContainer.leadingAnchor.constraint(equalTo: view.leadingAnchor),
             topOverlayContainer.trailingAnchor.constraint(equalTo: view.trailingAnchor),
             topOverlayContainer.bottomAnchor.constraint(equalTo: splitViewController.view.topAnchor),
```
alexaubry: Since the top overlay will be positioned on top of the content, shouldn't we constraint the top of the split view controller to the bottom of the top overlay container?

mikeger: This is exactly what we are doing in this line! It's topOverlay.bottom on the left :)

alexaubry: I meant, shouldn't we do this opposite and replace this line with

```swift
splitViewController.view.topAnchor.constraint(equalTo: topOverlayContainer.bottomAnchor) 
```

mikeger: Is there a difference in the arguments order in this case? I.e. what would it change.

PR #2175, opened 2018-05-23
Always reload using diff calculation system

daehn:
## What's new in this PR?

* Always reload using diff calculation system.


typfel:
Does this mean that all changes are now animated, even the initial one?

typfel:
But it looks alright I assume?

daehn:
@typfel yeah it looks better in my opinion.

PR #2176, opened 2018-05-24
ZIOS-9798: Add guest icon to guests in share screen

nicorsm:
## What's new in this PR?

I've added the guest icon inside `ShareDestinationCell`, the cell used in `ShareViewController`. To do so, I've refactored the design of the view by switching to `UIStackView`, so we can avoid constraint calculation for showing and hiding the icons inside the view. I've also added UI tests, similar to the ones already available.


Wire-iOS/Sources/Components/ShareViewController/ShareDestinationCell.swift
```diff
@@ -65,56 +85,54 @@ final class ShareDestinationCell<D: ShareDestination>: UITableViewCell {
...
+        self.contentView.addSubview(self.stackView)
+        
+        self.stackView.addArrangedSubview(avatarViewContainer)
+        constrain(self.contentView, self.avatarViewContainer) { contentView, avatarView in
+            avatarView.centerY == contentView.centerY
+            avatarView.width == self.avatarSize
+            avatarView.height == self.avatarSize
+        }
+        
+        //self.titleLabel.font = UIFont.systemFont(ofSize: 16.0, weight: UIFontWeightLight)
```
daehn: Can be removed?

nicorsm: Yes, little leftover from UI tests 😁

PR #2178, opened 2018-05-24
Remove diffing framework and participants change animations

daehn:
## What's new in this PR?

* Remove diffing framework and participants change animations.


typfel:
@daehn tests failed

PR #2179, opened 2018-05-24
Disable idle timer while video calling

daehn:
## What's new in this PR?

* Disable idle timer while video calling.

### This PR depends on:

- [x] Some final testing still has to be done.

### DIFF
--- a/Wire-iOS Tests/CallInfoTestFixture.swift
+++ b/Wire-iOS Tests/CallInfoTestFixture.swift
@@ -51,7 +51,8 @@ struct CallInfoTestFixture {
             isConstantBitRate: false,
             title: otherUser.displayName,
             isVideoCall: false,
-            variant: .light
+            variant: .light,
+            disableIdleTimer: false
         )
     }
     
@@ -70,7 +71,8 @@ struct CallInfoTestFixture {
             isConstantBitRate: false,
             title: otherUser.displayName,
             isVideoCall: false,
-            variant: .light
+            variant: .light,
+            disableIdleTimer: false
         )
     }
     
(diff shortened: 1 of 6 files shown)


Wire-iOS Tests/CallInfoTestFixture.swift
```diff
@@ -456,7 +477,8 @@ struct CallInfoTestFixture {
             isConstantBitRate: false,
             title: otherUser.displayName,
             isVideoCall: true,
-            variant: .light
+            variant: .light,
+            disableIdleTimer: false
```
typfel: Shouldn't `disableIdleTimer` be true in this case?

Wire-iOS Tests/Mocks/MockCallInfoViewControllerInput.swift
```diff
@@ -34,4 +34,5 @@ struct MockCallInfoViewControllerInput: CallInfoViewControllerInput {
     var title: String
     var isVideoCall: Bool
     var variant: ColorSchemeVariant
+    var disableIdleTimer: Bool
```
typfel: I guess this should be added to the Equality function?

Wire-iOS Tests/CallInfoTestFixture.swift
daehn: I just filled it in with false everywhere in the tests as it shouldn't have any effect there anyways.

daehn: Could maybe also generally add the implementation of this as a protocol extension.

daehn:
@typfel updated!

PR #2181, opened 2018-05-24
ZIOS-9944: Prevent audio messages and self profile picture changes while on call

nicorsm:
## What's new in this PR?

- Added logic to avoid the use of audio messages and camera on self profile view controller during a call
- Moved logic to a proper separate file, detached from `ConversationViewController`.
- Unified alert messages.


Wire-iOS/Sources/Components/CameraAccess.swift
```diff
@@ -0,0 +1,50 @@
...
+@objc enum CameraAccessFeature: Int {
+    case recordVideo
+    case recordAudioMessage
+    case takePhoto
+}
+
+public class CameraAccess: NSObject {
+
+    static var shouldBlockCallingRelatedActions: Bool {
+        guard let session = ZMUserSession.shared(),
```
typfel: We added a helper method on the user session `isCallOngoing`.

Wire-iOS/Sources/Components/CameraAccess.swift
```diff
@@ -0,0 +1,44 @@
...
+
+@objc enum CameraAccessFeature: Int {
+    case recordVideo
+    case recordAudioMessage
+    case takePhoto
+}
+
+public class CameraAccess: NSObject {
+
+    static func displayCameraAlertForOngoingCall(at point: CameraAccessFeature, from viewController: UIViewController) {
```
daehn: I'd maybe rename the parameter `point` to `feature`.

PR #2186, opened 2018-05-25
Convert CallInfoConfiguration to stored properties

daehn:
## What's new in this PR?

I was trying to reduce the number of UI reloads we do when the call changes and then realized that when checking if two instances of `CallInfoConfiguration` are equal it always gets reported as `true`. 

This issue was due to the fact that all of the properties of `CallInfoConfiguration` are computed and based on the stored `VoiceChannel` reference type, so when comparing two different instances they will internally access the same voice channel and return the same results for all variables. As this was not what we had in mind and want a immutable struct to be passed down as configuration I now changed this and all properties are now stored and only once computed in init.

Besides that 9d050cc fixes an issue with the timing of the mute overlay animation in the self preview which is now in sync with the general UI overlay show / hide animation.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
@@ -40,8 +40,9 @@ final class CallInfoRootViewController: UIViewController, UINavigationController
         }
     }
     
-    var configuration: CallInfoViewControllerInput {
+    var configuration: CallInfoViewControllerInput  {
         didSet {
+            guard !configuration.isEqual(toConfiguration: oldValue) else { return }
             updateConfiguration(animated: true)
         }
     }
--- a/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
@@ -40,7 +40,7 @@ final class CallInfoRootViewController: UIViewController, UINavigationController
         }
     }
     
-    var configuration: CallInfoViewControllerInput  {
+    var configuration: CallInfoViewControllerInput {
         didSet {
             guard !configuration.isEqual(toConfiguration: oldValue) else { return }
             updateConfiguration(animated: true)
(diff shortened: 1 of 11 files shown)


daehn:
Fixing tests right now

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -69,99 +59,117 @@ extension CallInfoConfiguration: CallInfoViewControllerInput {
...
+    let isMuted: Bool
+    let isTerminating: Bool
+    let canAccept: Bool
+    let mediaState: MediaState
+    let accessoryType: CallInfoViewControllerAccessoryType
+    let degradationState: CallDegradationState
+    let videoPlaceholderState: CallVideoPlaceholderState
+    let disableIdleTimer: Bool
+
+    let voiceChannel: VoiceChannel
```
typfel: Do we still need to keep a reference this the voice channel now?

typfel:
Tests doesn't compile.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
daehn: No, good point!

daehn: @typfel updated!

daehn:
This broke the call duration timer, investigating right now.

daehn:
@typfel start date makes it very annoying to test the status view in snapshot tests due to timing issues though. I fixed it and the `state` property is the only one that's computed now, but it's capturing relevant information in a snapshot struct so the only thing that can change is the call duration.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
```diff
@@ -40,8 +40,9 @@ final class CallInfoRootViewController: UIViewController, UINavigationController
         }
     }
     
-    var configuration: CallInfoViewControllerInput {
+    var configuration: CallInfoViewControllerInput  {
```
typfel: Double space

PR #2188, opened 2018-05-25
Revise ongoing call messages & fix typography

infotexture:
Minor copy changes


mikeger:
I need to fix the tests, please stay tuned.

PR #2191, opened 2018-05-28
Add initial call haptic feedback implementation

daehn:
## What's new in this PR?

* Add initial call haptic feedback implementation.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
@@ -1,0 +1,98 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
```diff
@@ -0,0 +1,98 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+final class CallHapticsController {
+    
+    private let hapticGenerator: CallHapticsGeneratorType?
+    private let voiceChannel: VoiceChannel
```
typfel: Is the `voiceChannel` used for anything since the state is passed in from the `CallViewController`?

daehn: Might not be used anymore, was using it before to get the participants states.

daehn: @typfel updated!

PR #2194, opened 2018-05-28
ZIOS-10038: iOS10 Self profile screen layout issues

nicorsm:
### Issues

Self profile view, on iOS 10:
- the content of the view overlaps the `titleView`
- the account selection view is aligned on the left, instead of center

### Solutions

- I've enabled the calculation of the top offset for iOS 10 (instead of iOS 11)
- I'm setting the frame of the `AccountSelectionController`, so it can take the size from the inner accounts view.


Wire-iOS/Sources/UserInterface/SelfProfile/AccountSelectorController.swift
```diff
@@ -58,6 +58,7 @@ final class AccountSelectorController: UIViewController {
     private func setShowAccounts(to showAccounts: Bool) {
         self.showAccounts = showAccounts
         accountsView.isHidden = !showAccounts
+        self.view.frame = accountsView.frame
```
mikeger: It would be formally correct to assign the size in the frame to accountsView size. Since those views belong to the different superviews (actually one is the superview of another), assigning the frame that has the position does not make sense (I know that in this case they both are starting from 0 position, just trying to bother you :) )

PR #2196, opened 2018-05-28
Fix: older photo in search feature in a group chat cannot be saved

billypchan:
### Issues

The older photo in search feature in a group chat cannot be saved without any error.

### Causes

SavableImage is released when its owner ConversationContentViewController is dismissed. The saving process return when imageData is nil.

### Solutions

Remove [weak self] in the SavableImage.saveToLibrary func's closures.

## Notes

Tests are added to check SavableImage object is not retained after saveToLibrary is called and save process is done after its owner is released.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+SaveImage.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ConversationContentViewController {
+    @objc(saveImageFromMessage:cell:)
+    func saveImage(from message: ZMConversationMessage?, cell: ImageMessageCell?) {
```
mikeger: I think from the way this method is called, the message should not ever be nil, so I would change the signature.

mikeger: How it's called: https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m#L371

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/SavableImage.swift
```diff
@@ -19,10 +19,27 @@
...
+}
+
+extension PHAssetChangeRequest: AssetChangeRequestProtocol {}
+
 @objc final public class SavableImage: NSObject {
 
+    /// protocols for inject mocking photo services
+    var photoLibrary: PhotoLibraryProtocol = PHPhotoLibrary.shared()
+    var assetChangeRequestType: AssetChangeRequestProtocol.Type = PHAssetChangeRequest.self
+    var applicationType: ApplicationProtocol.Type = UIApplication.self
```
mikeger: Would be cool to do the following:
1. I think that `wr_requestOrWarnAboutPhotoLibraryAccess` method is not coupled to the application in any way, so let's move it to be the part of `PhotoLibraryProtocol` and make it not static.
2. Remove `applicationType` property, since now you only need `photoLibrary`.
3. Optionally, do the same with `assetChangeRequestType`: implement the method with the same functionality in the category on `PHPhotoLibrary` and add the method to `PhotoLibraryProtocol`.

billypchan: @mikeger wr_requestOrWarnAboutPhotoLibraryAccess is coupled to UIApplication because it has to call several other static UIApplication.wr_warnAboutXXX functions to show some alert boxes. I would prefer to keep it for mocking injection.

Similarly, assetChangeRequestType is for mocking PHAssetChangeRequest, to bypass the creationRequestForAsset calling in tests.

mikeger:
@billypchan tests failed

PR #2203, opened 2018-05-29
Boostrap with CocoaPods 1.5.3

mikeger:
Do `gem install cocoapods` in order to update your cocoapods to the most recent one (1.5.3).


typfel:
@mikeger Is cocoa pods updated on Jenkins?

mikeger:
I updated cocoapods for user jenkins. I still see in the logs that it's using 1.3.x version 🙈 However the building is not distracted, since all jobs do `pod install` first and this replaces the lock file 🎊

PR #2205, opened 2018-05-29
ZIOS-9850: Refactor in-app link opening

alexaubry:
### Issues

The links in the settings were opened in Safari.app. It was requested to change this behavior and open those in a Safari view controller. Links in messages are not affected.

### Solutions

In the Settings factory, we return an external screen item that opens a Safari view controller on tap.

The API for opening URLs was also refined to be easier to use. When you want to open a URL, you can use a `BrowserViewController` which is now a subclass of `SFSafariViewController`. It automatically handles the black tint color requested by design + status bar appearance.

To open an URL in app, call `url.openInApp(above: viewController)`.

The Wire URLs are also now defined Swift, so that we don't need to cast them when creating the Safari view controller.


Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -46,3 +46,73 @@ extension NSURL {
         return URL.manageTeam(source: source)
     }
 }
+
+// MARK: - Standard URLS
+
+extension URL {
```
mikeger: Might it be slightly easier to re-implement the whole extension in Swift?

alexaubry: Agreed, but some URLs are still used by the Objective-C parts of the app though. What should we do about these? Leave them in Obj-C or move them to an NSURL extension in Swift?

alexaubry:
@daehn Yes, we're only using this for the settings / support links listed in the ticket, all the rest still opens in the browser.

PR #2211, opened 2018-05-30
Calculate footer size correctly to accomodate for dynamic type

daehn:
## What's new in this PR?

* Calculate footer size correctly to accomodate for dynamic type.


daehn:
@mikeger no, I don't think so 😬

PR #2212, opened 2018-05-30
Fix: Missing navigation button in device list screen

billypchan:
### Issues

Missing navigation button in device list screen when the user opens ProfileViewController from tapping "User started using a new device" link on the conversation screen.

### Causes

ProfileViewController is wrapped in a normal navigation controller without customised properties.

### Solutions

Wraps ProfileViewController with UIViewController.wrapInNavigationController() method.

## Notes

Added snapshot tests for ProfileViewController.


Wire-iOS Tests/Mocks/MockUserClient.m
```diff
@@ -39,4 +39,9 @@ - (void)resetSession
     // No-op
 }
 
+- (NSString *) deviceClass
```
mikeger: Why this is necessary?

mikeger: Called from the profile view I guess?

billypchan: For providing a mocking deviceClass string showing on the cell.

PR #2214, opened 2018-05-30
Removed the camera view controller: replaced with iOS UI

mikeger:
The Wire integrated camera UI was originally created back in the day when the iOS integrated camera capture did not support the number of features. Over the time unfortunately the great camera UI become rather outdated comparing to the iOS one. 

The effort to support the custom camera UI is unnecessary big, therefore I removed the old camera UI code and replaced it with the `UIImagePickerViewController` implementation.

Attention: needs to be tested on iPad.


mikeger:
@typfel yes, let me test it there first, had no iPad at home 🤷‍♂️

PR #2216, opened 2018-05-31
Fix: in-app browser is blank when open marketing consent dialog's privacy web page

billypchan:
### Issues
After https://github.com/wireapp/wire-ios/pull/2205 is merged, In-app browser is blank when open marketing consent dialog's privacy web page.

### Causes

SFSafariViewController cannot render correctly if it is not presented with the same window as the current view controller.

### Solutions

Present BrowserViewController from the same view controller instead of notificationsWindow.rootViewController.

## Notes

After testing, I found that BrowserViewController does not call completion closure. Rewrite BrowserViewController in a simpler way to fix it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.swift
@@ -21,32 +21,13 @@ import SafariServices
 
 @objc class BrowserViewController: SFSafariViewController {
 
-    // MARK: - Events
-
-    private class BrowserDelegate: NSObject, SFSafariViewControllerDelegate {
-
-        var completion: () -> Void
-
-        init(completion: @escaping () -> Void) {
-            self.completion = completion
-        }
-
-        func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
-            completion()
-        }
-
-    }
-
     @objc var completion: (() -> Void)? {
-        get {
-            return (delegate as? BrowserDelegate)?.completion
-        }
-        set {
-            guard let block = newValue else {
+        didSet {
+            guard let _ = completion else {
                 delegate = nil
                 return
             }
-            delegate = BrowserDelegate(completion: block)
+            delegate = self
         }
     }
 
@@ -77,4 +58,10 @@ import SafariServices
         UIApplication.shared.wr_setStatusBarStyle(originalStatusBarStyle, animated: true)
     }
 
+}
+
+extension BrowserViewController: SFSafariViewControllerDelegate {
+    func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+        completion?()
+    }
 }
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/UIAlertController+Newsletter.swift
```diff
@@ -25,7 +25,7 @@ extension UIAlertController {
     /// email regisration work flow: newsletter subscription dialog appears after conversation list is displayed.)
     static var newsletterSubscriptionDialogWasDisplayed = false
 
-    static func showNewsletterSubscriptionDialog(completionHandler: @escaping (Bool) -> Void) {
+    static func showNewsletterSubscriptionDialog(presentViewController:UIViewController, completionHandler: @escaping (Bool) -> Void) {
```
mikeger: Would be slightly more clear I think to create the signature like showNewsletterSubscriptionDialog(over viewController: UIViewController, ...`

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationFlowController.swift
```diff
@@ -123,9 +123,17 @@ extension TeamCreationFlowController {
         }
     }
 
+    fileprivate func showMarketingConsentDialog(presentViewController: UIViewController) {
+        UIAlertController.newsletterSubscriptionDialogWasDisplayed = false
+        UIAlertController.showNewsletterSubscriptionDialogIfNeeded(presentViewController: presentViewController) { [weak self] marketingconsent in
```
daehn: `marketingconsent` → `marketingConsent`

Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/Flow/TeamCreationFlowController.swift
```diff
@@ -123,9 +123,17 @@ extension TeamCreationFlowController {
...
+            self?.marketingConsent = marketingconsent
+        }
+    }
+
     fileprivate func pushController(for state: TeamCreationState) {
 
         var stepDescription: TeamCreationStepDescription?
+        var needToShowMarketingConsentDialog = false
```
daehn: I'd rename this `needsToShowMarketingConsentDialog`

Wire-iOS/Sources/UserInterface/Components/Controllers/BrowserViewController.swift
```diff
@@ -21,32 +21,13 @@ import SafariServices
...
-
-    }
-
     @objc var completion: (() -> Void)? {
-        get {
-            return (delegate as? BrowserDelegate)?.completion
-        }
-        set {
-            guard let block = newValue else {
+        didSet {
```
daehn: Could also override `viewDidLoad` and just set the delegate once in there. Then the whole `didSet` logic could go.

PR #2218, opened 2018-05-31
Fix: automation and accessibility fixes for self profile

mikeger:
### Issues

The user image view was not visible any more for the automation or accessibility.

### Solutions

Variety of fixes for the self profile screen.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -687,6 +687,8 @@
 "self.voiceover.hint" = "Open profile and settings";
 "self.new-device.voiceover.label" = "Profile, new devices added";
 
+"self.profile.change_user_image.accessibility" = "Tap to change user image";
```
alexaubry: Should we say "profile picture" instead of "user image"?

PR #2219, opened 2018-05-31
Improve corner radius performance + appearance

alexaubry:
### Issues

Right now, we use `cornerRadius` in most places to round the corners of views. At UIKonf, Luke Parham [explained](https://www.youtube.com/watch?v=gnGU5v3kVbI) (ca. 17min. in the video) that this property can be computationally intensive.

Instead of using this property to round corners, we introduce a `RoundedView` that rounds its corners with a layer mask and a `UIBezierPath`. By using it, we get:

- Better performance
- Slightly better looking corner radius ([source](https://twitter.com/argentumko/status/955773459463790592?lang=en))

### Solutions

A custom layer subclass (`ContinuousMaskLayer`) handles the rounding of the corners. You can specify the shape (either circle, rectangle or rounded with a radius). It automatically handles bounds changes, so subviews don't need to manually change the radius in `layoutSubviews`.

The `RoundedViewProtocol` protocol provides direct access  to the properties of the layer. This is useful if we ever decide to create a `RoundedButton` for example, we won't need to duplicate the code to cast the layer.

In `RoundedView`, we set the layer class to `ContinuousMaskLayer` and provide some utilities for Objective-C. 

## Notes

We should audit the codebase to see in which parts we need to use this new view. At the moment, it is used by the user image view, the thumbnail in video calls and the call survey.

In the future, we should use this view by default when we need to round the corners.


mikeger:
@alexaubry I am not 100% sure, but some of the screenshots might contain the user image views, so they would need to be re-rendered in case the layer masking is rendering the slightly different bitmap.

alexaubry:
@mikeger Yes, I noticed this. I am updating the snapshot tests at the moment 👍

PR #2220, opened 2018-05-31
Update tracking for GVC

typfel:
## What's new in this PR?

Update tracking for GVC.

## Note

- [x] Which call closed reasons should be included are still being discussed.


Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -0,0 +1,121 @@
...
+
+extension Analytics {
+    
+    func tag(callEvent: CallEvent, in conversation: ZMConversation, callInfo: CallInfo) {        
+        tagEvent(callEvent.eventName, attributes: attributes(for: callEvent, callInfo: callInfo, conversation: conversation))
+    }
+    
+    private func attributes(for event: CallEvent, callInfo: CallInfo, conversation: ZMConversation) -> [String : Any] {
+        var attributes = attributesForConversation(conversation)
+        attributes.merge(attributesForAccount(in: conversation)) { (_, new) in new }
```
mikeger: May be it's worth creating the predefined function that prefers the new values over the old ones, give it a name and pass it here in? Like `attributes.merge(something, .preferNew)`.

PR #2223, opened 2018-06-01
Some fixes for automation

mikeger:
- Fixed the accessibility ID of the call status view.
- Fixed usage of TARGET_OS_SIMULATOR in Swift.


Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallStatusView.swift
```diff
@@ -103,8 +103,7 @@ final class CallStatusView: UIView {
         stackView.translatesAutoresizingMaskIntoConstraints = false
         bitrateLabel.translatesAutoresizingMaskIntoConstraints = false
         stackView.spacing = 12
-        titleLabel.accessibilityIdentifier = "CallStatusLabel"
-        subtitleLabel.accessibilityIdentifier = "CallStatusSubtitleLabel"
```
typfel: They don't need  a `accessibilityIdentifier`anymore?

mikeger: So the whole view is visible as one accessibility element now, hiding the labels.

PR #2224, opened 2018-06-01
[Blocked] Add logic to show Airplay route picker in call UI

daehn:
## What's new in this PR?

* Add logic to show Airplay route picker in call UI.

### This PR depends on:

- [ ] The design is not final yet and a new position for the airplay button has to be found to make it always visible.


daehn:
I am closing this for now, but will keep the branch alive.

PR #2230, opened 2018-06-04
Don't re-display audio keyboard when there is still an ongoing call

daehn:
## What's new in this PR?

* Don't display the audio keyboard again when there is still an ongoing call.


Wire-iOS/Sources/UserInterface/Calling/CallActionsView/CallActionsView.swift
```diff
@@ -167,7 +167,7 @@ final class CallActionsView: UIView {
     
     // MARK: - State Input
     
-    // Entry single point for all state changes.
+    // Single entry point for all state changes.
```
mikeger: yoda-star-wars

PR #2236, opened 2018-06-05
Fix: add color variant support to SettingsClientViewController and its cells.

billypchan:
## What's new in this PR?

Add color variant support to SettingsClientViewController and its cells.

## Note
Fixed TextMessageCellTests's tearDown method to restore WRDateFormatter's static DateFormatters after modifying the locale for testing.


Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientColorVariantProtocol.swift
```diff
@@ -0,0 +1,66 @@
...
+//
+
+import Foundation
+
+protocol ClientListViewControllerDelegate: class {
+    func finishedDeleting(_ clientListViewController: ClientListViewController)
+}
+
+protocol ClientColorVariantProtocol {
+    var variant: ColorSchemeVariant? {get set}
```
daehn: I would add some spaces before and after the get/set e.g. `{ get set }`

PR #2240, opened 2018-06-05
Use module map cryptobox

typfel:
## What's new in this PR?

Just testing to build on CI


mikeger:
@typfel so what should we do with the drunken sailor?

typfel:
Close it for now

PR #2244, opened 2018-06-06
Fix: Disable edit mode in Devices menu if there is only one current device

billypchan:
## What's new in this PR?

Remove edit button when the user has one device only.


Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
```diff
@@ -83,6 +88,14 @@ class ClientListViewController: UIViewController,
         didSet {
             self.sortedClients = self.clients.filter(clientFilter).sorted(by: clientSorter)
             self.clientsTableView?.reloadData();
+
+            if clients.count > 0 {
+                // recreate the edit button
+                let editingList = self.editingList
+                self.editingList = editingList
```
mikeger: I think assigning the same value is not the most elegant way to update the button state. How about extracting the button state logic into the method and calling it from here (and from editingList.didSet) directly?

Wire-iOS Tests/DeviceManagement/ClientListViewControllerTests.swift
```diff
@@ -103,4 +103,13 @@ final class ClientListViewControllerTests: ZMSnapshotTestCase {
...
         self.verify(view: navWrapperController.view)
     }
+
+    func testForEditMode(){
+        prepareSut(variant: .light)
+        let navWrapperController = sut.wrapInNavigationController()
+        let editButton = sut.navigationItem.rightBarButtonItem!
+        UIApplication.shared.sendAction(editButton.action!, to: editButton.target, from: nil, for: nil)
+
+        self.verify(view: navWrapperController.view)
```
daehn: @billypchan the snapshot looks like it's been recorded on iPhone X?

billypchan: @daehn  Thanks for the reminder!

PR #2245, opened 2018-06-06
Fix audio call overlay being visible during second call

daehn:
## What's in this PR?

* Fix audio call overlay being visible during second call.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -199,21 +199,28 @@ extension ConversationInputBarViewController: AudioRecordViewControllerDelegate
     
 }
 
+// Counter keeping track of calls being made when the audio keyboard ewas visible before.
+private var callsInProgress = 0
```
mikeger: May be better to make it the class var? Now it's in global context, so I wonder what would happen on multi account setup.

daehn: Good point 👍

PR #2251, opened 2018-06-06
Call VoiceOver: come back to the call activation

mikeger:
### Issues

In VoiceOver, it was not possible to go back to the call.

### Causes

The `accessibilityActivate` method is not called on the view controller, but on it's view.


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -66,6 +68,7 @@ public extension ConversationViewController {
         button.adjustBackgroundImageWhenHighlighted = true
         button.setTitle("conversation_list.right_accessory.join_button.title".localized.uppercased(), for: .normal)
         button.accessibilityLabel = "conversation.join_call.voiceover".localized
+        button.accessibilityTraits = UIAccessibilityTraitStartsMediaSession
         button.titleLabel?.font = FontSpec(.small, .semibold).font
```
alexaubry: We probably shouldn’t clear the existing traits of the buttons here, otherwise the engine won’t know it’s a button.

We should be using a bitwise OR operator to add the trait.

mikeger: Good point!

PR #2261, opened 2018-06-08
Sort call participants list according to participants join timestamp

daehn:
## What's new in this PR?

* Sort call participants list according to participants join call timestamp.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
@@ -1,0 +1,52 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+final class CallParticipantTimestamps {
+
+    private var participants = Set<UUID>()
+    private var participantTimestamps = [UUID: Date]()
+    
+    func updateParticipants(_ newParticipants: [UUID]) {
+        let updated = Set(newParticipants)
+        let removed = participants.subtracting(updated)
+        let added = updated.subtracting(participants)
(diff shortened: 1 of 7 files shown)


daehn:
Still fixing tests.

Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+final class CallParticipantTimestamps {
+
+    private var participants = Set<UUID>()
```
mikeger: I think that may be there is no reason to keep the dedicated set, also I guess it's not too important to clean the table, since we only support up to certain amount of participants.

Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
```diff
@@ -0,0 +1,52 @@
...
+    private var participantTimestamps = [UUID: Date]()
+    
+    func updateParticipants(_ newParticipants: [UUID]) {
+        let updated = Set(newParticipants)
+        let removed = participants.subtracting(updated)
+        let added = updated.subtracting(participants)
+        Log.haptics.debug("updating participants list: \(newParticipants), old: \(participants)")
```
mikeger: Not sure if logging is necessary or just the debug leftover.

mikeger: Also it says haptics.

Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
daehn: It's used to calculate the added and removed subsets in a performant way.

Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
daehn: Yeah, it's a leftover ;)

Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallParticipantTimestamps.swift
mikeger: OK, I see, you need to know the added set. You can also just go over the new users and check if the timestamp already exists? I don't think that the set intersection is faster than `O(n)` anyways.

daehn: Sure, that's would also work and shouldn't make a big difference. I like cleaning the map up because it makes debugging way easier as people can leave and join again, and also there is no need to keep stuff around that's not needed right now.

PR #2262, opened 2018-06-08
Fix: Unexpected Photos permission appears after profile photo is set with Camera

billypchan:
### Issues

On iOS 11, photo permission dialog is still shown after picked a photo for the profile photo. We expectable the image picker improvements of iOS11 would remove that unnecessary dialog.
### Causes

UIImagePickerController imageFromMediaInfo and imageDataFromMediaInfo will triggers permission dialog. 

### Solutions

Get photo form UIImagePickerController directly instead of getting it form imageFromMediaInfo method.
The original implementation will be kept for iOS 10 or earlier since the fix does not remove the photo permission dialog on earlier iOS versions.

## Todo list
- [x] test with photos stored in iCloud
- [x] ~~review: 'UIImagePickerControllerReferenceURL' is deprecated: first deprecated in iOS 11.0 - Replace with public API: UIImagePickerControllerPHAsset~~ (The original implementation will be kept for iOS 10 or earlier.)
- [ ] ~~test with iOS10, pick an iCloud Photo~~
- [x] test with iOS10 (photo permission dialog shows)
- [x] test with iOS9 (photo permission dialog shows)
- [ ] fix the case: user changing his photo in self profile screen


Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/Steps/ProfilePictureStepViewController+UIImagePickerControllerDelegate.swift
```diff
@@ -0,0 +1,39 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ProfilePictureStepViewController: UIImagePickerControllerDelegate {
+    public func imagePickerController(_ picker: UIImagePickerController, didFinishPickingMediaWithInfo info: [String : Any]) {
+        guard let image = info[UIImagePickerControllerEditedImage] as? UIImage ?? info[UIImagePickerControllerOriginalImage] as? UIImage else {
+
+            self.dismiss(animated: true)
```
mikeger: I think we should still try to load the picture from the metadata info, since some pictures are not downloaded to the user device (they exist in iCloud and the device is only holding the preview). We would display the permissions dialogue, but I think this is better than failing silently.

mikeger: FYI: imageFromMediaInfo is our custom method, may be it's worth fixing it and not rewriting?

billypchan: @mikeger I will check the better way of accessing iCloud photos in iOS11.

Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/Steps/ProfilePictureStepViewController+ImagePicker.swift
```diff
@@ -0,0 +1,67 @@
...
+        showController(picker, inPopoverFromView: sender)
+    }
+}
+
+extension ProfilePictureStepViewController: UIImagePickerControllerDelegate & UINavigationControllerDelegate {
+    public func imagePickerController(_ picker: UIImagePickerController,
+                                      didFinishPickingMediaWithInfo info: [String : Any]) {
+
+        // For iOS 11, simply get single image form dictionary instead of PHImageManager, to get rid of photo access permission dialog
+        if #available(iOS 11.0, *) {
```
billypchan: @mikeger  checked that on iOS11 UIImagePickerController takes care about iCloud photos picking. I keep the original code of iOS 10- below.

PR #2264, opened 2018-06-10
ZIOS-10057: Fix VoiceOver and audio session for calls

alexaubry:
### Issues

When joining a call, the system will play the VoiceOver announcements at a very low level (basically inaudible unless you put the speaker next to your ear).

### Causes

After investigation, it appeared that the system audio was not mixed correctly with the call audio.

### Solutions

We set the system`AVAudioSession` mode to voice/video chat depending on the call type before joining the voice session. This allows the system to know it needs to mix important system announcements such as VoiceOver with the call audio.

We enable this mode when joining a call, and return to the default mode when terminating it.

## Notes

We need to check if we need to set the mode from `acceptDegradedCall` and `terminateDegradedCall`.

### DIFF
--- a/Wire-iOS/Sources/Helpers/sound/AVSMediaManager+Additions.h
+++ b/Wire-iOS/Sources/Helpers/sound/AVSMediaManager+Additions.h
@@ -50,4 +50,7 @@ FOUNDATION_EXPORT void MediaManagerPlayAlert(void);
 - (void)configureSounds;
 - (void)configureDefaultSounds;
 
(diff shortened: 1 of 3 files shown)


typfel:
Nice finding! We should show this to the AVS team so they can update `mode` property on the audio session. Changing it on the client side will lead to issues since AVS assume they are the owner of the audios session.

mikeger:
@alexaubry great finding, we'll integrate it from the AVS side, since they are in charge of the audio session settings!

PR #2267, opened 2018-06-11
ZIOS-9798: Add guest icon to guests in share screen, personal accounts

nicorsm:
### Issues

Personal accounts could see the Guest icon inside `ShareViewController`, which should be visible only to team accounts.

### Solutions

I've added a `hasTeam` clause to the conditions of the `showsGuestIcon` variable.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
@@ -25,7 +25,10 @@ extension ZMConversation: ShareDestination {
     
     public var showsGuestIcon: Bool {
         return self.conversationType == .oneOnOne &&
-            self.activeParticipants.first { $0 is ZMUser && ($0 as! ZMUser).isGuest(in: self) } != nil
+            self.activeParticipants.first {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -25,7 +25,10 @@ extension ZMConversation: ShareDestination {
     
     public var showsGuestIcon: Bool {
         return self.conversationType == .oneOnOne &&
-            self.activeParticipants.first { $0 is ZMUser && ($0 as! ZMUser).isGuest(in: self) } != nil
+            self.activeParticipants.first {
+                guard let user = $0 as? ZMUser else { return false }
+                return user.hasTeam && user.isGuest(in: self)
```
mikeger: Not sure about the logic here, as I understood the problem we must check if the current user (i.e. ZMUser.selfUser) has the team. Otherwise we are checking that SOME user is guest and a team user.

nicorsm: Yep, right, the problem here is that I'm currently check if a user inside the conversation has the team. I should check if I'm inside a team, not the users in the convos.

PR #2270, opened 2018-06-13
ZIOS-ZIOS-10119: Redesign ephemeral message countdown design

alexaubry:
### Issues

A new design was introduced for the ephemeral destruction indicator:

- Circular *pie-like* shape
- Animates smoothly
- Fades out text during last 10 seconds

### Solutions

- Use a shape layer with a moving circle stroke to produce the pie chart
- Use a `CABasicAnimation` to animate the line along with progress
- Refactor countdown start/stop in message cell to integrate with the new animations
- Change constraints in message cell to position the countdown under the avatar
- Do not display the countdown if the message is a ping (because the countdown circle overlaps the ping icon)

## Notes

- [x] Colors and animations need to be confirmed by design


Wire-iOS/Resources/Classy/stylesheet.cas
```diff
@@ -321,6 +318,11 @@ UnknownMessageCell {
     }
 }
 
+DestructionCountdownView {
```
typfel: Why are trying to phase out usage of Classy so I would move this styling into the code.

WireExtensionComponents/Utilities/ColorScheme.m
```diff
@@ -93,6 +93,9 @@
 NSString * const ColorSchemeColorTokenFieldBackground = @"token-field-background";
 NSString * const ColorSchemeColorTokenFieldTextPlaceHolder = @"token-field-text-placeholder";
 
+NSString * const ColorSchemeCountdownElapsedTime = @"countdown-elapsed-time";
+NSString * const ColorSchemeCountdownRemainingTime = @"countdown-remaining-time";
```
mikeger: So are the colors the same in black and white theme?

alexaubry: Oliver confirmed that they are the same in both modes.

mikeger: Ok, then may be it's fine to hardcode the color values as constants in the code, not to overload this class.

alexaubry:
@typfel @mikeger Updated the PR with your suggestions and feedback from design.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -36,6 +36,8 @@
 const NSTimeInterval ConversationCellSelectionAnimationDuration = 0.33;
 static const CGFloat BurstContainerExpandedHeight = 40;
 
+static NSString * const kConversationCellImminentExpirationAnimationKey = @"ImminentExpiration";
```
mikeger: I am not 100% sure where it is used. Might be possible to remove?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -281,8 +283,17 @@ - (void)createBaseConstraints
     
     [self.likeButton autoAlignAxis:ALAxisHorizontal toSameAxisOfView:self.toolboxView];
     [self.likeButton autoAlignAxis:ALAxisVertical toSameAxisOfView:self.authorImageContainer];
-    
-    [self.countdownContainerView autoPinEdgeToSuperviewEdge:ALEdgeTrailing withInset:8];
+
+    NSArray *countdownContainerConstraints =
+    @[
+      [self.countdownContainerView.topAnchor constraintEqualToAnchor:self.authorImageView.bottomAnchor constant:6],
+      [self.countdownContainerView.centerXAnchor constraintEqualToAnchor:self.authorImageView.centerXAnchor],
```
typfel: So the new timer is displayed over the avatar of sender, what happens when the avatar is not visible when one participant sends multiple messages in a row?

typfel: Found the design :) So it's below the avatar, but have you verified that the constraints are valid also in this case?

alexaubry: Yes, I checked, and it works both when the avatar is displayed and when it's not.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell.m
```diff
@@ -681,23 +684,43 @@ - (void)startCountdownAnimationIfNeeded:(id<ZMConversationMessage>)message
...
+    NSTimeInterval duration = self.message.destructionDate.timeIntervalSinceNow;
+
+    if (!self.countdownView.isAnimatingProgress && duration >= 1) {
+        NSTimeInterval progress = [self calculateCurrentCountdownProgress];
+        [self.countdownView startAnimatingWithDuration:duration currentProgress:progress];
+        self.countdownView.hidden = NO;
+    }
+
+    if (duration <= 10 && self.messageContentView.alpha > 0.5) {
+        self.messageContentView.alpha = 0.5;
```
billypchan: For the 10 sec option, the cell is faded immediately.

PR #2271, opened 2018-06-13
ZIOS-10059: Soundcloud still plays when switching accounts

nicorsm:
### Issues

1. If Soundcloud is playing and the user wants to switch to other accounts, the player still plays the song. If the user comes back to the previous account is not possible to stop the music.
2. In that case, Soundcloud crashes when the song ends.

### Causes

1. We weren't specifying the song to stop playing on account switch
2. We are trying to remove the observer for keypath `status` in `self.avPlayer.currentItem`, which wasn't added before in any point of the code (but still calls the observation methods).

### Solutions

1. I'm calling `stop()` on the shared instance of `mediaPlaybackManager`.
2. I've removed the call to `removeObserver` on `self.avPlayer.currentItem`. Question: is it enough removing the observer about `currentItem` inside `avPlayer`, like we're doing in the line before?


mikeger:
@typfel I don't remember this code any more, may be you have an idea?

typfel:
@nicorsm I assume removing the observer of the currentItem was an attempt clean up the observer which is added to the `audioTrack`. This should, I think,  already be handled by setting `self.audioTrack = nil` in `dealloc`.

nicorsm:
Thanks for the review @typfel !

PR #2273, opened 2018-06-13
ZIOS-10054: German L10N doesn't distinguish singular/plural string form

nicorsm:
## What's new in this PR?

Two issues fixed:
1. "warten" was duplicated in the plural version and inside the key (so it appeared like "2 Kontakte warten warten")
2.  Fixed singular case of the string by appending "wartet"


mikeger:
@nicorsm this file is overwritten by the Crowdin fetch, I think it is better to update it there.

PR #2274, opened 2018-06-13
Feature/zios 10130 adjust time values presets

billypchan:
## What's new in this PR?

Adjust timed messages expiry time to 10 seconds, 5 minutes, 1 hour, 1 day, 1 week or 4 weeks.

## Dependencies

Needs releases with:

- [ ]  update WireStyleKit for week icon
- [ ]  https://github.com/wireapp/wire-ios-data-model/pull/488


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -164,9 +164,16 @@ - (instancetype)initWithConversation:(ZMConversation *)conversation
         if (conversation != nil) {
             self.conversation = conversation;
             self.sendController = [[ConversationInputBarSendController alloc] initWithConversation:self.conversation];
-            self.conversationObserverToken = [ConversationChangeInfo addObserver:self forConversation:self.conversation];
-            self.typingObserverToken = [conversation addTypingObserver:self];
-            self.typingUsers = conversation.typingUsers;
+
+            if ([self.conversation respondsToSelector: @selector(managedObjectContext)]) {
```
mikeger: So this is just to allow testing? The code is not looking too user-friendly, may be there is some other way to make it work in tests? I think you can create the actual CoreData objects when sub-classing from `CoreDataSnapshotTestCase`.

Wire-iOS Tests/Mocks/MockConversation.m
```diff
@@ -72,4 +72,19 @@ - (BOOL)isCallingSupported;
     return NO;
 }
 
+- (NSTimeInterval)messageDestructionTimeout
+{
+    return 4;
```
alexaubry: I don't think we should return a value here. Maybe it's better to delete this getter to allow the consumers of the mock API to set the value they want through the property (ex: `conversation. messageDestructionTimeout = 15`)

billypchan:
@alexaubry I undo the changes in MockConversation after replace it with CoreDataSnapshotTestCase's converation. Thanks!

PR #2277, opened 2018-06-14
Swift 4 support

vytis:
## What's new in this PR?

Converted all code to use Swift4. However there might be some UI glitches still there because Classy relies on Swift code visible to Objective-C which now has to be explicitly annotated. If Classy cannot see the property it is simply not styled. Snapshot tests definitely helped to catch some of these issues, but I guess there should be plenty left...

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/56
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/770
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/53
- [x] https://github.com/wireapp/wire-ios-canvas/pull/9


Wire-iOS Tests/AccountViewTests.swift
```diff
@@ -27,7 +27,7 @@ class AccountViewTests: ZMSnapshotTestCase {
     }
     
     func testThatItShowsBasicAccount_Personal() {
-        // GIVEN
+        // GIVENO
```
typfel: Typo

Wire-iOS Tests/MarkdownTextViewTests.swift
```diff
@@ -98,70 +98,70 @@ final class MarkdownTextViewTests: XCTestCase {
...
-    func attrs(for markdown: Markdown) -> [String: Any] {
+    func attrs(for markdown: Markdown) -> [NSAttributedStringKey: Any] {
         switch markdown {
         case .none, .oList, .uList:
             return [
-                MarkdownIDAttributeName: markdown,
-                NSFontAttributeName: style.baseFont,
-                NSForegroundColorAttributeName: style.baseFontColor,
-                NSParagraphStyleAttributeName: style.baseParagraphStyle
+                .markdownID: markdown,
```
typfel: 🌟

WireExtensionComponents/Utilities/FontScheme.swift
```diff
@@ -45,30 +45,30 @@ public enum FontWeight: String {
 @available(iOSApplicationExtension 8.2, *)
 extension FontWeight {
     static let weightMapping: [FontWeight: CGFloat] = [
```
typfel: Can't the type be `[FontWeight : UIFont.Weight]` now? Would get rid of the `.rawValue`.

PR #2278, opened 2018-06-14
Dev build: allow arbitrary ephemeral timer value

mikeger:
## What's new in this PR?

This would allow QA to test the clients behavior with the defined destruction time interval.


typfel:
✅

PR #2285, opened 2018-06-18
Adjust SavableImage to work with animated GIFs

daehn:
## What's new in this PR?

* Adjust SavableImage to work with animated GIFs.


daehn:
I am also looking into using the mime type of the image to always store it on disk in the correct format, which would make this work with all kinds of image formats.

PR #2288, opened 2018-06-19
Extract call overlay to use its own window to support chatheads above it

daehn:
## What's new in this PR?

* Extract call overlay to use its own window to support chatheads above it.
* Clean up window creation code.


daehn:
@typfel had to fix one small issue where the notifications where not shown when the call overlay was above the conversation list.

PR #2290, opened 2018-06-19
ZIOS-10130: Support new timeout values in ephemeral keyboard picker

alexaubry:
### Issues

We support the new list of ephemeral timeout for messages in the ephemeral picker.

### Solutions

- Bump `WireSyncEngine`


typfel:
@alexaubry flaky test

PR #2310, opened 2018-06-25
Fix: Create a new day formatter for customized hour and minute style

billypchan:
## What's new in this PR?

Create a customized Day formatter to produce a formatted string like "27 days 23:43 left".


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
```diff
@@ -19,6 +19,49 @@
...
+            return formatter
+        }()
+
+
+        /// return a string with day in full style and hour/minute in positional style e.g. 27 days 23:43 left
+        ///
+        /// - Parameter timeInterval: timeInterval to convert
+        /// - Returns: formatted string
+        open func string(from timeInterval: TimeInterval) -> String? {
+            let day = floor(timeInterval / 86400)
```
vytis: Converting from seconds to days is generally a bad idea, because not all days are created equal (e.g. when daylight saving kicks in). There are plenty more subtle issues on the same topic outlined in the list - https://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time

In this particular case maybe it's not too bad as we are just formatting the time interval, maybe there is a way to keep this API private and use only in context of ephemeral timeouts?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
```diff
@@ -19,6 +19,49 @@
...
+            let timeWithoutDay = timeInterval - day * 86400
+
+            if let dayString = dayFormatter.string(from: timeInterval), let hourString = hourFormatter.string(from: timeWithoutDay) {
+
+                if hourString != "0" {
+                    return dayString + " " + hourString
```
vytis: Maybe it's better to use one formatter for cases where it's days + hours + mins and another where it's hours + mins?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
```diff
@@ -19,6 +19,49 @@
...
+            formatter.zeroFormattingBehavior = .pad
+            return formatter
+        }()
+
+
+        /// return a string with day in full style and hour/minute in positional style e.g. 27 days 23:43 left
+        ///
+        /// - Parameter timeInterval: timeInterval to convert
+        /// - Returns: formatted string
+        open func string(from timeInterval: TimeInterval) -> String? {
```
vytis: In some cases it breaks down, such as `string(from: 86501) = "1 day 1"`. It really needs some tests just to see what is the expected output of this function.

billypchan: @vytis  I changed the formatter.zeroFormattingBehavior form .dropAll to .pad to fix this.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/EphemeralTimeoutFormatter.swift
billypchan: The new DayFormatter class is specific for timeInterval >= 1 day only, for other cases would use hourFormatter/minuteFormatter/secondsFormatter.

billypchan:
@vytis I removed the hard-code time definition in the code and will clean up the similar case in the data model.

vytis:
We are a bit on minute-saving on CircleCI, let's try to add "[skip ci]" to misc formatting and other commits that do not make any difference in tests...

PR #2315, opened 2018-06-25
Fix/zios 10200 fix disabled time option button title color

billypchan:
### Issues

The color of disabled ephemeralIndicatorButton does not match the background image's tinted gray.

### Causes

Classy file color definition overrides the color setting in the code. The color is set as Graphite instead of light Graphite

### Solutions

Set the correct color and remove Classy color value.


mikeger:
@billypchan please make sure not to trigger the builds for the intermediate commits to save the CI work time.

PR #2317, opened 2018-06-26
Fix NSRangeException for DayFormatter

billypchan:
### Issues

App crashes when DayFormatter going to process the case its hourFormatter does not output the string with prefix like "xd xx:xx".

### Causes

I should consider other languages's case, e.g. in Chinese it outputs "x天 xx:xx".

### Solutions

Improve the regex for matching the day component. Add fail-save logic for regex not match case.


mikeger:
@billypchan can you explain a little bit about why the regular expression matching is necessary here?

billypchan:
This PR is a follow-up of https://github.com/wireapp/wire-ios/pull/2310.

The target is create a formatted time string for the remaining time display e.g. "11 months, 30 days 23:59 left"

Since the iOS DateFormatter doesn't allow created our own formatter string, the solution is create two DateComponentsFormatters for different styles in the year/month/day and hour/minute parts. The DateComponentsFormatter for hour/minute part's allowUnits have to include .day unit otherwise the hour part would have a value > 23. 

The Regex is for removing the day component of the sting generated by hourFormatter, then combine with the output of dayFormatter and we have the sting we want.

PR #2325, opened 2018-06-26
Go back to the conversation details after conv timer selected

mikeger:
### Issues

User don't have to stay on the timer selection screen after the timer was selected.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationTimeoutOptionsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation Options/ConversationTimeoutOptionsViewController.swift
@@ -49,6 +49,8 @@ class ConversationTimeoutOptionsViewController: UIViewController {
     fileprivate let userSession: ZMUserSession
     fileprivate var observerToken: Any! = nil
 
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation Options/ConversationTimeoutOptionsViewController.swift
```diff
@@ -162,12 +164,16 @@ extension ConversationTimeoutOptionsViewController: UICollectionViewDelegateFlow
...
+                return
+            }
+            
+            self.showLoadingView = false
             switch result {
             case .success:
-                self?.updateItems()
-                self?.collectionView.reloadData()
-            case .failure(let error): self?.handle(error: error)
+                self.dismisser?.dismiss(viewController: self, completion: nil)
```
daehn: Do we need the dismisser or could we also call `self.dismiss`?

mikeger: I think no, because I think it's more generic that way: controller must not care how it is being presented.

PR #2327, opened 2018-06-26
UI fixes for timer system messages

nicorsm:
## What's new in this PR?

Two little UI fixes:

- added non-breaking space replacement to timer descriptions
- removed grey hairline on the right-hand side of the cell
- (+) updated test screenshots


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/MessageTimerUpdateCell.swift
```diff
@@ -43,12 +43,13 @@ class MessageTimerUpdateCell: IconSystemCell {
             let name = (user.isSelfUser ? "content.system.you_started".localized : user.name)
             else { return }
         
-        let timerString = "\(displayString)"
+        let timerString = "\(displayString)".replacingOccurrences(of: " ", with: " ")
```
mikeger: May be worth putting the other non-breakable space to the descriptively named constant, like String.nonBreakingSpace, so it would explain the code logic better?

daehn: Or use the unicode syntax.

nicorsm: Unicode syntax is formally correct, but I think that a constant would make the code more clear.

daehn: Yeah I'd still use the constant, but in the definition of the constant use the unicode symbol otherwise the issue just moved to a different place in my opinion.

PR #2328, opened 2018-06-27
Revert "Include self in participants lists"

mikeger:
Reverts wireapp/wire-ios#2324

We have to revert it in order to make the current release and wait for other teams to do the changes.


typfel:
Why is this reverting seemingly unrelated changed to wireapp/wire-ios#2324 ?

mikeger:
@typfel there are some bogus commits in the history displayed, but I think it's some GH issue, since if you check the actual changes they seem fine.

PR #2331, opened 2018-06-27
Fix the status icons initial style

mikeger:
### Issues

In the start UI, the status icons if user has the light color scheme are displayed dark on dark initially. If user scrolls the start UI then the icons are coming back to normal.

### Causes

The color scheme on the cells is set via `UIAppearance`. For the first rendered cells the setting of user is happening before the appearance is setting the color scheme.

### Solutions

When color scheme is set, re-render the user status icon with the correct color.

## Notes

`SearchResultsViewController` was still created with the color scheme variant.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
@@ -321,10 +321,8 @@ extension UIViewController {
             teamContacts = teamContacts.filter({ !filteredParticpants.contains($0) })
         }
         
-        contactsSection.contacts = contacts.filter {
-            $0.displayName.starts(with: "B")
-        }
-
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -321,7 +321,9 @@ extension UIViewController {
             teamContacts = teamContacts.filter({ !filteredParticpants.contains($0) })
         }
         
-        contactsSection.contacts = contacts
+        contactsSection.contacts = contacts.filter {
+            $0.displayName.starts(with: "B")
```
typfel: Debug code?

mikeger: Yeah already removed :hurtrealbad:

PR #2338, opened 2018-06-28
ZIOS-10189: Handle asset expiration in collection and preview

alexaubry:
### Issues

When displaying message collections, we displayed the loading indicator or didn't obfuscate the content. 

### Solutions

In the collection cell class, we introduce an obfuscation view, that displays the placeholder when needed, and a "secure contents view" which contains all the subviews that need to be hidden to honor the expiration settings. When the message expires, we hide the secure contents and show the placeholder. When the user reopens the collection, the expired messages are gone.

In the full screen image, we replace the image by an obfuscation view when the timer expires.

We add a `isObfuscated` property to the search result label, that sets the font to the redacted version when the text needs to be obfuscated.


daehn:
Seems like the vertical offset changed:

screen shot 2018-06-28 at 17 34 19

daehn:
@alexaubry for some reason the cell content in the file cells is still shifted down quite a bit,.

PR #2340, opened 2018-06-28
Fix: image viewer toolbar not refreshed after scrolling form a timed image to non timed image or vice versa

billypchan:
### Issues

In ConversationImagesViewController, its toolbar is not refreshed and the user can still perform actions such as like/download a timed image, which should not be allowed.

### Causes

Buttons should be refreshed according to the image message is ephemeral or not.

### Solutions

When the user scroll to next page in ConversationImagesViewController, refresh the buttonsBar's buttons.


Wire-iOS/Sources/UserInterface/Collections/ConversationImagesViewController.swift
```diff
@@ -217,59 +217,64 @@ final class ConversationImagesViewController: TintColorCorrectedViewController {
         
         return nextIndex
     }
-    
-    private func createControlsBar() {
-        
+
+    private var controlsBarButtons: [IconButton] {
```
daehn: I would still name this `createControlBarButtons` and make it a function instead of a var as it does a lot of work and creates a lot of objects.

PR #2349, opened 2018-06-29
Add accessibility ID to the cell: device verified

mikeger:
### Issues

The toggle is visible for the automation as object without ID, this is happening because the accessibility is flattening the cell to be the toggle. This means that we need to add the id to the cell, too.


vytis:
I guess just not completely sure about all the `self.` removals. Currently it's quite a bit different in various parts of the app, so maybe let's think of what's a good approach there?

mikeger:
@vytis I think in the new code we always emit self when possible

vytis:
We probably do, but it's not something that we ever talked about (BTW I am personally also up for removing extra `.self`). Some other issues might be more polarizing ( `nil == foo` or `foo == .none` etc.) and maybe worth discussing about.

PR #2350, opened 2018-06-29
Update message timer update cells to work with you and off case

daehn:
## What's new in this PR?

* Update message timer update cells to work with you and turned off case.


daehn:
Closed in favour of https://github.com/wireapp/wire-ios/pull/2351

PR #2352, opened 2018-06-29
Architecture: Create custom column layout for GIF search

alexaubry:
### Issues

We were previously using the `ARCollectionViewMasonryLayout` pod to display the GIFs in multiple columns. As we're moving away from CocoaPods, we needed an alternative to this library.

### Solutions

We're removing the dependency on `ARCollectionViewMasonryLayout`, and introducing a new custom layout (`VerticalColumnCollectionViewLayout`) within the app to display the GIFs in columns.

We set the number of columns depending on the screen size, and provide the original of the GIF through a delegate method to allow the layout to calculate the position and proportional size of items.


vytis:
Some snapshot tests would be good here!

Wire-iOS/Sources/UserInterface/Components/CollectionLayouts/VerticalColumnCollectionViewLayout.swift
```diff
@@ -0,0 +1,137 @@
...
+            let adjustedWidth = columnWidth - itemHorizontalSpacing
+            let adjustedHeight = ((itemSize.height * adjustedWidth) / itemSize.width)
+
+            let frame = CGRect(x: xOffset[currentColumn], y: yOffset[currentColumn],
+                               width: adjustedWidth, height: adjustedHeight)
+
+            let positionTransform = CGAffineTransform(translationX: itemHorizontalSpacing,
+                                                      y: itemVerticalSpacing)
+
+            let insetFrame = frame.applying(positionTransform)
```
mikeger: I am not sure if it's necessary to use the transition, or you can just operate on the frame's `origin`?

Wire-iOS/Sources/UserInterface/Components/CollectionLayouts/VerticalColumnCollectionViewLayout.swift
```diff
@@ -0,0 +1,137 @@
...
+
+class VerticalColumnCollectionViewLayout: UICollectionViewLayout {
+
+    // MARK: - Configuration
+
+    /// The object providing size information to the layout.
+    weak var delegate: VerticalColumnCollectionViewLayoutDelegate?
+
+    /// The number of columns to use to organize the content.
+    var numberOfColumns: Int = 1
```
mikeger: Should setting the number of columns automatically invalidate the layout? Not sure how it's done in other `UICollectionViewLayout` implementations.

Wire-iOS/Sources/UserInterface/Components/CollectionLayouts/VerticalColumnCollectionViewLayout.swift
```diff
@@ -0,0 +1,137 @@
...
+        let columnWidth = contentWidth / CGFloat(numberOfColumns)
+
+        let xOffset = (0 ..< numberOfColumns).map { CGFloat($0) * columnWidth }
+        var yOffset: [CGFloat] = Array(repeating: 0, count: numberOfColumns)
+
+        // Separate the items by column
+
+        var currentColumn: Int = 0
+
+        for i in 0 ..< collectionView.numberOfItems(inSection: 0) {
```
mikeger: I feel like this method is OK, but it can be improved for the better architecture, since it's long and it is doing a lot of things, including the frame calculations. 

May be it can be decoupled from the collection view and wrapped in the struct with the method like `advance()` which would create the new instance of the struct advanced by one step of the iteration given the next cell frame?

alexaubry:
@vytis Added snapshot tests 👍

alexaubry:
@mikeger I have made some changes following your review:

- We invalidate the layout when changing properties of the columns
- We decouple the computing logic from the layout class itself

To decouple the logic from the layout, we created an object named `VerticalColumnPositioning` that contains the layout informations for the cells and columns. It gets initialized with a `VerticalColumnPositioningContext` object, that contains the layout properties of the collection view. 

For every cell in the collection, we call `insertItem(ofSize:at:)`, to incrementally update the positioning object with the layout attributes of this cell.

We also made a change in how we choose the column to put the item in. Before, we were just choosing the next column on the right. Now, we add the new item to the column with the smallest height. This allows us to get better layout on large screens like iPad, with columns approximately the same height on the bottom.

WireExtensionComponents/CollectionLayouts/VerticalColumnCollectionViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+/**
+ * A number of columns that changes that changes depending on
```
vytis: "that changes" x 2

WireExtensionComponents/CollectionLayouts/VerticalColumnCollectionViewController.swift
```diff
@@ -0,0 +1,136 @@
...
+ * the width of the container.
+ */
+
+public struct AdaptiveColumnCount {
+
+    /// The number of columns in a compact-width container.
+    let compact: Int
+
+    /// The number of columns in a regular-width container smaller than 1024pt.
```
vytis: As I understand it is related to both size class and width of the view? Maybe comment could be more explicit about this.

mikeger:
@alexaubry may be worth creating the ticket, so QA can give it one more round of testing?

alexaubry:
@mikeger Will do.

PR #2353, opened 2018-06-29
Enable swiping between tab bar view controllers

daehn:
## What's new in this PR?

* Enable swiping between tab bar view controllers.

ezgif-2-d4c9a3c18b


mikeger:
@daehn may be worth creating the ticket, so QA can give it one more round of testing? Would be valuable for them to know where we are using this controller.

PR #2358, opened 2018-07-02
ZIOS-10165: Audio message lost when you hit the 25Mb limit

nicorsm:
### Issues

Audio messages that exceed the limit of 25MB are discarded without displaying any message, due to a lack of error handling.

### Solutions

In my solution, I'm checking the file size inside the `displayLinkDidFire()` method of `AudioRecord`. The computation is made inside the `audioSizeIsCritical` variable, where I'm checking if the file size exceeds the threshold of 99% of allowed size (which is 24,75MB for a 25MB limit).

The error flow and the alert display logic is the same of the time limit, so I've moved the generation of the `UIAlertController` instance inside `AudioRecorder` (alert is still shown from view controllers).


Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
```diff
@@ -64,12 +64,13 @@ public protocol AudioRecorderType: class {
...
     var fileURL: URL? { get }
     var maxRecordingDuration: TimeInterval? { get set }
+    var maxFileSize: UInt64? { get set }
     var currentDuration: TimeInterval { get }
     var recordTimerCallback: ((TimeInterval) -> Void)? { get set }
     var recordLevelCallBack: ((RecordingLevel) -> Void)? { get set }
     var playingStateCallback: ((PlayingState) -> Void)? { get set }
     var recordStartedCallback: (() -> Void)? { get set }
-    var recordEndedCallback: ((Bool) -> Void)? { get set } // recordedToMaxDuration: Bool
+    var recordEndedCallback: ((Bool, Bool) -> Void)? { get set } // recordedToMaxDuration: Bool, recordedToMaxSize: Bool
```
mikeger: I think it would be more consistent to create the enum and wrap it into `VoidResult`:
`enum RecordingError: Error { toMaxDuration, toMaxSize };`

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecorder.swift
```diff
@@ -119,18 +122,22 @@ public final class AudioRecorder: NSObject, AudioRecorderType {
     public var recordLevelCallBack: ((RecordingLevel) -> Void)?
     public var playingStateCallback: ((PlayingState) -> Void)?
     public var recordStartedCallback: (() -> Void)?
-    public var recordEndedCallback: ((Bool) -> Void)? // recordedToMaxDuration: Bool
+    public var recordEndedCallback: ((Bool, Bool) -> Void)? // recordedToMaxDuration: Bool, recordedToMaxSize: Bool
     public var fileURL: URL?
+    public var maxFileSize: UInt64?
+    private var fileDescriptor: Int32?
+    private var fileSizeMonitor: DispatchSourceFileSystemObject?
```
mikeger: I think this property is not used at the moment.

mikeger: Edit: same goes for `fileDescriptor`.

nicorsm: Exactly, leftover from my previous implementation

Wire-iOS Tests/AudioRecorderTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        let expectation = self.expectation(description: "Wait for recorder to stop")
+        
+        recorder?.recordEndedCallback = { result in
+            expectation.fulfill()
+            let error = result.error
+            XCTAssertNotNil(error)
+            XCTAssertEqual(error as! RecordingError, RecordingError.toMaxDuration)
+        }
+        
+        self.waitForExpectations(timeout: 5.0, handler: nil)
```
alexaubry: It seems that these expectations are never fulfilled. Is `startRecording` starting the display link?

nicorsm: Yes, it calls `setupDisplayLink()` correctly and the test is working fine locally 🤔

Wire-iOS Tests/BackupPasswordViewControllerTests.swift
```diff
@@ -8,7 +8,7 @@
 // (at your option) any later version.
 //
 // This program is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// but WITHOUT ANY WARRANTY; without even the implied warranty of-
```
mikeger: Some random key most likely!

PR #2359, opened 2018-07-03
Embed SCSiriWaveformView in app bundle and remove pod

alexaubry:
## What's new in this PR?

We remove the `SCSiriWaveformView` dependency from the app and embed it into the source code directly.


vytis:
This added a warning (the last one from Danger output), when it used to be in CocoaPods it was all ignored.

alexaubry:
@vytis I fixed the warning.

PR #2362, opened 2018-07-03
Attempt to fix TabBar crash, animate tabbar selection line

daehn:
## What's new in this PR?

* Attempt to fix `TabBar` crash.
* Animate `TabBar` selection indicator line when swiping.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/TabBar/Tab.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/TabBar/Tab.swift
@@ -20,7 +20,6 @@ import UIKit
 import Cartography
 
 class Tab: Button {
-    private let selectionLineView = UIView()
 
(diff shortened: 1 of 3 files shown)


typfel:
Can you point out the crash fix?

daehn:
It's the `Dispatch.main.async` as it seems this method isn't always called from the main thread, wasn't able to reproduce it though, will have to see if it still occurs in internal builds.

typfel:
Ok, weird that it's not always called from the main thread though. I guess the real bug is somewhere else then.

daehn:
@typfel yeah I am also not too happy with this workaround to be honest, not sure if it's worth it spending more time on this though (assuming this workaround actually fixes the crash for good).

daehn:
Fixing snapshots.

PR #2363, opened 2018-07-03
ZIOS-10256: Fetch YouTube preview with REST API instead of Google SDK

alexaubry:
### Issues

As we are phasing away our usage of CocoaPods, we needed a replacement for the Google SDK that we currently use to fetch the media previews for YouTube.

### Solutions

We are going to take the same approach that is currently used for SoundCloud previews. 

In `YouTubeService`, we use the YouTuve V3 Data API to fetch the snippet of the video, instead of the Google SDK.

This request is authenticated and performed by the backend, and the result is returned to the app. We decode the resulting JSON using the Swift `JSONDecoder`, via `Decodable` structures modelling the YouTube response format.

We added unit and integration tests for these YouTube previews.


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -3455,6 +3465,35 @@
...
+				5EC16B8E20EB8C6F00955156 /* YouTubeService.swift */,
+				5EC16B8D20EB8C6F00955156 /* YouTubeService+DataTypes.swift */,
+			);
+			path = YouTube;
+			sourceTree = "<group>";
+		};
+		5EC16BA220EBBEE000955156 /* Resources */ = {
+			isa = PBXGroup;
+			children = (
+				5EC16BA520EBBF8400955156 /* yt_response_multiple_videos.json */,
```
vytis: Not clear from the project XML, but these should go to tests bundle, not the main app as we don't want to ship them to users.

alexaubry: That's already the case (this Resources folder points to `Tests/YouTube/Resources`), but I just renamed it to `MockResponses` to avoid confusion.

vytis:
Forgot the pleasure of actually removing CocoaPods integration :)

alexaubry:
@vytis I will do it in a separate PR :)

Wire-iOS/Sources/UserInterface/Components/MediaPreview/MediaPreviewViewController.m
```diff
@@ -119,7 +118,8 @@ - (void)fetchAttachment
         };
         
         if (self.linkAttachment.type == LinkAttachmentTypeYoutubeVideo) {
-            [[YoutubeService sharedInstance] mediaPreviewDataForYoutubeVideoURL:self.linkAttachment.URL completion:mediaPreviewResponseHandler];
+            YouTubeService *service = [[YouTubeService alloc] initWithRequester:[ZMUserSession sharedSession]];
```
daehn: Not sure if we should create a new `YouTubeService` instance every time, also not the biggest fan of singeltons, if there is a way to create one instance higher up and pass it down it would be perfect  (otherwise a singleton I guess).

PR #2364, opened 2018-07-03
Fix: [iPad] Conversation InputBar's popover position

billypchan:
### Issues

On iPad, when opening a popover form conversation input bar, the popover's arrow does not point to the icon.

### Causes

The sourceRect of UIPopoverPresentationController is calculated in several ways but not consider the case that the source view is embedded several levels of subviews.

### Solutions

Calculate the target button's frame's position corresponding to popover's present view.

### Attachments

Snapshots after the fix:

screen shot 2018-07-03 at 18 01 20


billypchan:
@alexaubry there is no snapshots test for the iPad popover currently since we run snapshot tests on iPhone simulator only and need some mocking injection to snapshot a popover. I can add try to add some later.

PR #2367, opened 2018-07-04
Fix: Do not show buttons except deleteAction for ephemeral image on quick menu

billypchan:
### Issues

The user is able to copy/like/save/share a timed image via force touch menu.

### Causes

Missing checking of currentMessage.isEphemeral when creating the previewActionItems 

### Solutions

Check currentMessage.isEphemeral and create different lists for both cases.


billypchan:
Yes, align the UI to the image viewer and image long-tap menu. 

Maybe we should define these rules in data-model level, i.e. : create a property in ZMConversationMessage allowActions, and it returns a list of enum such as [copy, like, save, share, delete] for a normal message or [delete] for an ephemeral message.

PR #2368, opened 2018-07-04
ZIOS-10195: Remove CocoaPods

alexaubry:
### Issues

We decided to stop using CocoaPods as a dependency manager. This pull requests removes all traces of CocoaPods from the project and the build script.

### Solutions

- Deintegrate from `xcodeproj`
- Update Circle CI manifest
- Update setup script
- Update README

## Dependencies

Needs releases with:

- [x] Build script update from `wire-ios-shared-resources` ([#22](https://github.com/wireapp/wire-ios-shared-resources/pull/22))


Scripts/recordSnaphots
```diff
@@ -19,7 +19,7 @@
  #
 
  xcodebuild clean test \
- -workspace Wire-iOS.xcworkspace \
```
vytis: I think this file is super out of date and not really used for anything. Worth looking later how many of the scripts there are still relevant

Gemfile
```diff
@@ -1,8 +0,0 @@
-source 'https://rubygems.org'
-
-gem 'cocoapods', '~> 1.3.1'
-gem 'xcpretty', '~> 0.2.4'
-gem 'danger', '~> 5'
```
daehn: Not sure about removing the complete Gemfile, but I guess as we download the `Gemfile.lock` from our shared ressources folder we don't need it here.

Scripts/recordSnaphots
daehn: I think we might be able to remove this file, I personally think it might be dangerous to run it as it will record all Snapshots again. The only use case I see is OS level font rendering changes that require a lot of changes, but even then I think we should do it manually as thinks might brake and need inspection.

Gemfile
alexaubry: As @vytis suggested, since the Gemfile is identical for frameworks and UI now, we can just reuse the same one from the shared resources (because we only need it in CI).

Scripts/recordSnaphots
alexaubry: As discussed, I removed this script.

PR #2376, opened 2018-07-05
Implement async image loading with URLSession

alexaubry:
### Issues

As discussed in architecture issue 36, we want to remove the `AFNetworking` dependency, since it is only used to asynchronously download images. As this is a simple task, we can write the infrastructure for that ourselves.

### Solutions

We add an an extension on `UIImageView` to display the image asynchronously:

- `displayImageAtURL:onSuccess:onError:`, with success and failure callbacks
- `displayImageAtURL:`, if callbacks are not needed

This method uses an `ImageDownloadCache`. This class holds the cache, dispatches the network requests, and decodes the images in the background for optimal performances. A `ResourceDownloadHandler` is associated with the download cache, and is set as the delegate of its `URLSession`.

The download handler builds up the data for the response, and changes the caching headers if caching is not supported on the remote server.

For caching, we used a standard `URLCache` from Foundation. The shared cache has a memory limit of 100MB, and a disk limit of 200MB. It adds missing cache headers for 2 hours, if the server does not provide caching information.


Wire-iOS/Sources/Components/ImageCache/DecodeImageOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        let imageBounds = CGRect(x: 0, y: 0, width: rawImage.width, height: rawImage.height)
+        context.draw(rawImage, in: imageBounds)
+
+        guard let rawDecodedImage = context.makeImage() else {
+            return
+        }
+
+        decodedImage = UIImage(cgImage: rawDecodedImage)
```
typfel: What do we have achieve by re-drawing the image?

alexaubry: This operations decodes the image in a background queue, instead of on the main thread. We theoretically gain some performance. ([reference](http://www.lukeparham.com/blog/2018/3/14/decoding-jpegs-with-the-best))

typfel: Yes but I mean ` UIImage(data: imageData)` already decodes the image, no?

typfel: Ah saw the blog post now. Sneaky `UIImage`.

Wire-iOS/Sources/Components/ImageCache/ImageDownloadCache.swift
```diff
@@ -0,0 +1,131 @@
...
+                completionHandler(image)
+            }
+        }
+
+        // Attempts decoding an image as local,
+
+        downloadHandler.schedule(downloadTask) { data, _, _ in
+
+            guard let responseData = data else {
+                resultHandler(nil)
```
daehn: Maybe worth forwarding the error in case there is one (could use `Result<UIImage>` as type in the closure for example).

daehn: Ok I see this is wrapped later on and we don't forward the error there either (like in the old API I assume), but maybe it's still good to have it so it _could_ be inspected in case something goes wrong.

alexaubry:
@daehn I was thinking whether more tests were needed with `ImageDownloadCache`. If you don't think it's necessary, then yes its' ready to be merged.

vytis:
Maybe some tests for image cache would be good? A good place to start is see what lines are not covered by tests from the coverage report - https://codecov.io/gh/wireapp/wire-ios/pull/2376/diff

Wire-iOS Tests/ImageCache/ImageDownloadCacheTests.swift
```diff
@@ -0,0 +1,182 @@
...
+
+        cache = URLCache(memoryCapacity: 100 * 1024 * 1024, diskCapacity: 0, diskPath: nil)
+        session = MockURLSession(delegate: resourceDownloadHandler, cache: cache, delegateQueue: delegateQueue)
+        imageDownloadCache = ImageDownloadCache(downloadHandler: resourceDownloadHandler, session: session)
+
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        cache.removeAllCachedResponses()
```
daehn: Please also nil all local variables (`cache`, `session` and `imageDownloadCache`) as a new instance of the test class will be created for each test being executed.

PR #2384, opened 2018-07-10
Don’t crash when audio message exceeds timer

johnxnguyen:
### Issues
When the timeout of an audio message is extended, the app crashes when the original timer would have fired.

### Causes
The timestamp label was still being updated in the extended period, however, to guard against a negative timer string, the client was force-crashed if the destruction date was exceeded. Extending the timer of an audio message doesn't change the original destruction date, it simply replaces the existing timer with a new one. Hence, one the destruction date was exceeded, the app was force-crashed.

### Solutions
Add extra conditions when constructing the timeout string: if the destruction date is exceeded and the message is an audio message, then fallback to the default timestamp string. If the destruction date is exceeded, but the message is not an audio message, force-crash the app.


johnxnguyen:
There's one (flaky) failing snapshot test, but it passes locally. Merging without retesting on CI.

johnxnguyen:
@mikeger yes only on internal. What do you mean when the controller is not visible, the conversation/message? If the conversation is covered with a presented view controller for video/documents, the timestamp string isn't being updated anyway.

We don't extend the timer for video message of documents because we don't need to (those messages are visible as long as the presented view controller is open). These messages are quietly deleted underneath.

PR #2386, opened 2018-07-10
Update participants presentation in group details view controller

daehn:
## What's new in this PR?

The participants presentation in group details view controller has been updated and now follows the following rules:

* We limit the amount of group participants we show to 7.
* If the amount of participants exceeds 7, we only show the first 5 and a "Show All" cell.
* The "Show All" cell opens a new view controller showing all participants and a search bar.


Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
```diff
@@ -0,0 +1,38 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
typfel: 2016 > 2018

daehn:
@typfel updated, also added a dismiss button to the details screen and adjusted the snapshot test to include a navigation bar

Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewModel.swift
```diff
@@ -0,0 +1,78 @@
...
+
+fileprivate extension String {
+    var isValidQuery: Bool {
+        return !isEmpty && self != "@"
+    }
+}
+
+class GroupParticipantsDetailViewModel: NSObject, SearchHeaderViewControllerDelegate {
+
+    private let __participants: [ZMBareUser]
```
mikeger: @daehn not sure if we have anywhere the double (or single) underscore prefix, what is the meaning of it?

mikeger:
We have the very similar UI in `ReactionsListViewController`, may be worth unifying it?

daehn:
@mikeger AFAIK design has some plans for `ReactionsListViewController` as well, so I am not sure if would make sense right now. But I'll check with them.

Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewModel.swift
daehn: Good point I think we use one underscore in some places, but maybe better to just prefix it with `internal` instead.

daehn: @mikeger done in https://github.com/wireapp/wire-ios/pull/2388

PR #2390, opened 2018-07-11
Fix: Image edit screen - send button is invisible in light theme

billypchan:
### Issues

In sketch screen and image editing screen, send button is invisible in light theme.

### Causes

Removed Classy styling for the send button 

### Solutions

Create a factory method for the send button.


Wire-iOS/Sources/UserInterface/Helpers/IconButton+Factory.swift
```diff
@@ -0,0 +1,108 @@
...
+            icon: .endCall,
+            size: .small,
+            accessibilityId: "LeaveCallButton",
+            backgroundColor: [.normal: ZMAccentColor.vividRed.color],
+            iconColor: [.normal: .white],
+            width: IconButton.width
+        )
+    }
+
+    class var sendButton: IconButton {
```
daehn: Maybe it's better to make this a function as well? It's creating and returning a new instance every time.

PR #2393, opened 2018-07-11
Add thumbnail creation method taking max shorter side length instead of max length

daehn:
## What's in this PR?

**Add thumbnail creation method taking max shorter side length instead of max length**

When creating a thumbnail using CoreImage the max length is specified, however for non-square and especially panoramic images this means the longer side will have a maximum length of the specified value. In the collections view we display th eimages in a square format in image views using the aspect fill content mode, we thus want the shorter side of the image to be the one with the specified max length when creating a thumbnail. This PR adds a new method which takes a the shorter side length as parameter and calculates the max size based on the images aspect ratio.

As an example before when creating a max size 100pt thumbnail for a panomara with 4:1 aspect ratio the resulting height would be 25pt, now it will be 100pt.

---

**Before:**

screen shot 2018-07-11 at 15 11 53

**After:**

screen shot 2018-07-11 at 15 49 12


WireExtensionComponents/Utilities/UIImage+ImageUtilities.m
```diff
@@ -178,6 +172,76 @@ + (UIImage *)imageFromData:(NSData *)imageData withMaxSize:(CGFloat)maxSize
...
+    CGImageSourceRef source = CGImageSourceCreateWithData((CFDataRef)imageData, NULL);
+    if (source == NULL) {
+        return nil;
+    }
+    
+    CGSize size = [self sizeForImageSource:source];
+    CGFloat longSideLength = shorterSideLength;
+
+    if (size.width > size.height) {
+        longSideLength = shorterSideLength * (size.width / size.height);
```
vytis: `height` or `width` could be 0 in malicious/corrupted image, check for that or it will 💥

vytis:
Some snapshot tests for this would be good

daehn:
@vytis  good point, will add some!

daehn:
@vytis updated!

Wire-iOS Tests/Mocks/MockCollection.swift
```diff
@@ -1,9 +1,19 @@
 //
-//  MockCollection.swift
-//  Wire-iOS
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
 //
-//  Created by Vytis ⚪️ on 2017-01-11.
```
vytis: Oooops. Was my 3rd day at Wire 🤷‍♂️

PR #2395, opened 2018-07-11
Fix the participant details list cells width on iPad

daehn:
## What's new in this PR?

* Fix the participant details list cells width on iPad.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewController.swift
@@ -18,7 +18,7 @@
 
 import UIKit
 
-final class GroupParticipantsDetailViewController: UIViewController, UICollectionViewDelegate, UICollectionViewDataSource {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewController.swift
```diff
@@ -106,5 +105,10 @@ final class GroupParticipantsDetailViewController: UIViewController, UICollectio
         guard let user = viewModel.participants[indexPath.row] as? ZMUser else { return }
         delegate?.presentDetails(for: user)
     }
+    
+    
+    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
+        return .init(width: view.bounds.size.width, height: 56)
```
alexaubry: Is there a reason why we use `.init(width:height:)` instead of `CGSize(width:height:)` ?

daehn: No, it's the same

PR #2402, opened 2018-07-12
 Initial version: removed unused events

mikeger:
### Issues

Some events leftovers from Localytics times are not enabled on Mixpanel. Other teams removed them, so it's time for them to go from our codebase too.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/798
- [x] https://github.com/wireapp/wire-ios-data-model/pull/517


vytis:
I guess would be good to verify that the current events still appear in Mixpanel. Great work!

PR #2407, opened 2018-07-17
ZIOS-10301: Add always-failing shared ID requester for testing

alexaubry:
### Issues

SE introduces the `SharedIdentitySessionRequester` protocol, that will be used to fetch SSO users from their login code. We need a concrete implementation of this protocol to handle the user input in the SSO alert.

As the current scope does not include making the network request to BE, we need an object that constantly returns an error to implement the negative flow.

### Solutions

We create the `TimeoutIdentitySessionRequester`, that always returns an error message after the timeout specified when creating it. This is a temporary solution that will be replaced by an actual concrete implementation in SE when BE specs are available.

## Dependencies

Needs releases with:

- [x] [SE #802](https://github.com/wireapp/wire-ios-sync-engine/pull/802) - Initial Shared Identity Session Infrastructure


Wire-iOS/Sources/UserInterface/Registration/SharedIdentity/TimeoutIdentitySessionRequester.swift
```diff
@@ -0,0 +1,39 @@
...
+
+import Foundation
+
+struct SharedIdentitySessionRequesterTimeoutError: LocalizedError {
+    let errorDescription: String = "The Single Sign-On provider did not respond to the authentication request. Contact your team administrator for further details."
+}
+
+class TimeoutIdentitySessionRequester: SharedIdentitySessionRequester {
+
+    let delay: Int
```
daehn: Maybe better to make this a `TimeInterval` if it represents seconds (I think then we can also omit the `.seconds()` below).

alexaubry:
@daehn Bumped SE and changed the timeout representation to `TimeInterval`.

PR #2413, opened 2018-07-17
Fix: Audio message is stopped recording if device entering screenlock mode

billypchan:
### Issues

The audio record keyboard is dismissed after the app is unlocked. The user can not save his audio record.

### Causes

Actually, the audio record keyboard is still present but not visible after the app called endEditing method. The user can not see the audio record keyboard with audio filter UI.

### Solutions

In ConversationInputBarViewController, observer the .appUnlocked notification form AppLockView. Display the audio record keyboard when the notification is observed. 

## Notes

This PR changes the keyboard visibility when the application is active and the lock setting is disabled. The app preserves the keyboard state instead of always resign the keyboard.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
@@ -22,6 +22,10 @@ import WireExtensionComponents
 
 private let zmLog = ZMSLog(tag: "UI")
 
+extension Notification.Name {
+    static let appUnlocked = Notification.Name("AppUnlocked")
+}
+
 @objcMembers final class AppLockViewController: UIViewController {
     fileprivate var lockView: AppLockView!
     fileprivate static let authenticationPersistancePeriod: TimeInterval = 10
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -170,6 +176,5 @@ extension AppLockViewController {
     
     @objc func applicationDidBecomeActive() {
         self.showUnlockIfNeeded()
-        self.resignKeyboardIfNeeded()
```
mikeger: Not sure why it's not necessary not to hide the keyboard?

billypchan: For the case app lock is enabled, the keyboard is dismissed when applicationWillResignActive. No need to hide it again when applicationDidBecomeActive. This line is removed since it is hit after AppLock.evaluateAuthentication succeed, which hides the keyboard again.

For the case app lock is disabled, the keyboard's visibility is not affected by the app state anymore.

mikeger: OK, I see.

PR #2420, opened 2018-07-18
[WIP] Implement custom in-app browser for SSO purposes

alexaubry:
### Issues

The app needs a custom browser to perform the single sign-on requests, instead of using the system authentication session.

### Solutions

We define a view controller that wraps a `WKWebView`. We can use the delegate methods to observe navigation events, e.g. cookies or redirections.

This is not an optimal solution regarding user experience, as in-app browsers do not share login sessions with the system (which means users will have to login with their IdP every tiume), but there is no way around it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Registration/Company/TimeoutIdentitySessionRequester.swift
+++ b/Wire-iOS/Sources/UserInterface/Registration/Company/TimeoutIdentitySessionRequester.swift
@@ -35,6 +35,14 @@ class TimeoutIdentitySessionRequester: SharedIdentitySessionRequester {
     }
 
     func requestIdentity(for token: UUID, _ completion: @escaping (SharedIdentitySessionResponse) -> Void) {
+//        let browser = TaskBrowserViewController(nibName: nil, bundle: nil)
+//        let controller = UINavigationController(rootViewController: browser)
+//
+//        UIApplication.shared.wr_topmostController(onlyFullScreen: true)?.present(controller, animated: true) {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Registration/Company/TimeoutIdentitySessionRequester.swift
```diff
@@ -35,6 +35,14 @@ class TimeoutIdentitySessionRequester: SharedIdentitySessionRequester {
     }
 
     func requestIdentity(for token: UUID, _ completion: @escaping (SharedIdentitySessionResponse) -> Void) {
+//        let browser = TaskBrowserViewController(nibName: nil, bundle: nil)
+//        let controller = UINavigationController(rootViewController: browser)
+//
+//        UIApplication.shared.wr_topmostController(onlyFullScreen: true)?.present(controller, animated: true) {
+//            UIApplication.shared.wr_updateStatusBarForCurrentControllerAnimated(true)
+//            browser.open(URL(string: "https://sso.wire.com/initiate-login/\(token.transportString())")!)
```
daehn: Can we theoretically call `open` on the browser before the presentation animation is finished?

alexaubry: Yes, we could do it. I just thought it would be less heavy on the UI thread if we did it after the animation ends, but maybe I'm wrong.

alexaubry:
This is no longer needed, as it was decided to use the Safari view controller for the company login flow.

PR #2425, opened 2018-07-19
ZIOS-10311: Implement infrastructure for browser company login flow

alexaubry:
## What's new in this PR?

In this PR, we introduce the browser flow for company login. When the user enters the code, we redirect them to the browser so that they can finish authentication.

### Solutions

- We remove the loading indicator, as indicated in the specs
- The `CompanyLoginFlowHandler` opens the URL in the appropriate container, in a backwards-compatible way.

### To-Do

- [x] Handle iOS 9 / 10 case
- [x] Propagate errors from App Delegate to current login/registration flow

## Dependencies

Needs releases with:

- [x] [SE #807](https://github.com/wireapp/wire-ios-sync-engine/pull/807) - Implement standard company login requestor
- [x] [DM #516](https://github.com/wireapp/wire-ios-data-model/pull/516) - Store push token inside user client


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -612,6 +612,14 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
             alert.addAction(cancelAction)
             
             self.present(alert, animated: true, completion: nil)
+
+        case .companyLoginFailure:
+            print("Handle error")
+            callback(true)
```
mikeger: Should it be callback(false) ?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -1234,6 +1234,55 @@
 			remoteGlobalIDString = 1EE85C4F1B5F96FB008E084B;
 			remoteInfo = ZetaFramework;
 		};
+		5E0EB2152101F9F000B5DC2B /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 5E0EB20F2101F9EF00B5DC2B /* WireSyncEngine.xcodeproj */;
```
daehn: Should be removed (I guess these are references to a dragged in SE).

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -613,11 +613,30 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
...
+            guard case .unauthenticated = appStateController.appState else {
+                callback(false)
+                return
+            }
+
+            let alert = UIAlertController(title: "general.failure".localized,
+                                          message: message,
+                                          preferredStyle: .alert)
+
+            let doneAction = UIAlertAction(title: "general.ok".localized, style: .default) { _ in
```
daehn: There is a static convenience function on `UIAlertController` that you can use for this, e.g. 
```swift 
alert.addAction(.ok { completion(false) })
```

(The same is available for cancel as well.)

alexaubry:
@daehn @mikeger Updated 👍

PR #2434, opened 2018-07-23
Fix: crash when collectionView.insertItems()

billypchan:
### Issues

On iOS9/10, the app crashes when the user scrolling down for several pages on GiphySearchViewController.

### Causes

Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'no UICollectionViewLayoutAttributes instance for -layoutAttributesForItemAtIndexPath: A_RANDOM_INDEX_PATH'

When the layout attributes of a certain indexPath are nil, an exception is thrown. (This issue is fixed on iOS11). 

### Solutions

Update Ziphy with new Ziph property previewImage to return a suitable preview image to display on the collection view.

### Notes
- [x] check Ziphy for the source for  0-height frame.


billypchan:
@mikeger Updated this PR with the ziphy fix.

PR #2437, opened 2018-07-24
Use backendhost from current env in CompanyLoginController

daehn:
## What's in this PR?

* Use backendhost from current env in CompanyLoginController.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
+++ b/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
@@ -58,7 +58,9 @@ import Foundation
 
     /// Create a new `CompanyLoginController` instance using the standard detector and requester.
     @objc public override convenience init() {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -58,7 +58,9 @@ import Foundation
 
     /// Create a new `CompanyLoginController` instance using the standard detector and requester.
     @objc public override convenience init() {
-        self.init(detector: .shared, requester: CompanyLoginRequester(backendHost: "staging-nginz-https.zinfra.io", callbackScheme: "wire"))
+        let environment = ZMBackendEnvironment(userDefaults: .standard)
```
alexaubry: Do we need to use the `standard` or `shared` user defaults?

daehn: We store the env type in the standard defaults in the other places I checked.

daehn: (or maybe even both because we probably need it in extensions as well)

PR #2438, opened 2018-07-24
Use sso login url scheme from xcconfig

daehn:
## What's in this PR?

* Use sso login url scheme from xcconfig.
* Disable company logins on public builds for now.


Wire-iOS/Sources/UserInterface/Company Login/URLSchemeHelper.h
```diff
@@ -0,0 +1,21 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+#import <Foundation/Foundation.h>
+
+NSString *wr_URLScheme(void);
```
alexaubry: Should we mention SSO in this variable, as it fetched `WIRE_SSO_URL_SCHEME`?

Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -57,9 +61,17 @@ import Foundation
     // MARK: - Initialization
 
     /// Create a new `CompanyLoginController` instance using the standard detector and requester.
-    @objc public override convenience init() {
+    @objc(initWithDefaultEnvironment) public convenience init?(withDefaultEnvironment: ()) {
```
alexaubry: This looks odd in Swift. Could this be a factory method instead?

daehn: Yeah looks a bit odd in Swift, but is a practice also present in the standard lib to have different initializer that do not take any parameters (necessary here because the regular ini't isn't failable). E.g. like the `ExpressibleByNilLiteral` initializer: `init(nilLiteral: ())`.

alexaubry: I see, that makes sense. I guess we can keep it this way then.

ConfigurationSets/Development.xcconfig
```diff
@@ -38,7 +38,7 @@ CODE_SIGN_ENTITLEMENTS_SHARE_EXT = Wire-Sharing-Extension-iOS/Wire-Sharing-Exten
 //Bundle ID and Version number
 WIRE_BUNDLE_ID = com.wearezeta.zclient-alpha
 WIRE_BUILD_NUMBER = 999999
-
+SSO_URL_SCHEME = wire-sso-develop
```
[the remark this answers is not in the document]
daehn: Double checking this right now, but I think so, yes.

PR #2448, opened 2018-07-25
Share the default font scheme so we don't re-create it unnecessarily

typfel:
### Issues

A `FontScheme` gets re-created on every `FontSpec.font` invocation which is happens quite often.

### Solutions

Introduce a shared FontScheme using the configured dynamic type setting.


mikeger:
I think the font scheme must be re-created if we detect the system font size change.

typfel:
@mikeger I assumed that `onContentSizeCategoryChange` gets called when font size changes.

mikeger:
Ah, true, but may be worth testing it once just in case.

PR #2452, opened 2018-07-26
Fix: snapshot test fail for testThatItRendersLinkPreviewMessagePreview

billypchan:
### Issues

Sometimes testThatItRendersLinkPreviewMessagePreview fails for the link preview image is not loaded.

### Causes

The image caching logic takes some time and the UIView for snapshot taking is freed before the image is loaded. 

### Solutions

Ref to ArticleViewTests, wait for a short time before the image is loaded.
Retain the view for snapshot taking, free it in tearDown.


mikeger:
If the testing device / computer would be slow enough it can also render the image after 200ms. I think the better solution is to use dispatch groups, like we do in the sync engine.

PR #2453, opened 2018-07-26
Asynchronous fetching of images 

typfel:
## What's new in this PR?

Update cells to use new asynchronous image methods.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/527
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/813


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -421,6 +421,6 @@ class ShareExtensionViewController: SLComposeServiceViewController {
 
 private func titleForMissingClients(users: Set<ZMUser>) -> String {
     let template = users.count > 1 ? "meta.degraded.degradation_reason_message.plural" : "meta.degraded.degradation_reason_message.singular"
-    let allUsers = (users.map { $0.displayName ?? "" } as NSArray).componentsJoined(by: ", ") as NSString
+    let allUsers = (users.map { $0.displayName } as NSArray).componentsJoined(by: ", ") as NSString
```
daehn: You can use `users.map(\.displayName)` instead (defined in `wire-ios-system`).

mikeger:
@typfel do you know why the share controller fixtures are changed?

typfel:
@mikeger which file a you referring to?

mikeger:
For example testThatItRendersCorrectlyShareViewController_Photos_{1024, 768}@2x.png and some around it

mikeger:
I mean, the change is minimal, I am just wondering :)

typfel:
They were failing due to sub pixel changes. Maybe because new image decoding code is slightly different.

typfel:
I'd say the pictures are less blurry now.

Wire-iOS/Sources/Helpers/syncengine/ZMTextMessageData+Images.swift
```diff
@@ -0,0 +1,120 @@
...
+fileprivate protocol ImageMessage {
+    
+    var cacheIdentifier: String? { get }
+    
+    func requestImageDownload()
+    func fetchImageData(queue: DispatchQueue, completionHandler: @escaping (_ imageData: Data?) -> Void)
+    
+}
+
+extension ImageMessage {
```
mikeger: We can potentially create the `UIImageView` subclass that would only care about loading an image from given `ImageMessage`.

mikeger:
Huge change, I really like the new version, thanks!

Wire-iOS/Sources/Helpers/syncengine/ZMTextMessageData+Images.swift
typfel: I think we could. What would this subclass be responsible for? Showing a loading indicator until the image is displayed. Images come many different places though and ImageMessageCells have separate logic since they deal with GIFs as well.

How did you imagine the API would look like?

`UIImageVIew.loadImage(from: ImageMessage)`

mikeger: I was thinking to have something simple, where you could just say `imageView.message = imageMessage` and it would show loading, load GIF or static, etc. May be it is not that easy, also nothing that we must do just now.

typfel: Yeah, might worth unifying the GIF and non gif image paths would enable nicer API.

daehn:
@typfel conflicting

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -154,17 +157,29 @@ final public class BackgroundViewController: UIViewController {
...
-            self.setBackground(imageData: imageData)
-        } else {
-            if let searchUser = user as? ZMSearchUser, let userSession = self.userSession {
-                searchUser.requestMediumProfileImage(in: userSession)
+        updateForUserImage()
+        updateForAccentColor()
+    }
+
+    private func updateForUserImage() {
+        dispatchGroup.enter()
```
billypchan, line 165: @typfel why dispatchGroup is needed here? I found in the tests that the `dispatchGroup` leaks in Xcode 12 test set.

PR #2455, opened 2018-07-26
New Conversation System Messages

johnxnguyen:
## What's in this PR?

**This PR updates the "XYZ started a new conversation" and "XYZ added ZYX" system message and is still WIP**

### This PR depends on:

- [x] A new release of `wire-ios-data-model` after https://github.com/wireapp/wire-ios-data-model/pull/515 has been merged.

### The following tasks still need to be done:

- [x] Add logic to present the participants details screen when tapping on the "and 3 others" link.
- [x] Update the "XYZ started the conversation" system message to collapse the names into a link.
- [x] Update the "XYZ started the conversation" system message to collapse all names into a link of the form "with everyone" when all team members have been added.
- [x] Update the "XYZ started the conversation" system message to collapse all names into a link of the form "with everyone and XYZ" guests when all team members and XYZ guests have been added.
- [x] Update the "XYZ started the conversation" system message the same way for the case that the conversation does not have a name (this is only for backwards compatibility as old clients might still me able to create conversations without names).
- [x] Animate and select the added / selected users in the participants detail screen after tapping on the "and 3 others" link in the new system message. The users that should be selected are already being passed into the correct controller, but the logic to highlight them is not there yet.
- [x] Ensure the logic for the naming when the self user is included "XYZ and you" instead of "XYZ, you" works the same way for the added / removed system messages.


johnxnguyen:
Single failing flaky test... all tests passing locally.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsCellViewModel.swift
```diff
@@ -17,138 +17,186 @@
...
+    /// Users not displayed in the system message but collapsed into a link.
+    /// E.g. `and 5 others`.
+    private var collapsedUsers: [ZMUser] {
+        let users = sortedUsersWithoutSelf()
+        guard users.count > maxShownUsers, action.allowsCollapsing else { return [] }
+        return Array(users.dropFirst(maxShownUsersWhenCollapsed))
+    }
+    
+    /// The users to display when opening the participants details screen.
+    var selectedUsers: [ZMUser] {
```
daehn: These are the users highlighted when tapping the "and X other", aren't they? If so shouldn't it only be the `collapsedUsers` and also for cases other than `.added`? I am probably missing something and looking in the wrong place.

johnxnguyen: Yes this is the set of users to be highlighted when tapping on the link, but According to the spec (might have be communicated to me in person though), all users added should be highlighted, and only adding users leads to highlighting.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsStringFormatter.swift
```diff
@@ -0,0 +1,234 @@
...
+        let names: [String]
+        let collapsed: Int
+        let selfIncluded: Bool
+        
+        var totalUsers: Int {
+            return names.count + collapsed
+        }
+    }
+    
+    private let kYouStartedTheConversation = "content.system.conversation.with_name.title-you"
```
daehn: I would propose to move these constants into their own internal struct and to drop the `k` prefix to make it a bit more Swifty, e.g.
```swift 
struct Key {
    static let xStartedTheConversation = "content.system.conversation.with_name.title"
}
```

Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewController.swift
```diff
@@ -54,6 +59,20 @@ final class GroupParticipantsDetailViewController: UIViewController, UICollectio
...
+        
+        if firstLayout {
+            firstLayout = false
+            scrollToFirstHighlightedUser()
+        }
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        firstLoad = false
```
daehn: Do we need to set `firstLoad` to false in two places? As we use it to decide whether we should scroll which we do from `viewDidLayoutSubviews` I think we can remove it from here.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsStringFormatter.swift
johnxnguyen: giphy

Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewController.swift
daehn: Ah sorry, missed that 👍

PR #2457, opened 2018-07-27
ZIOS-10339: Do not allow starting the auth flow with invalid SSO flow

alexaubry:
### Issues

We only validated the format of the SSO code before starting the auth flow, which caused the BE to display a 404 error page when starting it for codes that were not registered with it. We also started the flow without an internet connection, which caused Safari to show an error message.

As these errors are preventable, we need to handle them before starting the flow.

### Solutions

- We verify that there is an internet connection before starting the flow, and show an error otherwise
- We make a HEAD request to BE to verify that the SSO code exists. During this time, we show the spinner
    - If the code exists, we start the flow by opening the in-app browser
    - If the code does not exist, we show an alert

## Dependencies

Needs releases with:

- [x] #2453


Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -158,15 +162,46 @@ import Foundation
...
+
         guard let uuid = CompanyLoginRequestDetector.requestCode(in: code) else {
             return requireInternalFailure("Should never try to login with invalid code.")
         }
 
-        requester.requestIdentity(for: uuid)
+        delegate?.controller(self, showLoadingView: true)
+
+        requester.validate(token: uuid) {
+            guard !self.handleValidationErrorIfNeeded($0) else { return }
```
daehn: Should we hide the loading view?

daehn: Ah just saw it's done in `handleValidationErrorIfNeeded` method, but I would propose to move it here, as it's unrelated to the error handling and then showing and hiding it are together in the same method.

alexaubry: We hide it in `handleValidationErrorIfNeeded`

alexaubry:
@daehn updated!

PR #2458, opened 2018-07-27
Fix: Speed difference between conversation and keyboard when swipe to dismiss the keyboard

billypchan:
### Issues

When the user swipe to dismiss the keyboard on the conversation screen, the speed of the table view is faster than the keyboard. As the result, they are not moving at the same speed.

### Causes

The contentOffset's y value of the tableView increases when the user pans on the top edge of the keyboard.

### Solutions

Override contentOffset of the tableView. Do not change the offset if the user is panning at the edge of the keyboard.


mikeger:
Thanks, that was bothering people since the release date!

PR #2461, opened 2018-07-27
Remove iOS 9 version checking

billypchan:
## What's new in this PR?

Remove the unnecessary iOS version checking for the iOS8 era.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
@@ -202,12 +202,10 @@ private struct InputBarConstants {
         
     override public func didMoveToWindow() {
         super.didMoveToWindow()
-        // This is a workaround for UITextView truncating long contents.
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -202,12 +202,10 @@ private struct InputBarConstants {
...
     override public func didMoveToWindow() {
         super.didMoveToWindow()
-        // This is a workaround for UITextView truncating long contents.
-        // However, this breaks the text view on iOS 8 ¯\_(ツ)_/¯.
-        if #available(iOS 9, *) {
-            textView.isScrollEnabled = false
-            textView.isScrollEnabled = true
-        }
+
+        textView.isScrollEnabled = false
```
typfel: Maybe let's keep comment: `// This is a workaround for UITextView truncating long contents.` since it's not obvious why we toggle this on and off.

PR #2463, opened 2018-07-30
ZIOS-10341: Tracking is set to No after accessing New Team page

nicorsm:
### Issues

Tracking was disabled every time the team creation flow was shown.

### Solutions

I've moved the deactivation to `LandingViewController` only in the case there are no accounts left logged into the app.


Wire-iOS/Sources/UserInterface/Registration/RegistrationTeam/LandingViewController.swift
```diff
@@ -285,6 +285,7 @@ final class LandingViewController: UIViewController, CompanyLoginControllerDeleg
     private func updateBarButtonItem() {
         if SessionManager.shared?.firstAuthenticatedAccount == nil {
             navigationBar.topItem?.rightBarButtonItem = nil
+            TrackingManager.shared.disableCrashAndAnalyticsSharing = true
```
typfel: Doing this in `updateBarButtonItem` seems hard to explain. Maybe copy the `if`-statement into it's own method?

PR #2468, opened 2018-07-31
Fix: Prevent showing keyboard when conversation list is revealed

billypchan:
### Issues
1. The user toggles audio recording keyboard
2. The user receives a call
3. The user minimizes the call UI, 
4. The user opens the conversation list
5. The call ends

unexpected result: the audio keyboard shows over the conversation list

### Causes

missing checking of conversation screen is visible before revealing audio recording keyboard.

### Solutions

Add the checking before showing the keyboard.

### Todo
- [ ] check the case conversation details screen is on top and the audio keyboard would not be restored.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -234,6 +234,10 @@ extension ConversationInputBarViewController: WireCallCenterCallStateObserver {
     }
 
     private func displayRecordKeyboard() {
+        // do not show keyboard if conversation list is shown
+        guard let splitViewController = self.wr_splitViewController,
+              !splitViewController.isLeftViewControllerRevealed else { return }
+
```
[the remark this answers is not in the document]
billypchan: Good point! I will check it.

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+rightView.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    /// return true if right view (mostly conversation screen) is visible
+    /// for iPad regular mode the right view is always partially or fully visible
+    var isRightViewControllerRevealed: Bool {
+        switch self.layoutSize {
+        case .compact:
+            return !isLeftViewControllerRevealed
+        case .regularLandscape, .regularPortrait:
+            return true
+        }
```
alexaubry: I think in portrait we can collapse the left view controller? Maybe it's still worth checking for this case.

billypchan: In regular portrait mode, the conversation list is shown along with the conversation view. The keyboard is active in this case.

billypchan:
@mikeger Added isVisible checking for the rightViewController

PR #2470, opened 2018-07-31
Fix: It is necessary to click Audio Call button one more time after microphone alert is accepted

billypchan:
### Issues

The user can not start an audio call after a new installation. He/she has to press the phone button again after agreed microphone usage permission.

### Causes

 For a known reason, when the app is first-run and the user permitted microphone access, CXCallController's request method returns CXErrorCodeRequestTransactionErrorMaximumCallGroupsReached even maximumCallGroups is set already. 
### Solutions

Call RunLoop.run(until:) to wait for processesing data from all attached input sources.


Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
```diff
@@ -76,7 +76,12 @@ extension ZMConversation {
                     onGranted(granted)
                 }
             } else {
-                onGranted(granted)
+                // Hack: For a known reason, when the app is first-run and the user permitted microphone access,
+                // CXCallController's request method returns CXErrorCodeRequestTransactionErrorMaximumCallGroupsReached
+                // even maximumCallGroups is set already. Delay for a short moment solves the issue.
+                delay(0.1) {
```
mikeger: Frequently it's enough to call something on the next run loop, may be this would also work in this case? Something like 
```swift
DispatchQueue.main.async() {
    onGranted(granted)
}
```

billypchan: @mikeger checked that the caller of this line (wr_requestOrWarnAboutMicrophoneAccess) has put the closure in the main thread already.

mikeger: True, but I proposed it not to make it happen on the main thread, but rather to execute it on the next run loop spin.

Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
```diff
@@ -76,7 +76,12 @@ extension ZMConversation {
                     onGranted(granted)
                 }
             } else {
-                onGranted(granted)
+                // Hack: For a known reason, when the app is first-run and the user permitted microphone access,
```
vytis: you probably mean "for unknown reason" ?

PR #2477, opened 2018-07-31
ZIOS-10319: Improve share extension for links

nicorsm:
### Issues

Currently, the Share Extension appends the original URL to the content shared through the extension, even if the user decides to hide or change it. 

### Solutions

I'm updating the URL attachments of the current `PostContent` object before sending it. I've also removed the old code that appends the original URL.

### DIFF
--- a/Wire-iOS Share Extension/ShareExtensionViewController.swift
+++ b/Wire-iOS Share Extension/ShareExtensionViewController.swift
@@ -177,6 +177,8 @@ class ShareExtensionViewController: SLComposeServiceViewController {
     @objc func appendPostTapped() {
         navigationController?.navigationBar.items?.first?.rightBarButtonItem?.isEnabled = false
 
+        updateUrlAttachments()
+        
         postContent?.send(text: contentText, sharingSession: sharingSession!) { [weak self] progress in
             guard let `self` = self, let postContent = self.postContent else { return }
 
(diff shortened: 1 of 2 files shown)


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -223,6 +225,15 @@ class ShareExtensionViewController: SLComposeServiceViewController {
             }
         }
     }
+    
+    private func updateUrlAttachments() {
+        let urlDetector = try! NSDataDetector(types: NSTextCheckingResult.CheckingType.link.rawValue)
+        let matches = urlDetector.matches(in: contentText, options: [], range: NSMakeRange(0, contentText.count))
```
johnxnguyen: it looks like `contentText` is a `String`. Be careful, because the `count` is not the same as the `length` of an `NSString`. It could happen (if the content has certain emojis) that the upper bound of your range is actually lower than you expect, and so a url might not be detected even though it is there.

johnxnguyen: I would do this: `NSMakeRange(0, (contentText as NSString).length)

nicorsm: Thanks @johnxnguyen for the feedback!

PR #2478, opened 2018-08-01
Feature: Zoom in small image in full screen image viewer

billypchan:
## What's new in this PR?

Allow the user to zoom in a small image in full screen image viewer to fit the shorter side of the screen.

## Notes

- [x] check zoom scale is updated correctly after the device is rotated.


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController+Zoom.swift
```diff
@@ -32,18 +32,37 @@ extension CGSize {
 
         return minZoom
     }
+
+
+    /// returns the longest length among width and height
+    var longestLength: CGFloat {
```
mikeger: I think it might sound better in English to say `dimension`, may be @johnxnguyen can help us.

billypchan: or just call it maxSide?

johnxnguyen: I would say *extent*.

PR #2488, opened 2018-08-02
Fix build errors with Xcode 10

alexaubry:
### Issues

The frameworks and other Carthage dependencies build with Xcode 10 beta 5. It should be possible to build the UI as well. This PR fixes the problems encountered with building on Xcode 10, while maintaining compatibility with Xcode 9.4, so that migrating will be faster in the future. The Swift version is still Swift 4.1.

### Causes

Several build issues were risen by the new Swift compiler:

- Cannot make `let` variable `open` on `final` classes (replace with `public`)
- Some type casting needed to be made optional
- Deprecation alert for hiding the status bar (fixed using the `wr_` utility method for that)
- Fixed expressions that couldn't be parsed (too complex or ambiguous)
- Fix build cycle
- Fix overflow in Hashable implementation


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -7703,6 +7703,7 @@
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks @loader_path/Frameworks $(PROJECT_DIR)/Carthage/Build/iOS";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SKIP_INSTALL = YES;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
```
mikeger: May be worth putting to the xcconfig file?

mikeger:
I see the swift is not as open as it used to be :) great job making it compile with the new Xcode. Is the compilation performance getting better there?

WireExtensionComponents/Utilities/FontScheme.swift
```diff
@@ -157,14 +157,16 @@ extension FontSpec {
     }
 }
 
-extension FontSpec: Hashable {
+#if !swift(>=4.2)
```
vytis: Can't we just use default Hashable implemetation also on 4.1? I think it is there already

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
```diff
@@ -206,7 +206,13 @@ extension ChatHeadsViewController {
         
         case .changed:
             // if pan left, move chathead with finger, else apply pan resistance
-            let viewOffsetX = offset.x < 0 ? offset.x : (1.0 - (1.0/((offset.x * 0.15 / view.bounds.width) + 1.0))) * view.bounds.width
+            var viewOffsetX: CGFloat = 0
```
daehn: I think before it returned `offset.x` in case `offset.x` is smaller than 0, now it returns 0?

ConfigurationSets/Development.xcconfig
```diff
@@ -68,7 +68,7 @@ STRIP_INSTALLED_PRODUCT = NO
 // Code Generation
 //
 GCC_OPTIMIZATION_LEVEL = 0
-
+SWIFT_OPTIMIZATION_LEVEL = -Onone
```
daehn: Development.xcconfig is not included in all others AFAIK, maybe it's better to put it into Base.xcconfig instead.

Wire-iOS.xcodeproj/project.pbxproj
alexaubry: Updated 👍

ConfigurationSets/Development.xcconfig
alexaubry: `Development.xcconfig` is included in all the targets, so it should be fine?

alexaubry:
@mikeger A clean build took 2min 45s on Xcode 10, vs. 4min 10s on Xcode 9.4. Definitely feels better!

Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift
alexaubry: Fixed

WireExtensionComponents/Utilities/FontScheme.swift
alexaubry: We can't, because the struct contains Optional values, and Optional is Hashable only from Swift 4.2.

Maybe we could make Optional Hashable manually (by mimicking the implementation from Swift 4.2 https://github.com/apple/swift/blob/f1e9a048e59d8e1aa498dee4880f5d0754c3e13d/stdlib/public/core/Optional.swift#L409)?

vytis: A simpler thing would be to add default parameters and get rid of optionals here altogether. There is no good reason to have it I think? @mikeger might know more about it.

mikeger: I don't remember exactly why it's optional, logically does not seem to be necessary, worth trying to remove optionals.

alexaubry: It seems to be working without optionals, see the new changes.

alexaubry:
Some tests are failing, will investigate.

mikeger:
@alexaubry what's the status of this one?

mikeger:
@vytis @alexaubry this is getting too much diverged from the main branch. Should we spend some time on fixing and merging it? Or cannot be merged until we switch to the neXcode

vytis:
I was trying to fix this in spare time, there are still some issues with FontSpec that I am investigating. Maybe will spend more time today to get it to a reasonable shape!

vytis:
I slightly decreased scope by not doing automatic Hashable conformance. This was the cause of most test failures are quite tricky to find the reason when there are so many other changes.

PR #2490, opened 2018-08-02
Feature/conversation overflow alerts

johnxnguyen:
## What's new in this PR?

This PR provides changes for dealing with participant overflow when creating a conversation and adding users to an existing conversation. These are:

1. When selecting participants for a new conversation, it is only possible to select 299 (other) participants. Selecting the 300th participant presents an alert.
2. When adding participants to an existing conversation, it is only possible to select `x` participants, where x is the number of remaining free spaces in the conversation. Selecting the `(x + 1)`'th participant also shows the alert.
3. If the conversation is full, the *add participants* button in the group details screen is disabled.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -98,8 +98,8 @@
 "add_participants.all_contacts_added" = "Everyone’s here.";
 
 "add_participants.alert.title" = "Full house";
-"add_participants.alert.message.new_conversation" = "Up to 300 people can join a conversation.";
-"add_participants.alert.message.existing_conversation" = "Up to 300 people can join a conversation. Currently there is only room for %@ more people.";
+"add_participants.alert.message.new_conversation" = "Up to %d people can join a conversation.";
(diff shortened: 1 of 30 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -97,6 +97,10 @@
 
 "add_participants.all_contacts_added" = "Everyone’s here.";
 
+"add_participants.alert.title" = "Full house";
+"add_participants.alert.message.new_conversation" = "Up to 300 people can join a conversation.";
```
mikeger: I think it would be cool to have the format replacement in this string (like %d) and insert the value of `maxParticipants` dynamically. Also to https://github.com/wireapp/wire-ios/blob/f0423f679c9f5c9bc739acef05f4e4b9f497c557/Wire-iOS/Resources/Base.lproj/Localizable.strings#L590

johnxnguyen: @mikeger is it convention to always pass in Strings as arguments? Or we just never pass in digits?

mikeger: We are free to do as we pleased, since those strings are only for us :)

PR #2492, opened 2018-08-03
ZIOS-10388: Fix e-mail sign in lock when SSO is disabled

alexaubry:
### Issues

In public builds where company login is disabled, we always disabled the e-mail sign-in field.

### Causes + Solutions

E-mail autofill lock was mistakenly tied to the detection of the SSO feature, which is disabled on public builds. We fixed this by decoupling the logic of auto-fill lock and company login flow button.


alexaubry:
Closed in favor of #2493 (wrong branch)

PR #2500, opened 2018-08-03
Fix: Register screen animation issues

billypchan:
### Issues

- When Register screen appears, some of the subviews are resizing at the same time.
- When switching between registration/log in tabs, the subviews moves down and then moves up.

### Causes

In some view controllers, after the constraints are created the layout is not updated.
The KeyboardAvoidingViewController always starts an animation when keyboard's frame is updated.  When the first responder changes, two keyboard frame change notifications are fired (the keyboard dismisses and the reveals with another height) and the KeyboardAvoidingViewController "jumps".

### Solutions

Add [self.view layoutIfNeeded] in viewWIllAppear in several classes which found the frame is zero when viewDidAppear.

Rewrite KeyboardAvoidingViewController's animation code adapting iOS 10's UIViewPropertyAnimator, which allows stops running animation before starting a new one.

### TODO
- [x] fix the broken constraints in RegistrationRootViewController


mikeger:
@billypchan has conflict with other changes now.

PR #2501, opened 2018-08-06
Fix: Audio recorder overlay is shown on top of calling overlay if I accept the second call

billypchan:
### Issues

Audio keyboard reveals over the call window.

### Causes

When the call terminates, missing a checking for mainWindow is the top window or not before restoring audio keyboard.

### Solutions

Add the checking for the current top window is the main window before revealing the keyboard.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
@@ -239,7 +239,7 @@ extension ConversationInputBarViewController: WireCallCenterCallStateObserver {
               let rightViewController = splitViewController.rightViewController,
               splitViewController.isRightViewControllerRevealed,
               rightViewController.isVisible,
-              UIApplication.shared.topMostWindow == AppDelegate.shared().window
+              UIApplication.shared.topMostVisibleWindow == AppDelegate.shared().window
             else { return }
 
         self.wasRecordingBeforeCall = false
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Audio.swift
```diff
@@ -238,7 +238,9 @@ extension ConversationInputBarViewController: WireCallCenterCallStateObserver {
         guard let splitViewController = self.wr_splitViewController,
               let rightViewController = splitViewController.rightViewController,
               splitViewController.isRightViewControllerRevealed,
-              rightViewController.isVisible else { return }
+              rightViewController.isVisible,
+              UIApplication.shared.topMostWindow == AppDelegate.shared().window
```
daehn: I am a bit surprised this works, I was under the impression that the notification window (the one we display chatheads in) or the call overlay window would always be the topmost window.

mikeger: @daehn check the code, we are skipping call window if no call is active (method `topMostWindow` is our own one).

billypchan: The UIApplication.topMostWindow is a new property extracted from wr_topmostController method. It checks:
1. the windows has rootViewController
2. voice channel controller is active or not
3. the window's rootViewController is AppRootViewController 

notification window cannot pass #3.

daehn: Skipping the call window when no call is active makes sense, but the naming of the method might still be a bit confusing in my opinion. As the notification window can't be returned from calling `topMostWindow` but in theory it can be the top most window.

billypchan: I will rename it to topMostVisibleWindow

PR #2505, opened 2018-08-07
ZIOS-10393: Create modular authentication flows

alexaubry:
### Issues

In its current form, the authentication (login/registration) flows are not modular and require heavy modification when the spec changes (ex: adding a backup step, changing the flow for SSO, ...). This also introduces potential points of failure, when for example and error is handled by two different view controllers, or ignored by all.

We want view controllers to follow the single responsibility principle as much as possible, and avoid having multiple levels of delegation.

### Solutions

Instead of putting the burden of logic decision and state transition on the authentication view controllers, we introduce a single `AuthenticationCoordinator`, that handles events from the backend, the user session and input from the user.

The coordinators observes all the notifications from the unauthenticated session and slow sync completion to determine what the next step is. Steps are defined inside an enum that stores associated context objects (ex: authentication credentials, verification code, ...).

Steps can advertise themselves as non-interface (ex: only alerts). If the step says it needs an interface, the coordinator will generate one and push it to the navigation stack, using its `presenter`. Steps can also decide to wipe the entire stack (ex: it should not be possible to go back when removing a client).

Generated view controllers have a reference to the `authenticationCoordinator`. They can use this property to perform actions after user confirm inputs (e-mail credentials verification, complete backup, ...). The coordinator also holds a reference to the current view controller, and can ask it to execute some action (ex: showing the guidance dot when the password is invalid).

The coordinator shows the loading view when appropriate, and presents alert view controllers when needed.

The coordinator has a delegate: the app state controller. The coordinator notifies its delegate when authentication completes. It also queries the delegate for information (ex: number of accounts already registered) instead of using singletons or shared state to allow testability.

We create the coordinator and start the flow in the app root view controller, when the state changes to unauthenticated. The error provided by the sync engine will be used to determine the first step.

### To-Do

- [x] Create core infrastructure
- [x] Port login with e-mail
- [x] Port team registration
- [x] Port user registration with e-mail
- [x] Port user registration with phone number
- [x] Port user settings

### Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/542
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/69
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/835
- [x] https://github.com/wireapp/wire-ios-share-engine/pull/62


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -205,56 +206,42 @@ var defaultFontScheme: FontScheme = FontScheme(contentSizeCategory: UIApplicatio
...
+                authenticationCoordinator = AuthenticationCoordinator(presenter: navigationController,
+                                                                      unauthenticatedSession: UnauthenticatedSession.sharedSession!,
+                                                                      sessionManager: SessionManager.shared!)
+
+                authenticationCoordinator?.delegate = appStateController
+
+                updatedViewController = navigationController
+            }
+
+            authenticationCoordinator!.startAuthentication(with: error, numberOfAccounts: SessionManager.numberOfAccounts)
```
daehn: Maybe optional chaining is better here even though it never should be `nil`.

daehn:
@alexaubry the check list has a bullet point "port team registration" which I don't think is necessary for now.

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -116,7 +116,7 @@
          </CommandLineArgument>
          <CommandLineArgument
             argument = "-ZMBackendEnvironmentType staging"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
mikeger: Please disable that one.

Wire-iOS/Sources/Authentication/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -0,0 +1,736 @@
...
+        }
+    }
+
+    // MARK: - Add Email And Password
+
+    /**
+     * Skips the add e-mail and password step, if possible.
+     */
+
+    @objc func skipAddEmailAndPassword() {
```
mikeger: Is the method necessary, even if it is not doing anything?

Wire-iOS/Sources/Authentication/RegistrationPersonal/NagScreen/AddPhoneNumberViewController.m
```diff
@@ -203,111 +179,11 @@ - (void)updateViewConstraints
...
-    }
-    
-    [self showAlertForError:error];
-}
-
-- (void)phoneNumberChangeDidFail:(NSError *)error
-{
-    self.showLoadingView = NO;
-    [self showAlertForError:error];
+    // TODO: Remove
```
mikeger: Can be removed.

Wire-iOS/Sources/Permissions/ShareContactsViewController.m
```diff
@@ -177,7 +177,7 @@ - (IBAction)shareContacts:(id)sender
     [AddressBookHelper.sharedHelper requestPermissions:^(BOOL success) {
         if (success) {
             [[AddressBookHelper sharedHelper] startRemoteSearchWithCheckingIfEnoughTimeSinceLast:self.uploadAddressBookImmediately];
-            [self.formStepDelegate didCompleteFormStep:self];
+            // [self.formStepDelegate didCompleteFormStep:self];
```
mikeger: Can be removed?

Wire-iOS/Sources/Permissions/ShareContactsViewController.m
```diff
@@ -187,7 +187,7 @@ - (IBAction)shareContacts:(id)sender
 - (IBAction)shareContactsLater:(id)sender
 {
     [AddressBookHelper sharedHelper].addressBookSearchWasPostponed = YES;
-    [self.formStepDelegate didSkipFormStep:self];
+    // [self.formStepDelegate didSkipFormStep:self];
```
mikeger: Can be removed?

mikeger: (2x times more in this file)

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+ShareContacts.m
```diff
@@ -52,6 +51,7 @@ - (void)dismissChildViewController:(UIViewController *)viewController
     }];
 }
 
+// TODO: Remove
```
mikeger: Still needs to be removed I guess? We have no steps delegate any more.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -116,11 +116,10 @@ extension SettingsCellDescriptorFactory {
                 return ChangeEmailViewController()
              } else {
                 let addEmailController = AddEmailPasswordViewController()
-                addEmailController.showsNavigationBar = false
                 let stepDelegate = DismissStepDelegate()
                 stepDelegate.strongCapture = stepDelegate
-                
-                addEmailController.formStepDelegate = stepDelegate
+
+                // TODO: Integrate with new login controller
```
mikeger: Needs to be done still?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -149,7 +148,7 @@ extension SettingsCellDescriptorFactory {
                     let stepDelegate = DismissStepDelegate()
                     stepDelegate.strongCapture = stepDelegate
                     
-                    addController.formStepDelegate = stepDelegate
+//                    addController.formStepDelegate = stepDelegate
```
mikeger: Can be removed.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -24,8 +24,9 @@ import SafariServices
     static let settingsDevicesCellIdentifier: String = "devices"
     let settingsPropertyFactory: SettingsPropertyFactory
     
-    class DismissStepDelegate: NSObject, FormStepDelegate {
+    class DismissStepDelegate: NSObject {
         var strongCapture: DismissStepDelegate?
+        // TODO: Remove
         @objc func didCompleteFormStep(_ viewController: UIViewController!) {
```
mikeger: I think FormStepDelegate has to go.

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientUnregisterFlowViewController.swift
```diff
@@ -83,7 +76,7 @@ import Cartography
     
     fileprivate func setupNavigationController() {
         let invitationController = ClientUnregisterInvitationViewController()
-        invitationController.formStepDelegate = self
+        // invitationController.formStepDelegate = self
```
mikeger: Can be removed.

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientUnregisterFlowViewController.swift
```diff
@@ -116,7 +109,8 @@ import Cartography
     }
 
     // MARK: - FormStepDelegate
-    
+
+    // TODO: Remove
```
mikeger: ✂️

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientUnregisterInvitationViewController.swift
```diff
@@ -132,9 +132,9 @@ class ClientUnregisterInvitationViewController: RegistrationStepViewController {
     // MARK: - Actions
     
     @objc func openManageDevices(_ sender : UIButton!) {
-        if let formStepDelegate = self.formStepDelegate {
-            formStepDelegate.didCompleteFormStep(self)
-        }
+//        if let formStepDelegate = self.formStepDelegate {
```
mikeger: 🔪

mikeger:
I think it can be useful to separate the logic and observation, moving it to the sync engine or data model. We can do some brainstorming about it later, however can improve our SE public API. This can be done as the second iteration I guess.

Wire-iOS/Sources/Authentication/RegistrationPersonal/RegistrationViewController.m
```diff
@@ -0,0 +1,139 @@
...
+@property (nonatomic) BOOL registeredInThisSession;
+
+@property (nonatomic) RegistrationRootViewController *registrationRootViewController;
+@property (nonatomic) AuthenticationFlowType flowType;
+
+@end
+
+
+
+@implementation RegistrationViewController
```
vytis: Maybe some of these controllers could be re-written in Swift? Probably not all, but there are a few that are below 200 lines ObjC and almost no logic.

Wire-iOS/Sources/Authentication/Authentication/Coordinator/AuthenticationCoordinator.swift
alexaubry: Right, this is not necessary so I removed it.

Wire-iOS/Sources/Authentication/RegistrationTeam/AccessoryTextField.swift
```diff
@@ -136,14 +136,24 @@ class AccessoryTextField: UITextField {
...
             autocapitalizationType = .none
             accessibilityIdentifier = "EmailField"
-        case .password:
+            textContentType = .emailAddress
+        case .password(let isNew):
             isSecureTextEntry = true
             accessibilityIdentifier = "PasswordField"
-        case .name:
+            if #available(iOS 12, *) {
+                textContentType = isNew ? .newPassword : .password
```
vytis: 🥇🥈

Wire-iOS/Sources/Authentication/RegistrationPersonal/RegistrationViewController.m
alexaubry: Agreed. Might be worth doing in a second phase.

Wire-iOS Tests/AccessoryTextField/AccessoryTextFieldTests.swift
```diff
@@ -65,7 +65,7 @@ final class AccessoryTextFieldTests: ZMSnapshotTestCase {
 
     func testThatItShowsPasswordInputedAndConfrimButtonIsEnabled() {
```
typfel: typo: `testThatItShowsPasswordInputedAndConfrimButtonIsEnabled` >  `testThatItShowsPasswordInputedAndConfirmButtonIsEnabled`

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -267,8 +249,10 @@ var defaultFontScheme: FontScheme = FontScheme(contentSizeCategory: UIApplicatio
 
             viewController = clientViewController
         case .headless:
+            authenticationCoordinator = nil
             viewController = LaunchImageViewController()
         case .loading(account: let toAccount, from: let fromAccount):
+            authenticationCoordinator = nil
```
typfel: Since `authenticationCoordinator = nil` is happening for call cases except `.unauthenticated` we could break this out of the switch statement.

```
if case .unauthenticated = appState {} else {
    authenticationCoordinator = nil
}
```

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+ShareContacts.m
alexaubry: Replace the steps delegate with a dedicated delegate (see 8b4b8fc).

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -4758,7 +5050,6 @@
 				87775AB21D95138D000CAD4F /* avs-versions */,
 				872AC9C21B397E6A00575CF3 /* Cartfile */,
 				87A418BD1C0CAB1A006FA803 /* Cartfile.private */,
-				1E88CC4D1BEB63020020B853 /* Cartfile.resolved */,
```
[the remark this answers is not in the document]
alexaubry: Ah I hadn't noticed. Put it back.

Wire-iOS Tests/Authentication/AuthenticationStateControllerTests.swift
```diff
@@ -0,0 +1,151 @@
...
+    override func setUp() {
+        delegate = MockAuthenticationStateControllerDelegate()
+        stateController = AuthenticationStateController()
+        stateController.delegate = delegate
+    }
+
+    override func tearDown() {
+        delegate = nil
+        stateController = nil
+    }
```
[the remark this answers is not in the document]
alexaubry: Fixed

Wire-iOS Tests/Authentication/AuthenticationStateControllerTests.swift
```diff
@@ -0,0 +1,151 @@
...
+
+}
+
+class AuthenticationStateControllerTests: XCTestCase {
+
+    var stateController: AuthenticationStateController!
+    var delegate: MockAuthenticationStateControllerDelegate!
+
+    override func setUp() {
+        delegate = MockAuthenticationStateControllerDelegate()
```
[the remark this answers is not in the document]
alexaubry: Fixed

Wire-iOS Tests/AccessoryTextField/AccessoryTextFieldTests.swift
alexaubry: Fixed

Wire-iOS/Sources/AppRootViewController.swift
alexaubry: I added a method: `resetAuthenticationCoordinatorIfNeeded(for state: AppState)` that we call before the state transition. We check for the test case in it.

Wire-iOS/Sources/Authentication/Authentication/AuthenticationFlowStep.swift
```diff
@@ -0,0 +1,128 @@
...
+    case incrementalUserCreation(UnregisteredUser, IntermediateRegistrationStep)
+    case createUser(UnregisteredUser)
+
+    // MARK: - Properties
+
+    /// Whether the step can be unwinded.
+    var allowsUnwind: Bool {
+        switch self {
+        case .landingScreen, .clientManagement, .noHistory, .addEmailAndPassword, .incrementalUserCreation: return false
+        case .enterLoginCode: return true
```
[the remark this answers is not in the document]
alexaubry: I removed the true cases altogether as `true` is the default value, and expanded the false values for more clarity.

Wire-iOS/Sources/Authentication/Authentication/Event Handlers/Post-Login/Success/AuthenticationSuccessPhoneRegistrationEventHandler.swift
```diff
@@ -0,0 +1,29 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+class AuthenticationSuccessPhoneRegistrationEventHandler: AuthenticationEventHandler {
```
typfel: Does this handler have a purpose since it always returns nil?

Wire-iOS/Sources/Authentication/Authentication/Event Handlers/Initial Sync/AuthenticationInitialSyncEventHandler.swift
```diff
@@ -0,0 +1,71 @@
...
+        guard let selfUser = statusProvider?.selfUser, let profile = statusProvider?.selfUserProfile else {
+            return nil
+        }
+
+        // Check if the user needs email and password
+        let isRegistered = statusProvider?.authenticatedUserWasRegisteredOnThisDevice ?? false
+        let needsEmail = statusProvider?.authenticatedUserNeedsEmailCredentials ?? false
+
+        switch (isRegistered, needsEmail) {
+        case (true, false):
```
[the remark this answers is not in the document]
alexaubry: Good idea! Done in f8cca5d.

Wire-iOS/Sources/Authentication/Authentication/Event Handlers/AuthenticationEventResponderChain.swift
```diff
@@ -0,0 +1,223 @@
...
+
+    /// The handle to the OS log for authentication events.
+    private let log = ZMSLog(tag: "Authentication")
+
+    /// The object assisting the responder chain.
+    weak var delegate: AuthenticationEventResponderChainDelegate?
+
+    // MARK: - Configuration
+
+    var flowStartHandlers: [AnyAuthenticationEventHandler<(NSError?, Int)>] = []
```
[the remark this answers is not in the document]
alexaubry: I'm not sure about this, because we might want to use the `Int` both when there is an error and when there isn't (we use it to get the number of currently saved accounts). I think `Result` is more for mutually exclusive values?

Wire-iOS/Sources/Authentication/Authentication/Event Handlers/Post-Login/Success/AuthenticationSuccessPhoneRegistrationEventHandler.swift
alexaubry: That was a leftover file from the time phone and email registration were separate code paths. It wasn't used in the responder chain anyway. I removed the file.

Wire-iOS/Sources/Authentication/Authentication/AuthenticationFlowStep.swift
```diff
@@ -0,0 +1,130 @@
...
+    // Sign-In
+    case provideCredentials
+    case sendLoginCode(phoneNumber: String, isResend: Bool)
+    case enterLoginCode(phoneNumber: String)
+    case authenticateEmailCredentials(ZMEmailCredentials)
+    case authenticatePhoneCredentials(ZMPhoneCredentials)
+    case companyLogin
+
+    // Post Sign-In
+    case noHistory(credentials: ZMCredentials?, type: Wire.ContextType)
```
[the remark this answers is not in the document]
alexaubry: Good call, I'll check this in SE.

alexaubry:
> `isRegistered` could be `_` as we handle both cases with identical bodies:
> 
> ```
> case (true, false, let nextStep?):
>         return [.hideLoadingView, .transition(nextStep, resetStack: true)]
> ```
> and
> 
> ```
> case (false, false, let nextStep?):
>          return [.hideLoadingView, .transition(nextStep, resetStack: true)]
> ```

Updated, that looks nicer indeed 👍

Wire-iOS/Sources/Authentication/Authentication/AuthenticationFlowStep.swift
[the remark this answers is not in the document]
alexaubry: I added the implementation to `LoginCredentials`. The only two associated values that don't have manual implementations of `isEqual` are `UserClient` and `ZMUser`. Though it should be fine, as most of the time, the two steps that use them won't be compared to each other directly (only the first part is check in a switch or if/guard pattern matching).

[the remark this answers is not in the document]
vytis: I think `UserClient` and `ZMUser` are fine because they are `NSManagedObject`s and CoreData makes sure there is only one instance of it in memory, so pointer equality works in this case.

Wire-iOS/Sources/Authentication/Authentication/Event Handlers/AuthenticationEventResponderChain.swift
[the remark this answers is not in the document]
vytis: That makes sense, I thought that it's some sort of "success" value and is mutually exclusive with error.

PR #2506, opened 2018-08-07
WIP: iOS 10

johnxnguyen:
## What's new in this PR?

This is the working branch for all changes related to dropping iOS 9 support.


vytis:
RE .xcconfig files - it would be best to move all configuration from both general project and all targets to corresponding .xcconfig files. It's easy to see which ones are defined inside .xcodeproj by going to build settings (of each target and the project itself) and selecting "Customized". There are not that many there, but maybe this tool can help to factor them out - https://github.com/dempseyatgithub/BuildSettingExtractor

johnxnguyen:
@vytis right, so we would ideally have one config file per target for all of those customised values?

vytis:
It depends, maybe the current system will work fine - one set for development another for release. If not then yes, adding a file for each target will be the other option.

typfel:
@mikeger @vytis let's do the xcconfig extraction in a separate PR. Also John updated the PR to only modify the top project deployment target so it shouldn't even be in the xcconfig.

PR #2513, opened 2018-08-08
Fix: Picture with ending .heic

billypchan:
### Issues

Non iOS client may be unable to open saved HEIF image file introduced in iOS 11 release.

### Solutions

Convert the file to JPG format for cross-platform compatibility. 

## Notes

We should review this conversion is necessary in the future when HEIF becomes more popular.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
```diff
@@ -205,15 +205,32 @@ open class CameraKeyboardViewController: UIViewController {
         self.collectionViewLayout.invalidateLayout()
         self.collectionView.reloadData()
     }
-    
+
     fileprivate func forwardSelectedPhotoAsset(_ asset: PHAsset) {
         let manager = PHImageManager.default()
 
+        ///TODO: jpg?
         let options = PHImageRequestOptions()
```
alexaubry: Can be removed?

PR #2514, opened 2018-08-08
Modular authentication: isolate event observing stack 

alexaubry:
### Issues

As explained in #2505, the new `AuthenticationCoordinator` observes events from the Sync Engine and performs actions and state transitions as appropriate.

However, this approach is not optimal, as it makes the coordinator quite big. In the current implementation, with only login, it was spanning over 700 lines of code.

We needed to do the event handing in a separate place to keep the coordinator modular and scalable.

### Solutions

In an approach related to the request strategies, we define sub-coordinators, with the `AuthenticationEventHandler` protocol.

This protocol has concrete implementations, that handle only one type of events each. They declare which context type they want.

These handlers are stored in the `AuthenticationEventHandlingManager`, and are sorted by event type. The authentication coordinator notifies the handling manager when a supported event occurs, and provides the context object.

Then, the handling manager loops through its compatible event handlers and asks for the actions / side effects of the event, by passing it the current step and the context object. 

If the handler returns nil, we go to the next handler, until an handler returns a list of actions (defined in `AuthenticationCoordinatorAction`). Handlers use the current step and context object to determine if they can handle the event.

The action is then executed in the authentication coordinator.

This approach:
- adds type safety to the event error handling
- makes it obvious where to search when we want to change the flow behavior
- reduces the size of the authentication coordinator
- makes it easier to step every event handler individually or together

We currently have these new handlers for the entire post-login flow and the landing screen:
- backup is ready to be imported
- client registration failed
- client registration succeed
- initial sync completed
- user becomes unauthenticated without error
- user becomes unauthenticated with error

## To-Do

- [ ] Define the behavior when no handler supports the action

## Notes

This PR will be merged in the WIP branch for #2505. Complete testing will be provided in a subsequent PR.


Wire-iOS/Sources/UserInterface/Registration/Authentication/Event Handlers/AuthenticationEventHandlingManager.swift
```diff
@@ -0,0 +1,145 @@
...
+        case clientRegistrationSuccess
+    }
+
+    // MARK: - Properties
+
+    weak var delegate: AuthenticationEventHandlingManagerDelegate?
+
+    // MARK: - Configuration
+
+    var flowStartHandlers: [AnyAuthenticationEventHandler<(NSError?, Int)>] = []
```
daehn: Not sure this is really type erased and needs to be named `AnyAuthenticationEventHandler` when the associated type is just moved to be a generic class. Maybe it's fine to rename `AnyAuthenticationEventHandler` to `AuthenticationEventHandler` and the protocol to `AnyAuthenticationEventHandlerType` or something like that. I would expect a type erased box to not give me any information about the type it contains (so that it can be stored in collections, having one collection of type `[AnyAuthenticationEventHandler]`), but maybe I am missing something there.

daehn: Seems like I was mistaken, `AnySequence` does the same!

alexaubry: Yeah, basically I need an array of `AuthenticationEventHandler` that have the same `Context` associated type, so that I can safely iterate over them in `handleEvent`, when we detect an event with the context type they support.

The `AnyAuthenticationEventHandler` holds the generic constraint for the `Context`, that we need if we want to store objects that conform to the `AuthenticationEventHandler` protocol in the sequence. I could also have used subclassing, but it wasn't necessarily the best option for testing.

PR #2516, opened 2018-08-09
fix: No border and background color for transparent images

billypchan:
## What's new in this PR?

Remove border and background color of image message cell for the transparent images.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ImageMessageCell+BackgroundColor.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ImageMessageCell {
+    @objc func updateBackgroundColor() {
+        if let isTransparent = fullImageView?.image?.isTransparent(), isTransparent {
```
alexaubry: If we don't show a background for transparent images, in which cases do we need to set a background color for the image?

billypchan: @alexaubry After the discussion with the design team, the background color is shown for all cases(image placeholder, non-transparent image and etc) except transparent images.

alexaubry: Ah for the placeholder, that makes sense. 👍

PR #2517, opened 2018-08-09
Migrate pre-login to modern event handler + add logs

alexaubry:
## What's new in this PR?

- We migrate the pre-login authentication event handling to the next model described in #2514 
- We add alert actions to the coordinator to be able to display alerts related to login
- We add an `Auth` log to catch state and error from the coordinator and the event observer


Wire-iOS/Sources/UserInterface/Registration/Authentication/Event Handlers/AnyAuthenticationEventHandler.swift
```diff
@@ -26,6 +26,7 @@ class AnyAuthenticationEventHandler<Context> {
 
     private let statusProviderGetter: () -> AuthenticationStatusProvider?
     private let statusProviderSetter: (AuthenticationStatusProvider?) -> Void
+    private let nameGetter: () -> String
```
mikeger: Not sure if there is a nicer way to get the class name of particular object, I guess you've already tried doing something like "\(type(of: self)" ?

Wire-iOS/Sources/UserInterface/Registration/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -145,29 +146,43 @@ extension AuthenticationCoordinator {
 
 // MARK: - Event Handling
 
-extension AuthenticationCoordinator {
+extension AuthenticationCoordinator: SessionManagerCreatedSessionObserver {
+
+    func sessionManagerCreated(userSession: ZMUserSession) {
+        log.info("Session manager created session: \(userSession)")
+
```
daehn: Not sure if this can potentially be called multiple times and we might end up having multiple observers, in that case it would be better to keep a single reference to this one and overwrite it instead of appending.

Wire-iOS/Sources/UserInterface/Registration/Authentication/Event Handlers/Flow Start/AuthenticationStartClientDeletionErrorHandler.swift
```diff
@@ -30,6 +30,11 @@ class AuthenticationStartClientDeletionErrorHandler: AuthenticationEventHandler
     func handleEvent(currentStep: AuthenticationFlowStep, context: (NSError?, Int)) -> [AuthenticationCoordinatorAction]? {
         let (optionalError, _) = context
 
+        // If we already are in the client management step, do nothing
+        if case .clientManagement = currentStep {
+            return []
```
daehn: Not sure when we return `nil` and when an empty array here, maybe we can change the signature to be non optional and always return an empty array (or we always try to return nil if there are no actions, but if possible it would be nice to avoid this ambiguity).

Wire-iOS/Sources/UserInterface/Registration/Authentication/Event Handlers/AnyAuthenticationEventHandler.swift
alexaubry: Good call. I am now using `String(describing: Handler.self)` to get the class name; much nicer indeed.

Wire-iOS/Sources/UserInterface/Registration/Authentication/Coordinator/AuthenticationCoordinator.swift
alexaubry: Didn't think about this. Just fixed it 👍

Wire-iOS/Sources/UserInterface/Registration/Authentication/Event Handlers/Flow Start/AuthenticationStartClientDeletionErrorHandler.swift
alexaubry: Returning `nil` asks the coordinator to try getting the actions from the next handler in the list, because the context is not valid. Returning an empty array means that the context is valid for this handler, but there is nothing to do, and we shouldn't query the next handler for actions (basically a `break` statement in the for loop). Maybe we should have something more clear, agreed. Or even better, not having situations where we do nothing.

daehn: Ah I see, yeah maybe it's worth introducing an enum for this so the enum and case names can make this more clear.

PR #2521, opened 2018-08-10
Services: Copy changes and respecting allow guests toggle

vytis:
## What's new in this PR?

* Services are treated as guests, so disabling guests should disable adding services as well.
* Copy changes reflect the fact that now conversation can contain both people and services.


mikeger:
@vytis no screenshots affected?

vytis:
Spoiler alert: they were affected.

vytis:
@mikeger updated snapshots

PR #2526, opened 2018-08-13
Add tests for service participants system messages

daehn:
## What's new in this PR?

* Add tests for service participants system messages.
* Use different copy for added services and the self user being added to a conversation containing services.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ParticipantsCell/ParticipantsCellViewModel.swift
```diff
@@ -88,14 +114,12 @@ class ParticipantsCellViewModel {
...
         return conversation.canManageAccess && conversation.allowGuests
     }
-    
-    var showServiceUserWarning: Bool {
-        guard case .added = action, let systemMessage = message as? ZMSystemMessage, let conversation = message.conversation else { return false }
-        let selfAddedToServiceConversation = systemMessage.users.any(\.isSelfUser) && conversation.areServicesPresent
-        let serviceAdded = systemMessage.users.any(\.isServiceUser)
-        return selfAddedToServiceConversation || serviceAdded
+
+    var serviceUserWarningType: ServiceUserWarningType {
```
vytis: Maybe it would make more sense to have this as optional instead of `serviceUserWarningType.displayString` being optional?

vytis: Ah OK, didn't notice that `ServiceUserWarningType` is actually an enum, now it makes sense!

daehn:
The new design spec changes some things related to this, will adjust this PR accordingly.

PR #2532, opened 2018-08-14
Fix: Audio message reproduction is interrupted by incoming call

billypchan:
### Issues

When the user is playing an audio message and a call rings, the audio message is stopped and reset the start position after the user rejects a call.

### Causes

The AVPlayer is stopped and wiped when the call rings.

### Solutions

Pause the audio player when the call rings, then play it when the call is terminated and it was paused because of the incoming call.

### Todo

- [x] tests for the different cases which would pause the audio player.

### DIFF
--- a/Wire-iOS Tests/AudioMessageViewTests.swift
+++ b/Wire-iOS Tests/AudioMessageViewTests.swift
@@ -19,13 +19,42 @@
 import XCTest
 @testable import Wire
 
-final class AudioMessageViewTestsTests: XCTestCase {
+final class MockAudioTrack: NSObject, AudioTrack {
+    var title: String!
+
+    var author: String!
+
+    var artwork: UIImage!
+
+    var duration: TimeInterval = 0.0
+
+    var artworkURL: URL!
+
+    var streamURL: URL!
+
+    var previewStreamURL: URL!
+
+    var externalURL: URL!
+
+    var failedToLoad: Bool
+
+    func fetchArtwork() {
+        // no-op
+    }
+
+    override init() {
+        failedToLoad = false
+        super.init()
+    }
+}
+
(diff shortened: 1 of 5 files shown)


Wire-iOS Tests/AudioMessageViewTests.swift
```diff
@@ -0,0 +1,156 @@
...
+        sut.configure(for: audioMessage, isInitial: true)
+    }
+    
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+
+    /// Example checker method which can be reused in different tests
+    fileprivate func checkerExample(file: StaticString = #file, line: UInt = #line) {
```
daehn: Can be removed as it doesn't do / check anything?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
```diff
@@ -462,3 +478,32 @@ private let zmLog = ZMSLog(tag: "UI")
...
+
+        // Pause the audio player when call is incoming to prevent the audio player is reset.
+        // Resume playing when the call is terminating (and the audio is paused by this method)
+        switch (previousCallState, callState) {
+        case (_, .incoming(_, _, _)):
+            if player.isPlaying {
+                player.pause()
+                isPausedForIncomingCall = true
+            }
+        case (.incoming(_, _, _)?, .terminating(reason: _)):
```
johnxnguyen: I think you can you can write `.incoming` instead of `.incoming(_, _, _)`, and likewise `.terminating` instead of `.terminating(reason: _)`. If so, I'd do that to improve readability.

PR #2533, opened 2018-08-14
De-register ui menu observers after succesfull callback

daehn:
## What's new in this PR?

* De-register ui menu observers after successful callback.


daehn:
@mikeger also surprised me, but we registered the same cells again and again without ever de-registering, which for me led to multiple cells being selected.

PR #2536, opened 2018-08-14
Fix manage services button color scheme

mikeger:
### Issues

The button to manage services appeared white on white in the conversation details.

### Causes

The button was not updated for the current color scheme via UIAppearance.

### Solutions

Make sure the button (the cell in particular) is functioning correctly.

Additionally updated the logic for initial state of the no results view.


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.m
```diff
@@ -124,6 +124,8 @@ - (instancetype)init
         [[SectionHeader appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setColorSchemeVariant:ColorSchemeVariantDark];
         [[GroupConversationCell appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setColorSchemeVariant:ColorSchemeVariantDark];
         [[GroupConversationCell appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setContentBackgroundColor:UIColor.clearColor];
+        [[OpenServicesAdminCell appearanceWhenContainedInInstancesOfClasses:@[StartUIView.class]] setColorSchemeVariant:ColorSchemeVariantDark];
```
daehn: This `init` is getting quite large, I think we should extract this at some point.

PR #2545, opened 2018-08-15
Fix: Crash on iOS10 for undocumented UIViewAnimationCurve type

billypchan:
### Issues

App crashes when the keyboard appears on iOS 10 devices.

### Causes

iOS returns an undocumented type 7 animation curve raw value, which causes crashes on iOS 10 if it is used as an argument in UIViewPropertyAnimator init method. 

### Solutions

Workaround: assign a fallback value for iOS10.


Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Animation.swift
```diff
@@ -44,7 +44,21 @@ extension KeyboardAvoidingViewController {
 
             bottomEdgeConstraint.constant = bottomOffset
             self.view.setNeedsLayout()
-            animator = UIViewPropertyAnimator(duration: duration, curve: animationCurve, animations: {
+
+            var fixedAnimationCurve: UIViewAnimationCurve
+            fixedAnimationCurve = animationCurve
+            if #available(iOS 11.0, *) {
```
alexaubry: Couldn't we move the check for iOS 11 higher up (line 33), and always fallback to `UIView.animate` on iOS 10 and below instead? I'm not sure it's the best visual experience to have a different animation curve.

Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Animation.swift
```diff
@@ -18,6 +18,28 @@
 
 import Foundation
 
+extension Int {
+    var toAnimationCurve: UIViewAnimationCurve {
```
mikeger: I would implement `UIViewAnimationCurve ` extension with init that take int as an argument; currently adding the method to `Int` namespace makes it apply for any Int, so the autocomplete would always show this method, I think it can get annoying.

PR #2557, opened 2018-08-17
[WIP] Adjust call overlay to be interactively dismissible

daehn:
## What's in this PR?

* Adjust call overlay to be interactively dismissible.
* There is still work to be done in `ActiveVoiceChannelViewController` to adjust the presentation logic of the active `CallViewController` to work with incoming calls during an active call, as well as dismissing it properly when a chathead is being tapped.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
@@ -36,6 +36,7 @@ final class CallViewController: UIViewController {
     private var videoConfiguration: VideoConfiguration
     private let videoGridViewController: VideoGridViewController
     private var cameraType: CaptureDevice = .front
+    private var isInteractiveTransition = true
 
     var conversation: ZMConversation? {
         return voiceChannel.conversation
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -36,6 +36,7 @@ final class CallViewController: UIViewController {
     private var videoConfiguration: VideoConfiguration
     private let videoGridViewController: VideoGridViewController
     private var cameraType: CaptureDevice = .front
+    private var isInteractiveTransition = true
```
daehn: Right now this is needed because we use the `dismisser` to dismiss the active view controller, but I think it would make sense to change this, let the CallViewController dismiss itself and notify a delegate that it has dismissed instead.

Wire-iOS/Sources/UserInterface/Overlay/ActiveVoiceChannelViewController.swift
```diff
@@ -141,47 +141,56 @@ class ActiveVoiceChannelViewController : UIViewController {
...
         
         if let toViewController = to, let fromViewController = from {
+            let modalVC = ModalPresentationViewController(viewController: toViewController)
+            modalVC.view.alpha = 0
+            present(modalVC, animated: false)
             
-            toViewController.view.frame = view.bounds
-            toViewController.view.autoresizingMask = [.flexibleHeight, .flexibleWidth]
-            addChildViewController(toViewController)
+            _ = fromViewController
```
daehn: This part of the code still needs to be properly adjusted. Before we were using view controller containment, now we use proper view controller presentation which is easier to handle when we only have one `CallViewController`. However the case when we have an incoming call during an active call is not properly handled or tested yet.

typfel:
I solved the transition between calls by only letting `ActiveVoiceChannelViewController` handle transitions between two calls. Initial presentation. and dismissal is controlled by the new `CallController`. The call quality survey is now also controlled by a separate controller.

PR #2562, opened 2018-08-20
Cross fade avatars and images

typfel:
## What's new in this PR?

Avatars and images are cross faded when they are loaded from network/disk.

This affects:

- All avatars
- Image messages
- Link previews

Still missing:

Video thumbnails, since the loading indicator has a slightly different logic there.


vytis:
Does this avoid flickering of user images or it's for some other reason?

typfel:
Now it will cross fade instead of flicker so in a way yes, but I also wanted to add this since both web and android does it.

PR #2565, opened 2018-08-21
Feature/iOS 10 AVFoundation

johnxnguyen:
## What's new in this PR?

This PR implements updates to `AVFoundation` for iOS 10. The `CameraController` class has been rewritten in Swift and heavily simplified; it is only being used in the `CameraCell` of the keyboard and many of it's implemented features were not being used.


Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
```diff
@@ -0,0 +1,300 @@
...
+            setupResult = .failed
+            return
+        }
+        
+        // try to connect the preferred camera
+        switch currentCamera {
+        case .front: if !canAddFrontInput { currentCamera = .back }
+        case .back:  if !canAddBackInput  { currentCamera = .front }
+        }
+
```
vytis: The code in this function seems a bit repetitive, it should be possible to simplify it with something like this (not tested): 
```
let availableCameras = 
[.front, .back]
.compactMap { cameraDevice(for: $0 ) }
.compactMap { try? AVCaptureDeviceInput(device: $0) }
.filter { session.canAddInput($0) }
```

Then check if `currentCamera` is in `availableCameras` and if not choose another available one.

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
```diff
@@ -0,0 +1,300 @@
...
+     * The capture device for the given camera position, if available.
+     */
+    private func cameraDevice(for position: AVCaptureDevice.Position) -> AVCaptureDevice? {
+        return AVCaptureDevice.default(.builtInWideAngleCamera, for: .video, position: position)
+    }
+    
+    /**
+     * The device input for the given camera, if available.
+     */
+    private func input(for camera: SettingsCamera) -> AVCaptureDeviceInput? {
```
vytis: This can be a `var` on `SettingsCamera`

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    /**
+     * Connects the input for the given camera, if it is available.
+     */
+    private func connectInput(for camera: SettingsCamera) {
+        guard
+            let input = input(for: camera),
+            session.canAddInput(input)
```
vytis: This seems a bit redundant since we already checked this before

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
```diff
@@ -0,0 +1,300 @@
...
+            self.session.commitConfiguration()
+        }
+    }
+    
+    /**
+     * Disconnects the current camera and connects the given camera, but only
+     * if both camera inputs are available. The completion callback is passed
+     * a boolean value indicating whether the change was successful.
+     */
+    func changeCamera(to newCamera: SettingsCamera, completion: @escaping (_ success: Bool) -> Void) {
```
vytis: I guess we could just have a method `switchCamera` that would not have the parameter and the transition would be done depending on what is the current camera.

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
```diff
@@ -0,0 +1,300 @@
...
+            
+            handler(PhotoResult(imageData, nil))
+        }
+    }
+}
+
+
+private extension AVCaptureVideoOrientation {
+    
+    /// The video orientation mmatches against first the device orientation,
```
typfel: typo: `mmatches`

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
johnxnguyen: True, I'll give it a try!

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
johnxnguyen: I found that if this method returns true at some point in time, it doesn't mean that it will always return true later on. For example, you can't add an input to the session if it's already connected, and doing so will raise an exception.

Wire-iOS/Sources/UserInterface/Camera/CameraController.swift
johnxnguyen: Wait... how would you do this?

vytis: Ah yeah, this doesn't make sense actually. I somehow thought it's converting from one enum to another...

PR #2567, opened 2018-08-21
Remove Classy styling for ConversationCellBurstTimestampView

billypchan:
## What's new in this PR?

Remove Classy styling for ConversationCellBurstTimestampView

## Notice
For an unknown reason, Classy's character spacing in Label is narrower(caused by the label's width is 1px narrower) than normal styling method.


vytis:
@mikeger do you have an idea why the font spacing could be different here?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCellBurstTimestampView.swift
```diff
@@ -17,17 +17,19 @@
...
 @objcMembers final public class ConversationCellBurstTimestampView: UIView {
 
     public let unreadDot = UIView()
-    public let label = UILabel()
-
-    public var separatorColor: UIColor?
-    public var separatorColorExpanded: UIColor?
+    public let label: UILabel = {
+        let label = UILabel()
+        label.textColor = UIColor(scheme: .textForeground)
```
typfel: Does this update if you change color scheme?

billypchan: I will add a snapshot test for both color variants, thanks!

billypchan:
@vytis It seems that Classy styled label is 1pt narrower than the normal label, I had encountered this issue before. like this: 
https://github.com/wireapp/wire-ios/pull/2319/files?short_path=bf05f3c#diff-bf05f3c6881c1418a3e789e872439bc2

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCellBurstTimestampView.swift
typfel: I'm pretty sure it will work what I was wondering was if it the style is updated when the cell is re-used.

billypchan: I added a prepareForReuse method in ConversationCellBurstTimestampView to refresh the color variant when its parent ConversationCell's prepareForReuse is called.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCellBurstTimestampView.swift
```diff
@@ -126,4 +128,9 @@ import Cartography
         }
     }
 
+    func prepareForReuse() {
+        label.textColor = UIColor(scheme: .textForeground)
+        separatorColor = UIColor(scheme: .separator)
```
typfel: These colors are now set in two places would be less error prone to have it all in one place.

billypchan: Removed the init color settings in the variables' declaration.

typfel:
@billypchan sorry didn't check it before but we actually reload the conversation view controller when the color scheme changes.

ZClientViewController.m:534

billypchan:
@typfel prepareForReuse is removed and the tests are updated.

PR #2583, opened 2018-08-24
Fix/translate remain tests

billypchan:
## What's new in this PR?

Translate DeleteMessageTests, ProfilePictureStepViewControllerTests, ConversationTitleViewTests, ConversationListBottomBarTests & ArchivedNavigationBarTests into Swift in order to remove the dependency of "Wire_iOS_Tests-Swift.h"


Wire-iOS Tests/ConversationListBottomBarControllerTests.swift
```diff
@@ -0,0 +1,144 @@
...
+        snapshotBackgroundColor = UIColor(white: 0.2, alpha: 1) // In order to make the separator more visible
+        accentColor = .brightYellow
+        mockDelegate = MockConversationListBottomBarDelegate()
+        UIView.performWithoutAnimation({
+            self.sut = ConversationListBottomBarController(delegate: self.mockDelegate)
+        })
+    }
+    
+    override func tearDown() {
+        sut = nil
```
vytis: `mockDelegate = nil` should also be added here. Basically you need to nil everything that was created in `setUp()`, easiest way to check is to see if all `var`s in the class are nil'ed

Wire-iOS Tests/ConversationTitleViewTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        super.setUp()
+        conversation = MockConversation()
+        conversation.relatedConnectionState = .accepted
+        conversation.displayName = "Alan Turing"
+        sut = ConversationTitleView(conversation: conversation as Any as! ZMConversation, interactive: true)
+        snapshotBackgroundColor = UIColor.white
+    }
+    
+    override func tearDown() {
+        sut = nil
```
vytis: Missing cleanup of conversation

Wire-iOS Tests/ProfilePictureStepViewControllerTests.swift
```diff
@@ -0,0 +1,53 @@
...
+
+            weakSelf.sut.loadViewIfNeeded()
+            weakSelf.sut.viewWillAppear(false)
+            weakSelf.sut.showLoadingView = false
+            weakSelf.sut.profilePictureImageView.image = weakSelf.image(inTestBundleNamed: "unsplash_matterhorn.jpg")
+        }
+    }
+    
+    override func tearDown() {
+        sut = nil
```
vytis: same with `configurationBlock`, or even simply move it to the test function itself since we don't use it anywhere else anyway.

PR #2584, opened 2018-08-27
Fix: remove Classy styling for ViewControllers

billypchan:
## What's new in this PR?

Remove Classy styling of viewControllers.


Wire-iOS/Sources/UserInterface/Helpers/UIColor+Scheme.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension UIColor {
+
+    static var textDimmed: UIColor {
```
typfel: Is this for Ojb-C compatibility? Because I think adding every single enum as a static property defeats the purpose of the enum.

billypchan: I would prefer do not include objc support for these helper methods. For objc this is equivalent:

[UIColor wr_colorFromColorScheme:ColorSchemeColorTextDimmed]

typfel:
Was reviewing the snapshot diffs I found:

* `FullscreenImageViewController`:  missing background color in view controller
* ` ProfileViewControllerTests`: background color is dimmed compared to previous snapshots

Wire-iOS/Sources/UserInterface/Helpers/UIColor+Scheme.swift
typfel: In Swift you could already do `UIColor(scheme: .textDimmed)`

billypchan:
@typfel 
For ProfileViewControllerTests, the snapshots are matching the appearance on the app. The background of ProfileViewController should not be pure white.
Similarly, FullscreenImageViewController have a non-clear background color.

typfel:
@billypchan ok I see 👍

Wire-iOS/Sources/UserInterface/Helpers/UIColor+Scheme.swift
billypchan: Yes, but the helpers reduce the characters to type.

typfel: We still need to supply the color variant in many cases like in the search UI for example which is always dark.

PR #2589, opened 2018-08-27
snapshot test for ConversationListViewController

billypchan:
## What's new in this PR?

Create a snapshot test for ConversationListViewController with no conversations. Added account property to ConversationListViewController for injecting mock Account object for testing.


typfel:
Maybe set the snapshot background color to black to make the snapshots more readable?

billypchan:
@typfel This case is a little bit different, ConversationListViewController is transparent with a blurry image behind. I would add a mock blur view in the snapshot.

typfel:
@billypchan but we aren't testing the blurry picture, black background would be good enough in my opinion. There's even a property for this `snapshotBackgroundColor`

PR #2590, opened 2018-08-27
Fix: Invalid Tab button normal state color

billypchan:
### Issues

In the registration screen, tab button's color is changed after https://github.com/wireapp/wire-ios/pull/2574

### Causes

It uses a color rgb(255, 255, 255, 0.56) which is not defined in the color scheme.

### Solutions

Add this color to the scheme. 

### Note
- [x] snapshot test for the registration screen.


Wire-iOS/Sources/Components/Buttons/Button.m
```diff
@@ -43,23 +43,6 @@ @implementation Button
 + (void)initialize
 {
     if (self == [Button self]) {
-        CASObjectClassDescriptor *classDescriptor = [CASStyler.defaultStyler objectClassDescriptorForClass:self.class];
```
typfel: Is there no `Button` in stylesheet.cas anymore? Until that's the case this can't be removed.

billypchan: Yes, all Button related Classy style are gone.

PR #2591, opened 2018-08-27
Clean up objects created in tests to reduce peak memory usage

vytis:
### Issues

We are not very careful in `nil`-ing all the objects created in test `setUp()` which means the objects stay in memory until the testsuite has finished running.

### Solutions

The changes are quite straightforward, except for changes in [ConversationInputBarViewControllerTests.swift](https://github.com/wireapp/wire-ios/compare/develop...fix/tests-teardown#diff-5f85e1f8d69d9e7f9cdc4093a794aebf) which cause unrelated test to fail, but only when running the full test-suite. The reason is still unclear, but will investigate further and hopefully remove this limitation.


mikeger:
@vytis I see some tests are still failing on CI, could be that this happening because we are not using any groups on UI for async operations, before it was never deallocated, so not an issue I guess?

vytis:
I am investigating - there were some crashing tests locally, so maybe got messed up on CI somehow.

PR #2597, opened 2018-08-28
Store mixpanel ID to the file for simulator and automation

mikeger:
### Issues

The automation cannot catch the Mixpanel client ID from the logs.

### Causes & Solutions

Not clear, but as the system showed itself fragile, it was decided to write the ID to the file instead.


Wire-iOS/Sources/Analytics/AnalyticsMixpanelProvider.swift
```diff
@@ -146,6 +146,12 @@ final class AnalyticsMixpanelProvider: NSObject, AnalyticsProvider {
         if DeveloperMenuState.developerMenuEnabled(),
             let uuidString = mixpanelInstance?.distinctId {
             zmLog.error("Mixpanel distinctId = `\(uuidString)`")
+            
+            #if targetEnvironment(simulator)
+            let tempFilePath = URL(fileURLWithPath: "/var/tmp/mixpanel_id.txt")
+            try? FileManager.default.removeItem(at: tempFilePath)
+            try! uuidString.data(using: .utf8)?.write(to: tempFilePath)
```
alexaubry: Maybe we can use `Data(uuidString.utf8)` here, as this method cannot fail and does not return an optional?

PR #2603, opened 2018-08-29
Feature: Dedicated mute icon when video call overlay is hidden

billypchan:
## What's new in this PR?

In video calls, we now show a little indicator that you are muted at the bottom of the screen.
We do not show a mute icon on top of your own video anymore. This indication fades in when the controls fade out and fade back out when the controls fade in again.


Wire-iOS Tests/CallViewControllerTests.swift
```diff
@@ -38,23 +38,86 @@ extension XCTestCase {
     }
 }
 
+extension XCTestCase {
+    static func CallViewControllerWithMocks(mediaManager: ZMMockAVSMediaManager) -> CallViewController{
```
mikeger: I think the usual style in Swift is to name the methods and variables with small letter first, class names with the capital letter first.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/MuteIndicatorViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+
+extension CGFloat {
+    enum MuteIndicator {
+        static let containerHeight: CGFloat = 32
+        static let containerHorizontalMargin: CGFloat = 12
+        static let iconLabelSpacing: CGFloat = 8
+    }
+}
+
+final class MuteIndicatorViewController: UIViewController {
```
mikeger: Not sure if the thing has to be the controller, I guess it can also be just the view subclass, since it's not holding up to any logic?

billypchan: @mikeger I will rewrite it as a UIView's subclass.

mikeger: @billypchan I believe there is no difference in layout of view and view controller, since the view controller does not play the role in the layout.

PR #2604, opened 2018-08-29
ZIOS-10470: Call state in conversation list, top overlay: overhaul of labels and statuses	

nicorsm:
## What's new in this PR?

- The new message for missed calls in group conversations is "Missed call from (name)"
- Ongoing calls have a new white "phone" icon
- Changed icon for missed calls to ".endCall"
- Removed the "Ongoing call" label


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -208,6 +207,7 @@
 "conversation.status.message.file" = "Shared a file";
 "conversation.status.message.knock" = "Pinged";
 "conversation.status.message.missedcall" = "Missed call";
+"conversation.status.message.missedcall.groups" = "Missed call from";
```
alexaubry: Do we need a format string here to indicate who missed the call?

nicorsm: Good idea, since we don't use anymore the combination of bold and standard fonts for this kind of messages.

PR #2622, opened 2018-09-03
Bump SE to use new rejected elsewhere call end reason

alexaubry:
### Issues

The web client sends WCALL_REASON_REJECTED call termination reason messages when the user rejects the call on one client so that other clients stop ringing.

As discussed with @typfel, while we are not going to handle this case yet, we don't want it to appear as an unknown error.

### Dependencies

- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/847


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -3538,6 +3567,16 @@
 			path = Animations;
 			sourceTree = "<group>";
 		};
+		5E566B92213D646600A5AA9C /* Products */ = {
+			isa = PBXGroup;
+			children = (
+				5E566B98213D646600A5AA9C /* WireSyncEngine.framework */,
```
vytis: Dragged in SyncEngine...

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -1290,6 +1290,34 @@
...
+			remoteGlobalIDString = 3E186088191A56F6000FE027;
+			remoteInfo = "WireSyncEngine Test Host";
+		};
+		5E566B9D213D64BB00A5AA9C /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 5E566B91213D646600A5AA9C /* WireSyncEngine.xcodeproj */;
+			proxyType = 1;
+			remoteGlobalIDString = 549815921A43232400A7CE2E;
+			remoteInfo = "WireSyncEngine-ios";
+		};
```
billypchan: These project references should be excluded from the PR.

Wire-iOS.xcodeproj/project.pbxproj
alexaubry: Will remove when the SE PR is merged 👍

Wire-iOS.xcodeproj/project.pbxproj
alexaubry: Will remove when the SE PR is merged 👍

PR #2626, opened 2018-09-04
Update email address when sending calling debug report

typfel:
## What's new in this PR?

Send debug calling debug report to calling-ios@wire.com. Remove localization since they all go the same address.


Wire-iOS/Sources/UserInterface/Settings/SettingsTechnicalReportViewController.swift
```diff
@@ -69,7 +69,7 @@ class SettingsTechnicalReportViewController: UITableViewController, MFMailCompos
         let mailRecipient = NSLocalizedString("self.settings.technical_report.mail.recipient", comment: "")
```
alexaubry: Should `mailRecipient` be removed too, since we removed the localized string?

PR #2631, opened 2018-09-05
Fix: Translate MenuConfigurationProperties to Swift

billypchan:
## What's new in this PR?

Translate MenuConfigurationProperties to Swift to define its properties' nullability. Create a test to check the menu items of PingCell is created correctly.


Wire-iOS Tests/ConversationCell/PingCellTests.swift
```diff
@@ -46,5 +46,16 @@ final class PingCellTests: ZMSnapshotTestCase {
     func testForYouPinged() {
         verify(view: sut.tableViewForSnapshot())
     }
+
+    // MARK: shortcut menu
+
+    func testThatDeleteItemIsInMenu(){
```
typfel: I think we can find a better place for this test since this is a snapshot test case. As I understood it the logic for showing the menu is not implemented in the PingCell but in the `ConversationCell+Menu` and is shared between all the cells.

Wire-iOS Tests/ConversationCell/PingCellSnapshotTests.swift
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+final class PingCellSnapshotTests: ZMSnapshotTestCase {
```
typfel: Wouldn't you need to re-generate all the snapshots if you rename this class?

Wire-iOS Tests/ConversationCell/PingCellTests.swift
```diff
@@ -19,22 +19,15 @@
 import XCTest
 @testable import Wire
 
-final class PingCellTests: ZMSnapshotTestCase {
+final class PingCellTests: XCTestCase {
```
typfel: I though we could have a more generic test case for all cells like `ConversationCellMenuTests`

Wire-iOS Tests/ConversationCell/PingCellSnapshotTests.swift
billypchan: The snapshots are moved to the folder with a new name.

https://github.com/wireapp/wire-ios/pull/2631/files#diff-b680e7b994abf59d6e95c2919213000f

Wire-iOS Tests/ConversationCell/PingCellTests.swift
billypchan: @typfel I will add other cells' tests later.

typfel: That's fine.

PR #2632, opened 2018-09-05
Fix: mute indication visibility 

billypchan:
## What's new in this PR?

The isMute property of VideoGridConfiguration is not updated immediately and cause the mute indication's incorrection animation.  Get mute status from AVSMediaManager.sharedInstance() instead.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
@@ -81,6 +81,7 @@ class VideoGridViewController: UIViewController {
                 return
             }
 
+            muteIndicatorView.isHidden = false
             muteIndicatorView.alpha = isCovered ? 1 : 0
 
             UIView.animate(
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -75,23 +76,23 @@ class VideoGridViewController: UIViewController {
...
-                    completion: nil
-                )
-            } else {
+            guard mediaManager.isMicrophoneMuted else {
                 muteIndicatorView.isHidden = true
+                return
             }
+
+            muteIndicatorView.isHidden = false
+            muteIndicatorView.alpha = isCovered ? 1 : 0
```
typfel: Isn't this value being set again in the animation block?

billypchan: This is the init value, muteIndicatorView.alpha is flipped in the animation block.

typfel: Right ok.

typfel:
I think this can also be solved by moving the `isCovered` state into the configuration object but let's go with this for now.

PR #2633, opened 2018-09-05
Use Personal/Pro on account/team creation screen

infotexture:
See related ticket ZIOS-10510 for color change.


infotexture:
Should this PR target the release branch instead of `develop`?

billypchan:
@infotexture We need to update the localization string resource before adding the change to release branch.

PR #2634, opened 2018-09-05
Fix audio controls showing up during a call

typfel:
### Issues

Audio controls (play/play/skip) were showing up on the lock screen and in control center when on a call.

### Causes

`AudioTrackPlayer` was registering these actions with `MPRemoteCommandCenter`.

### Solutions

`AudioTrackPlayer` is registering these commands only when it's playing something and deregistering them when it stops.


billypchan:
Should we also hide those buttons when not playing a audio?

typfel:
@billypchan what do you mean by not playing? If the audio is just paused I think we want them there.

billypchan:
@typfel I checked that the system media player is gone and no more issue about the buttons.

PR #2637, opened 2018-09-06
ZIOS-10509: Add the list of open-source licenses inside the app

alexaubry:
### Issues

Yesterday, we noticed that the list of OSS licenses hosted on the Wire website, which we use as our acknowledgements page in the client settings, returned a 404. This revealed that using this technique is a fragile solution for legal license compliance:

- we don't embed the license text inside the package, as required by some licenses we depend on
- if the Wire website goes down, or if the GitHub repository for the licensed project goes down, the the license file is gone, even though we still need to abide by it
- when we update the dependencies on the iOS app, we need to update them on the website as well
- when we update the website, we need to remove unused dependencies, so the list can get out of sync with older versions of the app that use removed dependencies

As most iOS apps do (WhatsApp, Microsoft To-Do...), this PR proposes to embed the licenses in a readable and offline format inside the application settings.

### Solutions

To make this approach work, we needed several components

#### Displaying the list of open source software

- We load the list of libraries from a Plist resource file using `Codable`
- We update the settings description factory to return a list of libraries instead of opening the browser when the user taps `About > License Information`
- We add a detail screen where we show the license text and a button to open the GitHub repo for each dependency
- We enable self-sizing table view cells for the settings table view by using `UITableViewAutomaticDimension` and a flexible height constraint
- We add a type of cell for multiline text

licensedetails

#### Generating the Plist of Codable Items

One of the benefits of this PR is that the list of open source software will always be up to date with the dependencies that were built to use the app. To achieve this, we use this strategy:

- We create a `EmbeddedDependencies.plist` that contains the dependencies that are dragged into the app, and that don't come from Carthage. This Plist conforms to the same Codable scheme as the one used in the app 
- We create a build script (a Swift package) that reads the `Cartfile.resolved` file to find the GitHub dependencies, and looks in the `Carthage/Checkouts` folder for the LICENSE file of each repository
- Every time it runs, this script generates the Plist that will be used by the app in the Resources folder. It is encoded in the binary format, for more efficiency when decoding

### Note

The update licenses script does not require any maintenance. It will be executed at every run on CI and Jenkins, and incrementally on development machines (when the Cartfile.resolved file changes locally).

However, the `EmbeddedDependencies.plist` must be updated every time the dragged-in dependencies are changed. This will have a minimal impact though, as this technique is used minimally in the app (only 3 dependencies).


typfel:
Very nice!

PR #2646, opened 2018-09-10
Fix: cursor position on RegistrationTextField

billypchan:
### Issues

After the Classy is removed, the spacing between the cursor and the left edge of RegistrationTextField is reduced.

### Causes

Unknown, seems like Classy has overridden the editingRect of the textfield and adding inset for it.

### Solutions
implement override open func editingRect(forBounds bounds: CGRect) -> CGRect, set the corrent left edge inset

### Note
I tried to replace RegistrationTextField with UItextfield, but the editingRect is still overridden.


typfel:
This was the change which was made after Classy was introduced:
https://github.com/zinfra/rip/commit/d46544c74d4c60d006a156c7dc938145e201ecd5#diff-5052a96c12a4c705d1986c9c5a629523

This is basically what you are doing this PR except you use the `placeholderInsets` instead of the `textInsets`

PR #2660, opened 2018-09-13
Displaying Mentions

mikeger:
## What's new in this PR?

Support of showing the incoming mentions. Some initial snapshot tests added.


Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
```diff
@@ -0,0 +1,111 @@
...
+        
+        let message = otherUserConversation.appendMessage(withText: messageText, mentions: [mention1, mention2, mention3],
+                                                          fetchLinkPreview: false)
+        
+        sut.configure(for: message, layoutProperties: layoutProperties)
+        verify(view: sut.prepareForSnapshot())
+    }
+    
+    func testThatItRendersMentions_SelfMention() {
+        recordMode = true
```
vytis: Record mode everywhere!

mikeger:
@vytis added more markdown tests.

mikeger:
@typfel I think it would be great from DM side to show the API user not `Mention` but `MentionWithUser` in this PR terminology, then the API is more consistent. I can move it afterwards. Then we also don't need the `ZMMessage *backingMessage = (ZMMessage *)message;` which I don't enjoy.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
```diff
@@ -37,4 +37,26 @@ extension NSMutableAttributedString {
...
         return result
     }
+    
+    /**
+     * Parses the markdown and adding the @-mentions highlights.
+     * The logic is the following:
+     * 1. Replace @-mention ranges with UUID strings.
+     * 2. Process markdown styling.
+     * 3. Replace UUID with mentions.
+     * This is necessary, since markdown parsing might change the string length.
```
vytis: Actually the main reason is to avoid injecting markdown syntax from the name. For example if my name is `_vytis_` it should not render the mention as italic, but as "@_vytis_". This is what I was referring to in my previous comment about missing tests.

mikeger: The issue you mentioned is avoided by the fix, but unfortunately I found another issue when link is directly adjasent to the mention. I created the test (that is failing now). Will fix it tomorrow.

mikeger: This is how it's looking when the user has Markdown in the name: https://github.com/wireapp/wire-ios/pull/2660/files?short_path=8b3b20c#diff-8b3b20cad430e055785dd89c573ed595

cardassian-tailor: @mikeger I'm trying to look through the code to see how mentions will be handled by the notification system. When I mention you - how will you be notified ... that aspect. Does this portion of the code exist yet and can you point me to the area I should be looking in?

vytis: @cardassian-tailor we will start working on it soon, there is no code yet. However when we do write the code, it is just the initial version. What will be shipped to clients might evolve when we have a working prototype and iterate on it. So stay tuned!

mikeger:
@vytis @daehn please review

Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
```diff
@@ -0,0 +1,141 @@
+//
+//  TextMessageMentionsTests.swift
+//  Wire-iOS-Tests
+//
+//  Created by Mihail Gerasimenko on 9/13/18.
```
vytis: 🙉

mikeger: Some self-promotion...

PR #2672, opened 2018-09-18
Mentions: display in the conversation list

mikeger:
### Issues

The text message must be treated in the special way in case it contains the mention for the self user.


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -77,7 +79,11 @@ extension StatusMessageType {
     ]
     
     init?(message: ZMConversationMessage) {
-        if message.isText, let textMessage = message.textMessageData {
+        
+        if message.textMessageData?.isMentioningSelf ?? false {
```
typfel: Couldn't this be inside the if `message.isText`? The you also wouldn't need to handle the optional `message.textMessageData`

mikeger: Indeed, more elegant in that way.

PR #2675, opened 2018-09-18
Add initial mention sending code to input field

vytis:
## What's new in this PR?

After typing '@' symbol it will "enter" the mentioning mode and exit it after space is tapped. Then it will use that for searching users and replace the search string with text attachment showing the mention.

## Note

The ranges seem to be messed up and when adding more than one mention weird things start to happen. I will add tests for this part to make sure it works as expected.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+UITextViewDelegate.swift
```diff
@@ -49,15 +49,25 @@ extension ConversationInputBarViewController: UITextViewDelegate {
...
-            let attributedString = NSAttributedString(attachment: attachment)
-            textView.attributedText = textView.attributedText + attributedString
-            return false
+            self.mentionsHandler = MentionsHandler(atSymbolRange: range)
+        } else if let handler = mentionsHandler, let previousText = textView.text {
+            // In mentioning flow
+            let currentText = previousText.replacingCharacters(in: Range(range, in: previousText)!, with: text)
+            if handler.shouldReplaceMention(in: currentText) {
+                let searchString = handler.searchString(in: currentText)
+                let fetchRequest = ZMUser.sortedFetchRequest(with: ZMUser.predicateForConnectedUsers(withSearch: "@" + searchString))
```
typfel: I guess we should consider moving the search logic to SE later.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+UITextViewDelegate.swift
```diff
@@ -49,15 +49,25 @@ extension ConversationInputBarViewController: UITextViewDelegate {
...
-        // Start mentioning
+
+        // Enter mentioning flow
         if text == "@" {
-            let user: UserType = conversation.connectedUser ?? conversation.sortedOtherParticipants.randomElement() ?? ZMUser.selfUser()
-            let attachment = MentionTextAttachment(user: user)
-            let attributedString = NSAttributedString(attachment: attachment)
-            textView.attributedText = textView.attributedText + attributedString
-            return false
+            self.mentionsHandler = MentionsHandler(atSymbolRange: range)
```
typfel: I suspect that the `range` coming from `shouldChangeTextIn` is not in utf16 offsets.

typfel: But again we can fix these thing later.

vytis: All NSString APIs use UTF16, so I would imagine `range` is also indexing to a NSString. Have not checked though...

mikeger: @typfel we are implying that NSString APIs are all in UTF16 ranges / offsets, if not then we might reconsider quite some things.

typfel: Oh ok didn't know :)

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+UITextViewDelegate.swift
vytis: Yeah this is just a placeholder for now, the search will be handled by the popover.

PR #2682, opened 2018-09-19
Fix: settings.opted out tracking

billypchan:
### Issues

Event "settings.opted_out_tracking" is not sent

### Causes

If Mixpanel instance is freed immediately after that event is sent, it would not be sent.

### Solutions

Call Mixpanel's flush method and free the Mixpanel instance in the completion block.


marcoconti83:
Just to clarify, is "settings.opted_out_tracking" a custom event that we send to keep track of how many users opted out from the settings screen after opting in?

marcoconti83:
Clarified in person with @billypchan . The issue was that we never have a chance to send this last custom event before shutting down. It has been fixed by waiting for this last event to be sent before shutting down. The previous behavior was to shut down immediately.

PR #2684, opened 2018-09-20
Convert mentions to plain text when copied

typfel:
### Issues

When a mention was copied from the input bar it copy the invisible text attachment

### Solutions

Convert  any selected mentions to plain text when copied.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
@@ -17,6 +17,7 @@
 //
 
 import Foundation
+import MobileCoreServices
 import Down
 
 extension Notification.Name {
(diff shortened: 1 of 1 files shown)


cardassian-tailor:
Lots of changes have occurred since 3.20. Any plans to push this all into the next release sooner than later  (3.21) ?

vytis:
@cardassian-tailor the plan is not clear yet, but it's coming along nicely. Cannot really comment much on the timeline unfortunately...

cardassian-tailor:
> 
> 
> @cardassian-tailor the plan is not clear yet, but it's coming along nicely. Cannot really comment much on the timeline unfortunately...

Understandable. Look forward to the release. Wishing there was a way to test the beta other than joining testflight. 

Has anyone tried to see what occurs when one uses a conversation/channel search feature and searches for a mention? Has there been any discussion as to how that "should" work (returns all of that users sent messages or just the mentions of that user) ?

PR #2691, opened 2018-09-21
Fix: Put video button at the last of the input bar

billypchan:
## What's new in this PR?

Undo the video button removal.


vytis:
You will have to update snapshot tests for this one I think

billypchan:
@vytis Ran the full test on my local machine already and no change is needed.

vytis:
Ah great, we probably have only first line of toolbar in the screenshots then

PR #2698, opened 2018-09-24
Connect @ button to start mentioning flow

vytis:
### Issues

We have a button that should open the mentioning flow. Because we use regexps for parsing the string and find mentions we need to add additional spaces if needed. Also the cursor must be in the correct position after adding the @ sybmol.

To illustrate (`|` is the cursor position):

* `AB|CD` -> `AB @| CD`
* `AB| CD` -> `AB @| CD`
* `AB |CD` -> `AB @| CD`


vytis:
@typfel updated!

cardassian-tailor:
the suspense is killing me. Go ahead and push dev branch to production : )

PR #2699, opened 2018-09-24
UI fix for suggestions popup

nicorsm:
## What's new in this PR?

UI fixes reported by @eike96 in the main ticket (ZIOS-10527):

- removed separator
- matches were in the wrong order (top matches should start from the bottom)
- fixed UI tests


Wire-iOS/Sources/UserInterface/Conversation/Mentions/MentionsSearchResultsViewController.swift
```diff
@@ -138,7 +138,7 @@ extension MentionsSearchResultsViewController: MentionsSearchResultsViewProtocol
         
         rules.forEach { rule in
             let matches = usersToSearch.filter({ rule($0) }).filter { !foundUsers.contains($0) }
-                .sorted(by: { $0.name < $1.name })
+                .sorted(by: { $0.name > $1.name })
```
billypchan: @nicorsm I solved the same issue in https://github.com/wireapp/wire-ios/pull/2694. It also scrolls the list to the bottom after a new search query is typed.

nicorsm: Ok, so I'll rollback this logic! Thanks for reviewing.

PR #2700, opened 2018-09-24
Fix: self mention is not highlighted properly if it contains multibyte unicode chars

billypchan:
### Issues

self mention is not highlighted properly if it contains multibyte unicode chars

### Causes

NSMutableString's substring method does not give correct output for the string contains multibyte unicode chars.

### Solutions

Rewrite using Swift's String's subscript method.

## Notes

- [ ] Move String.substring to Wire Utility


vytis:
@typfel rewrote parts of the code you changed in #2696. Maybe the original issue was fixed?

billypchan:
Tested with the build after https://github.com/wireapp/wire-ios/pull/2696, the username with emoji has no link attribute. I will try another approach without using the String class's methods.

PR #2703, opened 2018-09-25
ZIOS-10592: The sequence of subtitle items is wrong

nicorsm:
## What's new in this PR?

The order of importance for the subtitle items in the conversation list is wrong.
Current one is mentions, messages, pings, missed calls. It should be mentions, missed calls, pings, messages.


mikeger:
@nicorsm I think we can add the test for this, to make sure it's correct in the future.

PR #2711, opened 2018-09-25
Fix crash when adding mention

vytis:
### Issues

Sometimes when inserting a mention the app would crash.

### Causes

It was not reproduced, but crash log shows out of bounds access when inserting space symbol.

### Solutions

Rewrote this place in a safer way, but appending space to the mention text attachment if needed.


cardassian-tailor:
> 
> ## What's new in this PR?
> ### Issues
> 
> Sometimes when inserting a mention the app would crash.
> ### Causes
> 
> It was not reproduced, but crash log shows out of bounds access when inserting space symbol.
> ### Solutions
> 
> Rewrote this place in a safer way, but appending space to the mention text attachment if needed.

keep up the great work

PR #2714, opened 2018-09-25
ZIOS-10616: Fix crash in mentions suggestion when typing

alexaubry:
### Issues

When you type your own username in the input field with the mentions suggestion controller displayed, the app crashed.

### Causes

When updating the list of mention suggestions, we automatically scroll to the last item, by doing `searchResults.count - 1`. However, when there are no search results, we attempt to scroll to the item at index `-1`; which causes the crash.

### Solutions

We check that the index path of the last item is valid before scrolling to it.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Mentions/MentionsSearchResultsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Mentions/MentionsSearchResultsViewController.swift
@@ -18,7 +18,6 @@
 
 import UIKit
 import Cartography
-import WireExtensionComponents
 
 @objc protocol MentionsSearchResultsViewControllerDelegate {
     func didSelectUserToMention(_ user: ZMUser)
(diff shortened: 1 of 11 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Mentions/MentionsSearchResultsViewController.swift
```diff
@@ -87,7 +88,12 @@ class MentionsSearchResultsViewController: UIViewController {
         
         collectionView.reloadData()
         collectionView.layoutIfNeeded()
-        collectionView.scrollToItem(at: IndexPath(item: searchResults.count - 1, section: 0), at: .bottom, animated: false)
+
+        let firstMatchIndexPath = IndexPath(item: searchResults.count - 1, section: 0)
+
+        if collectionView.containsCell(at: firstMatchIndexPath) {
```
typfel: It's not enough to check if `!searchResults.isEmpty`?

typfel: But I suppose checking if the cell row exists covers more cases.

alexaubry: It could work, I didn't think about it. Should I change?

typfel: This is also fine

PR #2719, opened 2018-09-26
 ZIOS-10575: Suggestions list is not expanded if the keyboard is hidden

nicorsm:
### Issues

Suggestions list is not expanded when the keyboard is hidden.

### Causes

This because the height constraint never gets updated after displaying the list.

### Solutions

I'm listening for `keyboardWillChangeFrameNotification` in order to adapt the height constraint to changes of the frame sizes.


mikeger:
@nicorsm I thought the container would be automatically resized, since we resize the conversation view that is containing suggestions list?

nicorsm:
@mikeger right now it can't get resized because I'm storing the height constraint and I'm updating that, but the keyboard resizing event was not observed - that's why it wasn't updating the size automatically.

mikeger:
@nicorsm I will fix the conflict here

PR #2725, opened 2018-09-27
Tapping on mention in input field should select it

vytis:
### Issues

When we insert a mention in input field it becomes a text attachment. Tapping on it would put cursor either before or after the mention. A better approach UX-wise would be to select the mention when tapping.

### Solutions

There is no useable delegate method for tapping on text attachments so we use the same approach as in 
`TokenizedTextView` and attach a gesture recognizer on whole input bar. The tap position is then used to character index on text view.


typfel:
@vytis what about `textView(_:shouldInteractWith:in:interaction:)`
https://developer.apple.com/documentation/uikit/uitextviewdelegate/1649336-textview

vytis:
@typfel talked face to face. Basically the delegate method does not get called if text view is editable.

PR #2728, opened 2018-09-27
Fix/mentions filter

johnxnguyen:
### Issues

Users containing a `@` symbol in their name (or any other punctuation for that matter) would not be found when mention searching using the query `@`.

### Causes

The search query and all candidate search material was _normalised for search_, which meant that all punctation was removed from these strings. Thus, the query would be empty, and even if it weren't, it would never match against any of the search material.

### Solutions

Don't normalised these strings.

## Needs Release With
- [x] https://github.com/wireapp/wire-ios-utilities/pull/57


typfel:
So I think the reason the strings were normalized was to handle case were you type `Bjo` and that should find `Björn`. Does that still work?

johnxnguyen:
@typfel I see.... yeah that is broken now...

PR #2731, opened 2018-09-27
Respect markdown attributes for mentions in the inputbar

typfel:
### Issues

The input bar does not respect the selected markdown setting when rendering mentions.

### Causes

Mentions in the input bar are text attachment which have a fixed font.

### Solutions

Retrieve the correct from the associated text storage within the text attachment.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
@@ -52,8 +52,8 @@ import Foundation
         return String(text[range])
     }
 
-    func replace(mention: UserType, in attributedString: NSAttributedString) -> NSAttributedString {
-        let mentionString = NSAttributedString(attachment: MentionTextAttachment(user: mention))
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
```diff
@@ -52,8 +52,8 @@ import Foundation
         return String(text[range])
     }
 
-    func replace(mention: UserType, in attributedString: NSAttributedString) -> NSAttributedString {
-        let mentionString = NSAttributedString(attachment: MentionTextAttachment(user: mention))
+    func replace(mention: UserType, in attributedString: NSAttributedString, typingAttributes: [NSAttributedString.Key: Any]) -> NSAttributedString {
+        let mentionString = NSAttributedString(attachment: MentionTextAttachment(user: mention)) && typingAttributes
```
vytis: Don't know if it even makes a difference, but in this case `typingAttributes` will not be assigned to `suffix` because it's appended later.

typfel: but suffix is just whitespace?

typfel: Will change

PR #2739, opened 2018-09-28
Preserving cursor position after inserting mention

vytis:
### Issues

When trying to insert a mention in the middle of the text the cursor would jump to the very end.

### Causes

Setting `attributedText` property for `UITextView` resets the cursor position to be at the end of text. 

### Solutions

There is a method `UITextView.replace(range:with:)` that preserves the cursor position, but it takes `String` as a parameter. This PR adds a similar method for `NSAttributedString` that tries to preserve the cursor position by putting it to the end of the replacement string.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
@@ -52,18 +52,17 @@ import Foundation
         return String(text[range])
     }
 
-    func replace(mention: UserType, in attributedString: NSAttributedString, typingAttributes: [NSAttributedString.Key: Any]) -> NSAttributedString {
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
```diff
@@ -52,18 +52,17 @@ import Foundation
         return String(text[range])
     }
 
-    func replace(mention: UserType, in attributedString: NSAttributedString, typingAttributes: [NSAttributedString.Key: Any]) -> NSAttributedString {
+    func replacement(forMention mention: UserType, in attributedString: NSAttributedString, typingAttributes: [NSAttributedString.Key: Any]) -> (NSRange, NSAttributedString) {
```
typfel: Since the attributed string is returned now we can apply `typingAttributes` from the `ConversationInputBarViewController` and avoid this parameter.

vytis: @typfel updated!

PR #2740, opened 2018-09-28
Truncate mentions that doesn't fit in one line inside the input bar

typfel:
### Issues

Mention with a very long name would get clipped and mess up the cursor position.

### Causes

We didn't limit the size of mention text attachment.

### Solutions

Truncate mention text attachment if it's bigger than the input area.


typfel:
@vytis updated

PR #2742, opened 2018-09-28
Align drafts placeholder with input bar copy

infotexture:


mikeger:
@infotexture so... the drafts are coming? :)

PR #2745, opened 2018-10-01
Updated StyleKit

vytis:
### Issues

There were some changes to Devices icon requested by design


mikeger:
@vytis the screenshot tests might be breaking 🙈

vytis:
I think it should be fine but let's see 🤞

PR #2749, opened 2018-10-02
Feature/in app notifications

johnxnguyen:
## What's new in this PR?

Now that we have dropped support for iOS 9, we can replace `ChatHeads` with native local notifications. This PR removes all ChatHead code and tests. Native local notification banners are now displayed everywhere in the app, except when the message for that notification is visible on screen, that is, when the conversation is active and scrolled to the bottom.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/864

## Notes

Currently, tapping on a notification banner while that conversation is active and visible (but not scrolled to the bottom) does nothing, since opening the message from the notification will just open the conversation. The `ZMRequestToOpenViewsDelegate` has a method to open a conversation to show a particular message, although the conforming object `ZClientViewController` just opens the conversation and ignores the message. I'd suggest to implement this in a separate PR.


typfel:
@johnxnguyen support for scrolling to a specific message will come with ZIOS-10529

PR #2753, opened 2018-10-04
Correctly mark failing CI build

vytis:
### Issues

At the moment when a CI build fails when running test it marks the whole step as a failure. This prevents subsequent steps from running and there is no chance for upload test results and failing snapshots.

### Solutions

Do not fail `fastlane test` step when we find failing tests. Leave it for CI to figure out we have something wrong. An example is - https://github.com/wireapp/wire-ios/runs/21496698

## Notes

1. I have added a failing test to see if it works fine, this will be reverted before merging.
2. It seems that `fail_build: false` flag does nothing, added a workaround for catching the exception manually.


vytis:
Hmm so it seems the pipeline is ✅ even if there are test failures...

PR #2756, opened 2018-10-04
Fix/self mention highlight

johnxnguyen:
## What's new in this PR?

Update the highlight color for self mentions as per the spec.


cardassian-tailor:
just a comment. i dont know if its the dark theme or just the implementation but the aesthetic presentation of a mention (bold letters) looks atrocious. How a mention looks in the search area looks much better. 

@mikeger

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Mentions.swift
```diff
@@ -137,3 +132,10 @@ extension NSMutableAttributedString {
         }
     }
 }
+
+private extension UIColor {
+    static func selfMentionHighlight(for variant: ColorSchemeVariant) -> UIColor {
+        let alpha: CGFloat = variant == .light ? 0.48 : 0.8
+        return UIColor(red: 254/255.0, green: 191/255.0, blue: 2/255.0, alpha: alpha)
```
typfel: Consider moving this color definition to `ColorScheme.m`. IMO it still makes sense to use the color scheme system until we've come up with an alternative.

johnxnguyen: I was going to do that at first, but I thought the scheme was more for reusable color definitions, rather than single use colors? No problem, I can change it 😄

PR #2758, opened 2018-10-04
Fix: Conversation title is obsolete on the action sheet inside of a conversation

billypchan:
## What's new in this PR?

Remove the title of the action menu of conversation list and adding a snapshot test.
Add showConverationNameInMenuTitle option to ConversationActionController.presentMenu method to allow choosing showing menu title or not.


mikeger:
@billypchan the conversation name must be shown when the action sheet is opened from the conversation list and not shown when it's opened from the conversation details.

billypchan:
@mikeger added the menu title display option for the presentMenu method for two situations.

PR #2761, opened 2018-10-05
Feature/notification settings action

mikeger:
### Issues

Allow to set the conversation to be notified only for mentions.


Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController+Notifications.swift
```diff
@@ -0,0 +1,94 @@
...
+        default: return .default
+        }
+    }
+    
+    func action(for conversation: ZMConversation, handler: @escaping (NotificationResult) -> Void) -> UIAlertAction {
+        let checkmarkText: String
+        
+        if let mutedMessageTypes = self.mutedMessageTypes, conversation.mutedMessageTypes == mutedMessageTypes {
+            checkmarkText = " ✓"
+        }
```
alexaubry: Could this be a problem for RTL languages to hard-code the check mark on the right side?

mikeger: Hey @jennifergoertzen , can you comment?

mikeger: We don't have the control that uses checkmark in the similar way at the moment. I would propose to put them on the right, since I was also unable to find on the internet what is the correct way.

cardassian-tailor:
Any plans to push these fixes out in a new release or will it mellow for the next few weeks? 

This and the mentions color fix seem like good candidates to push these out as a hotfix.

PR #2765, opened 2018-10-08
Include self mentions in badge count

typfel:
## What's new in this PR?

- Bump data model which now includes self mentions in the badge count
- Use `estimatedUnreadSelfMentionCount` instead of iterating over the unread messages to check if they contain a self mention while calculating the conversation status.


typfel:
@typfel in theory yes, both predicates have been updated.

PR #2770, opened 2018-10-09
Removed unnecessary methods

mikeger:
### Issues

PaintCode is generating 2 types of methods: one that returns the image and another that drawing the image on the current context. We only use images for logos and custom images and draw on the context icons with numbers (used with `ZetaIconType`).

Therefore we can remove a bunch of methods that we are never calling.


vytis:
Is that a manual change that we will need to do every time we update style kit?

mikeger:
@vytis this was changed in StyleKit file, no manual changes needed.

PR #2772, opened 2018-10-09
Test: git LFS

vytis:
### Issues

We store all snapshot tests as binary files in the repo which means that anytime they change we are adding another copy of it to git history. This bloats the repo and makes it slower to check it out.

### Solutions

Try out git LFS support which converts binary files to pointers and stores them separately from the repo.

## Notes

This is a proof of concept PR to check how it all works in practice.


Wire-iOS Tests/SearchResultLabelTests.swift
```diff
@@ -26,6 +26,7 @@ class SearchResultLabelTests: ZMSnapshotTestCase {
 
     override func setUp() {
         super.setUp()
+        recordMode = true
```
mikeger: Record mode ⏺

vytis:
To have most benefit of git LFS and remove all binary data we would need to rewrite history of the project. It probably not worth the effort at the moment, but can be reconsidered in the future.

PR #2780, opened 2018-10-10
Fix: Guests should not be shown in Add People list after guest access is disabled

billypchan:
### Issues

allow guest and service toggle has no effect after toggling if no guest or service in conversation.

### Causes

Missing handling in configuration.areGuestOrServicePresent == false case

### Solutions

clean up the logic and make sure all the cases calls _setAllowGuests()


billypchan:
@typfel yes, this PR corrects the alternate path, which did not call _setAllowGuests().

PR #2781, opened 2018-10-11
Fix/devices activated

johnxnguyen:
## What's new in this PR?

This PR updates the copy for the list of clients and new clients alerts. It also removes old broken code related to client activation location, since it is not needed anyway.


Wire-iOS/Sources/UserInterface/Settings/UIAlertController+NewSelfClients.m
```diff
@@ -36,7 +36,9 @@ + (UIAlertController *)alertControllerForNewSelfClients:(NSSet<UserClient *> *)c
             deviceName = userClient.type;
         }
         
-        NSString *deviceDate = [userClient.activationDate wr_formattedDate];
+        NSString *formatKey = NSLocalizedString(@"registration.devices.activated", nil);
```
typfel: You can pass the `formattedDate` argument in here directly.

PR #2785, opened 2018-10-12
Don't access lazy var in deinit

mikeger:
### Issues

Lazily loaded variable `audioRecorder` was accessed in `deinit`. It was working before, but with the new runtime version this is causing the cryptic crash. 

Looking on it logically, the purpose of the lazy var is to save the object creation time, if the creation is not necessary. When we are accessing the object in `deinit` every time we are still going to load it, which defeats the purpose of lazy.


typfel:
So we don't define `deinit`?

mikeger:
Looks like the Azure is not running on the branches pointing to non-default merge base :(

PR #2787, opened 2018-10-12
Fix: Conversation screen: Japanese hiragana is treat as emoji

billypchan:
### Issues

Japanese hiraganas are treated as emoji as the characters are enlarged in the message.

### Causes

The range of emoji Misc symbols are wrongly defined and includes those characters.

### Solutions

Fix the range and add more test cases for different languages.

## Notes

During the test process I found that Wire Utility's String.containsEmoji does not return correct result for Glagolitic and Georgian. I will move the emoji detection logic to Wire Utility and update the method.


billypchan:
The emoji related methods will be moved to Wire utility.

PR #2792, opened 2018-10-15
ZIOS-10698: An incorrect status is shown for active call after it is minimized

nicorsm:
## What's new in this PR?

Fixed the behavior of the calling indicator view in the conversation list. I'm now using the `canJoinCall` and `isCallOngoing` functions, already used in the handling of the navigation bar, in order to display the green "Join" button or the calling indicator properly. I've also added a preemptive check to the current status of the notifications, so I can discard any check if the user has chosen to mute the conversation.


nicorsm:
@mikeger Let me check...

PR #2795, opened 2018-10-15
ZIOS-10723: Remove PureLayout (part 1)

alexaubry:
### Issues

We decided to minimize the usage of 3rd party layout libraries. We start with a few files to avoid blocking other tasks.

### Solutions

We use layout anchors instead of PureLayout. Snapshot tests are used to make sure the layout does not break in the process.


Wire-iOS/Sources/UserInterface/Registration/RegistrationPersonal/SignIn/EmailSignInViewController.m
```diff
@@ -115,7 +112,7 @@ - (void)removeObservers
 
 - (void)createEmailField
 {
-    self.emailField = [[RegistrationTextField alloc] initForAutoLayout];
+    self.emailField = [RegistrationTextField new];
```
vytis: Absolute nitpick, but a more idiomatic way would be `[[foo alloc] init]`. From my former life as an ObjC developer I remember there were some subtle differences, but they are much less relevant nowadays probably...

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
```diff
@@ -208,23 +203,37 @@ - (void)viewWillLayoutSubviews
...
+    self.rightViewOffsetConstraint = [self.rightView.leadingAnchor constraintEqualToAnchor:self.view.leadingAnchor];
+    self.rightViewOffsetConstraint.priority = UILayoutPriorityDefaultHigh;
+
+    self.leftViewWidthConstraint = [self.leftView.widthAnchor constraintEqualToConstant:0];
+    self.rightViewWidthConstraint = [self.rightView.widthAnchor constraintEqualToConstant:0];
+
+    self.sideBySideConstraint = [self.rightView.leadingAnchor constraintEqualToAnchor:self.leftView.trailingAnchor];
     self.sideBySideConstraint.active = NO;
+
+    NSArray<NSLayoutConstraint *> *constraints =
```
vytis: A nitpick, but this could be a bit less messy:

```
[NSLayoutConstraint activateConstraints: @[
   // leftView
       [self.leftView.topAnchor constraintEqualToAnchor:self.view.topAnchor],
       [self.leftView.bottomAnchor constraintEqualToAnchor:self.view.bottomAnchor],
<...>
]];
```

I don't think it is worth the effort of changing it in this PR, but could be something to do in future PRs.

mikeger:
@alexaubry please don't change the alloc/init to new, it just adds changes noise I think. Also we never really use the `new` in Obj-C as far as I know.

billypchan:
ProfilePictureStepViewControllerTests failed for failed_testThatItRendersTheViewControllerCorrectlyInAllDeviceSizes.

failed_testthatitrenderstheviewcontrollercorrectlyinalldevicesizes_ 320 568 2x
reference_testthatitrenderstheviewcontrollercorrectlyinalldevicesizes_ 320 568 2x

alexaubry:
@mikeger I switched to using `new` because we can't use `initForAutoLayout` anymore, wasn't sure what syntax to use. I will revert to using `init` if it's better.

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.m
alexaubry: Noted 👍

PR #2801, opened 2018-10-17
Removed Mixpanel

mikeger:
### Issues

It was decided to remove the Mixpanel support from the client.


ConfigurationSets/Development.xcconfig
```diff
@@ -84,7 +84,7 @@ GCC_PREPROCESSOR_DEFINITIONS = $(GCC_PREPROCESSOR_DEFINITIONS_BASE) DEBUG=1 USE_
 
 // No analytics / Hockey for Debug builds:
 
-BASIC_GCC_PREPROCESSOR_DEFINITIONS =  ${GCC_PREPROCESSOR_DEFINITIONS_BASE} IOS USE_HOCKEY=0 ENABLE_DEVELOPER_MENU=1  LOG_LEVEL_CONFIG=0xFFFFFFFF WIRESTAN=1 YOUTUBE_API_KEY=$(YOUTUBE_KEY) VIMEO_API_KEY=$(VIMEO_KEY) VIMEO_API_SECRET=$(VIMEO_SECRET) ANALYTICS_API_KEY=$(ANALYTICS_KEY) MIXPANEL_API_KEY=$(MIXPANEL_KEY) HOCKEY_APP_ID_KEY=$(HOCKEY_APP_ID) WIRE_SSO_URL_SCHEME=$(SSO_URL_SCHEME)
+BASIC_GCC_PREPROCESSOR_DEFINITIONS =  ${GCC_PREPROCESSOR_DEFINITIONS_BASE} IOS USE_HOCKEY=0 ENABLE_DEVELOPER_MENU=1  LOG_LEVEL_CONFIG=0xFFFFFFFF WIRESTAN=1 YOUTUBE_API_KEY=$(YOUTUBE_KEY) VIMEO_API_KEY=$(VIMEO_KEY) VIMEO_API_SECRET=$(VIMEO_SECRET) HOCKEY_APP_ID_KEY=$(HOCKEY_APP_ID) WIRE_SSO_URL_SCHEME=$(SSO_URL_SCHEME)
```
alexaubry: Should we also remove `YOUTUBE_API_KEY`, `VIMEO_API_KEY` and `VIMEO_API_SECRET` from the config files? They don't seem to be used anywhere anymore.

Wire-iOS/Sources/Analytics/AnalyticsProviderFactory.swift
```diff
@@ -39,12 +39,6 @@ fileprivate let ZMEnableConsoleLog = "ZMEnableAnalyticsLog"
             zmLog.info("Creating analyticsProvider: AnalyticsConsoleProvider")
             return AnalyticsConsoleProvider()
         }
-        else if UseAnalytics.boolValue || AutomationHelper.sharedHelper.useAnalytics {
-            zmLog.info("Creating analyticsProvider: AnalyticsMixpanelProvider")
```
alexaubry: Maybe we should keep this `else if` case and replace it with a comment and return `nil`, so that we know where to create the provider if we decide to replace Mixpanel with another service eventually.

~~~swift
else if UseAnalytics.boolValue || AutomationHelper.sharedHelper.useAnalytics {
    // TODO: create the production analytics provider when it's available
    return nil
}
~~~

mikeger: You are right, let's keep it. I was thinking that it's OK, since people can just go back and check the PR that removed Mixpanel, but it would be more handy to keep the logic in.

ConfigurationSets/Development.xcconfig
mikeger: Right, they are also not used! Will remove.

PR #2802, opened 2018-10-17
fix: crash when generate preview in ShareViewController

billypchan:
## What's new in this PR?

Fix the crash related nil previewView generation in ShareViewController.

## Note 
Consider removing ShareViewController and related classes if it is not going to release to the public.


mikeger:
May be we should change the signature to be non-optional, if it (the preview) is expected to be generated in all cases?

PR #2804, opened 2018-10-17
ZIOS-10698: Join button fixes after design review

nicorsm:
## What's new in this PR?

- Join button is visible in the conversation list and detail if you receive a call when conversation is muted
- Fixed tests


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -271,18 +271,14 @@ final internal class CallingMatcher: ConversationStatusMatcher {
...
         
-        guard let conversation = conversation,
-            conversation.mutedMessageTypes == .none,
-            let state = state else {
-            return .none
-        }
+        guard let state = state else { return .none }
         
-        if state.canJoinCall(conversationType: conversation.conversationType) {
+        if state.canJoinCall() {
```
mikeger: So do you call `canJoinCall` from somewhere else where you have the `conversationType` parameter?

nicorsm: Yes, that was needed because the Join button was visible inside the conversation detail only for group conversations. According to design, we've removed this limit, so you can still join the call even if you're silenced on both sides (conversation list + detail) - I still need to implement that.

PR #2805, opened 2018-10-17
Fix: Setting: username text field has input character components when using Chinese Cangjie input method

billypchan:
### Issues

After https://github.com/wireapp/wire-ios/pull/1760, the fix https://github.com/wireapp/wire-ios/pull/1360 is no longer effective.

### Causes

The code self.text = self.text causes Chinese input method failing to combine character components to signal character.

### Solutions

After iOS 11 the fix in https://github.com/wireapp/wire-ios/pull/1760 is no longer necessary. Apply it only for iOS 10 devices.


billypchan:
This PR will be closed. This issue can be fixed in another approach with constraints and text labels/field's Compression Resistance Priority.

PR #2814, opened 2018-10-19
Feature/notification settings screen

johnxnguyen:
## What's new in this PR?

This PR introduces a dedicated screen to manage the conversation notification options. This is in addition to the existing action sheets used.

Additionally, the conversation options section is no longer shown if there are no options to show.

## Discussion
`GroupDetailsNotificationOptionsViewController` has a lot in common with `GroupDetailsTimeoutOptionsViewController`, they are visually the same. I initially wanted to try to refactor out the model logic so both options screens could use the same view controller, but I couldn't find a decent solution. Perhaps if one of you have some suggestions I'd be keen to hear them 😄


Wire-iOS/Sources/UserInterface/Conversation Options/ConversationNotificationOptionsViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+        CheckmarkCell.register(in: collectionView)
+        collectionView.register(SectionFooter.self, forSupplementaryViewOfKind: UICollectionView.elementKindSectionFooter, withReuseIdentifier: "SectionFooter")
+    }
+    
+    private func configureConstraints() {
+        
+        view.addSubview(collectionView)
+        collectionView.translatesAutoresizingMaskIntoConstraints = false
+        
+        collectionView.fitInSuperview()
```
alexaubry: Not sure whet this helper is defined, but we’re trying to move away from auto layout libraries. Maybe it’s worth doing the constraints explicitly here?

Wire-iOS/Sources/UserInterface/Conversation Options/ConversationNotificationOptionsViewController.swift
```diff
@@ -0,0 +1,174 @@
...
+            self.conversation.mutedMessageTypes = types
+        }
+        
+        self.dismisser?.dismiss(viewController: self, completion: nil)
+    }
+    
+    // MARK: Layout
+    
+    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
+        return CGSize(width: collectionView.bounds.size.width, height: 56)
```
alexaubry: Eventually, we should move away from fixed-height cells to support Dynamic Tyoe better.

Wire-iOS/Sources/UserInterface/GroupDetails/Cells/GroupDetailsNotificationOptionsCell.swift
```diff
@@ -0,0 +1,54 @@
...
+        super.applyColorScheme(colorSchemeVariant)
+        icon = UIImage(for: .alerts, iconSize: .tiny,
+                       color: UIColor(scheme: .textForeground, variant: colorSchemeVariant))
+    }
+
+}
+
+extension MutedMessageTypes {
+    var notificationString: String? {
+        switch self {
```
alexaubry: Do these need to be localized?

johnxnguyen: Ah oops, good catch!

Wire-iOS/Sources/UserInterface/Conversation Options/ConversationNotificationOptionsViewController.swift
johnxnguyen: So `fitInSuperview` is defined by us and already defines explicit constraints.

Wire-iOS/Sources/UserInterface/Conversation Options/ConversationNotificationOptionsViewController.swift
johnxnguyen: What would you do here? Also, I see we have hardcoded row heights throughout the app, perhaps we should support dynamic type in rows all together rather than in one place.

alexaubry: Right, I might be worth considering as a full standalone task.

PR #2820, opened 2018-10-23
Declare default init implementation for ObjC

mikeger:
### Issues

Start UI was crashing when trying to show the top people line.

### Causes

In `TopPeopleCell.createUserImageView` we have the following code:

```c
    self.badgeUserImageView = [[BadgeUserImageView alloc] init];
```

Seems like the `BadgeUserImageView` cannot be alloc-init'ed from Objective-C.

### Solutions

I was really surprised by that, since me and Obj-C compiler knew that any object inherited from NSObject has init method.

The solution is to declare the init method in swift.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/UserImageView/BadgeUserImageView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/UserImageView/BadgeUserImageView.swift
@@ -50,7 +50,7 @@ import WireExtensionComponents
 
     // MARK: - Initialization
 
-    @objc convenience init() {
+    @objc override convenience init(frame: CGRect) {
         self.init(size: .small)
     }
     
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/UserImageView/BadgeUserImageView.swift
```diff
@@ -50,6 +50,10 @@ import WireExtensionComponents
 
     // MARK: - Initialization
 
+    @objc convenience init() {
+        self.init(size: .small)
```
alexaubry: Does overriding `init(frame:)` fix the problem too? Maybe that's why the crash is happening?

mikeger: So it's calling UIView.init, that is in turn calling initWithFrame, that is not implemented in BadgeUserImageView. However the parent class of BadgeUserImageView is implementing the initWithFrame... 🤔 I think the compiler is inserting the crashing implementation of initWithFrame in this case. I will replace init with initWithFrame, sine this seems to work for both cases: when init and when initWithFrame is called ⚡️

PR #2821, opened 2018-10-23
Feature/support custom be

mikeger:
### Issues

This allows configuring Wire to be used against the custom backend.


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -148,7 +148,7 @@ class ShareExtensionViewController: SLComposeServiceViewController {
 
     private var authenticatedAccounts: [Account] {
         guard let accountManager = accountManager else { return [] }
-        return accountManager.accounts.filter(\.isAuthenticated)
+        return accountManager.accounts.filter { $0.isAuthenticated(with: BackendEnvironment.shared) }
```
typfel: We now need to pass in the backend environment. But isn't the `AccountManager` owned by the SessionManager which is turn gets the backend environment injected on launch. Maybe it makes sense to move around these methods?

`SessionManager.isAccountAuthenticated(_ account: Account)`

or maybe establish a delegate between `AccountManager` and `SessionManager`?

mikeger: I think AccountManager is created before session manager, I might be wrong, let me check.

mikeger: So the reason why I did not liked to initialize the account manager with backend spec is that it felt unnatural to bind those two things; AccountManager is very simple and to the point, and the only reason we need the backend info in it is because we need to find the cookie per server name, which is actually implemented as extension in SyncEngine: https://github.com/wireapp/wire-ios-sync-engine/commit/71c15b8b69706c3fe1a0ea90a17d416b71f414b9 

I am thinking now about the better way to structure it, I think your proposal to make it the method on SessionManager makes sense, I will update the SE and create the new PR.

PR #2822, opened 2018-10-23
ZIOS-10545: Share extension does not generate previews for URLs

nicorsm:
## What's new in this PR?

Share extension doesn't generate the preview for URLs if the user shares a text that contains a link - like "Hey! Check out www.wire.com".
I've added the possibility to fetch URLs if the string contains one or more URLs and there are no URL attachments.


mikeger:
May be we could reuse the code to generate the link previews from the normal message sending flow?
If not, I think we can add some tests for the method, providing different possible types of input.

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire Share Extension.xcscheme
```diff
@@ -80,7 +80,7 @@
       <RemoteRunnable
          runnableDebuggingMode = "1"
          BundleIdentifier = "com.apple.mobilesafari"
-         RemotePath = "/Applications/MobileSafari.app">
+         RemotePath = "/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot/Applications/MobileSafari.app">
```
vytis: Not sure if this is on purpose or was accidentally committed?

nicorsm: Accidentally commited, for sure 😁

PR #2829, opened 2018-10-25
SSO Login

mikeger:
### Issues

- Refactored the client management.
- Disabled the account settings change UI for SSO users.

### Prerequisites

- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/887


alexaubry:
@mikeger Are we planning to make SSO public in the next release? If yes, we should update `CompanyLoginController.companyLoginEnabled` to show it to public users.

mikeger:
@alexaubry good point, will enable it for the moment

Wire-iOS/Sources/Helpers/syncengine/UserClient+RemoveTask.swift
```diff
@@ -0,0 +1,134 @@
...
+        let alert = UIAlertController(title: "", message: message, preferredStyle: .alert)
+        let action = UIAlertAction(title: NSLocalizedString("general.ok", comment: ""), style: .default) { [unowned alert] (_) -> Void in
+            alert.dismiss(animated: true, completion: .none)
+        }
+        alert.addAction(action)
+        self.present(alert, animated: true, completion: .none)
+    }
+    
+    func requestPassword(_ completion: @escaping (ZMEmailCredentials?)->()) {
+        let passwordRequest = RequestPasswordViewController.requestPasswordController() { (result: Either<String, NSError>) -> () in
```
vytis: Maybe `Result` would be a bit more fitting here, `.right(let error)` seems a bit weird to be honest! Probably would need changes in `RequestPasswordViewController` as well.

mikeger: @vytis sure, but for the moment I also just moved that code, will update it in the upcoming PRs. Either is a bit weird because it's the guest from another world: https://www.schoolofhaskell.com/school/starting-with-haskell/basics-of-haskell/10_Error_Handling#either-may-be-better-than-maybe

vytis: Twin brother from another mother

PR #2830, opened 2018-10-25
Remove type Either in favor of Result

mikeger:
### Issues

The type `Result` is more specific to the given use case.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RequestPasswordViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RequestPasswordViewController.swift
@@ -19,18 +19,13 @@
 
 import Foundation
 
-enum Either<TLeft, TRight> {
-    case left(TLeft)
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/RequestPasswordViewController.swift
```diff
@@ -49,12 +44,12 @@ class RequestPasswordViewController: UIAlertController {
         let okAction = UIAlertAction(title: okTitle, style: .default) { [unowned controller] (action: UIAlertAction) -> Void in
             if let passwordField = controller.textFields?[0] {
                 let password = passwordField.text ?? ""
-                controller.callback?(Either.left(password))
+                controller.callback?(Result<String>.success(password))
```
vytis: This should do as well:
```suggestion
                controller.callback?(.success(password))
```

mikeger:
@vytis updated 🙇

PR #2832, opened 2018-10-25
Fix: broken constraints in conversation list cell when the app launches

billypchan:
### Issues

The menuView and the accessory view of conversation list cell are broken and it slows down app launch.

### Causes

The menuView's width is not defined when collapsed. 
Missing constraint to define accessory view's width when it is collapsed.

### Solutions
Rewrite these classes' constraint creation without 3rd party libraries:
- ConversationListItemView
- ConversationListCell
- SwipeMenuCollectionCell
- AnimatedListMenuView
- RoundedBadge
- ConversationListAccessoryView

Refactoring the conversaton list cell swiping animation.
Setup the constraint's constant instead of changing activation state to toggle the accessary view.

### Note:
TODO:
- [x] fix broken constraint with the cell has no accessories.
- [x] check some constraints are broken since the view's frame size was zero before app is active


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+leftView.swift
```diff
@@ -67,7 +67,7 @@ extension SplitViewController {
                 completionBlock()
             })
         } else {
-            self.view.layoutIfNeeded()
+//            self.view.layoutIfNeeded()
```
mikeger: This needs to be either removed or uncommented back.

Wire-iOS/Sources/UserInterface/Components/Views/RoundedBadge.swift
```diff
@@ -17,41 +17,58 @@
 //
 
 import Foundation
-import Cartography
 
 public class RoundedBadge: UIButton {
     public let containedView: UIView
     public var trailingConstraint: NSLayoutConstraint!
+    public var leadingConstraint: NSLayoutConstraint!
```
mikeger: Is there a reason to make this property public?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.h
```diff
@@ -23,6 +23,7 @@
 @class ConversationListItemView;
 @protocol ConversationListCellDelegate;
 
+static const CGFloat MaxVisualDrawerOffsetRevealDistance = 21;
```
mikeger: Why the value is changed from 48 to 21?

Wire-iOS/Sources/UserInterface/Components/Views/SwipeMenuCollectionCell.m
```diff
@@ -91,19 +84,21 @@ - (void)prepareForReuse
 
 - (void)setupSwipeMenuCollectionCell
 {
+    self.hasCreatedSwipeMenuConstraints = NO;
```
mikeger: All the ivars values in Obj-C are initialized to 0, this is the language convention therefore you don't have to explicitly set the value to NO.

Wire-iOS/Sources/UserInterface/Components/Views/SwipeMenuCollectionCell.m
```diff
@@ -367,8 +322,7 @@ - (void)setDrawerOpen:(BOOL)open animated:(BOOL)animated
     }
 
     dispatch_block_t action = ^() {
-        self.visualDrawerOffset = open ? self.drawerWidth : 0;
-        [self checkAndUpdateMaxVisualDrawerOffsetConstraints:self.visualDrawerOffset];
+        self.visualDrawerOffset = 0;
```
mikeger: Why this is changed now?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.m
```diff
@@ -79,20 +74,21 @@ - (instancetype)initWithFrame:(CGRect)frame
 
 - (void)setupConversationListCell
 {
+    self.hasCreatedInitialConstraints = NO;
```
mikeger: Same as before in regards of the default ivar value.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListItemView.m
```diff
@@ -67,28 +58,30 @@ - (instancetype)init
                                                    object:nil];
 
         [self setupStyle];
+
+        self.translatesAutoresizingMaskIntoConstraints = NO;
```
mikeger: I think setting the `translatesAutoresizingMaskIntoConstraints` to NO in our own classes is conflicting with the traditional UIKit control logic, where the default value is YES and user of the control decides to switch it off. I think it's easier to follow the platform's default logic in this area. I also find it suboptimal to set translatesAutoresizingMaskIntoConstraints to NO for every view used, but it's hard from the code to understand that the property is already set in this case.

mikeger:
@billypchan I don't think that the given big amount of work invested in this PR is comparable to the win that we get. I think in the future we should to minimize the changes done in the scope of one PR: if we see that the constraints are breaking it's good to have the PR that just fixes that in 1-2 lines if possible. The breaking constraints can be also fixed by decreasing the priority of some constraints that are breaking for valid reasons. Please consider this going forward.

Wire-iOS/Sources/UserInterface/Components/Views/RoundedBadge.swift
billypchan: The public property of the above trailingConstraint is not needed also. Thanks!

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell.h
billypchan: This fix corrects the frame size of a view on the left of the drawer, which had zero width. This constant is reduced.

Wire-iOS/Sources/UserInterface/Components/Views/SwipeMenuCollectionCell.m
billypchan: The "drawer close" animation is simplified. the setter of visualDrawerOffset calls checkAndUpdateMaxVisualDrawerOffsetConstraints and no need to call it again here.

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListItemView.m
billypchan: I checked translatesAutoresizingMaskIntoConstraints is called in ConversationListCell's updateConstraints again and it is not needed.

PR #2837, opened 2018-10-29
Fix: Login remove devices screen background is blue on devices with notch

billypchan:
### Issues

The navigation view controller of the remove devices screen does not cover the behind view.

### Causes

The navigation view controller is a subview of AdaptiveFormViewController and AdaptiveFormViewController 's top and bottom anchors were attached to to the safe area anchors.

### Solutions

Attach the navigation view controller's top and bottom anchors to its parents top and bottom anchors.
(NavigationController's constraints are fixed to prevent the confliction.)


mikeger:
@alexaubry can you look into this one?

Wire-iOS/Sources/Authentication/RegistrationPersonal/AdaptiveFormViewController.swift
```diff
@@ -95,8 +95,8 @@ class AdaptiveFormViewController: BlueViewController, AuthenticationCoordinatedV
         ]
 
         NSLayoutConstraint.activate([
-            childViewController.view.bottomAnchor.constraint(equalTo: safeBottomAnchor),
-            childViewController.view.topAnchor.constraint(equalTo: safeTopAnchor)
+            childViewController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
```
alexaubry: Did you try doing the full registration flow with this approach? The reason we use these safe anchors is to support the TOS and name screens. Do you think it would break these to change the constraints here?

billypchan: @alexaubry I checked the registration flow and it looks fine.

PR #2840, opened 2018-10-31
Fix: App crash when call replace emoji in message

billypchan:
## What's new in this PR?

Fix the crash caused by incorrect range calculation in replaceEmoticons() function.

### dependency 
- [x] https://github.com/wireapp/wire-ios-utilities/pull/64


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+MessageFormatting.swift
```diff
@@ -132,7 +132,7 @@ extension NSMutableAttributedString {
         let allowedIndexSet = IndexSet(integersIn: Range<Int>(wholeRange)!, excluding: excludedRanges)
         
         for range in allowedIndexSet.rangeView {
-            let range = NSRange(location: range.startIndex, length: range.endIndex - range.startIndex)
+            let range = NSRange(location: range.startIndex, length: range.endIndex - range.startIndex - 1)
```
mikeger: So... off-by-one error? Can you elaborate on the change?

billypchan: Swift's Range's endIndex indicates the non-inclusive upper Bound of the range. Take an example for the newly add testThatReplaceEmojiCorrectlyWhenTheRangeIsWithinTheStringLength:

We want to exclude 5..<20 from the string (The characters 1,2,3,4 will be replaced with emoji)
The allowedIndexSet is 0..<5, its endIndex is 5

We want to convert Range to NSRange with a location and a length argument. The length is 5 - 0 - 1 = 4 and the location is 0.

PR #2842, opened 2018-10-31
Fix: Hightlighted mention best match is not visible under dark mode

billypchan:
## What's new in this PR?

change selected mention user's background color to a color different from the non-hightlighted users.

### Note
This PR also improve the keyboard collapsion state checker. On iOS12.1 simulator when the keyboard is not toggled, it has the original size but it is positioned out of the screen bound.


WireExtensionComponents/Utilities/ColorScheme.h
```diff
@@ -75,6 +75,7 @@ extern ColorSchemeColor ColorSchemeColorTokenFieldBackground;
 extern ColorSchemeColor ColorSchemeColorTokenFieldTextPlaceHolder;
 
 extern ColorSchemeColor ColorSchemeColorSelfMentionHighlight;
+extern ColorSchemeColor ColorSchemeColorMentionBestMatchBackground;
```
typfel: I think this  color constant is something we could re-use for all cells that we want to show a highlighted state for. I would therefore name it something like `ColorSchemeColorCellHighlight`

PR #2843, opened 2018-11-01
Feature: Change highlight best match in suggestions list when hardware up/down arrow key is pressed

billypchan:
### Issues

Support mention list selection with hardware keyboard up/down keys. The collected view scrolls and keeps highlighted cell at the middle after selection updates.


Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -33,16 +33,25 @@ import Cartography
 
 @objc protocol UserList {
     var users: [UserType] { get set }
+    var bestSuggestionUser: UserType { get }
```
mikeger: I think the adjective `best` is not the best here. May be it's more to the point to call it `selectedUser`? I think it also needs to be optional, since the users list might be empty.

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -168,10 +180,54 @@ class UserSearchResultsViewController: UIViewController, KeyboardCollapseObserve
 extension UserSearchResultsViewController: Dismissable {
     func dismiss() {
         self.view.isHidden = true
+        collectionViewSelectedIndex = .none
     }
 }
 
 extension UserSearchResultsViewController: UserList {
+    var bestSuggestionUser: UserType {
```
mikeger: I think it is better to implement the setter on `collectionViewSelectedIndex` and clamp the value there. Also reload the collection view.

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -168,10 +180,63 @@ class UserSearchResultsViewController: UIViewController, KeyboardCollapseObserve
...
     func dismiss() {
         self.view.isHidden = true
+        collectionViewSelectedIndex = .none
     }
 }
 
 extension UserSearchResultsViewController: UserList {
+    var selectedUser: UserType? {
+
+        guard searchResults.count > 0 else {
```
vytis: Probably would make more sense to check `collectionViewSelectedIndex` instead. I assume that first user should always be pre-selected when the overlay opens? Or you need to tap the up arrow to select it first?

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -33,16 +33,25 @@ import Cartography
...
+    func selectNextUser()
 }
 
 class UserSearchResultsViewController: UIViewController, KeyboardCollapseObserver {
 
     private let collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewLayout())
-    private var searchResults: [UserType] = []
+    private var searchResults: [UserType] = [] {
+        didSet {
+            collectionViewSelectedIndex = searchResults.count - 1
```
vytis: This will be -1 if searchResults.count is empty, should probably set index to `nil` in that case.

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -168,10 +180,63 @@ class UserSearchResultsViewController: UIViewController, KeyboardCollapseObserve
...
 }
 
 extension UserSearchResultsViewController: UserList {
+    var selectedUser: UserType? {
+
+        guard searchResults.count > 0 else {
+            return nil
+        }
+
+        clampCollectionViewSelectedIndex()
```
vytis: This could I guess be moved to setter of `collectionViewSelectedIndex`. Also this method could simply be min and max:

```
let index = min(searchResults.count - 1, max(0, index))
```

mikeger:
@billypchan have you considered the comments from Vytis?

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -33,16 +33,43 @@ import Cartography
...
+            }
+        }
+    }
     private var query: String = ""
     private var collectionViewHeight: NSLayoutConstraint?
     private let rowHeight: CGFloat = 56.0
     private var isKeyboardCollapsedFirstCalled = true
+
+    private var _collectionViewSelectedIndex : Int? = .none
+    private var collectionViewSelectedIndex: Int? {
```
vytis: Not sure exactly how the UI works here, but it would make sense to have the selected index be non-optional and default to 0. So if you start the mentioning flow and there are some users tapping enter should highlight the current user. Tapping up/down would cycle through the list.

Wire-iOS/Sources/UserInterface/Conversation/Mentions/UserSearchResultsViewController.swift
```diff
@@ -168,10 +194,48 @@ class UserSearchResultsViewController: UIViewController, KeyboardCollapseObserve
...
 }
 
 extension UserSearchResultsViewController: UserList {
+    var selectedUser: UserType? {
+
+        guard let collectionViewSelectedIndex = collectionViewSelectedIndex else {
+            return .none
+        }
+
+        let index = searchResults.count - 1 - collectionViewSelectedIndex
```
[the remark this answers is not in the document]
billypchan: The logic is simplified, thanks!

PR #2844, opened 2018-11-02
Refactor color handling

vytis:
### Issues

Handling colour schemes was done when we had Classy and had quite outdated structure with multiple dictionaries. It was suspected to cause crashes sometimes when it was not initialized for some reason.

### Solutions

Moved colours to an enum and made dark/light colour pairs more explicit. The API of using the colors from the scheme changed slightly because convenience initializers must call other initializers so things like `UIColor(scheme: ColorSchemeColor)` is not possible unless we re-create the color every time.


WireExtensionComponents/Utilities/ColorScheme.swift
```diff
@@ -0,0 +1,248 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+
+public extension UIColor {
```
vytis: Main changes are in this file

WireExtensionComponents/Utilities/UIColor+AccentColor.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    static var softPink: UIColor  {
+        return UIColor(for: .softPink)!
+    }
+
+    static var violet: UIColor  {
+        return UIColor(for: .violet)!
+    }
+
+    @objc public convenience init?(for accentColor: ZMAccentColor) {
```
vytis: Accent color creation is failable initializer because there are some invalid values in the enum

WireExtensionComponents/Utilities/UIColor+WR_ColorScheme.h
```diff
@@ -23,8 +23,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface UIColor (WR_ColorScheme)
 
-+ (UIColor *)wr_colorFromColorScheme:(ColorSchemeColor)colorSchemeColor NS_SWIFT_NAME(init(scheme:));
-+ (UIColor *)wr_colorFromColorScheme:(ColorSchemeColor)colorSchemeColor variant:(enum ColorSchemeVariant)variant NS_SWIFT_NAME(init(scheme:variant:));
++ (UIColor *)wr_colorFromColorScheme:(NSInteger)colorSchemeColor;
++ (UIColor *)wr_colorFromColorScheme:(NSInteger)colorSchemeColor variant:(enum ColorSchemeVariant)variant;
```
vytis: This is the drawback of defining enum in Swift, no easy way to use it in Obj-C headers. There is a lot of code that uses this method, but I left it as it is.

WireExtensionComponents/Utilities/UIColor+AccentColor.swift
alexaubry: Isn't it easier to crash when we get the ZMAccentColorUndefined? This is pretty unlikely to happen unless something has already gone wrong?

WireExtensionComponents/Utilities/ColorScheme.swift
```diff
@@ -0,0 +1,248 @@
...
+    public static func from(scheme: ColorSchemeColor) -> UIColor {
+        return ColorScheme.default.color(named: scheme)
+    }
+
+    public static func from(scheme: ColorSchemeColor, variant: ColorSchemeVariant) -> UIColor {
+        return ColorScheme.default.color(named: scheme, variant: variant)
+    }
+}
+
+fileprivate struct ColourPair {
```
alexaubry: Maybe we should match the UIKit spelling and say Color?

WireExtensionComponents/Utilities/UIColor+WR_ColorScheme.h
alexaubry: Could we move those to Swift to avoid that?

WireExtensionComponents/Utilities/UIColor+AccentColor.swift
vytis: Yeah maybe it's actually easier to crash in initializer, not on call sites. Also maybe worth looking into removing these invalid enum values altogether.

WireExtensionComponents/Utilities/UIColor+WR_ColorScheme.h
alexaubry: Could you move those methods to Swift to avoid that?

vytis: Actually it is :) At some point during refactoring I was in a weird state where this needed lots of changes and decided against it, but it's not the case anymore.

WireExtensionComponents/Utilities/ColorScheme.swift
```diff
@@ -0,0 +1,248 @@
...
+
+    case selfMentionHighlight
+    case cellHighlight
+}
+
+extension UIColor {
+
+    public static func from(scheme: ColorSchemeColor) -> UIColor {
+        return ColorScheme.default.color(named: scheme)
+    }
```
billypchan: How about creating a UIColor from ColorSchemeColor with a init method like:
```
    public convenience init(scheme: ColorSchemeColor) {
        let cgColor = ColorScheme.default.color(named: .background).cgColor
        self.init(cgColor: cgColor)
    }
```

vytis: I did it like this initially, but I assume this would re-create the object every time we access it. Now all the colors used are cached in static vars.

vytis:
@alexaubry updated - call sites that were force-unwrapped now use colour constants directly and where it's taken from `ZMUser.accentColor` values goes through one single initializer that falls back to default blue on invalid values.

Wire-iOS/Sources/Helpers/syncengine/ZMAccentColor+Additions.swift
```diff
@@ -19,26 +19,18 @@
...
-    
-    static func allSelectable() -> [ZMAccentColor] {
-        return [.strongBlue,
-                .strongLimeGreen,
-                .vividRed,
-                .brightOrange,
-                .softPink,
-                .violet]
-    }
+//    var color: UIColor {
```
alexaubry: Should we delete this?

WireExtensionComponents/Utilities/ColorScheme.m
```diff
@@ -19,109 +19,45 @@
...
-}
-
-static NSString* dark(NSString *colorString) {
-    return [NSString stringWithFormat:@"%@-dark", colorString];
-}
-
-static NSString* light(NSString *colorString) {
-    return [NSString stringWithFormat:@"%@-light", colorString];
-}
+//static NSString * ColorSchemeNameAccentColorForColor(ZMAccentColor color);
```
mikeger: What's up with this one?

vytis: 🙈

WireExtensionComponents/Utilities/ColorScheme.swift
```diff
@@ -0,0 +1,250 @@
...
+
+    @objc(wr_colorFromColorScheme:variant:)
+    public static func from(scheme: ColorSchemeColor, variant: ColorSchemeVariant) -> UIColor {
+        return ColorScheme.default.color(named: scheme, variant: variant)
+    }
+}
+
+fileprivate struct ColorPair {
+    let light: UIColor
```
mikeger: This abstraction can be unfit for the possible case of adding more color schemes, like true dark or may be pumpkin color. But should be easy to change in case we will get there :)

vytis: It's all private to this file, just seemed like a most concise way to write out the combinations.

PR #2849, opened 2018-11-05
Fixed composer observation, added edited icon

mikeger:
### Issues

Observe the message being replied to, wether the message:
- Was edited: update the view.
- Was deleted: cancel the reply process.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
@@ -146,7 +146,7 @@ final class MessageThumbnailPreviewView: UIView {
             }
         }
         else if message.isVideo, let fileMessageData = message.fileMessageData {
-            let imageIcon = NSTextAttachment.textAttachment(for: .videoMessage, with: .textForeground)!
+            let imageIcon = NSTextAttachment.textAttachment(for: .videoCall, with: .textForeground)!
             let initialString = NSAttributedString(attachment: imageIcon) + "  " + "conversation.input_bar.message_preview.video".localized.localizedUppercase
             contentTextView.attributedText = initialString && attributes
             
--- a/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
@@ -236,7 +236,7 @@ final class MessagePreviewView: UIView {
         let attributes: [NSAttributedString.Key: Any] = [.font: UIFont.smallSemiboldFont,
                                                          .foregroundColor: UIColor.textForeground]
         
-        senderLabel.attributedText = ((message.sender?.displayName(in: message.conversation!) ?? "") && attributes) + self.editIcon()
+        senderLabel.attributedText = (message.senderName && attributes) + self.editIcon()
         
         if let textMessageData = message.textMessageData {
             let hasNoText = textMessageData.messageText == nil || textMessageData.messageText!.isEmpty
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -148,7 +146,7 @@ final class MessageThumbnailPreviewView: UIView {
             }
         }
         else if message.isVideo, let fileMessageData = message.fileMessageData {
-            let imageIcon = NSTextAttachment.textAttachment(for: .videoMessage, with: .textForeground, and: .medium)!
+            let imageIcon = NSTextAttachment.textAttachment(for: .videoMessage, with: .textForeground)!
```
alexaubry: The image here should be `.videoCall` according to the spec.

```suggestion
            let imageIcon = NSTextAttachment.textAttachment(for: .videoCall, with: .textForeground)!
```

Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -224,11 +222,22 @@ final class MessagePreviewView: UIView {
             ])
     }
     
+    private func editIcon() -> NSAttributedString {
+        if message.updatedAt != nil {
+            return "  " + NSAttributedString(attachment: NSTextAttachment.textAttachment(for: .pencil, with: .textForeground, iconSize: 8)!)
```
alexaubry: Is the double space intended?

Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -224,11 +222,22 @@ final class MessagePreviewView: UIView {
...
+        }
+    }
+    
     private func updateForMessage() {
-        senderLabel.text = message.sender?.displayName(in: message.conversation!)
+        
         let attributes: [NSAttributedString.Key: Any] = [.font: UIFont.smallSemiboldFont,
                                                          .foregroundColor: UIColor.textForeground]
         
+        senderLabel.attributedText = ((message.sender?.displayName(in: message.conversation!) ?? "") && attributes) + self.editIcon()
```
alexaubry: We added the `senderName` utility property with the cell refactoring and it could be easier to use it here.

It can compute the sender name in the conversation, and also takes into account the self user (it returns "You") and null sender (it returns "Someone").

```suggestion
        senderLabel.attributedText = (message.senderName && attributes) + self.editIcon()
```

Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
```diff
@@ -242,23 +251,20 @@ final class MessagePreviewView: UIView {
...
         }
         else if message.isAudio {
-            let imageIcon = NSTextAttachment.textAttachment(for: .microphone, with: .textForeground, and: .medium)!
+            let imageIcon = NSTextAttachment.textAttachment(for: .microphone, with: .textForeground)!
             let initialString = NSAttributedString(attachment: imageIcon) + "  " + "conversation.input_bar.message_preview.audio".localized.localizedUppercase
             contentTextView.attributedText = initialString && attributes
         }
         else if let fileData = message.fileMessageData {
-            let imageIcon = NSTextAttachment.textAttachment(for: .document, with: .textForeground, and: .medium)!
+            let imageIcon = NSTextAttachment.textAttachment(for: .document, with: .textForeground)!
```
alexaubry: Do we need a special case for videos here?

Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
mikeger: Yes, this is intended. Somehow it's looking good.

Wire-iOS/Sources/UserInterface/Components/MessagePreviewView.swift
mikeger: Videos are handled in another view, the one with the thumbnail.

PR #2851, opened 2018-11-06
Do not scroll conversation to the bottom when replying

mikeger:
### Issues

- The conversation used to have the logic to scroll the content to the bottom when user is focusing on the input field. This should not happen when user is editing the message or is replying to the message.
- ~~Fixed delegate call `delegate?.conversationCell?(self, didSelect: .save, for: self.message)` -> `delegate?.conversationCell?(self, didSelect: .save)` (message is inferred from the cell now).~~
- Fixed jumping to the cell (row index vs section index).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.m
@@ -69,7 +69,7 @@
 {
     NSUInteger index = [self.messageWindow.messages indexOfObject:message];
     if (index != NSNotFound) {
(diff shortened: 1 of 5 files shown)


alexaubry:
@mikeger We can't remove the `message` argument from the delegate method, because the new cells are not subclasses of the old `ConversationCell`, and therefore we can't infer the message, because they don't have the `message` property.

But when implementing the reply cell I realized I forgot to update the method in the content view controller, which caused a crash. I fixed it like this: https://github.com/wireapp/wire-ios/pull/2847/files#diff-473aefd08f4960861558569c7db525d8R820.

mikeger:
@alexaubry sorry, but I had no info about this, I think we need to improve the communication on this feature.

alexaubry:
@mikeger Sorry... I will update the architecture documentation with the info.

PR #2854, opened 2018-11-06
Feature/direct replies mute logic [Partial]

johnxnguyen:
## What's new in this PR?

This PR contains changes for the new DM API in the context of `MutedMessageTypes`. The changes include:

- Localized strings are updated for the notification settings.
- The group conversation option cells now display the option value as a subtitle.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/591

## Still Todo
- Add notification settings option to 1:1 conversation details/
- Remove the "Notifications..." conversation action from the "..." menu in the bottom right of both group and 1:1 conversation detail screens.


johnxnguyen:
Failing tests are due to reply cell snapshots.

PR #2855, opened 2018-11-06
ZIOS-10812: Remove reply option for unsent message

alexaubry:
### Issues

Before, it was possible to reply to a message that failed to deliver.

### Solutions

We check for the delivery status in the `canReply` method.

### DIFF
--- a/Wire-iOS Tests/ConversationCellActionControllerTests.swift
+++ b/Wire-iOS Tests/ConversationCellActionControllerTests.swift
@@ -1,0 +1,42 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


Wire-iOS Tests/ConversationCellActionControllerTests.swift
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class ConversationCellActionControllerTests: CoreDataSnapshotTestCase {
```
typfel: This can be a normal XCTestCase?

alexaubry: I'm using `otherUser` and `otherUserConversation` for building the message.

PR #2857, opened 2018-11-07
ZIOS-10817: Change behavior of fitting video and image messages

alexaubry:
### Issues

Following @typfel's remarks , we decided to change the behavior of the preview thumbnail for image and video assets inside a quote.

- For images: max height = 140, use the protobuf to determine aspect ratio, fit image
- For videos: use 16:9 aspect ratio, aspect fill the image

### Solutions

We add a separate protocol for images resources that have a size property (so we don't interfere with the link preview image resource type). We provide the content size and content mode for the image view and set those when we configure the quote view.


Wire-iOS/Sources/Helpers/syncengine/ImageResource.swift
```diff
@@ -67,13 +67,21 @@ struct LinkPreviewImageResourceAdaptor: ImageResource {
...
     var cacheIdentifier: String? {
         return fileMesssageData.imagePreviewDataIdentifier?.appending("-file")
     }
+
+    var contentMode: UIView.ContentMode {
+        return .scaleAspectFill
+    }
+
+    var contentSize: CGSize {
+        return CGSize(width: 250, height: 140)
```
billypchan: I think this number should be linked to the device's width. For iPhone Xs Max/iPad it should provide a large size.

alexaubry: According to the spec, the maximum height for the image is 140 for all devices, so we need to keep this width in order to maintain the 16:9 aspect ratio.

alexaubry: But maybe we can reduce the width of the quote bubble on iPad, I'll talk with design,

PR #2858, opened 2018-11-07
Message toolbox timer

mikeger:
### Issues

Implemented auto-updating countdown timer for the remaining timestamp for the ephemeral message. This one is shown inside the MessageToolboxView.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
@@ -60,6 +60,8 @@ import TTTAttributedLabel
     
     fileprivate var forceShowTimestamp: Bool = false
     private var isConfigured: Bool = false
+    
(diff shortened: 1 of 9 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
```diff
@@ -168,6 +170,19 @@ import TTTAttributedLabel
             self.configureTimestamp(message, animated: animated)
             self.tapGestureRecogniser.isEnabled = false
         }
+        
+        if message.shouldShowDestructionCountdown && timestampTimer == nil {
+            timestampTimer = Timer.scheduledTimer(withTimeInterval: 0.1, repeats: true) { [weak self] _ in
```
typfel: Should we stop the timer if the view is removed from the window?

mikeger: @typfel updated

PR #2859, opened 2018-11-07
Fix selection, menu and 3D Touch actions for cells

alexaubry:
### Issues

*This ticket covers ZIOS-10800, ZIOS-10805 and ZIOS-10814.*

After the cell refactoring, we needed to restore cell highlight upon selection, fix the order of the menu items for the legacy cells, fix the list of 3D Touch actions, and stop dismissing the keyboard when long pressing a cell.

### Solutions

- Port the legacy cells to use the new message action controller
- Observe appropriate menu events in the new cell to select and deselect the cell
- Update the new cell views to report their selectable area
- Use the action controller to generate the list of 3D Touch actions
- Disable drag and drop for the text message cell links to support 3D Touching them


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConversationCell+Menu.swift
```diff
@@ -19,12 +19,10 @@
 import UIKit
 
 public extension ConversationCell {
-    
+
     @objc public func showMenu() {
```
typfel: The legacy cells doesn't use `showMenu` method in the cell adapter?

typfel: Ah sorry I guess they aren't wrapped in the adapter.

alexaubry: Yeah that's correct, the legacy cells aren't wrapped.

PR #2860, opened 2018-11-07
Reply composer fixes

mikeger:
### Issues

- Fixed issue with the thumbnail for the panorama image.
- Fixed issue when starting the second reply while replying.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Reply.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Reply.swift
@@ -22,16 +22,24 @@ extension ConversationInputBarViewController: ReplyComposingViewDelegate {
     
     @objc(replyToMessage:composingView:)
     func reply(to message: ZMConversationMessage, composingView: ReplyComposingView) {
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Reply.swift
```diff
@@ -22,16 +22,24 @@ extension ConversationInputBarViewController: ReplyComposingViewDelegate {
...
-    func composingViewDidCancel(composingView: ReplyComposingView) {
+    private func removeReplyComposingView() {
         self.quotedMessage = nil
+        self.replyComposingView?.removeFromSuperview()
         self.replyComposingView = nil
-        composingView.removeFromSuperview()
+    }
+    
+    func composingViewDidCancel(composingView: ReplyComposingView) {
+        removeReplyComposingView()
```
alexaubry: Will this be animated still?

mikeger: No, for the moment there are no animations whatsoever. If we add the animations we'll need to think what should happen when the new reply is overtaking the old one.

PR #2862, opened 2018-11-07
Direct Replies

mikeger:
### Known issues

- Some timers in the conversation view are not yet hooked up.
- Some tests are not re-enabled yet.


Wire-iOS Tests/ConversationCell/TextMessageCellTests.swift
```diff
@@ -18,304 +18,305 @@
...
-extension TextMessageCell {
-
-    func prepareForSnapshot() -> UIView {
-        self.backgroundColor = .clear
-        return self.wrapInTableView()
-    }
-
-}
 
+// TODO 2838: - Migrate to new cell system
```
vytis: Some commented out code

Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
```diff
@@ -22,196 +22,159 @@ import WireDataModel
...
-        createSUT(for: .light)
-        let messageText = "Hello @Bill 👨‍👩‍👧‍👦 & @🏴󠁧󠁢󠁷󠁬󠁳󠁿🀄︎🧘🏿‍♀️其他人! I had some questions about your program. I think I found the bug 🐛."
-        let mention1 = Mention(range: NSRange(location: 6, length: 17), user: selfUser)
-        let mention2 = Mention(range: NSRange(location: 26, length: 28), user: otherUser)
-        let message = otherUserConversation.append(text: messageText, mentions: [mention1, mention2], fetchLinkPreview: false)
-
-        sut.configure(for: message, layoutProperties: layoutProperties)
-        verify(view: sut.prepareForSnapshot())
-    }
+//    /// "Saturday, February 14, 2009 at 12:20:30 AM Central European Standard Time"
```
vytis: Commented out here as well

Wire-iOS Tests/MessageDeletedCellTests.swift
```diff
@@ -20,53 +20,54 @@
...
-        cell.configure(for: message, layoutProperties: layoutProperties)
-        
-        let size = cell.systemLayoutSizeFitting(CGSize(width: 375, height: 0), withHorizontalFittingPriority: .required, verticalFittingPriority: .fittingSizeLevel)
-        cell.bounds = CGRect(x: 0.0, y: 0.0, width: size.width, height: size.height)
-        cell.setNeedsLayout()
-        cell.layoutIfNeeded()
-    }
-
-}
+// TODO 2838: - Migrate to new cell system
```
vytis: Commented out

Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
mikeger: @vytis this is mentioned in the description: 
> Some tests are not re-enabled yet.

vytis: It might be tricky to find the commented out code, maybe we can disable then in scheme settings? Or do some ifdef magic to ignore it if it does not compile?

mikeger: I will try to re-enable them now, don't feel good merging like that in dev.

Wire-iOS/Sources/Helpers/PerformanceDebugger.swift
```diff
@@ -0,0 +1,64 @@
...
+    private let log = ZMSLog(tag: "Performance")
+    private var displayLink: CADisplayLink!
+
+    override init() {
+        super.init()
+        displayLink = CADisplayLink(target: self, selector: #selector(handleDisplayLink))
+    }
+
+    deinit {
+        displayLink.remove(from: .main, forMode: .default)
```
vytis: `invalidate()` is more fool-proof - it removes display link from all run loops. Also there is a retain cycle here, because `CADisplayLink` retains the target strongly.

Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
mikeger: After some investigation/thinking:
Before we used to have the cell per message, so cell had to contain all the sender name, likes, timestamp, countdown timer, link preview, etc. etc.; therefore we had to have the general tests for every case for each message, so the image message is showing correctly the sender name, but also the text message must be tested for the same case, so on and so forth.

With the new paradigm we are moving form using one cell per message to use multiple cells per message. This would allow us to refactor the tests to test more specific aspect of every cell, where for example the sender is the separate cell. So we would only need to write one set of tests to verify that we are showing the sender correctly, and it would be sufficient for all types of messages.

From this point it looks like we can remove all the existing cell tests, since they are now not mapping 1-1 to the new cells structure.

@alexaubry @typfel thoughts / comments?

Wire-iOS/Sources/Helpers/PerformanceDebugger.swift
mikeger: @alexaubry can you fix this one quickly before we merge the stuff in?

Wire-iOS Tests/ConversationCell/TextMessageMentionsTests.swift
typfel: Yes, I think this makes sense, however the the rules for if the sender/toolbox/burst timestamp is visible has special rules for certain messages. But maybe it's better to test those cases using a unit tests instead of snapshots.

PR #2867, opened 2018-11-08
Restore search result highlight and flashing

alexaubry:
### Issues

- Restore flashing cells when the user reveals a message from the collection
- Restore highlighting the text search results
- Flash the cell when the user taps the reply quote


mikeger:
@alexaubry please try to rebase it on dev, now the diff seems off

alexaubry:
@mikeger I merged develop. Is that enough, or should I rebase?

mikeger:
Seems OK, commit history is _WASTED_, but we squash when merging anyways.

PR #2868, opened 2018-11-08
ZIOS-10831: Restore tapping on mention URL

alexaubry:
### Issues

It was not possible to open mentions or links from text messages.

### Solutions

This feature was a WIP item from the refactor. The text message cell is now the interaction delegate of the message text view, and makes appropriate calls to open mentions or links.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Text/ConversationTextMessageCell.swift
```diff
@@ -102,6 +129,13 @@ class ConversationTextMessageCellDescription: ConversationMessageCellDescription
         configuration = View.Configuration(attributedText: attributedString)
     }
 
+    func makeCell(for tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
```
mikeger: Not sure that I can see where this method is called from.

alexaubry: This is an override of the default implementation for this method in the cell description protocol. It's called by the section controller when creating the cell before it is configured.

mikeger: Ok, I see now!

PR #2875, opened 2018-11-09
ZIOS-10819: Dismiss keyboard when the conversation becomes read only

alexaubry:
### Issues

When replying to a message, and the user is removed from a conversation, the keyboard and the reply composer were still displayed.

### Solutions

When the conversation becomes read only, we resign first responder on the input bar and remove the composing view.


mikeger:
We need to potentially remove the mentions suggestions list, if visible.

alexaubry:
@mikeger I will add this too

alexaubry:
@mikeger Updated. When testing, I realized that if the last character in the draft is an `@`, we don't show the mentions suggestion when the user taps the text field initially, so I also fixed this.

PR #2876, opened 2018-11-09
Jump to the conversation end when sending the message

mikeger:
### Issues

In case we see the message from the current user appearing in the message window, the table view is scrolling to the very end of the conversation.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter+DataSource.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter+DataSource.swift
@@ -33,6 +33,12 @@ extension ConversationMessageWindowTableViewAdapter: ConversationMessageSectionC
     
 }
 
+extension MessageWindowChangeInfo {
+    var insertedMessages: [ZMConversationMessage] {
+        return self.insertedIndexes.compactMap { conversationMessageWindow.messages[$0] as? ZMConversationMessage }
+    }
+}
+
 extension ConversationMessageWindowTableViewAdapter: ZMConversationMessageWindowObserver {
     
     func reconfigureSectionController(at index: Int, tableView: UITableView) {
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter+DataSource.swift
```diff
@@ -46,10 +52,14 @@ extension ConversationMessageWindowTableViewAdapter: ZMConversationMessageWindow
         
         let isLoadingInitialContent = messageWindow.messages.count == changeInfo.insertedIndexes.count && changeInfo.deletedIndexes.count == 0
         let isExpandingMessageWindow = changeInfo.insertedIndexes.count > 0 && changeInfo.insertedIndexes.last == messageWindow.messages.count - 1
-        
+        var shouldJumpToTheConversationEnd = false
         stopAudioPlayer(forDeletedMessages: changeInfo.deletedObjects)
         
-        if isLoadingInitialContent || (isExpandingMessageWindow && changeInfo.deletedIndexes.count == 0) || changeInfo.needsReload {
+        if changeInfo.insertedMessages.any({ $0.sender?.isSelfUser ?? false }) {
```
typfel: Won't this jump to the end of you send a message from another device as well?

mikeger: This is bothering me as well, but not 100% percent sure it's something that can be bad for users.

typfel: Yes, maybe it's ok. What's the alternatives, trigger the scroll when inserting the message?

mikeger:
@typfel updated with the nicer solution, which can be condensed to 1 line change:
https://github.com/wireapp/wire-ios/pull/2876/files#diff-d0adf7620aab7082451935f800968beaR748

Basically when user clicks the send button we are bringing the input field into focus in order to finish the autocomplete. It was just broken because the call to scroll to the bottom with parameter animated true was somehow not happening to do anything.

mikeger:
@typfel please review when you have time.

PR #2877, opened 2018-11-09
ZIOS-10755: Fix hover state of the reply quote and composer

alexaubry:
### Issues

- When the user has the finger on the reply quote, we were using the wrong color
- When the user has the finger the reply composer, it was not highlighted 
- When the user tapped on the reply composer, the message was not highlighted after scrolling 

### Solutions

- Instead of setting the background color of the reply composer, we display a view on top
- We use touchesBegan/touchesEnded in the composer to detect the highlight instead of the tap gesture recognizer
- When scrolling from the reply composer, we use the scroll method that provides a completion handler, to flash the background of the message cell


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ReplyComposingView.swift
```diff
@@ -140,15 +139,30 @@ final class ReplyComposingView: UIView {
...
+
+    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        messagePreviewContainer.setHighlighted(true, animated: false)
     }
-    
-    @objc func onTap() {
+
+    override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
+        defer {
+            messagePreviewContainer.setHighlighted(false, animated: true)
```
mikeger: It could be slightly easier to convert the `messagePreviewContainer` to be button or to be hosted by the button; then assign the color for `highlighted` state, then it would be also semantically better, since the thing is clickable.

alexaubry:
@mikeger I updated the `ReplyRoundCornersView` to be a `UIControl`, so that it's in charge of the highlighting itself, and that the views that use it can just add a target for the `touchUpInside` event. What do you think?

mikeger:
@alexaubry seems good, one last thing I wanted to check, should the highlight be same on dark and light theme?

alexaubry:
@mikeger Good call, I messaged Eike and he asked for some changes.

alexaubry:
@mikeger Updated. We now remove the border when the cell is highlighted, and use a different color in dark mode.

PR #2878, opened 2018-11-09
Fix crash after editing message

typfel:
### Issues

If a message was edited while showing the timestamp (message is selected) it would crash due to a tableview update exception.

### Causes

When editing a message we change the nonce which means we will re-create the section controller since they are cached by nonce. The newly created section controller would be missing the timestamp cell since the selection was configured afterwards and thus crash.

### Solutions

Add the selection state to the constructor of the section controller.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter+DataSource.swift
```diff
@@ -83,7 +83,7 @@ extension ConversationMessageWindowTableViewAdapter: ZMConversationMessageWindow
     func reconfigureVisibleSections() {
         tableView.beginUpdates()
         if let indexPathsForVisibleRows = tableView.indexPathsForVisibleRows {
-            let visibleSections = indexPathsForVisibleRows.map({ $0.section })
+            let visibleSections = Set(indexPathsForVisibleRows.map({ $0.section }))
```
alexaubry: Can we use `.map(\.section)` here?

PR #2885, opened 2018-11-09
Fix: Replied quote message trim to 4 lines

billypchan:
### Issues

When a markdown list with more than 4 lines is quoted, the last line's line spacing is compressed and missing a trailing ellipsis.

### Causes

When an attribute text is truncated in inside a UITextView and the last line is shorter than its width, no ellipsis will be added.

### Solutions
Change the paragraph line breaks for to truncate tail, the long lines show ellipsis at the end.
Trim the attributed text and added an ellipsis at the end of the fourth line.

Before the fix, the line 3 to too close to the line 2 and no ellipsis is show:

step1

Replace the paraghaphy line break style to truncateTail, the ellipsis is show :

step2

Remove the lines other than first 4 lines, fixed the line spacing issue:

wire-ios tests_referenceimages_64_wire_ios_tests conversationreplycelltests_testthatitrendersmarkdownwith5linesandforthlineclipped_320 2x_before_bf3331ec5baf2cb9d06cba6f46f14c8e49a64d4b

## Notes

This solution has not yet considered the case that the cell's width varies during runtime, e.g. the user changes the app's width on iPad. The better solution is calculating the number of lines needed to render the attribute text and trim down the text to 4 lines.

## Dependency
- [x] https://github.com/wireapp/Down/pull/13


typfel:
I'm a bit bit wary adding this special formatting to fix layout bugs. I think we should consider just adding  the tail truncate option to `defaultPreviewParagraphStyle` and live with the fact the spacing it a bit off.

typfel:
@billypchan will you update this PR to add the tail truncate option to the `defaultPreviewParagraphStyle` and remove the new line formatting?

typfel:
Ah ok, but I don't want to add this kind of ad hoc logic to the markdown parsing. Maybe you and @johnxnguyen could sit down together and see if this line spacing issue could be fixed inside the `Down` framework. Perhaps there are some other paragraph style properties which can be changed to avoid the line spacing issue.

PR #2886, opened 2018-11-09
Fix: Link previews are compressed vertically

billypchan:
### Issues

Link preview's image is shorter than the current public release version.

### Causes

It seems that the current release version the image's height does not obey the constraint's constant. The rewriting of the conversation cell corrected this behaviour. 

### Solutions

Measure the current version's image size and set the height's constant.

## Notes

- [x] I will check ArticleView's constraints to see any conflicting constraints exist.


mikeger:
@billypchan are designders good with it?

billypchan:
@mikeger: This PR restores the original look of the cell on iPhone. I will confirm the iPad layout with design team again.

PR #2887, opened 2018-11-09
Add ephemeral timer for new cells

typfel:
## What's new in this PR?

We weren't displaying the ephemeral timer for the new cells (text, pings, ...). 

I decided to put the timer directly in the `ConversationMessageCellTableViewAdapter` which has the advantage the the content cells will get the timer without any modification except for an  optional `ephemeralTimerTopInset: CGFloat` which can be overridden to adjust the vertical position of the timer.


mikeger:
Not 100% sure why it’s necessary to run the display link, if you have the CAAnimation?

typfel:
I just ported the old code from `ConversationCell`, you're right it seems unnecessary.

typfel:
@mikeger updated

PR #2889, opened 2018-11-12
Only add a text message cell if the message has text content to display

typfel:
### Issues

Text messages containing only a link would display a link preview and an empty gap above it.

### Causes

We we always create a text message cell even if there's not text content to display.

### Solutions

Conditionally create text message cell.


mikeger:
@typfel somehow the build is failed.

PR #2892, opened 2018-11-12
fix: Automation AccessibilityId: likeButton is missing

billypchan:
## What's new in this PR?

Set MessageToolboxView's isAccessibilityElement property to false to allow like button can be accessed.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
```diff
@@ -66,7 +65,7 @@ import TTTAttributedLabel
     override init(frame: CGRect) {
         
         super.init(frame: frame)
-        self.isAccessibilityElement = true
+        isAccessibilityElement = false
```
mikeger: @alexaubry @billypchan would this be still OK for the actual VoiceOver users?

billypchan: The MessageToolboxView's accessibilityValue will be set to its subview's statusLabel's value in several places:
https://github.com/wireapp/wire-ios/blob/de6cc3a42ff36149c38e21384e021796be780430/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift#L419

After this fix the status label is accessible and I updated its accessibility value in the commit below.

alexaubry: @mikeger I will test on device

typfel: @alexaubry @billypchan  any update on this?

alexaubry:
@mikeger @typfel This sounds good on device. We will need to revisit the accessibility behavior of the message toolbox later anyway, as there is no way to show it in Voice Over at the moment.

PR #2896, opened 2018-11-13
Fix: Ping cell in dark mode snapshot test

billypchan:
## What's new in this PR?

Add snapshot for ping cell in dark mode.


typfel:
This is the old ping cell tests which should be deleted. The new ones are in `ConversationPingMessageTest`.

PR #2897, opened 2018-11-13
Fixes for handling links in reply messages and cells

alexaubry:
### Issues

We identified issues with links during review:

1. If the remote sender sends a link preview for a YouTube link, we'd show both the link preview and the media preview
2. The URL is shown twice in the reply composer and cell
3. If you long press a markdown link and lift your finger while the menu is visible, the system will ask to open the URL

### Solutions

1. Only show the link preview if the link attachment is of type `none`
2. Do not remove the trailing URL for link previews
3. Do not consider the tap to be inside the link text view if the menu is visible, to allow the user to dismiss it by tapping anywhere


mikeger:
@alexaubry the description contains numerous typos, can you re-check it?

alexaubry:
@mikeger done

PR #2898, opened 2018-11-13
Re-enable message snapshot tests

typfel:
## What's new in this PR?

- Fix ping tests not using correct spacing
- Don't rely on a dummy tableview when snapshotting messages, instead we put the cell view into a stack view.
- Re-add text message test
- Re-enable text message mention tests


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageCell.swift
```diff
@@ -136,10 +137,34 @@ extension ConversationMessageCellDescription {
     func didEndDisplayingCell() {
         
     }
-
+
```
alexaubry: Trailing whitespace

```suggestion

```

alexaubry:
@typfel Do we want to re-add the ping cell animation?

typfel:
@alexaubry let's discuss with design

PR #2899, opened 2018-11-13
ZIOS-10799: Added loading of quotes from draft messages

nicorsm:
## What's new in this PR?

Quotes weren't loading correctly when saved inside the `DraftMessage` object after switching the conversation or at startup.

### Needs release with

- [x] https://github.com/wireapp/wire-ios-data-model/pull/601


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -194,6 +194,20 @@ - (void)viewDidLoad
     [self.view addGestureRecognizer:pinchImageGestureRecognizer];
     
     [self createMentionsResultsView];
+    
+}
+
+- (void)loadQuotedMessageIfNeeded {
+    
+    NSUUID *nonce = self.conversation.draftMessage.quote.nonce;
```
typfel: Can't the loading of the message be done in data model?

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -194,6 +194,20 @@ - (void)viewDidLoad
...
+    NSManagedObjectContext *context = self.conversation.managedObjectContext;
+    ZMMessage *quote = [ZMMessage fetchMessageWithNonce:nonce
```
mikeger: May be this is better to do on the SE/DM side? Or as the extension. Like `self.conversation.draftMessage.quotedMessage?`. I think it's not the best to use the `NSManagedObjectContext` in UI directly.

typfel: Reading the data model code it looks like we are already loading the message.

typfel:
@nicorsm let's bump data model and merge this :)

PR #2900, opened 2018-11-13
Fix/dark mode bugs

johnxnguyen:
## What's new in this PR?

Various dark mode bug fixes:

### Issues

1. Ping cell text was dark on dark.
2. Share preview text was light on light.
3. Ephemeral keyboard picker background was light on dark.

### Causes

1. The attributed text didn't set the foreground color explicitly.
2. The background of the share view is explicitly set to white, but the content still obeys the color scheme.
3. Not entirely sure... the picker background color is set to clear during initialisation, but doesn't seem to take effect.

### Solutions

1. Use the color scheme.
2. In this case, force the color scheme variant to be `light` before rending the preview view, then restore the variant value afterwards.
3. Set the background color to clear after initialization.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -100,7 +100,13 @@ extension ZMMessage: Shareable {
 
 extension ZMConversationMessage {
     public func previewView() -> UIView? {
-        guard let view = self.preparePreviewView(shouldDisplaySender: false) else { return nil }
+        // the preview view will have a white bg color, so we need to
```
mikeger: @typfel was there a way to set the color scheme via UIAppearance?

typfel: Yes! Implement `Themeable` and mark it as `@dynamic`. 

examples: 
see: `ZClientViewControlller.setupAppearance` and `SeparatorCollectionViewCell`

Wire-iOS Tests/ConversationCell/PingCellSnapshotTests.swift
```diff
@@ -23,20 +23,20 @@ final class PingCellSnapshotTests: ZMSnapshotTestCase {
...
         sut = nil
-
+        ColorScheme.default.variant = .light
         super.tearDown()
     }
-
+    
     func testForYouPinged() {
+        sut = createPingCellForTest()
+        verify(view: sut.tableViewForSnapshot())
```
vytis: Would be good to have some background color in the snapshot

typfel: This is testing the old ping cells which is not in use anymore. We can delete this test class.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Forward.swift
```diff
@@ -100,15 +100,15 @@ extension ZMMessage: Shareable {
...
-        // temporarily force a light color scheme for all subviews.
-        let variant = ColorScheme.default.variant
-        defer { ColorScheme.default.variant = variant }
-        ColorScheme.default.variant = .light
-        
         let view = self.preparePreviewView(shouldDisplaySender: false)
         view.translatesAutoresizingMaskIntoConstraints = false
         view.backgroundColor = .white
+
+        if var themeable = view as? Themeable {
```
typfel: I thought you would use one of the `appearanceWhenContainedInInstancesOfClasses`

johnxnguyen: Ah right.. so it's going to be contained in a `UIView`. I suppose then I'd need to create a new view class as a container?

typfel: No it can also be UIViewController nowadays

https://developer.apple.com/documentation/uikit/uiappearance/1615013-appearance

PR #2904, opened 2018-11-13
Always show the message just sent

mikeger:
### Issues

We want to always show the message that was just sent from this device, even if user was looking on the older messages in the conversation history.

### Solutions

Message window observer is going to tell us if the new message is coming into the window, so we can check if this is the message that was just inserted.

Unfortunately the table view is showing the really weird behaviour w.r.t. scrolling position. Even if the position was set from the code, when some other part of the code is trying to enumerate the cells visible the table view layout is kicking in and moving the viewport back to the initial position.

The original fix is to blindly block the table view to accept any scroll position for the period of time.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.h
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.h
@@ -26,6 +26,7 @@
 @class ConversationMessageSectionController;
 @class AnyConversationMessageCellDescription;
 @class ConversationCellActionController;
+@class UpsideDownTableView;
 
 @interface ConversationMessageWindowTableViewAdapter : NSObject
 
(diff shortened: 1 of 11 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.h
```diff
@@ -37,7 +38,7 @@
 
 @property (nonatomic) NSArray<NSString *> *searchQueries;
 
-- (instancetype)initWithTableView:(UITableView *)tableView messageWindow:(ZMConversationMessageWindow *)messageWindow;
+- (instancetype)initWithTableView:(UpsideDownTableView *)tableView messageWindow:(ZMConversationMessageWindow *)messageWindow;
```
alexaubry: What is the benefit of passing `UpsideDownTableView` directly?

mikeger: It needs to have something that implements `lockContentOffset`.

PR #2907, opened 2018-11-14
WIP: Snapshot refactoring

billypchan:
### Issues

Refactoring snapshot test with these updates:

- remove view size in the snapshot file names and use the device instead.
- remove iPhone 3.5 inch snapshots
- add new snapshot size for 2018 new iPhones
- remove 3rd party layout constraints frameworks
- for verifying snapshot in all iPhones' width, test for widths 320, 375 and 414 pt and the suffix does not changes
- remove objc old style multiple interfaces of verify method. Refactor with Swift method with default argument values.

### TODO:
- [x] some of the snapshot tests cannot produce correct reference images when regenerating all snapshots. Maybe they are affected by other tests. I will run them separately again.
- [ ] some tests involve drawings produces snapshots with some subtle difference in runs. I will try to add tolerance argument to fix it.


Wire-iOS Tests/Utils/ZMSnapshotTestCase.h
```diff
@@ -21,6 +21,40 @@
...
+static CGSize const ZMDeviceSizeIPhone5 = (CGSize){ .width = 320, .height = 568 };
+static CGSize const ZMDeviceSizeIPhone6 = (CGSize){ .width = 375, .height = 667 };
+static CGSize const ZMDeviceSizeIPhone6Plus = (CGSize){ .width = 414, .height = 736 };
+static CGSize const ZMDeviceSizeIPhoneX = (CGSize){ .width = 375, .height = 812 };
+static CGSize const ZMDeviceSizeIPhoneXR = (CGSize){ .width = 414, .height = 896 };
+
+static CGSize const ZMDeviceSizeIPadPortrait = (CGSize){ .width = 768, .height = 1024 };
+static CGSize const ZMDeviceSizeIPadLandscape = (CGSize){ .width = 1024, .height = 768 };
+
+//static NSArray<NSValue *> * _Nonnull phoneSizes(void) {
```
mikeger: Can be removed?

Wire-iOS Tests/Utils/ZMSnapshotTestCase.m
```diff
@@ -193,34 +159,57 @@ - (UIView *)containerViewWithView:(UIView *)view
...
+              line:(NSUInteger)line
+        identifier:(NSString *)identifier
+        deviceName:(NSString *)deviceName
 {
     UIView *container = [self containerViewWithView:view];
     if ([self assertEmptyFrame:container file:file line:line]) {
         return;
     }
-    NSString *finalIdentifier = @"";
+    NSString *finalIdentifier = @""; ///TODO: replace size with  device type
```
mikeger: Do we need this TODO?

billypchan: It is cleared. Thanks!

billypchan:
@mikeger after refactoring layout constraint in ZMSnapshotTestCase I found the some snapshot produced has pixel offset. I will close this ticket and split this task into subtasks for renaming and refactoring ZMSnapshotTestCase.

PR #2911, opened 2018-11-14
ZIOS-10798: Use correct color when generating message preview

alexaubry:
### Issues

When the user was in dark mode, and sharing a media message, the media icon (location, image, video, file...) was shown white on white.

### Causes

The color of the image attachments was not updated when the override color variant changes.

### Solutions

Reload the entire attributed string with the override color variant when it changes.


alexaubry:
@mikeger I can add some.

PR #2914, opened 2018-11-14
ZIOS-10833: Fix accessibility identifier grouping for elements

alexaubry:
### Issues

The contents of the reply cell and composer could not be accessed by accessibility automation.

### Causes

1. The reply composer box was grouping the close button and the preview.
2. The individual contents of the reply cell were not accessible, because the rounded reply view was grouping the elements under the `ReplyCell` identifier.

### Solutions

1. Make the close button and the preview independent elements in the composer.
2. Set the accessibility properties (ID, label) on the table adapter cell, rather than on the contained view.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios/pull/2892


Wire-iOS Tests/ConversationMessageCell/MockCell.swift
```diff
@@ -47,6 +47,9 @@ class MockCellDescription<T>: ConversationMessageCellDescription {
     weak var delegate: ConversationCellDelegate?
     weak var actionController: ConversationCellActionController?
 
+    var accessibilityIdentifier: String?
```
mikeger: Very boring changes, may be we can provide the protocol default implementation?

alexaubry: Checked 1:1, we could revisit all the properties that can have a default implementation later.

PR #2915, opened 2018-11-14
ZIOS-10871: Scroll to top of message when opening quote

alexaubry:
### Issues

When the user was quoting a long message and they tapped the composer, we would scroll to the middle of the cell, which does not show the beginning of the original message.

### Solutions

Scroll to the top of the message.


mikeger:
🔝

PR #2916, opened 2018-11-14
Fix logic for when the sender and burst timestamp are displayed

typfel:
## What's new in this PR?

- Unread dot was not always displayed for unread messages
- Burst timestamp was not always displayed for the first message of the day
- Sender should always be displayed if the burst timestamp is displayed.


alexaubry:
@typfel Did you also mention in the ticket that we need to update the logic for when the avatar is shown?

typfel:
@alexaubry yep I mentioned it.

PR #2918, opened 2018-11-14
ZIOS-10877: Fix position of scroll when opening a conversation with unread messages

alexaubry:
### Issues

When the user opened a conversation with unread messages, we'd scroll them to the top of the last cell in the first unread message section (ex: the link preview) instead of the top of the first cell in the section (ex: the burst timestamp). 

### Solutions

Fix the index of the cell we scroll to in the section to be the last cell's index, because the table view is upside down.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.m
@@ -65,7 +65,8 @@
 {
     NSUInteger index = [self.messageWindow.messages indexOfObject:message];
     if (index != NSNotFound) {
-        [self.tableView scrollToRowAtIndexPath:[NSIndexPath indexPathForRow:0 inSection:index]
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.m
```diff
@@ -65,7 +65,8 @@ - (BOOL)scrollToMessage:(id<ZMConversationMessage>)message animated:(BOOL)animat
 {
     NSUInteger index = [self.messageWindow.messages indexOfObject:message];
     if (index != NSNotFound) {
-        [self.tableView scrollToRowAtIndexPath:[NSIndexPath indexPathForRow:0 inSection:index]
+        NSInteger rowIndex = [self.tableView numberOfRowsInSection:index] - 1;
```
mikeger: This can break easily if we change the logic, may be we can hook it up somehow to the data source?

alexaubry: Checked 1:1

PR #2922, opened 2018-11-15
Port image message cell to new cell setup

typfel:
## What's new in this PR?

Port image message cell to new cell setup


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageCellTableViewAdapter.swift
```diff
@@ -186,16 +191,32 @@ class ConversationMessageCellTableViewAdapter<C: ConversationMessageCellDescript
...
     }
 
     // MARK: - Standard Actions
+    
+    private func presentDetails() {
+        guard cellDescription?.supportsActions == true, let message = cellDescription?.message else {
+            return
+        }
+        
+        cellDescription?.delegate?.conversationCell?(cellView, didSelect: .present, for: message)
```
alexaubry: This probably will cause problems with file cells, as they use a single tap gesture to download files. I would suggest to add a method on the action controller, e.g. `actionForSingleTap() -> MessageAction` that would return the action to execute depending on the state of the message. For example, it would check if the file is downloaded, and return either the present or download action.

typfel:
@alexaubry updated

PR #2923, opened 2018-11-15
Fix missing edited icon accessibility identifier

alexaubry:
### Issues

The "Edited" icon was not visible to Voice Over and automation inside a quote displaying an edited message.

### Solutions

Add an accessibility label and make the element accessible when the label is set.


mikeger:
@alexaubry you mentioned the magic constant that would make iOS pick the english translation when the version is missing?

PR #2932, opened 2018-11-19
ZIOS-10855: Fix like button behavior

alexaubry:
### Issues

1. We would always show the like button when the toolbox is displayed, even if the user didn't select it and it shows the delivery receipt.
2. We always used the "liked" icon even if the message is unliked, instead of the "like" outline button asked by design.

### Solutions

1. Hide the like button when the toolbox isn't manually selected by the user.
2. Change the icon depending on the liked state of the message.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
```diff
@@ -241,10 +239,39 @@ import TTTAttributedLabel
...
+        }
+
+        // Change State and Appearance
+        likeButton.setIcon(message.liked ? .liked : .like, with: .like, for: .normal)
+        likeButton.setIcon(.liked, with: .like, for: .selected)
         likeButton.setSelected(message.liked, animated: false)
         self.reactionsView.likers = message.likers()
+
+        // Animate Changes
+        UIView.animate(withDuration: needsAnimation ? 0.2 : 0, animations: changes, completion: completion)
```
billypchan: When needsAnimation is false, we can call changes and completion directly to prevent unnecessary animate call.  

`
if needsAnimation {
  UIView.animate(withDuration: needsAnimation ? 0.2 : 0, animations: changes, completion: completion)
} else {
  changes()
  completion()
}
`

alexaubry: Updated.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxView.swift
```diff
@@ -236,15 +234,56 @@ import TTTAttributedLabel
...
+
+        // Change State and Appearance
+        likeButton.setIcon(message.liked ? .liked : .like, with: .like, for: .normal)
+        likeButton.setIcon(.liked, with: .like, for: .selected)
         likeButton.setSelected(message.liked, animated: false)
         self.reactionsView.likers = message.likers()
+
+        // Animate Changes
+        if needsAnimation {
+            UIView.animate(withDuration: needsAnimation ? 0.2 : 0, animations: changes, completion: completion)
```
billypchan: "needsAnimation ? 0.2 : 0" can be simpified to "0.2"

PR #2937, opened 2018-11-20
Fix: Emoji substition is not work when it is at the tail

billypchan:
## What's new in this PR?

This PR is a rework of https://github.com/wireapp/wire-ios/pull/2840

### Issues

The message with emoticon string at the end is not replaced.

### Causes

In https://github.com/wireapp/wire-ios/pull/2840 I mixed up the ClosedRange and Range and change the length of NSRange cause this new issue.

### Solutions

The real reason of https://github.com/wireapp/wire-ios/pull/2840 and this is the length of the string is shortened after each range is replaced. e.g.:

The original message string is:
```
"<3 Lorem @Ipsum Dolor }:-)Ame:) :)"
```
The @Ipsum is a mentioned name, it is excluded in the emoji substitution.

The original logic starts substitution from the beginning:

```
"❤ Lorem @Ipsum Dolor }:-)Ame:) :)"
```

The length of the string is shortened by one and the range of the mentioned name is no longer valid. It may access the element out of the string's length.

The solution is reverse the replacement order and start from the end of the string.

## Notes

- [ ] revise the emoticon substition and check how other platfrom works
- [ ] move emoticon substition code to Utilities to save building and testing time
- [ ] We have 2 tests using OCMock, should we retire it?


Wire-iOS Tests/EmoticonSubstitution/NSString+EmoticonSubstitutionTests.m
```diff
@@ -76,6 +76,24 @@ - (void)testThatSubstitutionInSpecificRangeWorks {
     XCTAssertEqualObjects(resolvedString, targetString);
 }
 
+- (void)testThatSubstitutionInTailRangeWorks {
+    // Given
+    NSString *targetString = @"<3 Lorem Ipsum Dolor 😈Ame😊 😊";
```
typfel: Why isn't the the `<3` substituted with ❤️ ?

typfel: Ah or we use a different substitution mapping in the test?

typfel:
So the `OCMock` is only used to replace the emoji replacement configuration which serves no purpose in the test? If so I would remove it.

billypchan:
@typfel Yes, I wrote a new test with OCMock which mocks EmoticonSubstitutionConfiguration. To replace it I have to write a new mock of EmoticonSubstitutionConfiguration.

PR #2941, opened 2018-11-21
Simplify build configuration 

vytis:
### Issues

We want to have more control on various ways to customize the app. We have several different .xcconfig files that are injected during Jenkins build and lets us customize Wire for testing. 

The configuration files were not touched in a long time and has a lot of redundancy and outdated flags. This PR improves the situation by:
* Moving all configuration options to .xcconfig files. From now on we should avoid adding any flags directly through build settings. It makes it hard to track what is the value of each setting when they are defined in multiple places.
* Organizing existing .xcconfig files in a cleaner structure.
* Adds a customization point for various settings.

#### Configuration files

They are located in `Wire-iOS/Resources/Configuration/`. There are several sets of configurations:

* Project. These are common for all targets and contains things like deployment version, SDK, Swift version etc.
* Per target. Things like Info.plist location, provisioning profile names and other flags for that target type.
* `Warnings.xcconfig`. All the warnings we have turned on are factored out into a separate file for better readability.
* `Version.xcconfig`. Contains the current Wire version number.

Because we have `Development` and `Release` configurations each set described above consists of 3 files:
* `-Base.xcconfig`. Contains settings common for both configurations.
* `-Development`. Contains development-only settings such as more warnings, less compiler optimizations etc.
* `-Release`. Release-only settings.

#### Customization points

Things that we want to customize are stored in a separate repository - https://github.com/wireapp/wire-ios-build-configuration. It is fetched when running `setup.sh`. It is located in `Configuration/` and for now contains only 3 .xcconfig files in a similar structure as above. It specifies app version, provision profiles, bundle ids etc. 

There are two ways to customize the app:
1. `./setup.sh --configuration_repo git@github.com:some/repo.git`. This will fetch the configuration files from the specified repository. This needs to be done only once, all subsequent runs of `./setup.sh` will only make `git pull` to that repository.
2. `./setup.sh --override_with some/path`. This will copy over contents of `some_path` to the `Configuration` directory. It should be used when we want to customize only some part of the configuration. In practice it will be used by CI or when trying to build over internal version of Wire with Xcode attached.

## Dependencies

To allow building by external people we need to open-source:

- [x] https://github.com/wireapp/wire-ios-build-configuration


typfel:
Just checking if I understand the change. We are making `wireapp/wire-ios-build-configuration` open source and it will now host the public configuration. If we want to have a private configuration we will create a separate repo for it?

vytis:
@typfel yes, that's the plan. We already have `wire-ios-build-assets` repo that is basically used for this, so we will have to adapt it with the structure in this PR.

vytis:
@typfel @mikeger updated the `download_assets.sh` script. "Customization points" section in the PR description explains how it will work.

PR #2943, opened 2018-11-21
WIP: Refactor snapshot

billypchan:
## What's new in this PR?

Refactoring ZMSnapshotTestCase in Swift and clean up the unnecessary code.

- remove objc old style multiple interfaces of verify method. Refactor with Swift method with default argument values.
- remove 3rd party layout constraints frameworks


billypchan:
I will fix the snapshot tests which produces results with pixel offset first in separate PR.

PR #2948, opened 2018-11-22
Fix: App crashes after choosing a phone number from

billypchan:
### Issues

Fixing crash caused by choosing autofill phone number/paste a phone number to ChangePhoneViewController.

### Causes

If a textField's keyboardType is .phonePad and textContentType = .telephoneNumber, the user is able to select a phone number from autoFill keyboard accessory and insert a phone number. The non numeric characters e.g. + or " " in the phone number would cause a crash in "main (main.m:29)
" without reason.

### Solutions

In PhoneNumberViewController we have a method to convert a phone number with country code to a number only format. Rewrite that method in Swift and apply it to ChangePhoneViewController.

The @optional signature of RegistrationTextFieldDelegate's method is removed. For the class do not implement its shouldPasteCharactersInRange method, conform to shouldPasteCharactersInRange instead.

## Notes

- [ ] check other textField with keyboardType is .phonePad and textContentType = .telephoneNumber has this issue or not.


Wire-iOS/Sources/Authentication/RegistrationPersonal/PhoneNumberViewController+Extension.swift
```diff
@@ -0,0 +1,36 @@
...
+    @objc
+    @discardableResult
+    func pastePhoneNumber(_ phoneNumber: NSString?) -> Bool {
+        guard let phoneNumber = phoneNumber else { return false }
+
+        return phoneNumber.shouldPasteAsPhoneNumber(presetCountry: self.country){country, phoneNumber in
+            if let _ /*country*/ = country, let phoneNumber = phoneNumber {
+
+                self.phoneNumberField.text = phoneNumber;
+                ///TODO: update country name and county code after a phone number with prefix is pasted
```
johnxnguyen: Todo

billypchan: @johnxnguyen This is for marking a known issue in the login/registration process. I will fix it after this PR is merged.

Wire-iOS/Sources/Authentication/RegistrationPersonal/RegistrationTextField.h
```diff
@@ -36,16 +36,14 @@ typedef NS_ENUM(NSUInteger, RegistrationTextFieldLeftAccessoryView) {
 
 
 @protocol RegistrationTextFieldDelegate <UITextFieldDelegate>
```
typfel: Can this protocol be deleted now that it's not exposed anymore `RegistrationTextFieldDelegate`

PR #2950, opened 2018-11-23
Revert configuration name back to Debug

vytis:
### Issues

We were having some issues after #2941 because configuration was renamed to Development. When building with other frameworks dragged in it would not be found because they are in "Debug" folder inside build products, but the app is being built in "Development"


typfel:
Shouldn't we rename the xcconfig file as well?

vytis:
@typfel yeah good point!

vytis:
@typfel updated!

typfel:
Submodule added?

vytis:
@typfel removed

PR #2952, opened 2018-11-23
Fix: when pasting a phone number with country code copied from phone app, the country is not updated

billypchan:
## What's new in this PR?

This PR is related to https://github.com/wireapp/wire-ios/pull/2948

### Issues

When a user pastes a phone number for login/registration, the country code is not updated.

### Causes

Missing a call to update the country code.

### Solutions

Add the missing call.

## Notes
The user can choose a phone number with autoFill after this PR.
Added unit test for inserting phone number.

## Todo

- [x] snapshot test
- [ ] consider moving NSString+PastePhoneNumber.swift and releted tests to Wire-Utilities


Wire-iOS Tests/String+PhoneNumberTests.swift
```diff
@@ -0,0 +1,102 @@
...
+        sut = "+41 86 079 209 36 37"
+
+        // WHEN
+        let presetCountry = Country(iso: "", e164: NSNumber(value: 49))
+        let result = sut.shouldInsertAsPhoneNumber(presetCountry: presetCountry) {
+            country, phoneNumber in
+            if let country = country, let phoneNumber = phoneNumber {
+                XCTAssertEqual(country.iso, "ch")
+                XCTAssertEqual(phoneNumber, "860792093637")
+            }
```
vytis: This test will pass if country and phoneNumber is nil - I think you can use the optional value inside `XCTAssertEqual` directly. Otherwise adding an `XCTFail()` in the else clause would help

Wire-iOS/Sources/Authentication/RegistrationPersonal/PhoneNumberViewController+Internal.h
```diff
@@ -17,7 +17,9 @@
 //
 
 @interface PhoneNumberViewController ()
+@property (nonatomic, readwrite) Country *country;
 
 - (void)updateRightAccessoryForPhoneNumber:(NSString *)phoneNumber;
+- (void)setCountry:(Country *)country;
```
vytis: No need for this method when the property is already `readwrite`

Wire-iOS/Sources/Authentication/RegistrationPersonal/PhoneNumberViewController+RegistrationTextFieldDelegate.swift
```diff
@@ -0,0 +1,71 @@
...
+        if textField.text?.count == 0 && string.count > 1 {
+            return insert(phoneNumber: string)
+        }
+
+        let number = PhoneNumber(countryCode: country.e164.uintValue, numberWithoutCode: newString)
+
+        switch number.validate() {
+        case .containsInvalidCharacters, .tooLong:
+            return false
+        case .tooShort:
```
vytis: If we don't do anything in this `case`, then handling it in `default` is fine as well

Wire-iOS/Sources/UserInterface/Settings/ChangingDetails/NSString+PastePhoneNumber.swift
```diff
@@ -18,29 +18,31 @@
...
         illegalCharacters.formUnion(CharacterSet.decimalDigits)
         illegalCharacters.formUnion(CharacterSet(charactersIn: "+-()"))
         illegalCharacters.invert()
         var phoneNumber: NSString = trimmingCharacters(in: illegalCharacters) as NSString
 
         if phoneNumber.trimmingCharacters(in: CharacterSet.whitespacesAndNewlines).hasPrefix("+") {
             if let country = Country.detect(forPhoneNumber: phoneNumber as String) {
-                let phoneNumberWithoutCountryCode = phoneNumber.replacingOccurrences(of: country.e164PrefixString, with: "")
+                /// remove the leading space and country prefix
+                let phoneNumberWithoutCountryCode = phoneNumber.replacingOccurrences(of: country.e164PrefixString, with: "").filter { !" ".contains($0) }
```
vytis: Maybe it's better to use `CharacterSet.whitespace` instead of directly comparing with " ". There are many different kinds of spaces out there!

Wire-iOS/Sources/UserInterface/Settings/ChangingDetails/NSString+PastePhoneNumber.swift
```diff
@@ -18,29 +18,31 @@
...
+    /// or  \u{e2}+49 123 12349999\u{e2}
     ///
-    /// - Parameter completion: completion closure with Country object and phoneNumber extracted from self
+    /// - Parameter completion: completion closure with Country object and phoneNumber extracted from self. country: a Country object parsed from self. phoneNumber: phone Number with no space
     /// - Returns: true if should paste as Phone number(not beginning with "+"). If self is prased as a phone number, reture false (it should the be pasted, the caller use the completion's data for further actions.)
     @discardableResult
-    @objc func shouldPasteAsPhoneNumber(presetCountry: Country,
-                                        completion: (_ country: Country?, _ phoneNumber: String?) -> Void)
-        -> Bool {
+    func shouldInsertAsPhoneNumber(presetCountry: Country,
```
vytis: Since this method is no longer user from Objective-C we can simply return an optional tuple of country and phone number. It's a bit weird in Swift to have a close and a Bool return when we have optionals

Wire-iOS/Sources/Authentication/RegistrationPersonal/PhoneNumberViewController+Internal.h
billypchan: The public header has defined the property as read-only so have to add an internal method for the setter.

vytis: But then above there is `@property (nonatomic, readwrite) Country *country;` which I think should be enough to have this

Wire-iOS/Sources/Authentication/RegistrationPersonal/PhoneNumberViewController+RegistrationTextFieldDelegate.swift
```diff
@@ -0,0 +1,72 @@
...
+        if textField.text?.count == 0 && string.count > 1 {
+            return insert(phoneNumber: string)
+        }
+
+        let number = PhoneNumber(countryCode: country.e164.uintValue, numberWithoutCode: newString)
+
+        switch number.validate() {
+
+            
+        case .containsInvalidCharacters, .tooLong:
```
mikeger: Indent is off.

Wire-iOS Tests/String+PhoneNumberTests.swift
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    func testThatPhoneNumberWithSpaceIsParsed(){
+        // GIVEN
+        sut = "+41 86 079 209 36 37"
+
+        // WHEN & THEN
+        let presetCountry = Country(iso: "", e164: NSNumber(value: 49))
+
+        if let (country, phoneNumberWithoutCountryCode) = sut.shouldInsertAsPhoneNumber(presetCountry: presetCountry) {
```
vytis: The test will pass if `sut.shouldInsertAsPhoneNumber` returns nil - either avoid having `if let` or add else clause with `XCTFail()`

Wire-iOS/Sources/UserInterface/Settings/ChangingDetails/NSString+PastePhoneNumber.swift
```diff
@@ -18,45 +18,48 @@
...
-    /// When pastedString is copied from phone app (self phone number section), it contains right/left handling symbols: \u202A\u202B\u202C\u202D
-    /// @"\U0000202d+380 (00) 123 45 67\U0000202c"
+    /// Notice: When pastedString is copied from phone app (self phone number section), it contains right/left handling symbols: \u202A\u202B\u202C\u202D or \u{e2}
+    /// e.g. @"\U0000202d+380 (00) 123 45 67\U0000202c"
+    /// or  \u{e2}+49 123 12349999\u{e2}
     ///
-    /// - Parameter completion: completion closure with Country object and phoneNumber extracted from self
-    /// - Returns: true if should paste as Phone number(not beginning with "+"). If self is prased as a phone number, reture false (it should the be pasted, the caller use the completion's data for further actions.)
+    /// - Parameter completion: completion closure with Country object and phoneNumber extracted from self. country: a Country object parsed from self. phoneNumber: phone Number with no space
+    /// - Returns: If the number can be prased, return a tuple of country and the phone number without country code. Otherwise return nil. country would be nil if self is a phone number without country
```
vytis: typo - prased

Wire-iOS Tests/String+PhoneNumberTests.swift
billypchan: updated all tests as suggested, thanks!

Wire-iOS/Sources/Authentication/RegistrationPersonal/RegistrationTextFieldDelegate.swift
```diff
@@ -0,0 +1,23 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol RegistrationTextFieldDelegate: UITextFieldDelegate {
+    func textField(_ textField: UITextField?, shouldPasteCharactersIn range: NSRange, replacementString string: String?) -> Bool
```
vytis: A nitpick, but `textField` should probably be not optional. Not sure if there is a valid scenario where it could be nil.

billypchan: removed all uncertain optional arguments from objc era, thanks!

mikeger:
@billypchan what's the status of this PR?

PR #2953, opened 2018-11-23
Configure custom backend endpoints

vytis:
### Issues

We need a better way to set different kinds of backend URLs without changes in the code. This PR adds ability to load them from JSON in https://github.com/wireapp/wire-ios-build-configuration.

### Solutions

A `Backend.bundle` was added to `Configuration` folder that contains JSON with all URLs we use in the app. 

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/111
- [x] https://github.com/wireapp/wire-ios-share-engine/pull/72


WireExtensionComponents/Utilities/Bundle+Backend.swift
```diff
@@ -0,0 +1,36 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension Bundle {
+    static var backendBundle: Bundle {
```
mikeger: What's the reasoning behind making it a folder, i.e. bundle? You planning to have there more than only the URLs going forward?

PR #2956, opened 2018-11-25
[WIP] ZIOS-10761: Fix actions in ephemeral collection items

alexaubry:
### Issues

We were displaying forbidden ephemeral actions in the collection (copy, like, etc).

### Solutions

To reduce code duplication and make sure we enforce the same behavior everywhere we use a menu controller, we use the new conversation message action controller (introduced with the cell refactor) in the collection and in the full-screen image controller.

We also refactor the collection view controller to simplify the handling of actions. Previously, there were 3 different entry points for actions, and they were always handled the same way.

Finally, we add a utility method for copying the message, so that all contents are copied the same way in different areas of the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,4 @@ Scripts/**/.build
 /fastlane/report.xml
 /test
 /fastlane/README.md
-Configuration/**
+Configuration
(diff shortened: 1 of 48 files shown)


.gitignore
```diff
@@ -25,4 +25,4 @@ Scripts/**/.build
 /fastlane/report.xml
 /test
 /fastlane/README.md
-Configuration/**
\ No newline at end of file
+Configuration
```
typfel: Intentional?

PR #2958, opened 2018-11-26
ZIOS-10914: Missing accessibility identifier for audio messages

nicorsm:
### Issues

Accessibility identifier was missing from the "Play" button of the audio messages cell.

### Solutions

- Converted it (together with `downloadProgressView` and `playerProgressView`) to identifier
- Added accessibility label to `playButton`


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/FileTransfer/AudioMessageView.swift
```diff
@@ -54,6 +54,7 @@ private let zmLog = ZMSLog(tag: "UI")
         label.numberOfLines = 1
         label.textAlignment = .center
         label.accessibilityLabel = "AudioTimeLabel"
```
mikeger: I think we should remove this label, since it's not working for the accessibility.

nicorsm:
Tests are working fine locally, I'll merge the PR.

PR #2964, opened 2018-11-27
Feature: Add read receipt toggle in group details screen

billypchan:
## What's new in this PR?

Create a new cell with to UISwitch for toggling read receipts setting of the conversation.

## Dependencies

Needs releases with:

- [ ] Data model from `feature/read-receipts`


johnxnguyen:
Isn't the read receipt option not available in non teams? The new cell appears in both snapshots.

johnxnguyen:
It would be nice to have a little bit of a description about the motivation to separate the `GroupDetailsOptionCell` into two protocols.

Wire-iOS/Sources/UserInterface/Components/Views/DetailsCollectionViewCell.swift
```diff
@@ -21,7 +21,21 @@ import UIKit
 class DetailsCollectionViewCell: SeparatorCollectionViewCell {
 
     private let leftIconView = UIImageView()
-    private let accessoryIconView = UIImageView()
+    public var rightAccessoryView: UIView? {
+        didSet {
```
johnxnguyen: Why did this change? It creates a lot of casting into `UIImageView`.

billypchan: Yes, it would be better to create two kinds of cell base classes.

johnxnguyen:
I'm working on the receipt option cell in the group creation flow. Visually it looks identical to the one in the group details, but it's different in that it's not actually tied to a conversation. Do you think we could refactor out the visual part of the cells, then subclass/implement a protocol (I think this was part of your original idea to separate the protocols?) to then configure that cell for the conversation? Otherwise we will end up implementing (almost) the same thing twice in two different places.

Wire-iOS/Sources/UserInterface/GroupDetails/Cells/GroupDetailsReceiptOptionsCell.swift
```diff
@@ -0,0 +1,59 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+import WireExtensionComponents
+
+class GroupDetailsReceiptOptionsCell: DetailsCollectionViewCell & ConversationOptionsConfigurable {
```
billypchan: @johnxnguyen When you need a cell with UISwitch but not link to ZMConversation, you can use GroupDetailsReceiptOptionsCell. I will move ConversationOptionsConfigurable to extension.

PR #2982, opened 2018-12-03
Possible fix for syncing issues

vytis:
### Issues

After #2970 we have noticed that app stays in syncing mode until the app is force-quit.

### Causes

From logs it seems that background tasks are not created when processing.

### Solutions

If some of the notifications processing expire, we should resume `BackgroundActivityFactory` after coming to foreground. Otherwise no background tasks will be created and we will be stuck.

## Dependencies

Not a direct dependency, but more logs for this are added in https://github.com/wireapp/wire-ios-transport/pull/115


Wire-iOS/Sources/AppDelegate.m
```diff
@@ -125,6 +125,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
 - (void)applicationWillEnterForeground:(UIApplication *)application
 {
+    [BackgroundActivityFactory.sharedFactory resume];
```
mikeger: I think we are observing the `applicationWillEnterForeground` in the sync engine, may be it would be more consistent to put the code there?

vytis: Yeah I agree that it makes sense to move it there

vytis:
Closed in favour of https://github.com/wireapp/wire-ios-sync-engine/pull/913

PR #2984, opened 2018-12-04
ZIOS-10888: User should be able to hide message status of last message in conversation.

nicorsm:
## What's new in this PR?

Implemented new logic to display the status for the last message - see ticket for details.
I've dropped `isLastMessageSentByUser` from `ZMConversationMessageWindow+Formatting`, removed it from the conditions that were forcing to display it everytime and added code to display the view for every new last message (from both sides) + after opening the current conversation.


typfel:
This is changing the behaviour from:
- Show status for last message sent be you
to
- Show status for last message in the conversation.

typfel:
@nicorsm oh ok, wasn't aware :)

typfel:
So maybe `ConversationMessageWindowTableViewAdapter.conversationWindowDidChange` is a better place to manage the selection. I'm a bit afraid of the side effects of changing the selection on `viewWillAppear`.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter+DataSource.swift
```diff
@@ -107,6 +107,29 @@ extension ConversationMessageWindowTableViewAdapter: ZMConversationMessageWindow
                 self.tableView.lockContentOffset = false
             }
         }
+        
+        if changeInfo.insertedIndexes.count > 0 {
```
typfel: Wouldn't it be possible to do:

```suggestion
        if changeInfo.insertedIndexes.count > 0 || isLoadingInitialContent {
```

and then remove `selectLastMessage` in `viewDidLoad`, then everything would be contained inside this method.

typfel: I see ok.

PR #2985, opened 2018-12-04
ZIOS-10939: Fix username not shown for added guest

alexaubry:
### Issues

When a wireless guest joined a conversation through a link, their name was not displayed in the system message cell on iOS clients.

### Causes

When the system message is inserted, it seems that the name is not fetched yet, which means that the generated cell won’t display a name either.

### Solutions

Prior to the cell refactoring, the update was performed in the message window change observer callback in the message window adapter. However, this part of the code was not updated to include the new section system, which means cells were not reloaded appropriately.

We fix this by using the new section controller logic in the message window observer.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
@@ -30,6 +30,12 @@
 
 #import "Wire-Swift.h"
 
+@interface ConversationMessageWindowTableViewAdapter () <ZMConversationObserver>
+
+@property (nonatomic, nullable) id conversationChangeObserver;
+
+@end
+
 @implementation ConversationMessageWindowTableViewAdapter
 
 - (instancetype)initWithTableView:(UpsideDownTableView *)tableView messageWindow:(ZMConversationMessageWindow *)messageWindow
@@ -43,9 +49,15 @@
         self.firstUnreadMessage = self.messageWindow.conversation.firstUnreadMessage;
         self.sectionControllers = [[NSMutableDictionary alloc] init];
         self.actionControllers = [[NSMutableDictionary alloc] init];
+        self.conversationChangeObserver = [ConversationChangeInfo addObserver:self forConversation:messageWindow.conversation];
     }
     
     return self;
+}
+
+- (void)dealloc
+{
+    self.conversationChangeObserver = nil;
 }
 
 - (void)stopAudioPlayerForDeletedMessages:(NSSet *)deletedMessages
@@ -136,6 +148,11 @@
             self.expandingWindow = NO;
         });
     }
+}
+
+- (void)conversationDidChange:(ConversationChangeInfo *)changeInfo
+{
+    [self reconfigureVisibleSections];
 }
 
 @end
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
@@ -55,11 +55,6 @@
     return self;
 }
 
-- (void)dealloc
-{
-    self.conversationChangeObserver = nil;
-}
-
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
```diff
@@ -43,11 +49,17 @@ - (instancetype)initWithTableView:(UpsideDownTableView *)tableView messageWindow
...
         self.actionControllers = [[NSMutableDictionary alloc] init];
+        self.conversationChangeObserver = [ConversationChangeInfo addObserver:self forConversation:messageWindow.conversation];
     }
     
     return self;
 }
 
+- (void)dealloc
+{
+    self.conversationChangeObserver = nil;
```
typfel: All properties are nil:ed out on dealloc

typfel:
Why can't we observe user directly?

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
alexaubry: What is the way to deal with these kind of observation tokens when we register them? I'm actually never sure how to do it properly.

typfel: They are just a empty objects which needs to be retained as long as you want the observer to be active. Usually you just store it in property and it will get released together with the object that does the observation.

See `SelfUnregisteringNotificationCenterToken` in WireUtilities

So essentially this is a wrapper around the `NotificationCenter.AddObserver`

https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver

alexaubry: I understand now. Thank you for the explanation.

alexaubry:
@typfel It's harder to observe the user directly, as we use the ParticipantsViewModel under the hood for that cell, and that object is used for multiple cases, including when we have a list of multiple users.

If this is the better option, should we add an observer for all the users in the view model?

typfel:
@alexaubry I misunderstood the problem, thought it was only about a single user.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationMessageWindowTableViewAdapter.m
```diff
@@ -138,5 +150,10 @@ - (void)expandMessageWindow
     }
 }
 
+- (void)conversationDidChange:(ConversationChangeInfo *)changeInfo
+{
+    [self reconfigureVisibleSections];
```
typfel: Let's add a condition that to only do this if `changeInfo.usersChanged`

alexaubry: Done

typfel: Ah wait. There's another observer.

alexaubry: Where?

typfel: So in `ConversationMessageWindowTableViewAdapter.m`
```
- (void)messagesInsideWindow:(ZMConversationMessageWindow *)window didChange:(NSArray<MessageChangeInfo *> *)messageChangeInfos
```

This will we called when a message in the conversation change and specifically `MessageChangeInfo` has the the `usersChanged.

typfel: `participantsChanged` won't be called when the name changes I think. It's just tracking who's in the group.

alexaubry:
@typfel I updated the pull request to use the observer you pointed out (cf. the updated description)

typfel:
@alexaubry 👍 perfect!

PR #2989, opened 2018-12-05
Fix: phone number is doubled when inserting form auto fill accessory

billypchan:
### Issues

 - the phone number is doubled when inserting form auto-fill accessory
 - app crashes when inserting a phone number with dashes

### Causes

 - When manually handles inserting a phone number to a text field, it should not let default handling doubles the job.
 - non-digit characters should be removed in a phone number type text field

### Solutions

 - Refactoring the insert method in ChangePhoneViewController and PhoneNumberViewController, unify the insert phone number logic.
 - remove all non-digit characters when inserting a phone number.


Wire-iOS Tests/ChangePhoneViewControllerSnapshotTests.swift
```diff
@@ -53,3 +53,38 @@ final class ChangePhoneViewControllerSnapshotTests: ZMSnapshotTestCase {
         verifyInIPhoneSize(view: sut.view)
     }
 }
+
+final class ChangePhoneViewControllerTests: ZMSnapshotTestCase {
```
typfel: This could be normal `XCTestCase`?

Wire-iOS Tests/ChangePhoneViewControllerSnapshotTests.swift
```diff
@@ -53,3 +53,38 @@ final class ChangePhoneViewControllerSnapshotTests: ZMSnapshotTestCase {
...
+        super.tearDown()
+    }
+
+    func testForADigitIsAllowed(){
+        // GIVEN
+        // call viewDidLoad
+        sut.loadViewIfNeeded()
+
+        // make table view's cells visible
+        sut.view.frame = CGRect(origin: .zero, size: defaultIPhoneSize)
```
typfel: Does this really affect the test?

billypchan: Yes, the below sut.tableView.cellForRow() call return nil if the frame is zero size.

PR #2990, opened 2018-12-05
Fix tests compilation

alexaubry:
### Issues

Following the merge of #2975, the app was not compiling on develop.

### Solutions

Revert a faulty merge conflict resolution. We also fix a bug where the message send timestamp was not displayed when the message was read.


mikeger:
@alexaubry it's hard to justify from the changeset how it is fixing the compilation, it seem to move stuff around and rephrase logic, may be I am missing something?

alexaubry:
@mikeger Some method was accidentally moved to an extension of Message during a merge conflict resolution, but it was supposed to be in the extension of ZMConversationMessage, which is right above.

alexaubry:
The second fix, where we use `isOneOf` just adds the case to check for the `read` delivery status, which wasn't checked before.

mikeger:
So what I see:
screen shot 2018-12-05 at 14 33 56
i.e. it was in Message, it is in Message?

mikeger:
Ah now I go it, following method is moved.

mikeger:
Sorry for the confusion @alexaubry

alexaubry:
Yes, `formattedAccessibleMessageDetails` was moved in `Message`

alexaubry:
Manual build https://dev.azure.com/wireswiss/Wire%20iOS/_build/results?buildId=1731 succeeded.

PR #2991, opened 2018-12-05
Read receipts group flow

typfel:
## What's new in this PR?

Allow the user to configure read receipts when creating a new group conversation.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -289,8 +291,12 @@
 "conversation.create.group_name.placeholder" = "Group name";
 "conversation.create.guidance.empty" = "At least 1 character";
 "conversation.create.guidance.toolong" = "Too many characters";
-"conversation.create.toggle.title" = "Allow guests and services";
-"conversation.create.toggle.subtitle" = "Open this conversation to services and people outside your team. You can always change it later.";
+"conversation.create.options.title" = "Conversation options";
+"conversation.create.options.subtitle" = "Allow guests: %@, Read receipts: %@";
```
mikeger: Not direct question to John or Jacob, but I guess this would be pretty hard to translate.

Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateErrorCell.swift
```diff
@@ -0,0 +1,44 @@
...
+        setup()
+    }
+    
+    fileprivate func setup() {
+        label.textAlignment = .center
+        label.font = FontSpec(.small, .semibold).font!
+        label.textColor = UIColor.Team.errorMessageColor
+        
+        contentView.addSubview(label)
+        label.translatesAutoresizingMaskIntoConstraints = false
```
mikeger: As far as I know this has to be called before adding to the superview.

Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateNameCell.swift
```diff
@@ -0,0 +1,58 @@
...
+        setup()
+    }
+    
+    fileprivate func setup() {
+        textField.isAccessibilityElement = true
+        textField.accessibilityIdentifier = "textfield.newgroup.name"
+        textField.placeholder = "conversation.create.group_name.placeholder".localized.uppercased()
+                
+        contentView.addSubview(textField)
+        textField.translatesAutoresizingMaskIntoConstraints = false
```
mikeger: Same as above.

Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
```diff
@@ -118,54 +157,43 @@ final public class ConversationCreationValues {
...
-        }
-        errorViewContainer.translatesAutoresizingMaskIntoConstraints = false
-
-        errorLabel.textAlignment = .center
-        errorLabel.font = TeamCreationStepController.errorFont
-        errorLabel.textColor = UIColor.Team.errorMessageColor
-        errorLabel.translatesAutoresizingMaskIntoConstraints = false
-        errorViewContainer.addSubview(errorLabel)
-        toggleSubtitleLabel.numberOfLines = 0
+        // TODO: if keyboard is open, it should scroll.
```
mikeger: Still relevant?

typfel: Yes, but not critical would do in separate PR or even create a separate ticket for it.

Wire-iOS/Sources/UserInterface/Conversation/Create/Cells/ConversationCreateErrorCell.swift
mikeger: Ok, from my limited testing this is not an issue.

typfel: I also noticed some constraints are breaking in the tests, will attempt to fix.

Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationCreateOptionsSectionController.swift
```diff
@@ -0,0 +1,48 @@
...
+import Foundation
+
+class ConversationCreateOptionsSectionController: ConversationCreateSectionController {
+    
+    typealias Cell = ConversationCreateOptionsCell
+    
+    var tapHandler: ((Bool) -> Void)?
+    
+    override func prepareForUse(in collectionView: UICollectionView?) {
+        collectionView.flatMap(Cell.register)
```
mikeger: I think using the optionals monad-style is pretty confusing.

Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationCreateSectionController.swift
```diff
@@ -0,0 +1,104 @@
...
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        assertionFailure("Must be overriden.")
+        return UICollectionViewCell()
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, viewForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> UICollectionReusableView {
+        switch kind {
+        case UICollectionView.elementKindSectionHeader:
+            let view = collectionView.dequeueReusableSupplementaryView(ofKind: kind, withReuseIdentifier: "SectionHeader", for: indexPath)
+            (view as? SectionHeader)?.titleLabel.text = headerText
```
mikeger: I think the type must be enforced if we want to be sure the code is not breaking in runtime.

Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
johnxnguyen: Not sure if @typfel looked at this. I believe design wants the bottom of the view to sit against the top of the keyboard, currently the keyboard sits over the bottom of the screen.

mikeger: I think we can use our famous `KeyboardAvoidingViewController` container.

Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationCreateSectionController.swift
johnxnguyen: I'm not sure what you mean... could you elaborate?

mikeger: If we dequeue the view, and it's not `SectionHeader`:
Should this potentially happen?
- If yes, I would do a condition, like if this view do this, if other do nothing.
- If no, I would assert the type.

johnxnguyen: I suppose we should assert the type. There are other places in the source code where we should apply this logic too.

Wire-iOS/Sources/UserInterface/Conversation/Create/Sections/ConversationCreateOptionsSectionController.swift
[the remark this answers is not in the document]
johnxnguyen: Or we could just register it the old fashioned way.

Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
typfel: it's already wrapped inside a `KeyboardAvoidingViewController` 🧐

mikeger: 🌯

mikeger:
@typfel are you checking the constraints?

typfel:
@mikeger they are fixed.

mikeger:
Ok, then we can merge?

Wire-iOS/Sources/UserInterface/Conversation/Create/ConversationCreationController.swift
typfel: So problem is that the text field loses focus when you scroll. I still say we can look into that in a separate PR.

PR #2995, opened 2018-12-06
Fix: create read receipt system message with correct source of sender

billypchan:
### Issues

App crash when showing a read receipt system message

### Causes

The sender of the system message is nil

### Solutions

Get the sender user from the correct source (ZMConversationMessage.sender)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationReadReceiptSettingChangedCellDescription.swift
```diff
@@ -18,59 +18,46 @@
...
         case .readReceiptsEnabled:
-            if let selfUser = systemMessage.involvesSelfUserOnly {
-                updateText = NSAttributedString(string: "content.system.message_read_receipt_on".localized(pov: selfUser.pov, args: "content.system.you_started".localized), attributes: baseAttributes)
-            } else if let otherUserName = systemMessage.otherUserName {
-                updateText = NSAttributedString(string: "content.system.message_read_receipt_on".localized(args: otherUserName), attributes: baseAttributes)
-                    .adding(font: .mediumSemiboldFont, to: otherUserName)
-            }
+            updateText = createSystemMessage(template: "content.system.message_read_receipt_on")
         case .readReceiptsOn:
             updateText = NSAttributedString(string: "content.system.message_read_receipt_on_add_to_group".localized)
```
typfel: This case has no font styling.

typfel: Also I think we are missing a snapshot test for `.readReceiptsOn`

billypchan: @typfel added the snapshot for that case.

PR #2997, opened 2018-12-06
Fix: Invite contacts does not work

billypchan:
### Issues

When the user has no email client installed, the screen jumps back to contacts overview and no one is invited/contacted. 

### Causes

The "Please configure your email..." dialog is not shown in this case for the warning "Warning: Attempt to present <UIAlertController: 0x7f9d0aa37800> on <InviteContactsViewController: 0x7f9d0c16dc00> whose view is not in the window hierarchy!
"

### Solutions

Present the dialog in NotificationsWindow.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.m
+++ b/Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.m
@@ -426,7 +426,8 @@ static NSString* ZMLogTag ZM_UNUSED = @"UI";
                                                                  [unableToSendController dismissViewControllerAnimated:YES completion:nil];
                                                              }];
             [unableToSendController addAction:okAction];
-            [self presentViewController:unableToSendController animated:YES completion:nil];
+
+            [[AppDelegate sharedAppDelegate].notificationsWindow.rootViewController presentViewController:unableToSendController animated:YES completion:nil];
             return;
         }
         


Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.m
```diff
@@ -426,7 +426,8 @@ - (void)inviteContact:(ZMAddressBookContact *)contact fromView:(UIView *)view
                                                                  [unableToSendController dismissViewControllerAnimated:YES completion:nil];
                                                              }];
             [unableToSendController addAction:okAction];
-            [self presentViewController:unableToSendController animated:YES completion:nil];
+
+            [[AppDelegate sharedAppDelegate].notificationsWindow.rootViewController presentViewController:unableToSendController animated:YES completion:nil];
```
typfel: Maybe it could also work by by setting the `modalTransitionStyle` of the alert view controller to `UIModalPresentationOverFullScreen`. It would then present on the root.

billypchan: @typfel I tested and it does not work. This problem is caused by this call in ContactsViewController's child InviteContactsViewController:

```
            navigationController.popToRootViewController(animated: false) {
                self.inviteUserOrOpenConversation(user, from:view) ///which may prsenet an alert
            }

```

In the future we should rewrite this, which should call inviteUserOrOpenConversation before poping the InviteContactsViewController.

PR #3002, opened 2018-12-07
Fix: do not create placeholder label when draw rect

billypchan:
## What's new in this PR?

This PR is part of the snapshot tests refactoring tasks.

### Issues

During refactoring snapshot tests, I found that the snapshots involving TextView's placeholder produce ambiguous results. The placeholder may have vertical offset or subtle colour differences in different runs

### Causes

The placeholder is created in overrode drawRect() method and its position is set by frame's value. I found it does not respect the property placeholderTextContainerInset.

### Solutions

Position the placeholder label with constraints

### Note

This fix updates the snapshots' content offset which having long text in TextView.


WireExtensionComponents/Views/TextView+Internal.h
```diff
@@ -0,0 +1,28 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+@interface TextView ()
+
+@property (nonatomic, nonnull) UILabel *placeholderLabel;
+
+@property (nonatomic, nullable) NSLayoutConstraint *placeholderLabelLeftAnchor;
```
mikeger: I think the name of the variable is confusing: it's not anchor but constraint.

WireExtensionComponents/Views/TextView+Placeholder.swift
```diff
@@ -0,0 +1,60 @@
...
+        placeholderLabel = UILabel()
+        placeholderLabel.font = placeholderFont
+        placeholderLabel.textColor = placeholderTextColor
+        placeholderLabel.textTransform = placeholderTextTransform
+        placeholderLabel.textAlignment = placeholderTextAlignment
+        placeholderLabel.isAccessibilityElement = false
+
+        addSubview(placeholderLabel)
+
+        placeholderLabel.translatesAutoresizingMaskIntoConstraints = false
```
mikeger: I think this must be called before the view is added to the superview, but I might be wrong.

PR #3004, opened 2018-12-07
Fixes for the Message Toolbox

alexaubry:
### Issues

1. It was not possible to distinguish the timestamp from the read receipts in automation (ZIOS-10886)
2. The animation when the user toggled likes/timestamp was slow
3. Some content was displayed incorrectly (after sending an image, the resend/delete buttons) were shown
4. Tapping on the toolbox of an image message opened the full screen view and 3D Touch (ZIOS-10958)

### Solutions

Previously, we were relying on some flags to determine the current state of the toolbox, which proved to be erratic at some times. We improved controlling the state of the contents by using an enum with associated values. The creation of the content is delegated to a new "message toolbox data source", that determines the state of the message and what text to display.

This enabled us to move away from the single text view, and display the contents as individual elements inside a stack view. With these changes, we can:

- enforce a spacing between elements without having to pad with spaces
- improve the animations, as hiding elements is managed by the stack
- have distinct identifiers on each part of the secondary line, which solves the automation issue
- avoid keeping link attributes cached in the text view, which caused the issue in automation where their accessibility identifiers could still be found

To fix the animation issue, we compare the new content to the old content when reloading. If the content is the same, we do not schedule another animation. We also add some semantics in the animation by reversing the swipe effect depending on the content on screen previously (ex: likes → timestamp: swipe down; timestamp → likes: swipe up).

Regarding the issue with the full screen image, we discovered that the "single tap action" in the cell adapter was called even though the cell description says it doesn't support actions. Disabling the gesture when actions are not supported by the cell solves the problem.

## To-Do

- [x] Disable 3D Touch on message toolbox


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -196,8 +196,6 @@ - (void)viewDidLoad
                                              selector:@selector(applicationDidBecomeActive:)
                                                  name:UIApplicationDidBecomeActiveNotification
                                                object:nil];
-
-    [self.conversationMessageWindowTableViewAdapter selectLastMessage];
```
typfel: Why remove this?

alexaubry: If the last message has likes, we will never show them because forceShowTimestamp will be true.

typfel: Ok we also do in another place so I guess we need to rethink the selection logic

alexaubry: Agreed.

PR #3012, opened 2018-12-10
ZIOS-7768: Added obfuscation of deleted ephemeral messages inside collection view

nicorsm:
### Issues

Ephemeral messages that are received in a conversation and still available inside the collection view were displayed as a grey box, instead of the proper obfuscation view.

### Causes

Ephemeral messages received in a conversation are deleted once the timer has expired. The obfuscation view was shown only when the message was obfuscated, and not when deleted. 

### Solutions

Right now, I'm displaying the obfuscation view in both cases (obfuscation and deletion).


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.m
```diff
@@ -231,7 +231,7 @@ - (void)setupScrollView
 
 - (void)updateForMessage
 {
-    if (self.message.isObfuscated) {
+    if (self.message.isObfuscated || self.message.isHiddenInConversation) {
```
typfel: It's not possible to use the already existing `hasBeenDeleted` property?

nicorsm: Totally missed that.. it's working as well 😁 I'll delete the DM-branch.

PR #3016, opened 2018-12-10
ZIOS-10914: Removed isAccessibilityElement from video, file and audio cells container

nicorsm:
## What's new in this PR?

The `isAccessibilityElement` setting on video, file and audio cell containers were causing the subviews not being accessible from the inspector, even if they had proper IDs setted. I've removed those assignments.


mikeger:
@alexaubry may be you can retroactively review this one tomorrow?

PR #3017, opened 2018-12-10
Message details: set bar background color

mikeger:
### Issues

Background color of the bar on the dark theme should not be 100% black.


mikeger:
I'll give it a try and merge it, this one-liner seems to have no consequences.

PR #3024, opened 2018-12-11
Fix for table view crash

typfel:
### Issues

App would sometimes crash when inserting a new message

### Causes

See [data model changes](https://github.com/wireapp/wire-ios-data-model/pull/635)

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/635
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/925


mikeger:
@typfel should we merge this to the release branch?

PR #3028, opened 2018-12-11
ZIOS-10985: Typing indicator won't dismiss after unsuccessful edit

nicorsm:
## What's new in this PR?

Typing indicator wasn't dismissed after closing the editing input bar in a conversation; this is because `conversation.setIsTyping(false)` was not called after the end of the editing.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Editing.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Editing.swift
@@ -50,6 +50,7 @@ extension ConversationInputBarViewController {
             self.conversation.draftMessage = nil
         }
         updateWritingState(animated: true)
+        conversation.setIsTyping(false)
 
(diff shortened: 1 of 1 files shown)


mikeger:
@nicorsm should we merge it to the release branch?

nicorsm:
@mikeger yeah sure!

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Editing.swift
```diff
@@ -50,6 +50,7 @@ extension ConversationInputBarViewController {
             self.conversation.draftMessage = nil
         }
         updateWritingState(animated: true)
+        conversation.setIsTyping(false)
```
typfel: Might need to be inside the `enqueueChanges `

typfel: Ah or maybe not since this is all one memory

PR #3030, opened 2018-12-11
Missing unread dot in the account switcher

typfel:
### Issues

If you were using Wire with multiple accounts the account switcher wouldn't indicate which accounts have unread messages.

### Causes

The `AccountView` used in the account switcher weren't initialised with a user since it can't observe the self user of a background account. The unread dot was only updated if a user was associated with the `AccountView`.

### Solutions

Re-write the update method to not require a user to be present.


mikeger:
@typfel standard comment should it go to the release branch

mikeger:
@typfel tests failed somehow?

typfel:
@mikeger code coverage upload failed. I'll restart.

typfel:
I'll merge to develop first and then cherry-pick

PR #3033, opened 2018-12-12
Fix: Invite contacts does not work (take 2)

billypchan:
## What's new in this PR?

This PR is a follow up of https://github.com/wireapp/wire-ios/pull/2997.
Fix other flow that presents alert in an already dismissed view controller.


Wire-iOS Tests/InviteContactsViewControllerSnapshotTests.swift
```diff
@@ -34,8 +34,7 @@ final class InviteContactsViewControllerSnapshotTests: ZMSnapshotTestCase {
         sut = nil
         super.tearDown()
     }
-
-    fileprivate func snapshotWithNavigationBarWithBackButton(file: StaticString = #file, line: UInt = #line) {
+    fileprivate func wrapInNavigationControllerWithDummyPerviosViewController() {
```
vytis: Typo - Pervios

Wire-iOS Tests/InviteContactsViewControllerSnapshotTests.swift
```diff
@@ -59,14 +62,23 @@ final class InviteContactsViewControllerSnapshotTests: ZMSnapshotTestCase {
...
-    func testForContactsWithoutSectionBar() { ///TODO: hide no contact label
+    func testForContactsWithoutSectionBar() {
         let mockUsers = MockUser.mockUsers()
         sut.dataSource?.ungroupedSearchResults = mockUsers
 
         snapshotWithNavigationBarWithBackButton()
     }
 
-    func testForContactsAndIndexSectionBarAreShown() { ///TODO: hide no contact label
+    /// CI server produce empty snapshot but it works on local machine. It seems the alert with type UIAlertControllerStyleAlert is not shown on CI server.
```
vytis: This is pretty unlikely - probably it's a timing issue. We make a screenshot before the screen/view is displayed correctly. We had issues like this with avatars - images were loaded asynchronously and sometimes did not show up. Few things could be done in this case - disable animations (`view.layer.speed = 0`) or try to debug it by printing view hierarchy with `view.value(forKey: "recursiveDescription")!` at the end of test.

PR #3035, opened 2018-12-13
 bump Data model

billypchan:
## What's new in this PR?

bump Data model for https://github.com/wireapp/wire-ios-data-model/pull/637


typfel:
We should not bump to `136.0.1` I'll make a hotfixa for 135.

PR #3042, opened 2018-12-17
Add missing icon for pending outoging connections

typfel:
### Issues

Outgoing pending connection were missing the clock icon on the conversation list.

### Causes

We were using `UserImageView` instead of `BadgeUserImageView` which adds the correct icon if necessary.

### Solutions

Replace `UserImageView` with `BadgeUserImageView`.


vytis:
No snapshot added?

typfel:
@vytis added

PR #3046, opened 2018-12-17
Fix: [iPad] When inviting users to group chat the keyboard makes the list items disappear

billypchan:
### Issues

On iPad, the add participant screen is not usable when the keyboard is shown. The search result is always empty.

### Causes

The SearchResultsView has an isContainedInPopover property which is set when init. However, its value is false.

### Solutions

Inject the parent view controller instead. Check the parent's isContainedInPopover method's return value when the keyboard is toggled.

## Notes

This PR rewrites the Cartography from SearchResultsView. A new excludeAnchor argument is added to UIView.fitInSuperview method.


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsView.swift
```diff
@@ -29,7 +28,7 @@ import Cartography
     var lastLayoutBounds : CGRect = CGRect.zero
     var accessoryContainerHeightConstraint: NSLayoutConstraint?
     var accessoryViewBottomOffsetConstraint : NSLayoutConstraint?
-    var isContainedInPopover : Bool = false
+    var parentViewController: UIViewController?
```
typfel: This needs to be weak to avoid a retain cycle.

PR #3048, opened 2018-12-18
move URLs to URL+Wire

billypchan:
## What's new in this PR?

Move URLs to URL+Wire.swift.

### DIFF
--- a/Wire-iOS/Sources/Helpers/URL+Wire.swift
+++ b/Wire-iOS/Sources/Helpers/URL+Wire.swift
@@ -45,7 +45,7 @@ extension URL {
 
 extension URL {
 
-    static var wr_itunes: URL {
+    static var wr_wireAppOnItunes: URL {
         return URL(string: "https://itunes.apple.com/us/app/wire/id930944768?mt=8")!
     }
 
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -45,6 +45,14 @@ extension URL {
 
 extension URL {
 
+    static var wr_itunes: URL {
```
vytis: maybe a bit more explanatory would be good like `wireAppOnItunes`

Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -45,6 +45,14 @@ extension URL {
...
+        return URL(string: "https://itunes.apple.com/us/app/wire/id930944768?mt=8")!
```
alexaubry: While we are at it. As Apple recommends in the Link Maker, maybe it would be more resilient to use a global link here, because it automatically redirects the user to the correct storefront (eg, if they're not on the US store, they will be redirected to the DE store). cc @marcoconti83 

```suggestion
        return URL(string: "https://geo.itunes.apple.com/us/app/wire/id930944768?mt=8")!
```

vytis: That sounds like a good idea I think!

PR #3051, opened 2018-12-18
Store URLs in a JSON file

billypchan:
## What's new in this PR?

Store hard-coded URLs in a JSON file.

### Dependencies
- [x] https://github.com/wireapp/wire-ios-build-configuration/pull/2


Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -29,15 +31,57 @@ import UIKit
     }
 }
 
+final class WireUrl {
```
typfel: Couldn't this be a `Codable` struct?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -1343,6 +1342,9 @@
 		EFD514AB21886C5D00D7C7E0 /* UIView+ConstraintsHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = EFD514AA21886C5D00D7C7E0 /* UIView+ConstraintsHelper.swift */; };
 		EFD514AD21886CC600D7C7E0 /* ConversationListItemView+Constraints.swift in Sources */ = {isa = PBXBuildFile; fileRef = EFD514AC21886CC600D7C7E0 /* ConversationListItemView+Constraints.swift */; };
 		EFD5CF4E2189C68D00BEC99E /* NSMutableAttributedString+ReplaceEmojiTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EFD5CF4C2189C5F700BEC99E /* NSMutableAttributedString+ReplaceEmojiTests.swift */; };
+		EFD5FC3B21C93FEB00026E3A /* url.json in Resources */ = {isa = PBXBuildFile; fileRef = EFD5FC3A21C93FEB00026E3A /* url.json */; };
+		EFD5FC3C21C93FEB00026E3A /* url.json in Resources */ = {isa = PBXBuildFile; fileRef = EFD5FC3A21C93FEB00026E3A /* url.json */; };
```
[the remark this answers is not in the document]
billypchan: The JSON file is move to wire-ios-build-configuration and these lines are removed.

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -7003,6 +7007,7 @@
 				EF21270B1FB9DFE300625A9B /* CountryCodes.plist in Resources */,
 				BF0BDF721DA7A9D8003C61DE /* RedactedScript-Regular.ttf in Resources */,
 				DB3E943A1B96EDBD00990482 /* LaunchScreen.xib in Resources */,
+				EF2B0DD121CA8EF1003B2882 /* url.json in Resources */,
```
vytis: Still seems to be also added in Resources...

billypchan: It is correct. The sibling file Backend.bundle is located at 4 lines above.

Wire-iOS Tests/UrlTests.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+
+import XCTest
+@testable import Wire
+
+final class UrlTests: XCTestCase {
+    func testForFingerprintLearnMoreURL(){
+        let sut = NSURL.wr_fingerprintLearnMoreURL
+
+        XCTAssertEqual(sut.absoluteString, "https://wire.com/privacy/why")
```
alexaubry: Will these tests break if we run the tests with another resource set (eg. when building for a customer)?

Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -29,15 +31,80 @@ import UIKit
     }
 }
 
+struct WireUrl: Codable {
```
alexaubry: I would capitalize URL.

Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -29,15 +31,80 @@ import UIKit
...
+    let emailInUseLearnMore: URL
+    let randomProfilePictureSource: URL
+
+    let termsOfServicesTeams: URL
+    let termsOfServicesPersonal: URL
+
+    let manageTeamBase: URL
+
+    static var shared: WireUrl! = {
+        guard let filePath = Bundle.main.url(forResource: "url", withExtension: "json") else {
```
alexaubry: Maybe we can explore generating these properties automatically at build time?

We could have a template that looks like this:

~~~swift
extension URL {

    {% for url in urls %}
    static var wr_{{ url.name }} : URL {
        return URL(string: {{ url.value }})!
    }

    {% endfor %}
}
~~~

We could have a build phase in the project that calls a script to generate that extension. As an input, we could pass the path to the JSON file. As the output, we could put the path to the resulting Swift file.

The advantage of this solution is that:
- decoding the JSON or finding the resource would never fail
- we don't use too much resources when setting up the app (especially since the decoding would be done on the main thread)
- new URLs are added automatically, and removed URLs are removed automatically when the JSON changes

Maybe this something we can discuss with @vytis.

Wire-iOS Tests/UrlTests.swift
vytis: It will, but for now it's not clear if we ever change it. It's better to be safe that the public one works and worry about other things later.

Wire-iOS/Sources/Helpers/URL+Wire.swift
vytis: Yeah this sounds like a good idea, the only problem is that we need to make sure Sourcery (or some other tool) is installed on all developer machines and CI. Might not be too bad and there are plenty more things that we could use it for, e.g. generating mock classes for protocols that we use in tests. Maybe you can make an architecture issue about all this?

PR #3055, opened 2018-12-20
Fix logic for showing/hiding read receipts tab 

typfel:
### Issues

Non team users couldn't see the read receipts tab for messages written as a guest in a team conversation.

### Causes

Read receipts tab was shown/hidden based on if the self user was part of a team.

### Solutions

Show the read receipt tab if the conversation was created in a team.


Wire-iOS/Sources/Helpers/syncengine/ZMMessage+Actions.swift
```diff
@@ -62,13 +62,11 @@ extension ZMConversationMessage {
         guard conversation.conversationType == .group else {
             return false
         }
-
+        
         // Show the message details in Team groups.
-        if ZMUser.selfUser()?.isTeamMember == true {
+        if conversation.teamRemoteIdentifier != nil {
             return true
```
alexaubry: Maybe it would actually be safer to do more checks in the case of team users. For example, returning true here means we would show the details for a ping that the user didn't send.

```suggestion
            return canBeLiked || isSentBySelfUser
```

PR #3070, opened 2019-01-04
Retire ocmock - refactor soundcloud service tests

billypchan:
## What's new in this PR?

Rewrite tests in SoundcloudServiceTests without ocmock. The userSession property of SoundcloudService is replaced with ProxiedURLRequester protocol type for injecting mock user session class.


typfel:
OCMock was still used by something?

billypchan:
@typfel WireTesting needs ocmock. I will check if it is possible to not to include wire-ios-testing when building wire-ios.

PR #3074, opened 2019-01-07
ArticleView: Retire TTTAttributedLabel and Cartography

billypchan:
## What's new in this PR?

Retire TTTAttributedLabel and Cartography in ArticleView. When the ArticleView is tapped, the only possible action is open the URL. 
A new UIView.pin() helper method is added to substitute Cartography's edges constraint creation.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/ArticleView.swift
```diff
@@ -217,21 +206,6 @@ import WireExtensionComponents
     
 }
 
-extension ArticleView : TTTAttributedLabelDelegate {
-    
-    func attributedLabel(_ label: TTTAttributedLabel!, didSelectLinkWith url: URL!) {
```
mikeger: Shouldn't the label allow opening the URLs?

billypchan: I checked this class only sets the label's text property only. No links should be shown in this label.

PR #3076, opened 2019-01-07
Refactor new conversation cell

typfel:
## What's new in this PR?

- Move the `.newConversation` system message to the new cell system
- Delete `ConversationCell` and related classes


mikeger:
Unfortunately this once again changes the files that I've deleted on my branch :(

typfel:
@mikeger yes it's unfortunate but this should be last time.

PR #3086, opened 2019-01-10
Observe changes in ShareExt

johnxnguyen:
### Issues

Sharing from Share Extension while the Wire Client is foregrounded does not update the UI.

### Causes

Core Data changes caused by the Share Extension are persisted to disk, which are only processed when the app comes into the foreground. This makes sense most of the time (it's unusual that the share extension is opened from within the client), however it we need a way to merge these changes when the app is already in the foreground.

### Solutions

Darwin Notifications offer very simple interprocess communication. Simple, because it only allows a C function callback to be invoked when the notification is fired, without any user info or context capturing. This means it is effectively a _ping_ that a certain event has occurred. 

`DarwinNotificationCenter` is a helper class implemented in `WireUtilities` that accepts standard Swift callback handlers for a `DarwinNotification`. In this case, the notification is called `shareExtDidSaveNote`, which is posted when Core Data changes from the share extension have been persisted to disk. When the application activates, we register a callback for this notification, which will ask the active user session to process these core data changes, if any. 

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-utilities/pull/71
- [ ] https://github.com/wireapp/wire-ios-share-engine/pull/80
- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/940


johnxnguyen:
Closing as it's no longer needed.

PR #3092, opened 2019-01-14
Fix/invite button

johnxnguyen:
### Issues

The guests allowed cell (label: "Services and people outside your team can join this conversation.", button: "Invite people") is visible in
- group conversations for non team users
- one to one conversation for team users

### Causes

Missing checks for these cases when constructing the new conversation system message cell.

### Solutions

Add the necessary checks


johnxnguyen:
@typfel regarding tests, what's the process for configuring the conversation when generating the mock system message (in `ConversationSystemMessageTests`)?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -347,9 +347,8 @@ class ConversationSystemMessageCellDescription {
             let startedConversationCell = ConversationStartedSystemMessageCellDescription(message: message, data: systemMessageData)
             cells.append(AnyConversationMessageCellDescription(startedConversationCell))
             
-            if conversation.allowGuests {
+            if (ZMUser.selfUser()?.isTeamMember ?? false) && conversation.conversationType == .group && conversation.allowGuests {
```
typfel: It's not enough to check if the self user is a team member because what's matters is if the self user is guest in this particular conversation, which might not be the case when two teams have a group conversation together.

See: `selfUser.isGuest(in: conversation)`

typfel:
@johnxnguyen and regarding configuring the conversation, take look at `StartedConversationCellTests.swift` for some examples.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
billypchan: Is it possible to just check:
  `ZMUser.selfUser()?.permissions?.contains(.addConversationMember)` ?

PR #3093, opened 2019-01-14
ZIOS-11057: Support the feature flag to disable phone login

alexaubry:
### Issues

We introduced a feature flag to disable phone login: `PHONE_AUTHENTICATION_DISABLED`. We need to remove the button to sign in by phone if that flag is present.

### Solutions

In the `SignInViewController`, we conditionally create the buttons depending on whether this flag is present. This PR also fixes a bug where we'd show the registration tab bar when tapping on the "log in" button even though `ACCOUNT_CREATION_DISABLED` was specified.


Wire-iOS Tests/SignInViewControllerTests.swift
```diff
@@ -19,7 +19,15 @@
...
-final class SignInViewControllerTests: XCTestCase {
+class EmailOnlySignInViewController: SignInViewController {
+
+    override var supportsMultipleFlowTypes: Bool {
+        return false
+    }
+
+}
+
+final class SignInViewControllerTests: ZMSnapshotTestCase {
```
billypchan: It would be better to create a new class SignInViewControllerSnapshotTests for the new snapshot tests.

Wire-iOS Tests/SignInViewControllerTests.swift
```diff
@@ -45,4 +53,42 @@ final class SignInViewControllerTests: XCTestCase {
...
+    func testThatItHidesPhoneNumberButtonIfNeeded() {
+        // GIVEN
+        let sut = EmailOnlySignInViewController()
+
+        // THEN
+        let form = wrap(sut)
+        verify(view: form.view)
+    }
+
+    func testThatItDoesNotSwitchToPhoneScreenIfPhoneDisabled() {
```
billypchan: This snapshot test produces the same image as the test above. Maybe create some checks like sut has no phone number text field is enough?

vytis:
Since this is a very isolated case, maybe it's better to have one flag that disables registration and allows only login flow - e.g. `ALLOW_ONLY_EMAIL_LOGIN` or something similar. The complexity increases with every flag added, so it's better to keep it to a minimum. We might re-visit it at a later point if we decide to move to a better feature flag infrastructure.

Wire-iOS Tests/SignInViewControllerTests.swift
alexaubry: I updated the test case, thanks.

Wire-iOS Tests/SignInViewControllerTests.swift
alexaubry: I personally don't think this is necessary, as the test cases are pretty short and simple, and will be removed eventually with the new UI PR. But I did split the logic and snapshot tests into two distinct sections of the class though, to make it more clear.

PR #3095, opened 2019-01-14
Fix: Long tap -> delete message does not work if last line of message is not in view

billypchan:
## What's new in this PR?

This PR is a follow-up of https://github.com/wireapp/wire-ios/pull/3085.

### Issues

A cell with very long content may have no action after the user tapped the delete menu item.

### Causes

If a cell is not full visible in the table view, UITableView.cellForRow() may returns nil.

### Solutions
Get the cell from the tableView's dataSource delegate method with given indexPath.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Message.swift
```diff
@@ -21,10 +21,18 @@ import Foundation
...
         }
-        
-        return self.tableView.cellForRow(at: indexPath)
+
+        // Notice: if the cell is not full visible in the table view, UITableView.cellForRow() may returns nil.
+        // To handle this case, get the cell from the tableView's dataSource delegate method.
+        var cell = tableView.cellForRow(at: indexPath)
+
+        if cell == .none {
+            cell = tableView.dataSource?.tableView(tableView, cellForRowAt: indexPath)
```
typfel: Why not always ask the data source?

vytis: Not sure if it's a good approach - you are asking data source to create a new cell in this case. If the cell is at least a bit visible it means we have already created it before. 
It might be related to the fact that we now have one section per message, not one cell per message. So maybe we don't get the right index path in this case? Maybe @alexaubry has ideas what could be happening?

billypchan: @typfel I was thinking about performance. If we always ask the data source for the cell it always creates a new cell. I will check if this method called frequently or just called in some cases.

alexaubry: These functions actually don't make sense anymore, because there cannot be a single index path or cell for a message, because the message is composed of multiple cells.

I would suggest to remove:

- `ConversationMessageWindowTableViewAdapter.indexPathForMessage`
- `ConversationContentViewController.cellForMessage`

And inspect why we need to call these in the first place at call site.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/ConversationMessageCell.swift
```diff
@@ -20,8 +20,14 @@ import UIKit
...
 
 
+@objc public protocol MessageActionResponder: NSObjectProtocol {
+    @objc(wantsToPerformAction:forMessage:)
+    func wants(toPerform action: MessageAction, for message: ZMConversationMessage!)
+}
+
 @objc protocol ConversationMessageCellDelegate: MessageActionResponder {
-    
+    var cellTappedForMenu: (UIView & SelectableView)? { get set }
```
billypchan: remember the cell tapped to open the menu, then use it for source view for showing the menu.

billypchan:
I will try another approach to fix this issue.

PR #3098, opened 2019-01-16
Feature: I don't want be able to create group conversation

billypchan:
## What's new in this PR?

Show or hide the create guestroom/ create group cells depends on the permission.


Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Create Group/CreateGroupSection.swift
```diff
@@ -26,7 +32,15 @@ class CreateGroupSection: NSObject, CollectionViewSectionController {
     }
     
     private var data: [Row] {
-        return ZMUser.selfUser().isTeamMember ? [Row.createGroup, Row.createGuestRoom] : [Row.createGroup]
+        if ZMUser.selfUser().isTeamMember {
+            if ZMUser.selfUserHas(permissions: requiredPermissions) {
```
johnxnguyen: Instead of calling `ZMUser.selfUserHas(permissions:)` you can use the protocol field `selfUserIsAuthorized`. It calls the method underneath, passing in the `requirePermissions` for you.

billypchan:
@johnxnguyen updated this code for your comments.

PR #3106, opened 2019-01-18
Feature/permissions conversation options

johnxnguyen:
## What's new in this PR?

In the group details screen, if the self user is a `collaborator`, then they should not see any UI that allows them to:

1. Rename the conversation
2. Change the group access (guests)
3. Change the timeout
4. Change the read receipts


Wire-iOS/Sources/UserInterface/GroupDetails/Cells/GroupDetailsRenameCell.swift
```diff
@@ -80,6 +84,11 @@ class GroupDetailsRenameCell : UICollectionViewCell {
         contentStackView.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -16).isActive = true
         contentStackView.spacing = 8
         
+        if !selfUserIsAuthorized {
+            titleTextField.isUserInteractionEnabled = false
+            accessoryIconView.isHidden = true
+        }
```
billypchan: We can move these check to func configure(for conversation: ZMConversation), so the UI will response to the role change event.

billypchan:
@johnxnguyen we can merge this PR after the conflicts are fixed.

Wire-iOS Tests/GroupDetailsViewControllerSnapshotTests.swift
```diff
@@ -22,45 +22,84 @@ import XCTest
 final class GroupDetailsViewControllerSnapshotTests: CoreDataSnapshotTestCase {
     
     var sut: GroupDetailsViewController!
+    var groupConversation: ZMConversation!
```
billypchan: This is a nice a helper and we can move it to CoreDataSnapshotTestCase later.

johnxnguyen: `CoreDataSnapshotTestCase` already has the `createGroupConversation()` helper, but I needed to configure my own group.

PR #3112, opened 2019-01-18
Convert MessageActionResponder and ConversationContentViewController+ConversationMessageCellDelegate to Swift 

billypchan:
## What's new in this PR?

This PR is a preparation of rework of https://github.com/wireapp/wire-ios/pull/3095. 

To get the source view when performing an action, I propose adding a sourceView argument in this method instead for recalculating the cell form the message argument.

```
protocol MessageActionResponder {
   func wants(toPerform action: MessageAction,
                       for message: ZMConversationMessage!,
                       sourceView: UIView!)
}
```


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/MessageActionResponder.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@objc public protocol MessageActionResponder: NSObjectProtocol {
+    @objc(wantsToPerformAction:forMessage:)
+    func wants(toPerform action: MessageAction, for message: ZMConversationMessage!)
```
typfel: So you'll add the `sourceView` argument here?

billypchan: I tried to add it but it affects quite a lots of caller/callee. Those changes will be done in seperated PR.

vytis: The naming maybe could be a bit more clear on the intention. For me it sounds somewhere between `should(perform action:)` (which then returns Bool depending if it should happen or not) and `perform(action)`. What's the actual purpose of the method? As I understand it should display the right popover over this message?

billypchan: @vytis about the naming we can refer to the original PR: https://github.com/wireapp/wire-ios/pull/552. I think shorten the method's name to `perform(action)` makes sences. And we can add sourceView argument to simpify the logic to obtain the view for UI logics.

The first method implement that method is:
```
- (void)wantsToPerformAction:(MessageAction)action forMessage:(id<ZMConversationMessage>)message
{
    if (MessageActionForward == action) {
        ConversationCell *cell = [self cellForMessage:message];
        [self showForwardForMessage:message fromCell:cell];
    }
}
```

vytis: Maybe `perform(action)` or `performIfNeeded(action)` would be more inline to the regular naming convention

billypchan:
@vytis I renamed the method and I will add the argument in a coming PR.

PR #3116, opened 2019-01-21
Feature/permissions invite button

johnxnguyen:
## What's new in this PR?

Hide the invite button at the start of a conversation for non authorized users.


Wire-iOS Tests/StartedConversationCellTests.swift
```diff
@@ -24,9 +24,9 @@ import XCTest
 class StartedConversationCellTests: ConversationCellSnapshotTestCase {
     
     override func setUp() {
+        selfUserInTeam = true
```
billypchan: It would be clearer to usee TeamTest helper method.

johnxnguyen: Ok, I would have to add it to all the tests (except one), but indeed it would be more clear :)

johnxnguyen:
Merging, tests passed but azure failed due to codecov step.

PR #3119, opened 2019-01-21
Calculate section and row update together using DifferenceKit

typfel:
### Issues

Tableview updates are not animating correctly when sending and receiving messages.

### Causes

We are updating the table view several times in a row for section and row updates. The reason for this is that the section updates comes from the fetch results controller and the row updates comes from the message section controller. 

### Solutions

Create a data struct for both section and rows and calculate the tableview diff using DifferenceKit (https://github.com/ra1028/DifferenceKit), which supports sections.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/651

### TODO

- [x] Search results highlights aren't cleared
- [x] Clean up `makeCell(for:)`


Cartfile
```diff
@@ -9,3 +9,4 @@ github "wireapp/HockeySDK-iOS" ~> 5.0
 github "wireapp/FLAnimatedImage" "1.0.12-wire"
 github "wireapp/PureLayout" "v3.0.0"
 github "wireapp/Down" "v2.1.0"
+github "ra1028/DifferenceKit" ~> 0.8
```
vytis: I guess needs to be moved to wireapp

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -1390,6 +1390,27 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
+		1642362D21F236DD00D810B5 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 1642362721F236DD00D810B5 /* WireDataModel.xcodeproj */;
```
vytis: Datamodel dragged in

Cartfile
typfel: 👍 I moved it just forgot to update the Cartfile

PR #3122, opened 2019-01-21
 Feature: permissions

billypchan:
## What's new in this PR?

Merge feature branch to develop branch.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/652

---
# Overview

These changes are for the **External Collaborator Feature**. A _collaborator_ is a new set of permissions that define a user role (existing roles include _owner_, _admin_, and _member_). Roles/permissions define which actions are available to a team user, where the owner has full permissions, and each subsequent role contains a subset of the permissions of the previous role. The _collaborator_ role has the fewest permissions, which, in a nutshell, allows them to do everything that a normal _member_ can do, except:

- create group conversations
- edit group details (conversation name, guests, timeout and read receipts)
- add or remove participants/services from a conversation.

## How Do We Manage Permissions?

As mentioned above, a _collaborator_ is simply a new set of permissions. This means we already have the data structures necessary contain permission information (see `Permissions` in DataModel). However, some superficial adjustments have been made in DataModel to provide a layer of abstraction that aims to clarify a mental distinction between **roles** and a set of individual **permissions**, whiles also improving testability and Objective C interoperability. This abstraction takes the form of a 1 to 1 mapping between four static variables of the `Permissions` option set and the `TeamRole` enum. There is now a `teamRole` property defined on the `UserType` protocol, which is concretely accessed through the potential `Member` object linked to a `ZMUser`.  See the link to the DataModel PR above for more information.

## What Does This Mean For The UI?

The _collaborator_ role does not allow any new behavior, only restrictions. As such, we need only to hide various UI components from collaborators to prevent them from executing actions. To do this, we will check the permissions of the self user and we will mark certain UI components as restricted.

## Checking Permissions 

Typically we're only concern about the self user and their role/permissions. To simplify the process of checking the role/permissions of the self user, we have provided some static properties/methods in an extension of the `ZMUser` class. Most useful is `ZMUser.selfUserHas(permissions:)`.

## Restricting Components

The `Restricted` protocol is designed to mark certain components as requiring "authorization" before being "used". It aims to bring all the permission checking logic into one place: defining what are the `requiredPermissions` and checking if the `selfUserIsAuthorized`. Admittedly, it is sometimes simpler to check the self user permissions directly rather than conforming to the `Restricted` protocol, but this, in my opinion, is due to some components being too closely tied to other components.

It is **important to note** that the `Restricted` protocol is only designed to restrict team users depending on their role. It does not attempt to restrict _team only features_ from non team members. Therefore, the usual `ZMUser.selfUserisTeamMember` checks are necessary to hide pro features.

## How Do We Test Permission Behavior?
Since we expose the team role on the `UserType` protocol, we can easily extend `MockUser` to include a team role property. By setting the role, we associate a set of permissions with the mock user, without having to mock team membership. 

If we're testing in the context of Core Data, one will need to set up the appropriate membership relationship between a `Team` and a `ZMUser`, then call the `setTeamRole()` method on the `Member` object. Remember, the individual permissions are accessible through the team role.

## Notes

While working on `GroupDetailsViewControllerSnapshotTests` it was found that most (possibly all) of the tests were incorrectly configured and recorded. Firstly they were configured with a one to one conversation, secondly they are missing the participants list. I still need to work on some of these issues, which will come later in a separate PR.


Wire-iOS Tests/ConversationDetail/GroupDetailsViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    func testForOptionsForNonTeamUser() {
+        nonTeamTest {
+            sut = GroupDetailsViewController(conversation: groupConversation)
+            verify(view: self.sut.view)
+        }
+    }
+
+    func testForActionMenu() {
+        // TODO: the menu is missing "mark unread" option at top
```
typfel: Still valid?

billypchan: It is a snapshot issue found by @johnxnguyen, we will check this later to see why the action menu missing an item in the test case.

johnxnguyen: Yeah, it was missing from the original snapshot tests (as well as the whole participants list). I will look into both of these issues tomorrow.

PR #3125, opened 2019-01-22
Fix: Partner can edit the Group name after navigate from other views(Notification/People/... button)

billypchan:
### Issues

Rename conversation cell is editable after the screen is reloaded.

### Causes

The textField's isUserInteractionEnabled property is set again after cell.configure() method is called.

### Solutions

Remove the unnecessary property settings.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/RenameGroupSectionController.swift
+++ b/Wire-iOS/Sources/UserInterface/GroupDetails/Sections/RenameGroupSectionController.swift
@@ -54,8 +54,6 @@ class RenameGroupSectionController: NSObject, CollectionViewSectionController {
         let cell = collectionView.dequeueReusableCell(ofType: GroupDetailsRenameCell.self, for: indexPath)
         cell.configure(for: conversation)
         cell.titleTextField.textFieldDelegate = self
-        renameCell?.titleTextField.isUserInteractionEnabled = conversation.isSelfAnActiveMember
-        renameCell?.accessoryIconView.isHidden = !conversation.isSelfAnActiveMember
         renameCell = cell
         return cell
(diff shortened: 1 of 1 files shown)


johnxnguyen:
@billypchan I'm wondering now if we actually needed this check `conversation.isSelfAnActiveMember`? Perhaps it's best to put this check in the `configure(for conversation:)` method.

Wire-iOS/Sources/UserInterface/GroupDetails/Sections/RenameGroupSectionController.swift
```diff
@@ -54,8 +54,6 @@ class RenameGroupSectionController: NSObject, CollectionViewSectionController {
         let cell = collectionView.dequeueReusableCell(ofType: GroupDetailsRenameCell.self, for: indexPath)
         cell.configure(for: conversation)
         cell.titleTextField.textFieldDelegate = self
-        renameCell?.titleTextField.isUserInteractionEnabled = conversation.isSelfAnActiveMember
-        renameCell?.accessoryIconView.isHidden = !conversation.isSelfAnActiveMember
```
billypchan: @johnxnguyen this check is duplicated in the `cell.configure(for: conversation)` above so it is removed.

johnxnguyen: Not exactly:

```
if !selfUserIsAuthorized {
    titleTextField.isUserInteractionEnabled = false
    accessoryIconView.isHidden = true
}
```

`selfUserIsAuthorized` just checks that the self user has permission, not if they're an active member in the conversation. I'm not entirely sure what that means, but it is stored in DM.

PR #3126, opened 2019-01-22
ZIOS-10740: Join call button is not visible if I leave a group call

nicorsm:
### Issues

Join button was not visible in conversation view and list when muted settings for the conversation was different than the default value (`.none`)

### Solutions

I've fixed the controls for the display logic, according to the specs. Still need to do a final test with all the cases.


Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -120,7 +120,7 @@
          </CommandLineArgument>
          <CommandLineArgument
             argument = "-ZMBackendEnvironmentType staging"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
vytis: Probably would be better to revert

nicorsm: Ouch

PR #3130, opened 2019-01-23
Fix: When I am downgraded from Member to Partner, I can still remove services

billypchan:
### Issues

The bottom button in ServiceDetailViewController is not hidden for partner role.

### Causes

The button is not conformed to Restricted protocol.

### Solutions

Create a new class RestrictedButton for this case. Refactor the RestrictedIconButton(which was named RestrictedButton).

## Todo

- [x] snapshot test for ServiceDetailViewController


Wire-iOS/Sources/Helpers/syncengine/ZMUser+Permissions.swift
```diff
@@ -71,3 +71,15 @@ extension Restricted {
         if selfUserIsAuthorized { onSuccess() }
     }
 }
+
+extension Restricted where Self: UIButton {
+    var shouldHide: Bool {
+        return ZMUser.selfUser().isTeamMember && !selfUserIsAuthorized
+    }
```
johnxnguyen: You don't need to check if the self user is a team member here, it's already covered by `selfUserIsAuthorized`.

PR #3131, opened 2019-01-24
Fix for un-sendable audio message

typfel:
### Issues

It's possible to trigger the hold & record audio message button in way that creates an audio message that can't be sent.

wire 2019-01-24 at 11 11 52

### Causes

The hold & record audio message feature is triggered by a long press which shows an inline audio recording UI. The UI is is not reflecting the truth because before we can start the recording we need to switch the audio session category to `.playAndRecord` which can take more than a second at the moment. The bug occurs if the user aborts the long tap within this second before the recording starts.

### Solutions

Don't show the recoding UI until the recording has a actually started.

### Notes

I think the delay when switching audio session category is not necessary and we should ask AVS to remove/reduce it.


vytis:
Also one thing I noticed is that the duration displayed in this UI turns out to be different than what displayed in conversation. Had trouble making a 10s audio clip for testing purposes, it was always 11s!

PR #3133, opened 2019-01-24
Support disabling newsletter and sharing data usage in settings

typfel:
## What's new in this PR?

Support `SEND_ANONYMOUS_DATA_DISABLED` and `NEWSLETTER_DISABLED` in settings by hiding the corresponding sections if these flags are present.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
@@ -33,12 +33,15 @@ extension ZMUser {
 extension SettingsCellDescriptorFactory {
 
     func accountGroup() -> SettingsCellDescriptorType {
-        var sections: [SettingsSectionDescriptorType] = [
-            infoSection(),
-            appearanceSection(),
-            privacySection(),
-            personalInformationSection(),
-            conversationsSection()]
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Account.swift
```diff
@@ -33,12 +33,15 @@ extension ZMUser {
...
-            infoSection(),
-            appearanceSection(),
-            privacySection(),
-            personalInformationSection(),
-            conversationsSection()]
+        var sections: [SettingsSectionDescriptorType] = [infoSection(),
+                                                         appearanceSection(),
+                                                         privacySection()]
+        
+        #if !SEND_ANONYMOUS_DATA_DISABLED || !NEWSLETTER_DISABLED
```
vytis: Maybe it's better to just keep one flag for all of this, I don't think we will need to toggle it separately. We are adding quite a lot of them already...

typfel: I was thinking about it, but I need a good name.

typfel: `DATA_COLLECTION_DISABLED`

PR #3137, opened 2019-01-24
Remove cartography form ServiceDetailViewController

billypchan:
## What's new in this PR?

A clean-up task of https://github.com/wireapp/wire-ios/pull/3130
Remove cartography form ServiceDetailViewController after the snapshot tests are created to confirm the UI constraints are correctly created.


Wire-iOS/Sources/UserInterface/Services/ServiceDetailViewController.swift
```diff
@@ -158,6 +137,29 @@ final class ServiceDetailViewController: UIViewController {
         }
     }
 
+    private func createConstraints() {
+        var topMargin: CGFloat = 16
```
alexaubry: I think we should avoid using the top margin here, and constrain to the safe area instead.

billypchan: @alexaubry updated as your suggestion and updated snapshot test with a navigation bar.

PR #3139, opened 2019-01-25
Fix: Change  ConversationContentViewController's wantsToPerformAction()'s argument type

billypchan:
## What's new in this PR?

This PR is the next step of  https://github.com/wireapp/wire-ios/pull/3112.

### Issues

ConversationContentViewController's `wantsToPerformAction()` method gets `cell:(UIView<SelectableView> *)cell` as an argument which causes some issues:
 - the cell may not be a UITableViewCell in all context e.g. it can be called from full screen image screen and there are no views on it conforms SelectableView protocol
- the cell argument is obtained from the line `dataSource.cell(for: message)`, but it may return nil in some cases, e.g. the cell is not fully visible.

### Causes

The method was originally designed for the conversation screen, but we use it in somewhere without UITableViewCell.

### Solutions

Change argument form SelectableView to UIView. The SelectableView protocol is for positioning the popover/activity view position on iPad. Optional casting the argument to SelectableView for those cases.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -0,0 +1,163 @@
...
+            })
+        case .delete:
+            assert(message.canBeDeleted)
+
+            deletionDialogPresenter = DeletionDialogPresenter(sourceViewController: presentedViewController ?? self)
+            deletionDialogPresenter.presentDeletionAlertController(forMessage: message, source: view) { deleted in
+                if deleted {
+                    self.presentedViewController?.dismiss(animated: true)
+                }
+                if !deleted {
```
vytis: This can probably go into `else` of the previous statement

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -0,0 +1,163 @@
...
+            delegate.conversationContentViewController(self, didTriggerEditing: message)
+        case .sketchDraw:
+            openSketch(for: message, in: .draw)
+        case .sketchEmoji:
+            openSketch(for: message, in: .emoji)
+        case .sketchText:
+            // Not implemented yet
+            break
+        case .like:
+            let liked = !Message.isLikedMessage(message)
```
vytis: it seems either the variable name should be `notLiked` or we don't need the negation there

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -0,0 +1,163 @@
...
+
+            if let indexPath = dataSource?.indexPath(for: message),
```
vytis: `guard let` would be a better choice and avoid one level of indentation in the rest of the code

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+SaveImage.swift
```diff
@@ -19,15 +19,23 @@
 import Foundation
 
 extension ConversationContentViewController {
-    @objc(saveImageFromMessage:cell:)
-    func saveImage(from message: ZMConversationMessage, cell: SelectableView?) {
+    func saveImage(from message: ZMConversationMessage, view: UIView?) {
         guard let imageMessageData = message.imageMessageData, let imageData = imageMessageData.imageData else { return }
         
         let savableImage = SavableImage(data: imageData, isGIF: imageMessageData.isAnimatedGIF)
```
vytis: typo here - should be `saveable`

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
```diff
@@ -609,6 +421,39 @@ - (void)activeMediaPlayerChanged:(NSDictionary *)change
...
+    }
+
+    if ([message isEqual:self.dataSource.selectedMessage]) {
+
+        // If this cell is already selected, deselect it.
+        self.dataSource.selectedMessage  = nil;
+        [self.dataSource deselectWithIndexPath:indexPath];
+        [tableView deselectRowAtIndexPath:indexPath animated:YES];
+
+        // Make table view to update cells with animation (TODO can be removed when legacy cells are removed)
```
vytis: I guess this was copied from other place, but do we still need this comment and the lines below?

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
billypchan: It is really annoying to read the code in this case. I will clean it up.

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+SaveImage.swift
billypchan: Will rename 52 occurrences in the project

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.m
vytis: I meant the comment here saying that "TODO can be removed when legacy cells are removed". What does it mean by legacy cells? Is it those from the recent refactoring or this comment is older than that? If it's recent refactoring maybe adding file names of what are those legacy cells at the moment would help. Otherwise we will keep this comment/TODO forever...

Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+MessageAction.swift
```diff
@@ -0,0 +1,162 @@
...
+            })
+        case .delete:
+            assert(message.canBeDeleted)
+
+            deletionDialogPresenter = DeletionDialogPresenter(sourceViewController: presentedViewController ?? self)
+            deletionDialogPresenter.presentDeletionAlertController(forMessage: message, source: view) { deleted in
+                if deleted {
+                    self.presentedViewController?.dismiss(animated: true)
+                } else {
+                    // TODO 2838: Support editing
```
vytis: Same thing here about the comment - it's a good opportunity to revisit this and either delete the comment or fix the issue

billypchan: I think we can delete it for now.

related PR:
https://github.com/wireapp/wire-ios/pull/1298
https://github.com/wireapp/wire-ios/pull/2862

PR #3147, opened 2019-01-30
ZIOS-11136: Removed setting of Audio Category for audio messages

nicorsm:
## What's new in this PR?

When we play an audio message we modify the audio session category, this confuses the media manager in AVS since it expects to be in control of the audio session.
I've removed the setting of audio category from `AudioMessageView` and `AudioRecorder` classes.


typfel:
Maybe we can simply disallow playing audio message while in a call.

PR #3153, opened 2019-02-01
Avoid save when fetching profile picture

vytis:
### Issues

When searching for people or scrolling a conversation we would make a save each time we fetch profile picture.

### Causes

We would wrap the method to request profile picture in `session.enqueueChanges`. This is not needed because the method simply sends a notification.


typfel:
That's true for V3 profile pictures but I think the save is currently necessary for V2 profile pictures since it relies on observing `localSmallProfileRemoteIdentifier` being set to nil.

PR #3155, opened 2019-02-01
Fix/ordered participants

vytis:
### Issues

Some APIs changed after merging https://github.com/wireapp/wire-ios-data-model/pull/666

### Solutions

Using unsorted participants where possible.


typfel:
Looks like tests failed to compile

PR #3157, opened 2019-02-04
Handle and test SSO URL actions

alexaubry:
### Issues

We want to be able to start the SSO flow when the user taps a special link. The detailed breakdown can be found at https://github.com/wireapp/ios-architecture/issues/64.

In this PR, we handle the changes from https://github.com/wireapp/wire-ios-sync-engine/pull/956 to handle new URL actions and tell the session manager if it should accept them or not.

### Solutions

We introduce a `CompanyLoginLinkResponseContext` protocol and a `CompanyLoginLinkResponseAction` enum. The context has some required state properties. An extension on the protocol returns the action to execute depending on this state. The actions are: allow starting flow, disallow starting flow and show error. This is the map of state and actions:

| Link | Number of accounts | User is adding account | Result |
|-----|:--------------------:|:----------------------:|--------|
| Valid | < 3 | No | Show the SSO browser |
| Valid | > 3 | No | Show "too many accounts" alert |
| Valid | -  | Yes | Ignore the link |
| Invalid | - | No | Show "invalid link" alert |
| Invalid | - | Yes | Ignore the link |

In the UI, we have a concrete implementation of the protocol that uses the state from the app root view controller to determine this context. This concrete implementation is used to provide the response to the session manager that calls our method to determine if it should run the action.

And in tests, we have a mock context that allows us to test the previously mentioned state → action map.

### Notes

This PR handles the negative cases (ZIOS-11164 and ZIOS-11165), but the positive case (opening the browser, ZIOS-11163) will require further SE work before it can work.

This PR will be merged to the feature branch.


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -627,11 +627,38 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
...
+
+    private func executeCompanyLoginLinkAction(_ action: CompanyLoginLinkResponseAction, callback: @escaping (Bool) -> Void) {
+        switch action {
+        case .allowStartingFlow:
+            callback(true)
+
+        case .disallowStartingFlow:
+            callback(false)
+
+        case.showDismissableAlert(let title, let message, let allowStartingFlow):
```
typfel: missing space after `case`

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -627,11 +627,38 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
...
+            executeCompanyLoginLinkAction(context.actionForInvalidRequest(error: error), callback: callback)
+        }
+    }
+
+    private func executeCompanyLoginLinkAction(_ action: CompanyLoginLinkResponseAction, callback: @escaping (Bool) -> Void) {
+        switch action {
+        case .allowStartingFlow:
+            callback(true)
+
+        case .disallowStartingFlow:
```
vytis: A nitpick, but maybe "prevent" would be a better verb?

alexaubry: Good call, I will change it.

PR #3160, opened 2019-02-05
Fix: iPad crashes when removing devices

billypchan:
### Issues

 iPad crashes when removing devices

### Causes

Some typo in the code of constraints

### Solutions

Fix the typo and create snapshots to check the appearance on all devices

## TODO

- [ ] Check iPhone 4 inch layout


Wire-iOS/Sources/Authentication/Interface/ViewControllers/RemoveClientStepViewController.swift
```diff
@@ -73,34 +70,48 @@ class RemoveClientStepViewController: UIViewController, AuthenticationCoordinate
...
+        updateConstraints(userInterfaceSizeClass: traitCollection.horizontalSizeClass)
     }
 
 }
 
+
+extension UITraitEnvironment {
+    func toggle(compactConstraints: [NSLayoutConstraint],
+                regularConstraints: [NSLayoutConstraint],
+                userInterfaceSizeClass: UIUserInterfaceSizeClass) {
```
alexaubry: It would be nice to pass `isRegular` in the function arguments directly, as this would let us customize the size class in tests.

alexaubry: I would also move this to a separate file in the Helpers, as it can be used in other view controllers too.

Wire-iOS/Sources/Authentication/Interface/ViewControllers/RemoveClientStepViewController.swift
```diff
@@ -43,6 +43,10 @@ class RemoveClientStepViewController: UIViewController, AuthenticationCoordinate
                                                         variant: .light)
 
         super.init(nibName: nil, bundle: nil)
+
+        title = "registration.signin.too_many_devices.manage_screen.title".localized(uppercased: true)
```
alexaubry: Is there a reason to do this outside `viewDidLoad`? I think it could be a bit expensive, as we would end up loading the view in the initializer by calling `configureSubviews`.

Wire-iOS/Sources/Authentication/Interface/ViewControllers/RemoveClientStepViewController.swift
billypchan: @alexaubry I prefer passing a UIUserInterfaceSizeClass since in past time I encountered some situation that it has a value .unspecified. I rewrite the logic with a switch for that.

PR #3165, opened 2019-02-05
ZIOS-11173: Search bar overlaps first item inside country code selection

nicorsm:
## What's new in this PR?

When searching for a country code inside the signup flow (`CountryCodeTableViewController`), the first item is always shown behind the search bar. I've added a header as padding.


typfel:
This can't be solved by `contentInsetAdjustmentBehavior` and `automaticallyAdjustsScrollViewInsets `?

nicorsm:
@typfel tested right now, doesn't seems to work

michaeldong:
#3204

PR #3169, opened 2019-02-06
Fix: Obfuscated view not shown after an image is expired

billypchan:
### Issues

Obfuscated view not shown after an image is expired.

### Causes

isObfuscated is got from local struct Configuration, it is not updated after message object is updated.

### Solutions

Get the isObfuscated from the message.isObfuscated.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Rich Content/ConversationImageMessageCell.swift
```diff
@@ -98,8 +98,17 @@ class ConversationImageMessageCell: UIView, ConversationMessageCell {
...
     
     func configure(with object: Configuration, animated: Bool) {
-        obfuscationView.isHidden = !object.isObfuscated
-        
+        let showObfuscationView: Bool
+
+        if let isObfuscated = message?.isObfuscated {
+            showObfuscationView = isObfuscated
+        } else {
+            showObfuscationView = object.isObfuscated
```
typfel: Why not just rely on `message?.isObfuscated` and default to false if message is nil?

PR #3173, opened 2019-02-07
ZIOS-11170: Do not show the consent alert if the user logs in again

alexaubry:
### Issues

Previously, we'd show the "Help us make Wire better" alert when the user logs in a second time with the same account on the same device. It was decided to not do this anymore.

### Solutions

We update the logic of checking if the alert needs to be shown. We only show it if you "add a new account" (login / registration).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
@@ -32,10 +32,7 @@ extension ConversationListViewController {
         guard needToShowDataUsagePermissionDialog else { return }
 
         // If the user registers, show the alert.
-        // If the user logs in and hasn't accepted analytics yet, show the alert.
-        guard isComingFromRegistration ||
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+DataUsagePermissions.swift
```diff
@@ -60,7 +57,4 @@ extension ConversationListViewController {
         #endif
     }
 
-    private var userAcceptedAnalytics: Bool {
```
typfel: Why is not necessary to query this anymore?

typfel: Because it could only happen during login?

alexaubry: That check made the alert show for every login or registration if the user didn't accept the analytics, regardless of this is the first or second time they log in, which is the behavior we wanted to change. Now we only ask them the first time, and if they refuse and log in a second time, we don't ask them again.

Issue #3156, opened 2019-02-03
When tapping an old search result in history, Wire stops responding


Dexwell:
## Checklist (please check before submitting)

* [X] I checked the existing issues, and the issue was not reported before.

## The problem

When searching my (large) chat history and tapping a result from e.g. May 2018, the modal closes and then the app stops responding until I force close it.

## Environment

* Wire version that exhibits the issue: All latest versions, for a while now
* Last Wire version that did not exhibit the issue (if applicable): Not sure, before Direct Replies though
* iOS version used to run Wire: 12.1.2
* iOS device type: iPhone XS
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE, Wi-Fi

billypchan:
We had a lot of fixes and improvements and I can not reproduce this issue and I close this issue.

PR #3174, opened 2019-02-07
Improve conversation list scrolling performance

vytis:
### Issues

On older devices scrolling conversation list is quite choppy.

### Solutions

Multiple small issues were isolated:
* Setting various values on `UserImageView` and `AvatarImageView` would always redraw everything.   
     * **Solution:** redraw only if the value being set is different from old one.
* Creating empty attributed strings with helpers. 
     * **Solution:** Use `NSAttributedString()` instead.
* Posting a notification when conversation list scrolls. 
     * **Solution:** Don't post it and remove unused code that was listening to it.
* `prepareForReuse` resets all values and triggers a redraw. 
     * **Solution:** Remove the methods as we set all the values when dequeuing the cells anyway.

### Result

On iPhone 6 scrolling conversation list up and down when from ~24fps to ~30fps.


cardassian-tailor:
is this related in any way to  #3156

vytis:
@cardassian-tailor no not related. The issue described there is something that we are working on separately. This should improve situation when scrolling through conversations, but does not touch what happens inside a conversation or search.

Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
```diff
@@ -47,7 +47,7 @@
             }
         }
         
-        guard let textColor = color else { return "".attributedString }
+        guard let textColor = color else { return NSAttributedString() }
```
billypchan: Maybe more make sense to change the return type of this method to `NSAttributedString?`. 3 callers of this method are use the return as NSAttributedString?.

vytis: Will take a look, thanks!

vytis: @billypchan updated!

PR #3176, opened 2019-02-07
Fix: Update ProfileSelfPictureViewController constraints and fix orientation can change issue

billypchan:
## What's new in this PR?

This PR is the first step of old constraint creation helper (NSLayoutConstraint+Helpers.h/m).

### Issues

ProfileSelfPictureViewController is possible to rotate on iPhone.

### Causes

supportedInterfaceOrientations is not implement in ProfileSelfPictureViewController.

### Solutions

Implement `supportedInterfaceOrientations` and return `wr_supportedInterfaceOrientations`

### Note
Pure layout and old helper methods are replaced with updated constraint helpers.
Update wr_supportedInterfaceOrientations with simpler logic.
Created snapshots for ProfileSelfPictureViewController.


Wire-iOS/Sources/UserInterface/Helpers/UIViewController+Orientation.swift
```diff
@@ -19,21 +19,15 @@
...
-                return .all
-            }
-        case .phone:
-            return .portrait
+    /// return the default supported interface orientations of a view controller
+    /// return .all only if the idiom is .pad and size class is .regular
+    @objc var wr_supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        switch (UIDevice.current.userInterfaceIdiom, traitCollection.horizontalSizeClass) {
+        case (.pad, .regular),
+             /// Notice: for iPad with iOS9 in landscape mode, horizontalSizeClass is .unspecified (it is .regular in iOS11).
```
vytis: We do not support iOS9 anymore, maybe it's worth removing this case alltogether if this comment is still valid (probably worth checking on iOS10 at least).

PR #3177, opened 2019-02-07
Improve image messages scrolling performance

vytis:
### Issues

When profiling the app some performance issues were isolated in conversation view while scrolling image messages.

### Causes

When cell was dequeued we were re-calculating `accessibilityCustomActions` unnecessary. Most of the work was done in `ZMConversationMessage.isImage`.

### Solutions

Cache possible actions for the message to avoid recalculating it.


typfel:
I can imagine we need to invalidate this cache for certain actions like `canBeDownloaded`.

vytis:
Good point, was too quick to solve it in a blunt way. Maybe it's better to address the underlying issue of why querying all of this is slow. Closing the PR until further investigation!

PR #3178, opened 2019-02-07
ZIOS-11188: Fix keyboard jumping and unresponsive button

alexaubry:
### Issues

- When pushing or popping a view controller in the login UI, the bottom buttons could be under the keyboard
- When popping a view controller in the login UI, the content was moved to the center and not adapted to the keyboard frame
- When submitting and email or phone number on the registration screen, the keyboard would appear again after accepting the TOS
- When registering with an existing email, the "log in" button would do nothing
- The wrong button was set as destructive in the logout popup

### Solutions

- Start observing keyboard changes after the view was presented but before we show the keyboard
- Remove multiple calls to the action handling method from email and phone fields
- Fix bottom padding calculation
- Add missing actions to handle logging in from registration


alexaubry:
Manual build passed (https://dev.azure.com/wireswiss/Wire%20iOS/_build/results?buildId=3094)

PR #3181, opened 2019-02-08
Feature: Conversation detail profile screen - unify bottom tool bar 

billypchan:
## What's new in this PR?

Unify the bottom bar UI design for both 1-to-1 conversation and group conversation user profile screen. Both of them shows a 3-dot button now. When pressing the button for group conversation, an action menu remove item shows.


alexaubry:
@billypchan We should merge this PR to a feature branch that’s branched off of the current release branch.

billypchan:
@alexaubry I will rebase this PR to the next coming release.

Wire-iOS/Sources/UserInterface/GroupDetails/ConversationActions/ConversationActionController.swift
```diff
@@ -16,17 +16,96 @@
 // along with this program. If not, see http://www.gnu.org/licenses/.
 //
 
-@objcMembers final class ConversationActionController: NSObject {
-    
-    struct PresentationContext {
-        let view: UIView
-        let rect: CGRect
+@objc protocol ActionController: NSObjectProtocol {
```
typfel: If the `ActionController` had 
```
var actions: [Conversation.Action]
func handleAction(_ action: Conversation.Action) 
```

then the methods for creating and presenting the `UIAlertViewController` could be defined on the protocol level.   it would also make it easier to test that the expected actions are present without having to do a snapshot test.

billypchan: I had thought about that before but there is a problem that ActionController must be defined as a `NSObjectProtocol` (a objc class ProfileDetailsViewController has a member of this type) and Swift enum type Conversation.Action is not allowed. I will refactor this after `ProfileDetailsViewController` is converted to Swift.

PR #3184, opened 2019-02-11
Feature: disable self-profile editing

billypchan:
### Issues

Disable self-profile setting email, name, phone number, handle, accent color & profile picture editing for certain type user.
In the self-user setting screen, when profile editing is disabled, the appearance section is hidden.
Created UserRight class to centralize the UI permission control.

## Dependencies

Needs releases with:

- [x] Update Data model.


alexaubry:
@billypchan Let's make a feature branch for this (branched off of the release branch), as we don't want to merge the locked profile changes in develop, or in the release branch until the SSO link feature is validated.

WireCommonComponents/SettingsPropertyName.swift
```diff
@@ -43,10 +43,14 @@ public enum SettingsPropertyName: String, CustomStringConvertible {
     case disableSendButton = "DisableSendButton"
     
     case disableLinkPreviews = "DisableLinkPreviews"
-    
+
```
alexaubry: Extra whitespace

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -23,9 +23,12 @@ import SafariServices
...
     static let settingsDevicesCellIdentifier: String = "devices"
     let settingsPropertyFactory: SettingsPropertyFactory
-        
-    init(settingsPropertyFactory: SettingsPropertyFactory) {
+    let userRightInterfaceType: UserRightInterface.Type
+
+    init(settingsPropertyFactory: SettingsPropertyFactory,
+		 userRightInterfaceType: UserRightInterface.Type = UserRight.self) {
         self.settingsPropertyFactory = settingsPropertyFactory
+		self.userRightInterfaceType = userRightInterfaceType
```
alexaubry: Extra indentation

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -0,0 +1,62 @@
...
+             .editProfilePicture,
+             .editAccentColor:
+			return isProfileEditable
+        }
+
+        return false
+    }
+	
+	private static var isProfileEditable: Bool {
+		return ZMUser.selfUser()?.managedBy == "wire"
```
typfel: `managedBy` is optional therefore it might be `nil`, shouldn't we default to allow editing if this field isn't set?

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -0,0 +1,62 @@
...
+             editAccentColor
+    }
+
+    static func selfUserIsPermitted(to permission: UserRight.Permission) -> Bool {
+        switch permission {
+        case .editEmail:
+        #if EMAIL_EDITING_DISABLED
+            return false
+        #else
+			return isProfileEditable
```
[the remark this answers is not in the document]
billypchan: side effect of using another computer :)

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
billypchan: Since all the case other than ZMUser.selfUser()?.managedBy == "wire" disables editing, set default to not allow editing is correct.

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -0,0 +1,62 @@
...
+        case resetPassword,
+             editName,
+             editHandle,
+             editEmail,
+             editPhone,
+             editProfilePicture,
+             editAccentColor
+    }
+
+    static func selfUserIsPermitted(to permission: UserRight.Permission) -> Bool {
```
typfel: I find it would be nicer to return a `OptionSet`of `UserEditPermission` from a method on the `UserType` or something similar.  But maybe we can discuss this after we've gotten the release out.

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
typfel: Do we re-fetch the self user when upgrading, otherwise this would disable editing of all regular users?

typfel: I also don't think it's a good idea to do this string comparison in the UI. The logic for deciding if the user is managed by  us or SCIM seem like something which should be done in the data model.

I propose making `managedBy` an internal data model property and expose a boolean property to the UI like `managedBySCIM` or similar.

billypchan: I will add a fallback value to allow editing in that case.

nicorsm: @billypchan @typfel since all other cases than "wire" should be considered as non-editable, should I create a `managedByUser` or `managedByWire` property instead?

typfel: Yes, in that case `managedByWire` sounds like a good name.

PR #3188, opened 2019-02-12
Add developer option for re-calculating your badge count

typfel:
## What's new in this PR?

Some internal users have an incorrect badge count, this PR adds a developer option to re-calculate it.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/677


vytis:
Not directly related, but one thing that bothered me with "find first unread conversation" was that tapping OK should probably take you to this conversation. Not sure if it's easy to do from there, but would make the option a bit easier to use.

PR #3189, opened 2019-02-12
Missing call overlay after first login

typfel:
### Issues

Call overlay is missing when making a call directly after logging in.

### Causes

`transitionToLoggedInSession()` is no longer called when transitioning to the authenticated state after logging in. The reason for this is that associated value `completedRegistration` has changed meaning to "added new account".

### Solutions

I don't see a reason why we should not always call `transitionToLoggedInSession()` when transitioning to the authenticated state, so I'll just ignore the value of `completedRegistration`.

## Notes

- Seems like good idea to rename `completedRegistration` to `addedAccount`


alexaubry:
@typfel Let's include it into the release, I will cherry-pick it into the release branch after you merge it.

PR #3194, opened 2019-02-13
ZIOS-11219: Update accessibility traits for disabled settings text field

alexaubry:
### Issues

The automation suite needs a way to check if a text field is disabled in the settings Account screen.

### Solutions

We change the accessibility properties of the text field cells used in the Accounts screen:

- The accessibility ID of the field will be `(property name)Field` (ex: `UsernameField`) where property name is the text of the label visible next to the field
- If the user can change the value, the element type is "text field" and it can be edited
- If the user cannot change the value, the element type is "static text" and it cannot be edited

### Note

We also add the missing logic to check if a user can reset their password (was TODO previously).

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
@@ -34,16 +34,18 @@ final class UserRight: UserRightInterface {
     }
 
     static func selfUserIsPermitted(to permission: UserRight.Permission) -> Bool {
+        let selfUser = ZMUser.selfUser()
+        let usesCompanyLogin = selfUser?.usesCompanyLogin == true
+        
         switch permission {
         case .editEmail:
         #if EMAIL_EDITING_DISABLED
             return false
         #else
-            return isProfileEditable
+            return isProfileEditable && !usesCompanyLogin
         #endif
         case .resetPassword:
-        	///TODO: For SSO user we don't allow setting or resetting the password
-            break
+            return isProfileEditable && !usesCompanyLogin
         case .editName,
              .editHandle,
              .editPhone,
@@ -51,8 +53,6 @@ final class UserRight: UserRightInterface {
              .editAccentColor:
 			return isProfileEditable
         }
-
-        return false
     }
 	
 	private static var isProfileEditable: Bool {
--- a/Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
@@ -42,10 +42,10 @@ final class UserRight: UserRightInterface {
         #if EMAIL_EDITING_DISABLED
             return false
         #else
-            return isProfileEditable && !usesCompanyLogin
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -34,25 +34,25 @@ final class UserRight: UserRightInterface {
...
         #if EMAIL_EDITING_DISABLED
             return false
         #else
-            return isProfileEditable
+            return isProfileEditable && !usesCompanyLogin
         #endif
         case .resetPassword:
-        	///TODO: For SSO user we don't allow setting or resetting the password
-            break
+            return isProfileEditable && !usesCompanyLogin
```
billypchan: @alexaubry I think `isProfileEditable` is not the necessary condition for reset password permission. And this user right checker I am going to apply here:

`SettingsCellDescriptorFactory+Account.swift`:
```
        if let user = ZMUser.selfUser(), !user.usesCompanyLogin {
            sections.append(actionsSection())
        }
```

alexaubry: So checking for `usesCompanyLogin` would be enough?

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -34,25 +34,25 @@ final class UserRight: UserRightInterface {
     }
 
     static func selfUserIsPermitted(to permission: UserRight.Permission) -> Bool {
+        let selfUser = ZMUser.selfUser()
+        let usesCompanyLogin = selfUser?.usesCompanyLogin == true
```
billypchan: I would prefer not using `let` but a static computed var here since it may be updated after the app started.

alexaubry: These are local variables though, they will be recalculated every time the function is called.

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
```diff
@@ -34,25 +34,25 @@ final class UserRight: UserRightInterface {
...
+        
         switch permission {
         case .editEmail:
         #if EMAIL_EDITING_DISABLED
             return false
         #else
-            return isProfileEditable
+            return isProfileEditable && !usesCompanyLogin
```
billypchan: Similar to below, `usesCompanyLogin` is not a pre-requirement to allow editing email. We can discuss the relationship between the newly introduced `managedBy` and `usesCompanyLogin` as they seem alike.

alexaubry: As discussed, updated to use `!usesCompanyLogin` only

Wire-iOS/Sources/UserInterface/Settings/UserRight.swift
alexaubry: As discussed, updated to use `!usesCompanyLogin` only

PR #3199, opened 2019-02-14
merge Release/3.27 to develop

billypchan:
## What's new in this PR?

merge Release/3.27 to develop


alexaubry:
We should not merge this until we have the release out. Maybe we should also revert your two latest commits, in case we need to make changes to the release branch after QA finishes testing.

PR #3203, opened 2019-02-15
Fix: bottom empty space issue when showing keyboard on iPad

billypchan:
### Issues

On entering the verification screen (on iPad), there is an empty space between the content and the keyboard

### Causes

KeyboardAvoidingViewController's bottom inset calculation does not consider the view's top offset > 0 cases.

### Solutions

Calculate bottom offset with handling of modalPresentationStyle == .formSheet case.


Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Extension.swift
```diff
@@ -61,4 +68,21 @@ extension KeyboardAvoidingViewController {
...
+    func createInitialConstraints() {
+        viewController.view.translatesAutoresizingMaskIntoConstraints = false
+        let constraints = viewController.view.fitInSuperview(with: EdgeInsets(top: topInset, leading: 0, bottom: .nan, trailing: 0),
+                                                             exclude: [.bottom])
+
+        topEdgeConstraint = constraints[.top]
+
+        let bottomEdgeConstraint = viewController.bottomLayoutGuide.bottomAnchor.constraint(equalTo: bottomLayoutGuide.bottomAnchor)
+
+        NSLayoutConstraint.activate([bottomEdgeConstraint])
```
alexaubry: Here, we can do:
```suggestion
        bottomEdgeConstraint.isActive = true
```

PR #3204, opened 2019-02-16
Maybe better solution: 'Search bar overlaps first item inside country code selection'  (#3165)

michaeldong:
Maybe not need to set tableview contentinset to fix this issue.


typfel:
Thanks @michaeldong this is indeed a nice solution.

PR #3206, opened 2019-02-18
ZIOS-11238: Improve profile and availability view configuration

alexaubry:
### Issues

We currently use two different views for the profile view: one in the conversation profile details, and one in the self profile. At the end of this run, we want to dosplay both as table view headers. To make this process easier, it is possible to reuse the same view in both contexts and adapt the elements we display depending on the context.

### Solutions

We introduce an `Options` option set to the `ProfileView` and the `AvailabilityTitleView`. We can use these to hide elements (such as the user name, handle, team name, availability, ...) and disable some user interactions (such as setting the availability).

We also add tests for the profile view.


Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -16,106 +16,156 @@
...
-            let variant = ColorSchemeVariant.dark
-            titleColor = UIColor.from(scheme: .textForeground, variant: variant)
-            titleColorSelected = UIColor.from(scheme: .textDimmed, variant: variant)
-        } else {
-            //otherwise, take the default variant
-            titleColor = UIColor.from(scheme: .textForeground)
-            titleColorSelected = UIColor.from(scheme: .textDimmed)
-        }
+        /// Whether to use the apparence for dark mode (light text).
+        static let useDarkAppearance = Options(rawValue: 1 << 1)
```
typfel: Maybe it's better to control this with a separate property. We already have `Themable` protocol which in other places we use in conjunction with `UIAppearance`.

alexaubry: @typfel Updated.

PR #3210, opened 2019-02-20
[Feature] Update profile details screen to display rich profile info

alexaubry:
### Issues

Users that log in with SCIM will be allowed to have extended metadata attached to their profile (see 
https://github.com/wireapp/wire-ios-data-model/pull/681). It will consist in a set of key/value pairs that we receive from the backend.

We want to display this data inside the profile details when the user taps on another user of their team in a the conversation details.

screenshot of the implemented feature

### Solutions

- We implement a `ProfileDetailsContentController` type that is responsible for calculating the content to display in the profile details: the SCIM info, the read receipts settings status, both or neither. (ZIOS-11237)
- We rewrite the `ProfileDetailsViewController` to use a table view, where the header is the profile view (reused from the settings), with the user profile image and availability, if applicable (ZIOS-11199). The content controller is the delegate and data source of this table view. The VC is only responsible for layout operations.
- We create a new cell type to display the extended metadata. (ZIOS-11236)
- We use a table view footer text to display the read receipts status. (ZIOS-11239)

Writing tests for this feature also required changes in the testing infrastructure:

- We added test cases for the matrix of permissions for seeing SCIM info and details, based on viewer/target/conversation type. This class uses Mock users directly, instead of using Core Data and `ZMUser`. This is possible because the profile details screen takes a generic `UserType` as an argument, instead of `ZMUser`.
- We create a base test class to facilitate creating Mock users.
- We fix a behavior in tests where the blue accent UIColor was always returned for Mock users, instead of reading it from the accent color value. This change has repercussions on tests that falsely displayed the blue accent color.

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/683

### Notes

One side effect of this PR is that it removes the action buttons from the profile details. As @nicorsm is working on the updates in this area separately (moving it to a floating bar), we will merge the current PR in a feature branch, until this part of the refactor is done.

Secondly, the format used in `ProfileDetailsContentController` to parse the dictionary (`key` and `value`) is not fixed, due to missing backend spec. We will probably provide an abstraction for it in data model once the structure is fixed.

Finally, due to missing BE support, we need to include some fake data for QA to be able to test their automation. The current process for making such builds is: turning on the flag, building from Xcode and sending them a custom-made IPA. The related code to support this will be removed when the BE is ready and they are able to configure the SCIM environment with the appropriate data in Appium.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -745,7 +745,9 @@
 
 "profile.read_receipts_enabled_memo.header" = "YOU HAVE ENABLED READ RECEIPTS";
 "profile.read_receipts_disabled_memo.header" = "YOU HAVE DISABLED READ RECEIPTS";
-"profile.read_receipts_memo.body" = "If both sides turn on read receipts, you can see when messages are read.\nYou can change this option in your account settings.";
+"profile.read_receipts_memo.body" = "If both sides turn on read receipts, you can see when messages are read.\n\nYou can change this option in your account settings.";
+
+"profile.extended_metatadata.header" = "Information";
```
typfel: typo `metatadata` > `metadata`

Wire-iOS Tests/Mocks/MockUserTests.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Creates a self-user with the specified name and team membership.
+     * - parameter name: The name of the user.
+     * - parameter inTeam: Whether the user is a member of the current team.
+     * - returns: A configured mock user object to use as a self-user.
+     * - note: The accent color of a self user is red by default.
+     */
+    
+    func createSelfUser(name: String, inTeam: Bool) -> MockUser {
```
typfel: If these methods were not in a XCTestCase subclass it would be easier to re-use them in other tests.

alexaubry: Would it be okay to make them static methods on `MockUser` instead?

alexaubry: Updated.

PR #3211, opened 2019-02-21
ZIOS-11230: Action Bar: Display as a Floating Toolbar

nicorsm:
### Issues

We wanted to update the design of the action buttons in the profile details to be a bar that floats at the bottom of the screen and is pinned between user details and devices screens.

### Solutions

First, we did a refactoring of both `ProfileFooterView` and `GroupDetailsFooterView`. Since `GroupDetailsFooterView` is what we want to achieve (a floating bar on top of "Profile" and "Devices" tabs of the Profile View Controller), I've started from there, moving the common logic into `ConversationDetailFooterView`. 

I've also moved `ProfileFooterView` to `ProfileViewController` instead of `ProfileDetailViewController`. In this way the footer view belongs to the parent view and is floating between views.

We also introduce an object, `ProfileActionsFactory`, that is responsible for returning the list of actions based on the relationship between the user and the viewer (self-user). We add unit tests for it, to make sure we respect the matrix of permissions.


Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsFooterView.swift
```diff
@@ -19,81 +19,49 @@
...
-        NSLayoutConstraint.activate([
-            heightAnchor.constraint(equalToConstant: 56),
-            addButton.centerYAnchor.constraint(equalTo: centerYAnchor),
-            moreButton.centerYAnchor.constraint(equalTo: centerYAnchor),
-            addButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 16),
-            moreButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -16),
-            moreButton.leadingAnchor.constraint(greaterThanOrEqualTo: addButton.leadingAnchor, constant: 16)
-        ])
+    func update(for conversation: ZMConversation) {
+        leftButton.isHidden = ZMUser.selfUser().isGuest(in: conversation)
```
alexaubry: Can partners also add to a group?

Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsFooterView.swift
```diff
@@ -19,81 +19,52 @@
 import UIKit
 
 protocol GroupDetailsFooterViewDelegate: class {
-    func detailsView(_ view: GroupDetailsFooterView, performAction: GroupDetailsFooterView.Action)
+    func footerView(_ view: GroupDetailsFooterView, performs action: GroupDetailsFooterView.Action)
```
typfel: Why the extra `s` in performs?

Wire-iOS/Sources/UserInterface/GroupDetails/ConversationDetailFooterView.swift
```diff
@@ -0,0 +1,124 @@
...
+            leftButton.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 16),
+            rightButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -16),
+            rightButton.leadingAnchor.constraint(greaterThanOrEqualTo: leftButton.leadingAnchor, constant: 16)
+        ])
+    }
+
+    // MARK: - Events
+    
+    @objc func setupButtons() {
+        // no-op
```
typfel: If it's essential that these methods are overridden we could add a `fatal("Should be overridden in subclasses")`

Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsFooterView.swift
alexaubry: I didn't notice it when reviewing, will fix.

Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsFooterView.swift
alexaubry: Apparently not, will update to avoid a regression.

PR #3223, opened 2019-02-28
Fix: remove mockUserFor method and correct argument types

billypchan:
## What's new in this PR?

This PR is part of the MockUser Swift conversion.

### Issues

MockUser has a `mockUserFor` method which is misleading and unnecessary.

### Causes

MockUser.mockUsers() actually returns a MockUser array, not ZMUser array so we have to cast it back to MockUser. The `mockUserFor` just cast it back to MockUser.

### Solutions
Refactor the code and remove `mockUserFor` method.

Change UserConnectionView's init method argument type to UserType. (In tests it accepts MockUser as an argument)


vytis:
So why do we have `MockUser.mockUsers` returning `ZMUser` instead of `MockUser`? Maybe it would be better to clean up the API there and we would not have to do this extra casting just to make Swift happy?

billypchan:
@vytis It is pretty tricky here. There are several places mix using MockUser and ZMUser, i.e. some method say it returns MockUser but actually it returns ZMUser and vice versa. I will clean up those methods.

PR #3225, opened 2019-02-28
ZIOS-10631: Deleted account shows Open Conversation button

nicorsm:
### Issues

The "Open Conversation" button was visible on the Profile footer view of a deleted user, due to an unhandled case in `ProfileActionsFactory`.

### Solutions

I'm returning an empty array if the user has the `isAccountDeleted` property set to true.


alexaubry:
@nicorsm Just FYI, we decided not to include this in 3.28.

PR #3234, opened 2019-03-06
Added build type parameter to prepare step

vytis:
### Issues

For internal builds we need to override app configuration by passing a parameter to `setup.sh`.

### Solutions

Added a build type parameter to Fastlane `prepare` stage that copies configuration from predefined folder. https://github.com/wireapp/wire-ios-build-assets should be checked out to project directory before running this step.


vytis:
CI is having some problems, tried locally and it seems to work fine 🤷‍♂️

PR #3240, opened 2019-03-11
Fix: Clean up PureLayout from Audio player

billypchan:
## What's new in this PR?

Clean up PureLayout from Audio player and its subviews.
Added snapshot tests for failed/empty/track info loaded states of AudioTrackViewController.

Fix the missing icon when the audio track fails to load.


Wire-iOS Tests/AudioPlayer/AudioTrackViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,71 @@
...
+        createSutWithTrack()
+
+        verify(view: sut.view)
+    }
+
+    func testForErrorView() {
+        createSutWithTrack()
+        
+        sut.audioTrackView.failedToLoad = true
+//        sut.audioTrackView.layoutIfNeeded()
```
vytis: Some commented out code left

Wire-iOS Tests/AudioPlayer/AudioTrackViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,71 @@
...
+    override func setUp() {
+        super.setUp()
+    }
+    
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+
+    func testForInitState(){
```
vytis: The snapshot is empty transparent image for this one - is that expected? Maybe we don't need the snapshot then?

vytis: A sorry the empty snapshot is `AudioErrorViewSnapshotTests/testForInitState@2x.png`

Wire-iOS/Sources/Components/AudioPlayer/AudioErrorView.m
```diff
@@ -18,29 +18,25 @@
...
+#import "AudioErrorView+Internal.h"
+#import "Wire-Swift.h"
 
-@interface AudioErrorView ()
-@property (nonatomic) UIImageView *errorIconView;
-@end
+@import WireCommonComponents;
 
-@implementation AudioErrorView
+@implementation AudioErrorView ///TODO: test
```
vytis: I think it's already tested?

Wire-iOS Tests/AudioPlayer/AudioTrackViewControllerSnapshotTests.swift
billypchan: @vytis from the source code I found it should show an icon on AudioErrorView. But I tried the original and fixed source code and that the icon is not drawn. I think it should be a UI issue?

Wire-iOS/Sources/Components/AudioPlayer/AudioTrackView+Constraint.swift
```diff
@@ -0,0 +1,52 @@
...
+            artworkImageView.fitInSuperview()
+            playPauseButton.fitInSuperview()
+        }
+
+        super.updateConstraints()
+    }
+
+    @objc
+    var errorView: AudioErrorView {
+        if nil == lazyErrorView {
```
vytis: Maybe we can use regular `lazy` variable here instead of having two properties? Or it doesn't work for `@objc` variables? If it does not then another option would be to have an `@objc` getter that simply returns the `lazy` swift variable.

billypchan: lazy is not allowed here for the error "'lazy' must not be used on a computed property". I would refactor this part into a method like `setupErrorView`. Lazy does not really save a lot in this case.

Wire-iOS/WireBridgingHeader.h
```diff
@@ -175,7 +175,14 @@
...
 #import "MediaPlaybackManager.h"
 #import "AudioTrackViewController.h"
+#import "AudioTrackViewController+Internal.h"
 #import "AudioPlaylistViewController.h"
+#import "AudioHeaderView.h"
+#import "AudioHeaderView+Internal.h"
+#import "AudioTrackView.h"
+#import "AudioTrackView+Internal.h"
+#import "AudioErrorView.h"
+#import "AudioErrorView+Internal.h"
```
vytis: Probably not relevant for this PR, but we should as much as possible avoid adding more things to bridging header. It has a major impact to compile times and sometimes it's maybe worth converting full class to Swift

Wire-iOS/Sources/Components/AudioPlayer/AudioTrackView+Constraint.swift
vytis: It should be initialized with a closure:
```
@objc lazy var errorView: AudioErrorView = {
    <code to create view>
    return view
}()

Wire-iOS Tests/AudioPlayer/AudioTrackViewControllerSnapshotTests.swift
billypchan: @vytis I found that there is a missing icon resource in the case and I will ask the design team for the update.

Wire-iOS/Sources/Components/AudioPlayer/AudioTrackView+Constraint.swift
billypchan: I will rewrite this after AudioTrackView is translated to Swift. (It does not compile for "Extensions must not contain stored properties")

vytis: I think the current approach with just calling `setupErrorView` is fine, we do that in quite a few places already anyway

PR #3241, opened 2019-03-12
Fix: text message cell - accessibilityIdentifier is not updated after timeout

billypchan:
### Issues

The text message cell's accessibilityIdentifier is not changed to "Obfuscated message" after the message is expired.

### Causes

The code was checking local member message.isObfuscated for updating accessibilityIdentifier, but the message object is set to nil after the message is expired and therefore it is not update.

### Solutions

Add an isObfuscated to Configuration struct and update accessibilityIdentifier depends on isObfuscated's value.


typfel:
Where is the message set to nil after it's obfuscated? I'm wondering because I remember that we did the reverse of this in https://github.com/wireapp/wire-ios/pull/3169/

billypchan:
@typfel The Configuration struct for ConversationTextMessageCell is different from the cases in https://github.com/wireapp/wire-ios/pull/3169. It does not have a member `let message: ZMConversationMessage` so the `isObfuscated` flag is necessary.

PR #3243, opened 2019-03-12
Fix: remove pureLayout from ConfirmAssetViewController

billypchan:
## What's new in this PR?

Remove pureLayout from ConfirmAssetViewController.
Create constraint helpers method for pinning edges or axes to super view.


Wire-iOS Tests/ConfirmAssetViewControllerTests.swift
```diff
@@ -18,7 +18,7 @@
 
 @testable import Wire
 
-class ConfirmAssetViewControllerTests: ZMSnapshotTestCase {
+final class ConfirmAssetViewControllerTests: ZMSnapshotTestCase {
```
alexaubry: Why make it final?

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -71,7 +82,64 @@ extension UIView {
         return constraints
     }
 
-    // MARK: - edge alignment
+    @discardableResult func pinToSuperView(axisAnchor: AxisAnchor,
```
nicorsm: Very little fix: usually `superview` is written in lowercase, so the name could become `pinToSuperview` (see also line 111)

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController+Constraint.swift
```diff
@@ -0,0 +1,136 @@
...
+         titleLabel,
+         bottomPanel,
+         imageToolbarView,
+         imageToolbarSeparatorView,
+         confirmButtonsContainer,
+         acceptImageButton,
+         rejectImageButton,
+         imagePreviewView,
+         playerViewController?.view,
+         imageToolbarViewInsideImage].forEach{$0?.translatesAutoresizingMaskIntoConstraints = false}
```
alexaubry: I'm personally not a big fan of this construct, because of the overhead of allocating an array with all these views. In the end, it's the same number of lines to do it one by one, and doesn't consume more resources than required.

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController+Constraint.swift
```diff
@@ -0,0 +1,136 @@
...
+        
+        let safeTopBarHeight = CGFloat(ConfirmAssetViewController.topBarHeight() + UIScreen.safeArea.top)
+        
+        // Top panel
+        topPanel.fitInSuperview(exclude: [.bottom])
+        topBarHeightConstraint = topPanel.heightAnchor.constraint(equalToConstant: safeTopBarHeight)
+        topBarHeightConstraint.isActive = true
+        
+        titleLabel.fitInSuperview(with: EdgeInsets(top: UIScreen.safeArea.top, leading: .nan, bottom: 0, trailing: .nan),
```
alexaubry: Why use nan here? I think it would be easier to not use `fitInSuperview` if this is required. Activating all constraints at once is also often a non-neglectable performance improvement.

Wire-iOS Tests/ConfirmAssetViewControllerTests.swift
billypchan: There are some discussions about `final` and performance:
https://developer.apple.com/swift/blog/?id=27

Wire-iOS/Sources/UserInterface/Components/Controllers/ConfirmAssetViewController+Constraint.swift
billypchan: I am writing a better helper method with better syntax. This line is rewritten with new method `pinToSuperview`

PR #3250, opened 2019-03-13
App building and release

vytis:
# What's new in this PR?

Building and releasing the app was done by old homegrown scripts that became hard to maintain. This PR tries to recreate the current setup while keeping most of the logic in code instead of Jenkins job configuration.

## How it works

The script uses Fastlane and contain several steps. All steps take the same set of parameters, but not everything is used in all steps.

### Parameters

|Parameter | Description|
|--|--|
|`build_type` | Type of build we are building. At the moment can be `Playground`, `Development`, `Internal`, `AppStore`, `AVS` and `RC` | 
| `build_number`  | This will be the build number of the released binary |
| `last_commit`  | We produce changelog from all commits added since this commit. Used when uploading to Hockey |
| `for_simulator`  | `true` if we want to build for simulator. Defaults to "false" |
| `configuration`  | What configuration to use when building the app. Can be "Release" or "Debug", defaults to "Release" |

### Steps

All steps are invoked by using fastlane command e.g:
```
fastlane prepare build_type:Development build_number:123
```

|Step  | Description| Parameters |
|--|--|--|
| `prepare` | Calls `setup.sh` script and if needed does other preprocessing needed for current build | `build_type`, `build_number` |
| `build` | Builds app for testing |  None |
| `test` | Runs tests | None |
| `post_test` | Upload code coverage data to CodeCov.io | None |
| `build_for_release` | Builds the app for a certain type of release | `build_type`, `build_number`, `for_simulator`, `configuration` | 
 | `release_app_store` | Uploads the app to AppStore. Only relevant if `build_type:AppStore` |`build_type`, `build_number`|
| `upload_s3` | Upload all relevant files generated in previous steps to S3 for automation | `build_type`, `build_number` |
| `upload_hockey` | Upload release build to Hockey | `build_type`, `build_number`, `last_commit` |


fastlane/Fastfile
```diff
@@ -44,30 +46,46 @@ platform :ios do
...
-            buildlog_path: "appstore",
-            output_directory: "appstore",
-            include_bitcode: false,
-            include_symbols: true,
-            xcargs: "BUILD_NUMBER=#{ENV["BUILD_NUMBER"]}"
-        )
     end
 
     desc "Upload to AppStore"
     lane :release_app_store do
```
typfel: Wouldn't `upload_app_store` be a more clear name, similar to `upload_s3`?

PR #3257, opened 2019-03-14
Fix: remove Purelayout form SketchColorPickerController

billypchan:
## What's new in this PR?

Remove Purelayout form SketchColorPickerController.


Wire-iOS/Sources/UserInterface/Sketchpad/SketchColorCollectionViewCell.swift
```diff
@@ -0,0 +1,81 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+// 
+import UIKit
+
+@objcMembers
+final class SketchColorCollectionViewCell: UICollectionViewCell {
+    var sketchColor: UIColor? {
+        didSet {
+            if sketchColor == oldValue {
```
typfel: A `guard` statement would be suitable for these cases.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -210,6 +217,8 @@ extension UIView {
     }
 
     func setDimensions(size: CGSize) {
+        translatesAutoresizingMaskIntoConstraints = false
```
typfel: It's a but unexpected to have this kind side effect when calling this method, I would remove it.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -145,11 +145,18 @@ extension UIView {
                                            with insets: EdgeInsets = .zero,
                                            exclude excludedAnchors: [Anchor] = [],
                                            activate: Bool = true) -> [Anchor: NSLayoutConstraint] {
+
         guard let superview = superview else {
             fatal("Not in view hierarchy: self.superview = nil")
         }
 
-        return pin(to: superview, safely: safely, with: insets, exclude: excludedAnchors, activate: activate)
+        translatesAutoresizingMaskIntoConstraints = false
```
typfel: Also here, let's not add side effects to the constraints helpers.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
billypchan: I think it makes sense to set `translatesAutoresizingMaskIntoConstraints` to false when creating constraints for width/height. PureLayout's UIView.initForAutoLayout method also set `translatesAutoresizingMaskIntoConstraints` to false.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -144,13 +144,14 @@ extension UIView {
     @discardableResult func fitInSuperview(safely: Bool = false,
                                            with insets: EdgeInsets = .zero,
                                            exclude excludedAnchors: [Anchor] = [],
+                                           translatesAutoresizingMaskIntoConstraints: Bool = false,
```
typfel: This will still be a side effect. Let's set `translatesAutoresizingMaskIntoConstraints` outside this method.

billypchan: I will undo these.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
typfel: But it also unexpected since we now write constraints without any helper library and Apple's methods does not implicitly modify `translatesAutoresizingMaskIntoConstraints`.

I'm arguing for keeping our helper methods very simply and not build our own constraints library.

PR #3258, opened 2019-03-14
Add test assertions for color contrast

alexaubry:
### Issues

We have added [Accessibility guidelines](
https://github.com/wireapp/wire-ios/wiki/Accessibility-Guidelines) for the project. In order to help ensuring we comply with section G (color contrast), we want to have test assertions to use to check the contrast of text and buttons during snapshot testing.

### Solutions

We add new test assertions: 

- `WRValidateContrast` (base assertion) to verify two colors against each other.
- `WRValidateLabelContrast` to verify the text color of a label against a given background. We get the font size and boldness automatically.
- `WRValidateTextButtonContrast` to verify the text color of a label against a given background. We get the font size and boldness automatically.
- `WRValidateIconButtonContrast` to verify the icon with the background color of the button. Both colors are read automatically from the view.

When building new interfaces, we should try to add tests using these. 

These rely on the algorithm defined by the [W3C](https://www.w3.org/TR/WCAG20-TECHS/G17.html) to verify that we comply to at least the AA standard. 

### Notes

We added a demo usage of these in `LandingViewControllerTests`. It raised two issues: one with the `loginHintsLabel`, and one with the create account button background color, whose green doesn't have enough contrast with the white icon.


vytis:
Great start! However it seems a bit tedious to properly check this - I assume we need to manually write out what to check. Maybe we could somehow automate it by iterating over accessibility elements? Not sure if it's easy to figure out what is the background for each thing though.

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -106,7 +106,7 @@ class LandingViewController: AuthenticationStepViewController {
         let label = UILabel()
         label.text = "landing.login.hints".localized
         label.font = LandingViewController.regularFont
-        label.textColor = UIColor.Team.subtitleColor
+        label.textColor = UIColor.Team.textColor
```
billypchan: Would it be better to define another color scheme set for higher contrast? I would suggest creating a new color scheme other than .dark or .light to make the whole app with higher contrast and it would not affect our current design.

Also when the user enabled Accessibility - Increase Contrast, we can check `UIAccessibilityDarkerSystemColorsEnabled` flag to change the color defination. https://developer.apple.com/documentation/uikit/1615087-uiaccessibilitydarkersystemcolor?language=occ

alexaubry:
@vytis This is a good point. I agree that manually writing these tests could be too much effort, and that we might end up missing some issues. At least for labels and buttons, it should be easy to figure out the foreground color. We could traverse the superview tree backwards to see what the background color is, and blending alpha colors if needed.

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
alexaubry: This is a good point, and I think we should discuss this with design. But however, we need to have a correct contrast by default, even when the darker color settings is not enabled, so little changes like this one might be required.

billypchan:
close for now til further discussion with the UX team

PR #3261, opened 2019-03-14
Fix: purelayout removal - MediaBar

billypchan:
## What's new in this PR?

Remove Purelayout from MediaBar.

## TODO
- [x] rewrite MediaBar in Swift


Wire-iOS/Sources/UserInterface/Overlay/MediaBar/MediaBar.swift
```diff
@@ -0,0 +1,119 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+@objcMembers
```
alexaubry: `@objcMembers` might be a bit of overkill here, as, we only need to expose the initializer, `titleLabel`, `playPauseButton` and `closeButton`.

According to https://docs.swift.org/swift-book/ReferenceManual/Attributes.html:

> Applying the objc attribute when it isn’t needed can increase your binary size and adversely affect performance.

Wire-iOS/Sources/UserInterface/Overlay/MediaBar/MediaBar.swift
```diff
@@ -0,0 +1,119 @@
...
+    private(set) var titleLabel: UILabel!
+    private(set) var playPauseButton: IconButton!
+    private(set) var closeButton: IconButton!
+
+    private var bottomSeparatorLine: UIView!
+    private let contentView = UIView()
+    private var initialConstraintsCreated = false
+
+
+    init() {
```
alexaubry: It would be more in line with other subviews to override `init(frame:)`. Then we get the default initializer for free.

Wire-iOS/Sources/UserInterface/Overlay/MediaBar/MediaBar.swift
billypchan: I compared the objc interface generated with @objcMembers/@objc has no difference in this case. The members marked private(set) are exposed and private member are not exposed.

alexaubry: Can you check that `canRespondToSelector` also fails for the private methods?

PR #3267, opened 2019-03-15
Fix: purelayout removal - ParticipantDeviceCell

billypchan:
## What's new in this PR?

Remove PureLayout from ParticipantDeviceCell

## discussion
After ParticipantDeviceCell is translated to Swift, injection of `MockUserClient` as an `UserClient` for the method `func configure(for client:)` is no longer allowed. I changed the argument type form `UserClient` to `UserClientType` but there is a missing method `attributedRemoteIdentifier` which is defined in `UserClient`'s extension. `attributedRemoteIdentifier` depends on UI project's localized string and it should not move to Data Model?


Wire-iOS/Sources/UserInterface/UserProfile/Views/ParticipantDeviceCell.swift
```diff
@@ -0,0 +1,104 @@
...
+        constraints.append(trustLevelImageView.pin(to: nameLabel, axisAnchor: .centerY, activate: false))
+
+        constraints.append(nameLabel.pinToSuperview(anchor: .top, inset: 16, activate: false))
+        constraints.append(nameLabel.leadingAnchor.constraint(equalTo: trustLevelImageView.trailingAnchor, constant: 16))
+
+        constraints.append(identifierLabel.pin(to: nameLabel, anchor: .leading, activate: false))
+        constraints.append(identifierLabel.topAnchor.constraint(greaterThanOrEqualTo: nameLabel.bottomAnchor))
+        constraints.append(identifierLabel.pinToSuperview(anchor: .bottom, inset: 16, activate: false))
+
+        NSLayoutConstraint.activate(constraints)
```
vytis: In other places we avoid having temporary array by creating it here. If some of the helper methods return more than one constraint you can use + to concatenate them:

```
NSLayoutConstraint.activate( 
   trustLevelImageView.setDimensions(length: 16, activate: false) + [
   trustLevelImageView.pinToSuperview(anchor: .leading, inset: 24, activate: false),
   <...>
   identifierLabel.pinToSuperview(anchor: .bottom, inset: 16, activate: false)
])

billypchan: The reason I create a typed array is trying to improve compilation time. I will compare both approaches to see is it still necessary for Swift 4.2.

PR #3270, opened 2019-03-18
Fix: PureLayout removal - LaunchImageViewController

billypchan:
## What's new in this PR?

Remove PureLayout from LaunchImageViewController and create snapshot tests.


Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// Convinience method for showing the @c activityIndicator and @c loadingScreenLabel and start the spinning animation
+    func showLoadingScreen() {
+        shouldShowLoadingScreenOnViewDidLoad = true
+        loadingScreenLabel.isHidden = false
+        activityIndicator.startAnimation(nil)
+    }
+
+    /// Convinience method for hiding all the animation related functionality
+    func hideLoadingScreen() {
+        activityIndicator.stopAnimation(nil != nil)
```
vytis: What is the reason behind this check? Can't we just use `false`?

Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+    private var activityIndicator: ProgressSpinner!
+
+    /// Convinience method for showing the @c activityIndicator and @c loadingScreenLabel and start the spinning animation
+    func showLoadingScreen() {
+        shouldShowLoadingScreenOnViewDidLoad = true
+        loadingScreenLabel.isHidden = false
+        activityIndicator.startAnimation(nil)
+    }
+
+    /// Convinience method for hiding all the animation related functionality
```
vytis: Typo - should be "Convenience". Another occurrence is in a method above.

Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
billypchan: See translation leave over. I will fix it.

Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
billypchan: Also some copy & paste error from @objc version :)

PR #3275, opened 2019-03-19
Feature: Go to a specific user profile when clicking on a link outside/inside the app

billypchan:
## What's new in this PR?

Support the new URL scheme to open a profile screen of a user.
The `ProfileViewController` is presented from ZClientViewController in this case. A new context `.ProfileViewer` for `ProfileViewController` is created for the case no conversation is linked to `ProfileViewController`.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/984

### Todo:
- [x] self-user handling
- [x] snapshot test for the new context
- [x] alert dialog for invalid link
- [x] handle link after refresh start
- [x] remove device tab for context `.ProfileViewer`
- [ ] show alert when there is no activated session (in launch screen)
- [ ] handle the case: self profile screen is shown, and the user open a link of other user
- [ ] handle the case: other user(A) profile screen is shown, and the user open a link of other user (B)


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -147,6 +147,18 @@ extension ZClientViewController {
...
+    // MARK: - present a ProfileViewController
+
+    func openProfileScreen(for user: ZMUser) {
+        let profileViewController = ProfileViewController(user: user, viewer: ZMUser.selfUser(), context: .profileViewer)
+        profileViewController.delegate = self
+
+        let navWrapperController: UINavigationController = profileViewController.wrapInNavigationController()
+        navWrapperController.modalPresentationStyle = .formSheet
+
+        present(navWrapperController, animated: true)
```
vytis: So it will present the profile screen on top of anything that is currently displayed? What about if we have an alert view or call screen?

billypchan: After the team discussion, this screen should have similar behaviour like the device list screen. I will check if it is necessary to dismiss the alerts/minimize the call screen when ZClientViewController wants to present a screen.

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -563,12 +565,45 @@ public extension SessionManager {
...
+                                      message: message,
+                                      cancelButtonTitle: "general.ok".localized)
+
+        present(alert, animated: true, completion: nil)
+    }
+
+    func sessionManagerShouldExecuteURLAction(_ action: URLAction, callback: @escaping (Bool) -> Void) -> Bool {
         switch action {
+        case .openConversation(_, _):
+        ///TODO: open a conversation if id is valid
```
alexaubry: Will it be done in another PR?

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -563,12 +565,45 @@ public extension SessionManager {
...
+        case .warnInvalidDeepLink(let error):
+            switch error {
+            case .invalidUserLink:
+                presentAlert(title: "url_action.invalid_user.title".localized,
+                             message: "url_action.invalid_user.message".localized)
+            case .invalidConversationLink:
+                presentAlert(title: "url_action.invalid_conversation.title".localized,
+                             message: "url_action.invalid_conversation.message".localized)
+            case .notLoggedIn:
+                ///TODO: alert for this case
```
alexaubry: Will it be done in another PR?

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
alexaubry: Were there discussions about reloading the UI from scratch? I think I remember it being on the table at some point.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController+Extension.swift
```diff
@@ -40,12 +40,37 @@ extension ProfileViewController {
...
+    func setupTabsController() {
+        var viewControllers = [UIViewController]()
+
+        if context != .deviceList {
+            let profileDetailsViewController = setupProfileDetailsViewController()
+            viewControllers.append(profileDetailsViewController)
+        }
+
+        if context != .profileViewer,
+            (fullUser().isConnected ||
```
alexaubry: We should probably have a `canSeeDevices` shortcut here. Otherwise it probably won't work with mock users.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController+Extension.swift
```diff
@@ -125,13 +151,30 @@ extension ProfileViewController: ProfileFooterViewDelegate, IncomingRequestFoote
...
+        case .openSelfProfile:
+            openSelfProfile()
+        }
+    }
+
+    private func openSelfProfile() {
+        ///do not reveal list view for iPad regular mode
+        let leftViewControllerRevealed: Bool
+        if let presentingViewController = presentingViewController {
+            leftViewControllerRevealed = !presentingViewController.isIPadRegular(device: UIDevice.current)
```
alexaubry: Is this a design requirement? Having cases for iPad only seems a bit unscalable.

Wire-iOS/Sources/AppRootViewController.swift
billypchan: Yes, conversation link PR is here: https://github.com/wireapp/wire-ios/pull/3276

Wire-iOS/Sources/AppRootViewController.swift
billypchan: Yes

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController+Extension.swift
billypchan: Yes, the requirement is simply showing the self profile screen without any other side effect(reveal conversation list)

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController+Extension.swift
billypchan: This method is translated from objc original code. We can change it after this PR.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
billypchan: For the case "Not logged in and show an info dialog" I am still waiting for the reply from design team. It will be included in next PR.

PR #3276, opened 2019-03-20
ZIOS-11279: Go to a specific conversation when clicking on a link outside/inside the app

nicorsm:
## What's new in this PR?

Added `.openConversation` action to load a conversation when opening a `wire://conversation/` link. It uses the same behavior of opening the conversation from a push notification. Based on https://github.com/wireapp/wire-ios/pull/3275.


billypchan:
@nicorsm I tested the case "launch with the deep link" and it passed.

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -576,9 +576,14 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
...
-        case .openConversation(_, _):
-        ///TODO: open a conversation if id is valid
-            break
+        case .openConversation(_, let conversation):
+            if let conversation = conversation,
+               let userSession = ZMUserSession.shared() {
+                sessionManager?.showConversation(conversation, at: nil, in: userSession)
+                return true
+            } else {
+                return false
```
typfel: @billypchan  we return `false` which if I remember correctly won't remove the pending url action. Won't we then potentially have a loop were we continuously try perform this action?

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -576,9 +576,14 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
 
     func sessionManagerShouldExecuteURLAction(_ action: URLAction, callback: @escaping (Bool) -> Void) -> Bool {
         switch action {
-        case .openConversation(_, _):
-        ///TODO: open a conversation if id is valid
-            break
+        case .openConversation(_, let conversation):
+            if let conversation = conversation,
+               let userSession = ZMUserSession.shared() {
+                sessionManager?.showConversation(conversation, at: nil, in: userSession)
```
typfel: Why is it necessary to go through the session manager here and not call `zClientViewController.showConversation()` like we do for user profiles?

billypchan: I think the reason is the `SessionManager` does not provide an interface without `userSession` to open a conversation. I will update the interface.

billypchan: @typfel After refactoring in https://github.com/wireapp/wire-ios-sync-engine/pull/988, both open user profile/conservation action will be handled by SessionManager's methods.

PR #3277, opened 2019-03-21
Fix: app freezes when jumping an old message from search

vytis:
### Issues

Jumping to an old message from search would freeze the app.

### Causes

When jumping to a message we would try to load all messages between the newest to the one we searched for. This would take a long time or would never finish because the app would be killed for consuming too much memory.

### Solutions

We use `NSFetchedResultsController` to fetch messages for the conversation. To jump to certain past message we would:
* Count the number of messages between the newest one and the one we are interested in. 
* Skip loading most of the messages by setting `fetchOffset` and load only a few above and below to give context.
* Update `fetchOffset` and `fetchLimit` when scrolling the list to load more messages.

When scrolling up from the position we would increase the `fetchLimit` so that more messages are put in the table view. This is exactly the same when we are not in a search mode.

To scroll down we need to change `fetchOffset` so that more messages are loaded from below. The tricky part is to keep the scroll position, because if we do nothing then it jumps to the bottom of the page and we completely loose context.


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
```diff
@@ -255,40 +303,41 @@ final class ConversationTableViewDataSource: NSObject {
...
+            // 1. Remember the newest message now
+            let newestMessageBeforeReload = messages.first!
+            // 2. Load more messages
+            loadNewerMessages()
+            // 3. Get the index path of the message that should stay displayed
+            let indexPath = self.indexPath(for: newestMessageBeforeReload)!
+            // 4. Get the frame of that message
+            let indexPathRect = tableView.rectForRow(at: indexPath)
+            // 5. Update content offset so it stays visible. To reduce flickering compensate for empty space below the message
+            scrollView.contentOffset = CGPoint(x: 0, y: indexPathRect.minY - 16)
```
typfel: Where does the number `16` come from?

vytis: I wrote in the comment a bit, but basically from the fact that each cell has 16px padding before the content starts. Without this the cell jumps a bit when scrolling. I could not really find a constant for this anywhere, but maybe adding a named variable would help

typfel: Right it comes from `ConversationMessageContext.spacing` I think.

PR #3279, opened 2019-03-21
Revise onboarding copy to clarify audiences

infotexture:
## What's new in this PR?

Revised copy pr 2019-03-21 discussion w/ JG, SY & RH.


alexaubry:
Hi Roger, I'll take over this PR to fix the layout issues caused by the new copy. We created a ticket in Jira to track in internally (ZIOS-11305).

alexaubry:
@infotexture fixed the issues. You can see how the longer copy looks like with this snapshot: https://github.com/wireapp/wire-ios/blob/d9ad996e790f9773e7e2e9667f7f55dc417265b3/Wire-iOS%20Tests/ReferenceImages_64/Wire_iOS_Tests.AuthenticationInterfaceBuilderTests/testLandingScreen%402x.png

typfel:
We are also adding a logout button to some screens?

alexaubry:
@typfel No, that's because I'm snapshotting the whole navigation controller instead of just the view controller. The logout button is in the nav bar. The reason I do the navigation controller now is because I want to verify that we ignore the navigation bar to position the Wire logo on the first part.

PR #3282, opened 2019-03-22
ZIOS-11296: Improve performance of YouTube and SoundCloud links

alexaubry:
### Issues

Due to compatibility issues between clients, we need to treat YouTube and SoundCloud links as a separate type of links, since they cannot be sent as link previews.

Their parsing was previously done in the UI and not persisted; which means we needed to recompute them for every rendering of the text message cells.

In addition to this performance bottleneck, we used the YouTube and SoundCloud API to manually fetch the details of the media, which this could have been achieved using the OpenGraph data provided by these websites at faster speeds (both for displaying the preview and general scrolling) and lower costs.

### Solutions

In the frameworks:

- Instead of using the YouTube and SoundCloud API, we parse the data from OpenGraph.
- Instead of checking every link when creating a text message cell, we persist them in the database as a transformable property

In UI:

- Remove the code related to the YouTube API and remove old view controllers used for YouTube previews
- Update YouTube preview cell to accept a generic link attachment
- Add necessary `ImageResource` adaptor to download and cache the images from the new `LinkAttachment` class
- Force refetching the attachments for legacy messages when scrolling
- Improve accessibility of previews

For now, only YouTube uses the new infrastructure by default. We introduce an internal feature flag to test the new infrastructure for SoundCloud.

### Dependencies

Needs release with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/991

### To-Do

- [x] More snapshot tests


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+MessageFormatting.swift
```diff
@@ -217,19 +203,18 @@ extension NSMutableAttributedString {
...
     }
     
-    func removeTrailingLink(for linkPreview: LinkMetadata, linkAttachments: [LinkAttachment]) -> [LinkAttachment] {
-    
-        // Don't remove trailing link if we embed content
-        guard linkAttachments.first?.type == LinkAttachmentType.none,
-              linkPreview.originalURLString.lowercased() != "giphy.com", // Don't remove giphy links
-              let linkPreviewAttachment = linkAttachments.reversed().first(where: { $0.string == linkPreview.originalURLString }),
-              linkPreviewAttachment.range.upperBound == self.length
+    func removeTrailingLink(for linkPreview: LinkMetadata) {
```
typfel: Don't see any calls to this method

alexaubry: That's right, I just noticed the test failures about this too, will fix.

alexaubry: Fixed.

PR #3284, opened 2019-03-22
Fix: Open deep link when profile screen is presented

billypchan:
## What's new in this PR?

Dismiss other presented view before presenting the user profile screen from deep link.


Wire-iOS/Sources/AppStateController.swift
```diff
@@ -28,13 +28,13 @@ protocol AppStateControllerDelegate : class {
...
 
-class AppStateController : NSObject {
+final class AppStateController : NSObject {
 
     /**
      * The possible states of authentication.
      */
 
-    enum AuthenticationState {
+    enum AuthenticationState: Equatable {
```
typfel: This seems unrelated to the fix.

PR #3287, opened 2019-03-25
Fix: Showing a dialog when opens a deep link with no logged in account.

billypchan:
### Issues

If a user opens a deep link with no logged in account, no info dialog is shown.

### Causes

The pending `URLAction` would not be executed in this case since there is no active user session.

### Solutions

When the app becomes active, AppRootViewController checks if the `authenticationState` is not `loggedIn` and having unhandled deep link stored, show a warning dialog.

### Dependencies
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/992
- [x] Wait for the text copy of the dialog.


Cartfile.resolved
```diff
@@ -23,7 +23,7 @@ github "wireapp/wire-ios-link-preview" "17.0.0"
 github "wireapp/wire-ios-protos" "19.0.0"
 github "wireapp/wire-ios-request-strategy" "140.0.1"
 github "wireapp/wire-ios-share-engine" "129.0.0"
-github "wireapp/wire-ios-sync-engine" "241.0.1"
+github "wireapp/wire-ios-sync-engine" "243.0.1"
```
typfel: This doesn't match the `Cartfile`, when bumping a major version you need to update major version in the `Cartfile` and run `carthage update`

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -708,3 +700,14 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
 extension Notification.Name {
     static let companyLoginDidFinish = Notification.Name("Wire.CompanyLoginDidFinish")
 }
+
+extension UIViewController {
```
typfel: This change is not necessary, let's create a generic `presentAlert()` method when we have need for it.

billypchan: @typfel There are several places in the project creating a simple alert with an OK button. I will clean them up.

typfel: Very well but do in a separate PR

Cartfile
```diff
@@ -1,6 +1,6 @@
 github "wireapp/wire-ios-ziphy" ~> 12.0
 github "wireapp/Cartography" "3.0.3"
-github "wireapp/wire-ios-sync-engine" ~> 240.0
+github "wireapp/wire-ios-sync-engine" ~> 243.0.1
```
billypchan: @typfel Cartfile is updated

typfel: 👍 No need specify minor & patch version in the `Cartfile` but it also doesn't hurt so it's fine to keep it.

Wire-iOS/Sources/AppRootViewController.swift
typfel: But let's revert this change for now.

PR #3289, opened 2019-03-25
Fix: User profile deep link does not show rich profile

billypchan:
### Issues

User profile deep link does not show rich profile.

### Causes

The rich profile section's visibility is despended on the `conversation` property, but it is nil in this case.

### Solutions

Add the `ProfileViewControllerContext` argument when creating `ProfileDetailsContentController`. Show the rich profile section when the `context` is `.ProfileViewer`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
@@ -115,7 +115,7 @@ final class ProfileDetailsContentController: NSObject, UITableViewDataSource, UI
     
     /// Updates the content for the current configuration.
     private func updateContent() {
-        switch (conversation?.conversationType, context) {
+        switch conversation?.conversationType ?? .oneOnOne {
         case (.group?, _),
              (_, .profileViewer):
             let richProfile = user.richProfile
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -107,8 +115,9 @@ class ProfileDetailsContentController: NSObject, UITableViewDataSource, UITableV
     
     /// Updates the content for the current configuration.
     private func updateContent() {
-        switch conversation?.conversationType {
-        case .group?:
+        switch (conversation?.conversationType, context) {
```
typfel:
```suggestion
        switch conversation?.conversationType ?? .oneOnOne {
```

We can achieve the same thing without the context if we default to `.oneOnOne` if the conversation is missing.

alexaubry: I agree this would be a better solution. We could update the existing pattern matching to `case .oneOnOne?, .none`.

Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -107,8 +115,8 @@ class ProfileDetailsContentController: NSObject, UITableViewDataSource, UITableV
     
     /// Updates the content for the current configuration.
     private func updateContent() {
-        switch conversation?.conversationType {
-        case .group?:
+        switch conversation?.conversationType ?? .group {
```
billypchan: @typfel @alexaubry It is refactored as your suggestion. (The default value should be `.group`, when `conversation` is nil, the UI should have no read receipt components.)

PR #3294, opened 2019-03-26
Fix: No Guest label in user profile deep link popup

billypchan:
### Issues

No Guest label in user profile deep link popup

### Causes

The guest label was depended on conversation, but it is nil when opening the profile screen with deep link.

### Solutions

When there is no conversation, a guest is defined by a user that do not belongs to the same team, or not a member of the team.


Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
```diff
@@ -203,9 +203,15 @@ class ProfileView: UIView, Themeable {
...
     // MARK: - Content and Options
     
-    func prepareForDisplay(in conversation: ZMConversation?) {
-        guestIndicatorStack.isHidden = conversation.map(user.isGuest) != true
-        
+    func prepareForDisplay(in conversation: ZMConversation?, context: ProfileViewControllerContext) {
+
+        switch context {
+            case .profileViewer:
+                guestIndicatorStack.isHidden = !(ZMUser.selfUser()?.isTeamMember ?? false && !user.isTeamMember)
```
alexaubry: This means we wouldn't show the guest label for a guest from another team. To fix this we could:

- add a viewer property (as `UserType`)
- check if viewer is a team member, and if yes, check if `viewer.canAccessCompanyInformation(of: user)`

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
```diff
@@ -203,9 +203,22 @@ class ProfileView: UIView, Themeable {
...
     
-    func prepareForDisplay(in conversation: ZMConversation?) {
-        guestIndicatorStack.isHidden = conversation.map(user.isGuest) != true
-        
+    func prepareForDisplay(in conversation: ZMConversation?, context: ProfileViewControllerContext) {
+
+        let guestIndicatorHidden: Bool
+        switch context {
+            case .profileViewer:
+                if let selfUser = ZMUser.selfUser() {
```
alexaubry: It would be better to not use the self user here. What if we add a viewer property like in the rest of the Profile section of the codebase? It think it would make it better for testing once we get the MockUser stack all figured out.

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
```diff
@@ -203,9 +203,22 @@ class ProfileView: UIView, Themeable {
...
+                    guestIndicatorHidden = !(selfUser.isAHost(of: user))
```
alexaubry: We need to check if selfUser is a team member before checking if two users are in the same team, because this will return false, even if the two users are both non-team users.

Wire-iOS/Sources/Helpers/syncengine/ZMUser+Permissions.swift
```diff
@@ -58,6 +58,11 @@ extension UserType {
         return otherUser.isConnected || self.canAccessCompanyInformation(of: otherUser) || otherUser.isWirelessUser
     }
 
+    /// Returns whether the current user (viewer) can see guest label in profile screen of another user.
+    func isAHost(of otherUser: UserType) -> Bool {
```
alexaubry: Why do we need a wrapper for this here? Can we just call the function it wraps?

billypchan:
@alexaubry This PR will be updated to match the design spec. I will create tests for all possible 20 combinations of self roles and other user roles, which 8 of them need to show the guest label.

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
```diff
@@ -203,9 +209,23 @@ class ProfileView: UIView, Themeable {
...
     
-    func prepareForDisplay(in conversation: ZMConversation?) {
-        guestIndicatorStack.isHidden = conversation.map(user.isGuest) != true
-        
+    func prepareForDisplay(in conversation: ZMConversation?, context: ProfileViewControllerContext) {
+
+        let guestIndicatorHidden: Bool
+        switch context {
+            case .profileViewer:
+                if let zmUser = user.zmUser,
```
alexaubry: We should not need to make this an extension on ZMUser. `UserType.canAccessCompanyInformation` already checks if the team identifier of the receiver and the viewer are the same. All we need to do is:

```swift
guestIndicatorHidden = !viewer.isTeamMember || viewer.canAccessCompanyInformation(of: user)
```

We hide the label:
- if self user is not a team member (non-team self users don't have the concept of guests in their team)
- or if self user can access the company info of the other user (not a guest in the team, because they are in the same team)

PR #3297, opened 2019-03-27
ZIOS-11310: Added support for .malformedLink error case

nicorsm:
## What's new in this PR?

Added UI support to the `.malformedLink` error case for deep links.

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/995


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1544,6 +1544,9 @@
 "url_action.authorization_required.title" = "Authorization required.";
 "url_action.authorization_required.message" = "You need to log in to view this content.";
 
+"url_action.invalid_link.title" = "Invalid link.";
```
alexaubry: We will need a strings update after this PR is merged (and coordinate for German translation)

PR #3299, opened 2019-03-27
Fix grouped call system messages not showing all timestamps

typfel:
### Issues

When multiple calls are made we group them together into one system message. When tapping on such a system message we should expand and show a timestamp for when each call was made but we only showed the first timestamp.

### Causes

Label was configured with `.numberOfLines = 1`

### Solutions

Configure label with `.numberOfLines = 0` when displaying the call list.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
@@ -109,17 +109,17 @@ class MessageToolboxDataSource {
            message.systemMessageData?.systemMessageType == .missedCall {
             content = .callList(makeCallList())
         }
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/MessageToolboxDataSource.swift
```diff
@@ -101,8 +104,13 @@ class MessageToolboxDataSource {
 
         // Determine the content by priority
 
+        // 1) Call list
+        if message.systemMessageData?.systemMessageType == .performedCall ||
+           message.systemMessageData?.systemMessageType == .missedCall {
+            content = .callList(makeCallList())
+        }
         // 1) Failed to send
```
alexaubry: Very nitpicky, but we could update the numbers in the list.

2) Failed to send
3) Likers
4) Timestamp

PR #3300, opened 2019-03-27
ZIOS-10257: Add a border to the timed message indicator

alexaubry:
### Issues

We want to add a 10% border around the ephemeral message indicator to improve legibility.

### Solutions

We offset the inner circle of the pie by 10% of the view's width. We also fix a bug where the indicator would stay visible after the message expires.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/DestructionCountdownView.swift
```diff
@@ -59,12 +59,16 @@ import Cartography
 
     public override func layoutSublayers(of layer: CALayer) {
         super.layoutSublayers(of: layer)
-        elapsedTimeLayer.frame = bounds
-        elapsedTimeLayer.path = makePath(for: bounds)
+        let backgroundFrame = bounds
+        let borderWidth = 0.10 * backgroundFrame.width
+        let elapsedFrame = CGRect(x: borderWidth, y: borderWidth, width: backgroundFrame.width - borderWidth * 2, height: backgroundFrame.height - borderWidth * 2)
```
typfel: Could we achieve the same thing with `insetBy`?

alexaubry: It works, good idea, thanks.

PR #3301, opened 2019-03-27
Show unread count for new conversations

typfel:
### Issues

When team 1:1 conversations are created they generate an unread dot. This unread state was however only indicated by the subtitle "x started a conversation with you", which wasn't very obvious and many people missed it. This resulted in users complaining about unread badges which won't go away after reading all messages.

### Solutions

Show an unread count for this scenario so that it's obvious that this conversation is unread.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
@@ -634,6 +634,10 @@ final internal class StartConversationMatcher: TypedConversationStatusMatcher {
         return (resultString && type(of: self).regularStyle).addAttributes(type(of: self).emphasisStyle, toSubstring: senderString)
     }
     
+    func icon(with status: ConversationStatus, conversation: ZMConversation) -> ConversationStatusIcon? {
+        return ConversationStatusIcon.unreadMessages(count: 1)
+    }
+    
     var combinesWith: [ConversationStatusMatcher] = []
 }
 


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -634,6 +634,10 @@ final internal class StartConversationMatcher: TypedConversationStatusMatcher {
         return (resultString && type(of: self).regularStyle).addAttributes(type(of: self).emphasisStyle, toSubstring: senderString)
     }
     
+    func icon(with status: ConversationStatus, conversation: ZMConversation) -> ConversationStatusIcon? {
+        return ConversationStatusIcon.unreadMessages(count: 1)
```
alexaubry: Is this called from some place?

typfel: Yes, it's starts from
https://github.com/wireapp/wire-ios/blob/a3db88762409ec842d3886dfed79a27b280f5156/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListItemView%2BUpdate.swift#L68

which in turn calls the above mentioned method on the top matcher

https://github.com/wireapp/wire-ios/blob/a3db88762409ec842d3886dfed79a27b280f5156/Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation%2BStatus.swift#L720

PR #3304, opened 2019-03-28
Fix: pureLayout removal - ConversationInputBarViewController

billypchan:
## What's new in this PR?

Remove pureLayout from `ConversationInputBarViewController`


alexaubry:
@billypchan While the two methods have the same compilation time, there can be performance impacts at runtime with appending to an array vs. using a constant in the `NSLayoutConstraint.activate`, because using the mutable version needs to reallocate memory for every item we append. It would rather keep using the `NSLayoutConstraint.activate` method going forward, but as we mentioned before, we should plan a team discussion about this.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Private.h
```diff
@@ -27,20 +25,28 @@ NS_ASSUME_NONNULL_BEGIN
...
-@property (nonatomic, nullable) IconButton *uploadFileButton;
-@property (nonatomic, nullable) IconButton *sketchButton;
-@property (nonatomic, nullable) IconButton *pingButton;
-@property (nonatomic, nullable) IconButton *locationButton;
-@property (nonatomic, nullable) IconButton *ephemeralIndicatorButton;
-@property (nonatomic, nullable) IconButton *emojiButton;
-@property (nonatomic, nullable) IconButton *markdownButton;
-@property (nonatomic, nullable) IconButton *gifButton;
-@property (nonatomic, nullable) IconButton *mentionButton;
+@property (nonatomic, null_unspecified) IconButton *audioButton;
```
alexaubry: What does `null_unspecified` do?

billypchan: These properties was equivalent to something like`var audioButton: IconButton?`. I change them to `var audioButton: IconButton!` by marking them `null_unspecified`. Since they are created when the view is loaded and no need to unwrap when accessing them.

typfel: But we can access these properties before the view is loaded so they are not guaranteed to be there. I think it would safer to specify them as nullable or does that cause a lot of issues?

billypchan: @typfel we have some discussion about objc header nullability here: https://github.com/wireapp/wire-ios/pull/3339#discussion_r273461685 . If we rewrite this class to Swift I would initialize it with a closure to prevent these optional members.

typfel: Right, but the public header is wrong since they can be null.

billypchan: @typfel I refactored the subviews creation(create them when init instead of viewDidLoad) and made them nonnull.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.m
```diff
@@ -176,15 +171,9 @@ - (void)viewDidLoad
...
-    
-    [self createInputBar]; // Creates all input bar buttons
-    [self createSendButton];
-    [self createEphemeralIndicatorButton];
-    [self createMarkdownButton];
 
-    [self createHourglassButton];
-    [self createTypingIndicatorView];
-    
+    [self setupViews];
```
typfel: The views are still `nullable` since they are created in `viewDidLoad`. But can't we just let them be nullable?

billypchan: @typfel I refactored the subview creation. I will convert the whole class to Swift soon to make it clearer.

PR #3308, opened 2019-03-29
Improve build time - AudioRecordViewController

billypchan:
## What's new in this PR?

Improve build time by removing Cartography from AudioRecordViewController.
Updated snapshot tests of AudioRecordViewController with time label filled with dummy time.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -39,8 +38,8 @@ private let zmLog = ZMSLog(tag: "UI")
     case recording, finishedRecording
 }
 
-@objcMembers public final class AudioRecordViewController: UIViewController, AudioRecordBaseViewController {
-    
+public final class AudioRecordViewController: UIViewController, AudioRecordBaseViewController {
```
alexaubry: Do we need to keep this public?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -52,9 +51,8 @@ private let zmLog = ZMSLog(tag: "UI")
     let topContainerView = UIView()
     let cancelButton = IconButton()
     let recordingDotView = RecordingDotView()
-    var recordingDotViewVisible: ConstraintGroup?
-    var recordingDotViewHidden: ConstraintGroup?
-    
+    var recordingDotViewVisible = [NSLayoutConstraint]()
```
alexaubry: Using the array literal is usually more idiomatic in Swift.

```suggestion
    var recordingDotViewVisible: [NSLayoutConstraint] = []
```

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -52,9 +51,8 @@ private let zmLog = ZMSLog(tag: "UI")
...
+    var recordingDotViewHidden = [NSLayoutConstraint]()
```
alexaubry: Using the array literal is usually more idiomatic in Swift.

```suggestion
    var recordingDotViewHidden: [NSLayoutConstraint] = []
```

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -179,80 +178,79 @@ private let zmLog = ZMSLog(tag: "UI")
...
+         topTooltipLabel,
+         buttonOverlay,
+         topSeparator,
+         timeLabel,
+         recordingDotView,
+         audioPreviewView,
+         cancelButton,
+         rightSeparator].forEach(){ $0.translatesAutoresizingMaskIntoConstraints = false }
+
+        var constraints = [NSLayoutConstraint]()
```
alexaubry: Using the array literal is usually more idiomatic in Swift.

```suggestion
        var constraints: [NSLayoutConstraint] = []
```

alexaubry: Also, why are you not activating all constraints at once with `NSLayoutConstraint.activate`?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -283,17 +281,17 @@ private let zmLog = ZMSLog(tag: "UI")
...
     
-    func setRecordingState(_ state: AudioRecordState, animated: Bool) {
+    private func setRecordingState(_ state: AudioRecordState, animated: Bool) {
         updateRecordingState(state)
         
         if animated {
-            UIView.animate(withDuration: 0.2, animations: {
-                self.view.layoutIfNeeded()
+            UIView.animate(withDuration: 0.2, animations: { [weak self] in
+                self?.view.layoutIfNeeded()
```
alexaubry: I'm not sure if we need to use weak here? Self should not be retained after the animation finishes anyway.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -201,6 +217,53 @@ extension UIView {
                    activate: activate)
     }
 
+    func pinArray(to view: UIView,
```
alexaubry: Isn't it more simple and easier to debug to declaratively write the constraints at creation site than using these helpers? It feels like we're basically re-writing PureLayout. I think we should all discuss the best way to create constraints soon.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
billypchan: No, we don't need it public. We have quite a lot of classes in the UI project are marked as public I think they are unnecessary.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
billypchan: I mark it as `private` first. I create this method for replacing its sibling which returns `[Anchor: NSLayoutConstraint]`. Maybe a struct is a better choice return type.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/AudioRecordViewController.swift
```diff
@@ -179,80 +178,79 @@ private let zmLog = ZMSLog(tag: "UI")
...
+                                        audioPreviewView.bottomAnchor.constraint(equalTo: bottomContainerView.bottomAnchor, constant: -12),
+                                        audioPreviewView.rightAnchor.constraint(equalTo: buttonOverlay.leftAnchor, constant: -12),
+
+                                        cancelButton.centerYAnchor.constraint(equalTo: bottomContainerView.centerYAnchor),
+                                        cancelButton.rightAnchor.constraint(equalTo: bottomContainerView.rightAnchor),
+                                        buttonOverlay.rightAnchor.constraint(equalTo: cancelButton.leftAnchor, constant: -12)])
+
+        constraints.append(contentsOf: cancelButton.setDimensions(length: 56, activate: false))
+
+        NSLayoutConstraint.activate(constraints)
```
vytis: We can probably append both of these arrays and activate constraints in one go

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
vytis: After looking into https://github.com/wireapp/wire-ios/pull/3322 a bit more I am also quite hesitant on expanding the helpers too much. In this case we basically make a homegrown version of PureLayout without any tests and with less consistent API. We should isolate the cases where the default NSLayoutConstraint methods are too verbose and add few helpers for it.

PR #3310, opened 2019-03-29
Fix: Collection detail view has a huge top bar

billypchan:
### Issues

Collection image view has a huge top bar.

### Causes

The navigation bar is doubled when `ConversationImagesViewController` is pushed to a navigation controller.

### Solutions

~~Check a new navigation bar when `viewWillAppear` instead of `viewDidLoad`.
Update `FullscreenImageViewController`'s top overlay visibility after `showCloseButton` flag is updated.~~
Remove the manually created navigation bar in `ConversationImagesViewController`. Always wraps `ConversationImagesViewController` in a navigation controller when it is presented.

### Notes

`Cartography` is removed from `ConversationImagesViewController`.


typfel:
Any reason why would couldn't always present the view controller in a navigation controller?

billypchan:
I think it is a legacy issue. I will check which view presents `ConversationImagesViewController` without wrapping in a navigation controller.

Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.m
```diff
@@ -120,6 +120,14 @@ - (instancetype)initWithMessage:(id<ZMConversationMessage>)message
     return self;
 }
 
+-(void)setShowCloseButton:(BOOL)showCloseButton
```
typfel: Is this method still called?

typfel: And if it's not do we still need the `closeButton` and `topOverlay`?

billypchan: I am removing them, thanks!

PR #3313, opened 2019-04-01
ZIOS-11296: Remove legacy SoundCloud playback

alexaubry:
### Issues

This is a followup of #3282. We want to make sure we don't send too much metadata to SoundCloud, and to simplify dealing with playback during calls etc.

### Solutions

Remove the SoundCloud API and playback related code and update snapshots.

### Notes

Needs design's clearance before merging.


billypchan:
so we have no more custom UI for soundCloud?

alexaubry:
@billypchan We are still waiting for the decision and some tweaks from design, but yes we want to remove the custom UI.

Wire-iOS Tests/ConversationMessageCell/ConversationCellSnapshotTestCase.swift
```diff
@@ -87,6 +88,12 @@ class ConversationCellSnapshotTestCase: CoreDataSnapshotTestCase {
                 XCTAssertTrue(waitForGroupsToBeEmpty([defaultImageCache.dispatchGroup]))
             }
 
+            if waitForTextViewToLoad {
+                // We need to run the run loop for UITextView to highlight detected links
+                let delay = Date().addingTimeInterval(1)
+                RunLoop.main.run(until: delay)
```
billypchan: Can we set this to 0.1? I checked other use cases waiting for UI operations use `Date(timeIntervalSinceNow: 0.1`

alexaubry: Done. Will wait to see if the tests still pass.

alexaubry: It looks like the tests fail because it's not long enough. Will revert.

PR #3314, opened 2019-04-01
ZIOS-11330: Add blacklist property to link metadata

alexaubry:
### Issues

We do not want to display YouTube/SoundCloud/Giphy link previews in the UI, or make them behave like link previews (i.g. with removing the trailing link from the text).

### Solutions

- Bump data model with auto exclusion of blacklisted link previews in the text message data.
- Implement the same behavior in the mock text message data.


vytis:
I see in the snapshots that URL is not coloured in blue, is that expected?

alexaubry:
@vytis The links are not colored because the UITextView detects the links on the main run loop, which we need to run manually in the tests. I made some fixes to support the highlighting in #3313.

PR #3321, opened 2019-04-03
ZIOS-11332: Added support for unsupported AVS version warning

nicorsm:
## What's new in this PR?

Added support to the new "unknown protocol" error provided by AVS and implemented in https://github.com/wireapp/wire-ios-sync-engine/pull/1003. 

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1003


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -583,6 +583,8 @@
 "voice.cancel_button.title" = "Cancel";
 "voice.call_button.title" = "Call";
 "voice.end_call_button.title" = "End Call";
+"voice.call_error.unsupported_version.message" = "You received a call not supported by the current version of Wire.\nGet the latest version of Wire in the App Store.";
```
alexaubry: Is this the final copy?

Wire-iOS/Sources/UserInterface/Calling/CallController/CallController.swift
```diff
@@ -25,7 +25,8 @@ class CallController: NSObject {
 
     fileprivate let callQualityController = CallQualityController()
     fileprivate var scheduledPostCallAction: (()->Void)?
-    fileprivate var token: Any?
+    fileprivate var callStateToken: Any?
```
alexaubry: Maybe we could use an array of Any to store all the tokens, e.g. `fileprivate var observerTokens: [Any] = []`. I think it would be more scalable if we add more observers in the future.

Wire-iOS/Sources/UserInterface/Calling/CallController/CallController.swift
```diff
@@ -183,3 +185,19 @@ extension CallController: CallQualityControllerDelegate {
     }
 
 }
+
+
+extension CallController: WireCallCenterCallErrorObserver {
+    
+    func callCenterCallError(_ error: CallError) {
```
alexaubry: See the comment about the name of the callback the in the SE PR.

Wire-iOS/Resources/Base.lproj/Localizable.strings
nicorsm: No, I need to check...

nicorsm: Copy is under review.

PR #3322, opened 2019-04-03
[POC] improve constraint helpers

billypchan:
### Issues

During the `PureLayout` removal tasks, a lot of similar repeated code are written due to following cases:

#### fitInSuperview
* When using `fitInSuperview` to create 4(or less) constraints, there are 3 cases
  1. we simply want to make a view fit in superview and enable the constraint
  2. we want to store one of four constraints and change it later
  3. we are at the middle of creating constraints batch and want to activate them at the same time

For case 2, `fitInSuperview` returns a dictionary `[Anchor: NSLayoutConstraint]`. But we have to convert the dictionary to `[NSLayoutConstraint]` for case 3.

#### pinToSuperview
Similar code is need when we want to pin a view to super view 

```
 loadingScreenLabel.pinToSuperview(axisAnchor: .centerX, activate: false)
                                 activityIndicator.pinToSuperview(axisAnchor: .centerX, activate: false),
 loadingScreenLabel.pinToSuperview(anchor: .bottom, inset: 40, activate: false)
:
:
```

#### fitInSuperview
Change the return type to a new struct `FittingConstraints`

```
func fitInSuperview(safely: Bool = false,
                        with insets: EdgeInsets = .zero,
                        exclude excludedAnchors: [Anchor] = [],
                        activate: Bool = true) -> FittingConstraints
```

which has a property `array` to return `[NSLayoutConstraint]`, and a `subscript` method for accessing specific constraint for edges.

#### pinToSuperview
Create a new `pinToSuperview` interface to accept an array of `AnchorInput`
```
func pinToSuperview(anchorInputs: [AnchorInput],
                        activate: Bool = true) -> [NSLayoutConstraint] 
```

### Note
`LaunchImageViewController` is updated to demo the use of above update.


Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -42,16 +42,40 @@ struct EdgeInsets {
...
 }
 
-enum AxisAnchor {
-    case centerX
-    case centerY
+struct AnchorInput {
+    let anchorType: AnchorType
+    let inset: CGFloat
+
+    init(anchorType: AnchorType, inset: CGFloat = 0) {
```
vytis: Instead of having a protocol `AnchorType` maybe it's better to have a nested enum that has two cases with associated values. So when new types are added we can get help from compiler to update the code where it is used.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -104,6 +128,33 @@ extension UIView {
...
+    func pinToSuperview(anchorInputs: [AnchorInput],
+                        activate: Bool = true) -> [NSLayoutConstraint] {
+        guard let superview = superview else {
+            fatal("Not in view hierarchy: self.superview = nil")
+        }
+
+        var constraints: [NSLayoutConstraint] = []
+
+        for anchorInput in anchorInputs {
+            if let anchor = anchorInput.anchorType as? Anchor {
```
vytis: Expanding the previous comment - this would become a switch then

Wire-iOS/Sources/UserInterface/Helpers/UIView+Constraints.swift
```diff
@@ -42,16 +42,40 @@ struct EdgeInsets {
...
+    let anchorType: AnchorType
+    let inset: CGFloat
+
+    init(anchorType: AnchorType, inset: CGFloat = 0) {
+        self.anchorType = anchorType
+        self.inset = inset
+    }
+}
+
+struct FittingConstraints {
```
vytis: This struct is equivalent to `[Anchor: NSLayoutConstraint]`, just instead of `constraints.array` we would use `constraints.values`. Do we really need a new type for this?

Wire-iOS/Sources/UserInterface/LaunchImage/LaunchImageViewController.swift
```diff
@@ -78,17 +78,14 @@ class LaunchImageViewController: UIViewController {
...
-        constraints += contentView.fitInSuperview(activate: false).values
-
-        constraints.append(loadingScreenLabel.pinToSuperview(axisAnchor: .centerX, activate: false))
-        constraints.append(loadingScreenLabel.pinToSuperview(anchor: .bottom, inset: 40, activate: false))
-
-        constraints.append(activityIndicator.pinToSuperview(axisAnchor: .centerX, activate: false))
-        constraints.append(activityIndicator.bottomAnchor.constraint(equalTo: loadingScreenLabel.topAnchor, constant: -24))
-
-        NSLayoutConstraint.activate(constraints)
+        let constraints: [Any] = [contentView.fitInSuperview(activate: false).array,
```
vytis: having `[Any]` is a bit troubling, especially when we have all code in Swift. If the problem is that some of these methods return array and some return plain constraints we can combine it like this:
```swift
var constraints = [loadingScreenLabel.pinToSuperview(anchorInputs:[AnchorInput(anchorType: AxisAnchor.centerX),
	                                                                       AnchorInput(anchorType: Anchor.bottom, inset: 40)],
	                                                                   activate: false)]
constraints += contentView.fitInSuperview(activate: false).array
<...>
NSLayoutConstraint.activate(constraints)
```

vytis:
We decided to remove all the helpers so this is not needed anymore!

PR #3324, opened 2019-04-04
Fix: Partner can see a team user profile which is not in conversation

billypchan:
## What's new in this PR?

Remove the user of `URLAction.openUserProfile`. Update `SearchUserViewConroller` to handle multiple calling of `handleSearchResult`.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1004


Wire-iOS/Sources/UserInterface/UserProfile/SearchUserViewConroller.swift
```diff
@@ -68,15 +71,28 @@ final class SearchUserViewConroller: UIViewController {
     }
 
     private func handleSearchResult(searchResult: SearchResult, isCompleted: Bool) {
-        showLoadingView = false
+        guard !resultHandled,
+            (searchResult.count > 0 ||
```
typfel: I think this `count > 0` is unnecessary since you only flip the `resultHandled`,if you find a user. If you remove it and `searchResult.count == 0` this method does nothing anyway.

Wire-iOS/Sources/UserInterface/UserProfile/SearchUserViewConroller.swift
```diff
@@ -93,3 +108,13 @@ final class SearchUserViewConroller: UIViewController {
     }
 }
 
+extension SearchResult {
```
typfel: This extension can be delete now

PR #3327, opened 2019-04-05
Fix: Cannot see self profile via deeplink

billypchan:
### Issues

Cannot see self-profile via deeplink.

### Causes

After https://github.com/wireapp/wire-ios/pull/3324, we always start searching for the userID fro user profile deep link, but self-user does not exist in search result.

### Solutions

directly open profile screen if the user id is self-user's id

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -580,7 +580,13 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
     func sessionManagerShouldExecuteURLAction(_ action: URLAction, callback: @escaping (Bool) -> Void) {
         switch action {
         case .openUserProfile(let id):
-            sessionManager?.showConnectionRequest(userId: id)
+            /// For self user, open the profile viewer without searching
+            if let selfUser = ZMUser.selfUser(),
+                id == selfUser.remoteIdentifier {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -580,7 +580,13 @@ extension AppRootViewController: SessionManagerURLHandlerDelegate {
     func sessionManagerShouldExecuteURLAction(_ action: URLAction, callback: @escaping (Bool) -> Void) {
         switch action {
         case .openUserProfile(let id):
-            sessionManager?.showConnectionRequest(userId: id)
+            /// For self user, open the profile viewer without searching
+            if let selfUser = ZMUser.selfUser(),
+                id == selfUser.remoteIdentifier {
+                sessionManager?.showUserProfile(user: selfUser)
+            } else {
+                sessionManager?.showConnectionRequest(userId: id)
```
typfel: This name is misleading since it's also showing user profile. Could be named `showUserProfile(userId: id)`. I'm with changing it in another PR though.

billypchan: @typfel I am also thinking about moving this checking back to sync engine and rename them. Will be done in another PR.

typfel: Ok 👍

PR #3329, opened 2019-04-05
ZIOS-11329: Fix "Someone joined" text when a user is added on another device

alexaubry:
### Issues

When a Wireless user, unconnected user or service was added by another user or by another device to the conversation, we would show the label "Someone joined". We would also not warn the user that the user is a service, or that services are present.

### Causes

The system message containing the participants change was received before we fetched the user. This is why the name was nil on the ZMUser object, and that we showed the "Someone" template name.

We did not observe the changes on the users of the system message, this is why the text was never updated.

We also did not observe the state of participants as guests / services on the conversation view controller, which means we didn't know a user was a service and we would show the wrong banner.

### Solutions

- Update data model with new observation for external participants. Update the guests bar when this changes.
- Observe the sender and the system message users in the conversation message section controller
- Check for equality of the participants system message instead of just checking for type: otherwise, the diffing algorithm wouldn't pick up the change and the cell wouldn't be reloaded.

### Dependencies

Needs release with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/712


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -390,6 +392,33 @@ - (void)presentParticipantsViewController:(UIViewController *)viewController fro
                                           contentViewController:viewController];
 }
 
+- (void)updateObservedUsersForParticipantsInConversation:(ZMConversation *)conversation
```
typfel: What's the goal of this observation? Maybe we could use the conversation observer instead and listen to `participantsChanged`?

alexaubry:
@typfel After your comments, I added a property on the conversation change info to detect when a service was added, and use it instead of my previous solution. You can check the related data model PR for the details.

PR #3339, opened 2019-04-08
pureLayout removal -  ContactsViewController

billypchan:
## What's new in this PR?

Remove pureLayout from `ContactsViewController` and `ContactsEmptyResultView`.
Remove `purelayout` framework form the project.


Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Constraints.swift
```diff
@@ -0,0 +1,111 @@
...
+         bottomContainerSeparatorView,
+         bottomContainerView].forEach(){$0.translatesAutoresizingMaskIntoConstraints = false}
+
+        let standardOffset: CGFloat = 24.0
+        let titleLabelConstraints = titleLabel.fitInSuperview(with: EdgeInsets(top: UIScreen.safeArea.top, leading: standardOffset, bottom: standardOffset, trailing: standardOffset), activate: false)
+
+        titleLabelTopConstraint = titleLabelConstraints[.top]
+        titleLabelBottomConstraint = titleLabelConstraints[.bottom]
+
+        var constraints: [NSLayoutConstraint] = titleLabelConstraints.values.map{$0}
```
vytis: I think it should work without the `map{$}`, values of the dictionary returned by `fitInSuperview` are already `NSLayoutConstraint`

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Internal.h
```diff
@@ -22,33 +22,41 @@ static NSString * const _Nonnull ContactsViewControllerSectionHeaderID = @"Conta
...
-@property (nonatomic, nullable) NSLayoutConstraint *titleLabelTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *titleLabelBottomConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *closeButtonTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *closeButtonBottomConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *topContainerHeightConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *searchHeaderTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *searchHeaderWithNavigatorBarTopConstraint;
-
-@property (nonatomic, nullable) NSLayoutConstraint *bottomEdgeConstraint;
+@property (nonatomic, null_unspecified) IconButton *cancelButton;
```
vytis: If we know that these things will not be nil when used, then we might as well make it all non-nullable with `NS_ASSUME_NONNULL_BEGIN`

Wire-iOS Tests/ContactsViewControllerSnapshotTests.swift
```diff
@@ -1,28 +1,39 @@
...
 
-#import <UIKit/UIKit.h>
-#import "Button.h"
+final class ContactsViewControllerSnapshotTests: ZMSnapshotTestCase {
+    
+    var sut: ContactsViewController!
+    
+    override func setUp() {
+        super.setUp()
+        sut = ContactsViewController()
```
vytis: The snapshot does not have any entries in it. It would be much more useful if it's pre-filled with some information

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Constraints.swift
billypchan: The compiler complain for "Expression type 'Dictionary<Anchor, NSLayoutConstraint>.Values' is ambiguous without more context" without `map{$}`.
A better version would be `.map{$0.value}` which maps the dictionary to `[NSLayoutConstraint]`

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Internal.h
billypchan: There is one array in the private header has to be `nullable` (`NSArray *actionButtonTitles`) so I prefer to define these properties one by one, to remind the Swift caller these UI related properties are optional.

Wire-iOS Tests/ContactsViewControllerSnapshotTests.swift
billypchan: I will remove this test since `InviteContactsViewControllerSnapshotTests` already tested the layout of `ContactsViewController`.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Constraints.swift
vytis: This is probably expected - the operators are defined on Collection level and works with multiple kinds of collections.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Internal.h
vytis: If `NS_ASSUME_NONNULL_BEGIN` is added you can override it by setting some properties to `nullable`. `null_unspecified` is the default when ObjC API is not optimized for Swift - most of the time you can know if property can be `nil` or not and mark accordingly. In case of nullable arrays we always rewrite it to be nonnull and just be empty instead.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+Internal.h
```diff
@@ -16,40 +16,54 @@
...
-@property (nonatomic, nullable) NSLayoutConstraint *titleLabelTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *titleLabelBottomConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *closeButtonTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *closeButtonBottomConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *topContainerHeightConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *searchHeaderTopConstraint;
-@property (nonatomic, nullable) NSLayoutConstraint *searchHeaderWithNavigatorBarTopConstraint;
-
-@property (nonatomic, nullable) NSLayoutConstraint *bottomEdgeConstraint;
+@property (nonatomic) TransformLabel *titleLabel;
```
vytis: So now all of it is `null_unspecified`? Or you missed adding `NS_ASSUME_NONNULL_BEGIN / NS_ASSUME_NONNULL_END`?

billypchan: Yes, I checked they are all `null_unspecified` after `NS_ASSUME_NONNULL_BEGIN` is removed. The `nullable` properties are moved back to .m file.

vytis: My question was more why you made it `null_unspecified` - this is basically only used for code that is old and has no nullability annotations. These properties might as well be `nonnull` because it's all populated before first use anyway.

billypchan: added `NS_ASSUME_NONNULL_BEGIN` for your suggestion. Thanks!

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.h
```diff
@@ -62,7 +62,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic) ColorSchemeVariant colorSchemeVariant;
 
 /// Button displayed at the bottom of the screen. If nil a default button is displayed.
-@property (nonatomic) Button *__nullable bottomButton;
+@property (nonatomic, null_unspecified) Button * bottomButton;
```
typfel: Let's keep this `nullable` because the `bottomButton` is optional it's a good indicator that we need to handle both cases.

PR #3340, opened 2019-04-09
ZIOS-11353: Add live validation label for password rules

alexaubry:
### Issues

We want to display a tooltip under the password text field to show the users the rules before they start typing (in gray), and show them why it failed (in red) as they're typing.

### Solutions

We replace the use of `TextFieldValidator.ValidationError` in the `valueValidated` callback to use an enum that has both `error` and `info`. We also specify if the error needs to be displayed as such, so that we can remove the red label when the user clears the text field. 

The side effect of this is that we needed to `TextFieldValidator.ValidationError.none`, which was not semantically correct inside the new enum (it was not technically an error).

We also add the `DefaultValidatingStepDescription` protocol, which is a description with an initial validation result that we should display when there is no error or info returned by the text field validator.

### Notes

This is a task related to the strong passwords implementation. After we merge this PR, we will be able to incorporate the rules in the validator and update the copy for them.


vytis:
Is it expected that the screenshots are shifted by few pixels?

alexaubry:
@vytis Yes it's expected, there were some issues with the spacing that I fixed (I was not hiding the right view on the stack view).

PR #3342, opened 2019-04-09
PureLayout remove - PermissionDeniedViewController

billypchan:
## What's new in this PR?

Remove `PureLayout` form `PermissionDeniedViewController`. The snapshot test is included in `testForPremissionDeniedViewController`.

### DIFF
--- a/Wire-iOS/WireBridgingHeader.h
+++ b/Wire-iOS/WireBridgingHeader.h
@@ -118,7 +118,10 @@
 #import "ContactsDataSource.h"
 #import "StartUIViewController.h"
 #import "Button.h"
+
 #import "PermissionDeniedViewController.h"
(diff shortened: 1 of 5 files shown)


Wire-iOS/WireBridgingHeader.h
```diff
@@ -118,7 +118,10 @@
 #import "ContactsDataSource.h"
 #import "StartUIViewController.h"
 #import "Button.h"
+
 #import "PermissionDeniedViewController.h"
+#import "PermissionDeniedViewController+Internal.h"
```
alexaubry: Is there a reason for creating this new header?

billypchan: It is for exposing objc internal interface for Swift extension since objc has no scope control feature.

PR #3337, opened 2019-04-08
ZIOS-11345: Support email field in ProfileDetailsViewController

nicorsm:
## What's new in this PR?

Added support for the new `emailAddress` field of user objects. The email address will looks like a rich profile field inside `ProfileDetailsViewController`.

PR #3344, opened 2019-04-09
ZIOS-11354: Add live validation of password in Add email settings

alexaubry:
### Issues

We want to warn the phone-based users about password rules when they set their initial password in the Settings.

### Solutions

- Create a cell to display the value of the ValueValidation
- Transition to autolayout automatic height
- Allow passing a UserType to construct the screen and add snapshot tests


Wire-iOS Tests/ChangeEmailViewControllerTests.swift
```diff
@@ -0,0 +1,55 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class ChangeEmailViewControllerTests: ZMSnapshotTestCase {
```
billypchan: After https://github.com/wireapp/wire-ios/pull/3337, `selfUser` of `CoreDataSnapshotTestCase` has default email address. You can replace `ZMSnapshotTestCase` with `CoreDataSnapshotTestCase` and user `selfUser` insteand of `mockUser`.

PR #3347, opened 2019-04-09
ZIOS-11350: Integrate new password rules

alexaubry:
## What's new in this PR?

We introduce new rules to make passwords stronger.

### Solutions

- Add a helper to load the rules from the JSON file in the configuration.
- Add a helper to generate the text for the rules dynamically (we won't need to update the string for on-premises customers if they decided to change them).
- Update the text field validator to use the new password rule set.
- Update the text field validator enum to handle the password-specific failures.


Wire-iOS/Sources/UserInterface/Settings/ChangingDetails/Views/ValueValidationCell.swift
```diff
@@ -87,7 +87,7 @@ class ValueValidationCell: UITableViewCell {
 
             label.accessibilityIdentifier = "validation-failure"
             label.text = error.errorDescription
-            label.textColor = UIColor.Team.errorMessageColor
+            label.textColor = UIColor(rgb: 0xfc3e37, alpha: 1)
```
typfel: This color doesn't have a  name?

alexaubry: I can add a name. But I tweaked the version from the login UI slightly so that there is more contrast with the background in the settings.

typfel: So it's a black/white theme issue

alexaubry: Pretty much. I can add `UIColor.Team.darkErrorMessageColor` then?

typfel: I'm ok with changing it in a different PR

PR #3348, opened 2019-04-10
ZIOS-11352: Add red dot on password field when the password is invalid

alexaubry:
### Issues

According to the spec, we want to display a red dot on the right of the text field as the user is typing their new password until it becomes valid.

### Solutions

- Replace the accessory view with a stack view that contains both the 
- Add necessary handlers to show and hide the dot
- Move colors to the color scheme following yesterday's comments from @typfel


Wire-iOS/Sources/UserInterface/Components/Views/EmailPasswordTextField.swift
```diff
@@ -140,6 +151,11 @@ class EmailPasswordTextField: UIView, MagicTappable {
         separatorContainer.view.backgroundColor = color
     }
 
+    func applyColorScheme(_ colorSchemeVariant: ColorSchemeVariant) {
+        emailField.colorSchemeVariant = colorSchemeVariant
```
typfel: I wouldn't except `applyColorScheme()` to set the `colorSchemeVariant`. Couldn't we set these in `colorSchemeVariant didSet {}`?

alexaubry: Updated.

PR #3353, opened 2019-04-11
Show an alert when the user changes their status

typfel:
## What's new in this PR?

When availability status starts influencing if you get notifications or not we would like to inform users about this when they change their status.

This reminder alert can be disabled by tapping the "don't remind me again".

### Notes

I'm not very happy about adding this logic to a view, but re-factoring this blew up so that will come in a separate PR.


Wire-iOS/Sources/Components/Settings+Availability.swift
```diff
@@ -0,0 +1,41 @@
...
+    }
+    
+}
+
+extension Settings {
+    
+    func shouldRemindUserWhenChanging(_ availability: Availability) -> Bool {
+        guard let defaults = defaults() else { return true }
+        
+        return !defaults.bool(forKey: availability.dontRemindMeUserDefaultsKey)
```
alexaubry: Small nitpick: we could use the nil coalescing, since you're doing it for the method under:

```swift
return defaults()?.bool(forKey: availability.dontRemindMeUserDefaultsKey) != true
```

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1521,6 +1521,17 @@
 "availability.away.placeholder" = "%@ is away";
 "availability.busy.placeholder" = "%@ is busy";
 
+"availability.reminder.away.title" = "You are now away";
+"availability.reminder.away.message" = "When you are away you won't be receving notification or calls";
```
billypchan:
```suggestion
"availability.reminder.away.message" = "When you are away you won't be receving notifications or calls";
```

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1521,6 +1521,17 @@
...
+"availability.reminder.busy.title" = "You are now busy";
+"availability.reminder.busy.message" = "When you are busy you will only receive notification about mentions or replies";
```
billypchan:
```suggestion
"availability.reminder.busy.message" = "When you are busy you will only receive notifications about mentions or replies";
```

PR #3354, opened 2019-04-11
fix: long user name breaks self profile view layout

billypchan:
### Issues

long user name breaks self-profile view layout

### Causes

The label's horizontal `ContentCompressionResistancePriority` is too high.

### Solutions

Lower horizontal `ContentCompressionResistancePriority` to `.defaultLow` and set `lineBreakMode` ` .byTruncatingTail` to limit the width of the label.


Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
```diff
@@ -62,7 +62,7 @@ final class ProfileView: UIView, Themeable {
     /// The view controller that displays the view.
     weak var source: UIViewController?
     
-    @objc dynamic var colorSchemeVariant: ColorSchemeVariant = ColorScheme.default.variant {
+    var colorSchemeVariant: ColorSchemeVariant = ColorScheme.default.variant {
```
typfel: Why remove the `dynamic` annotation?

Wire-iOS/Sources/Components/CustomSpacingStackView.swift
```diff
@@ -18,7 +18,7 @@
 
 import UIKit
 
-@objcMembers class CustomSpacingStackView: UIView {
+final class CustomSpacingStackView: UIView {
```
typfel: Let's not do unrelated changes.

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileView.swift
billypchan: It is no longer necessary after its callers are converted to Swift.

typfel: The dynamic annotation is there to support `UIAppearance` which relies on the method being dynamic.

PR #3358, opened 2019-04-12
Fix: Line height is not constant if a multiple-line message contains mention

billypchan:
### Issues

When a paragraph has a mention, the line height is different from other paragraphs.

### Causes

The mentioned text has no paragraph style attribute and it affects the whole paragraph.

### Solutions

Add the paragraph style attribute to the mentioned text. This fix would not apply to the quoted message since it affects the line break mode in this case.

before the fix:

Wire-iOS Tests_ReferenceImages_64_Wire_iOS_Tests MarkDownSnapshotTests_testMentionWithAtSign_375@2x_Before_3de72b1f22b9906df0995d33938a26f5e5a3a2a7

after the fix, notice the first paragraph's line height:

testMentionInFirstParagraph_375@2x


billypchan:
This PR affects the reply block line break mode. I will fix that.

PR #3359, opened 2019-04-15
ZIOS-11367: Refine icon infrastructure

alexaubry:
### Issues

In this PR, we tackle a long-running issue with icons and integrating with our Paint Code StyleKit (https://github.com/wireapp/ios-architecture/issues/40).

The main problems were:
- The ZetaIconType enum was not type safe. There was no guarantee that a case was mapped to an actual PaintCode rendering method. We were not notified when a PaintCode icon was removed.
- To add more sizes, we needed to have very specific cases in the size enum (e.g. ZetaIconSizeRegistrationButton)
- Relying on selectors and invocations prevented us from moving to a Swift implementation.

The goal of this PR is to introduce type safety and create a Swift-First API.

#### Swift-First API

- We rename `ZetaIconType` to `StyleKitIcon`
- We add the rendering method as a factory method on the icon type, instead of a convenience initializer on UIImage (e.g. `StyleKitIcon.camera.makeImage(size:color:)`)
- We rename `ZetaIconSize` to `StyleKitIcon.Size`. This enum has a minimal amount of predefined cases (tiny, medium, large) and allows using custom values, with a case that has an associated value (`Size.custom(CGFloat)`). We can also use Double and Integer literals without needing to add new cases.
- We use `UIGraphicsImageRenderer` to draw the images, instead of the old C-based API.
- We add a helper on `UIImageView` to set the icon.
- We refine the `ImageButton` methods used to set an image for  to take advantage of nullability and the rich size enum.

This is an example of how we use the new API compared to the old one:

**Before**

```swift
let imageView = UIImageView()
imageView.image = UIImage(for: .camera, iconSize: .large, color: .white)

navigationBar.image = UIImage(for: .camera, fontSize: 42, color: .white)

iconButton.setIcon(.camera, withSize: .registrationButton, color: .white)

```

**After**

```swift
// Helper to set the icon on an image view
let imageView = UIImageView()
imageView.setIcon(.camera, size: .large, color: .white)

// Creating from the icon directly, without needing a separate initializer for the custom value
navigationBar.image = StyleKitIcon.camera.makeImage(size: 42, color: .white)

// Setting on a button directly, without needing Objective-C compatibility
iconButton.setIcon(.camera, size: .custom(32), color: .white)
```

#### Type-Safety

To ensure type safety, we use code generation to create the enum of available icons every time we build the app.

To achieve this, we updated the PaintCode file with @eike96 to have the name included in the asset instead of identifiers, and we wrote a build script in Swift that parses the style kit file with a regex to get the list of icons. This script then generates the enum Swift file (which we don't add to source control).

This approach ensures that our enum is always up to date, and that we can catch unavailable images at build time, rather than getting an empty image at runtime.

Another structure that the script generates is the  `RenderingProperties` of the image, that maps a case to the function pointer it needs to call to draw the image in the current context. This means we don't rely on selectors anymore, which eliminates a lot of possible failures.

### Next Steps

We can identify the places where we need to use `StyleKitIcon.makeImage`, and make these properties take a `StyleKitIcon` directly.

### Notes

We also remove the emoji keyboard code from the conversation screen (it was hidden from there) and removed the deprecated drafts data as discussed with @marcoconti83.


vytis:
Great job doing all this tedious refactoring! Also I think the drafts CoreData store can be deleted.

alexaubry:
@vytis Done!

marcoconti83:
@alexaubry great PR description! 👍

PR #3360, opened 2019-04-15
Show alert informing about changes in availability behaviour

typfel:
## What's new in this PR?

Show alert informing about changes in availability behaviour.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1009
- [x] https://github.com/wireapp/wire-ios-data-model/pull/713

### Todo

- [x] Re-use alerts we display when the users changes their status


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1513,6 +1513,10 @@
 "availability.message.cancel" = "Cancel";
 "availability.message.set_status" = "Set a status";
 
+"availability.behaviour_change_alert.title" = "Changes to the status feature";
+"availability.behaviour_change_alert.message" = "From now on the your status will also affect your notification setting across all conversations. ";
+"availability.behaviour_change_alert.ok" = "No";
```
alexaubry: Is "no" the correct copy here? Would expect it to be "OK".

typfel: Typo :) But copy is not anyway not final.

PR #3368, opened 2019-04-17
Rewrite constraints without helpers + update UI

alexaubry:
### Issues

- Remove usage of layout helpers and adhere to the style we agreed upon
- Implement the new style of binary action buttons


vytis:
One thing that we could do to make the work a bit easier is to turn off warnings as errors and mark the constraint helper methods as deprecated. This would make a lot of warnings, but maybe they would get fixed faster.

alexaubry:
@vytis Good idea, will do in another PR.

PR #3369, opened 2019-04-17
Support for custom backends

vytis:
### Issues

We want to allow users to switch to a custom backend instance with a special deeplink.

### Solutions

Added handling of the deeplink and showing all relevant errors and alerts when changing backend.
Updated landing view controller UI to show the name and configuration JSON of the custom backend.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-build-configuration/pull/8
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1011
- [x] https://github.com/wireapp/wire-ios-transport/pull/135

### To Do

- [x] Build the various links the app opens (`URL+Wire.swift`) with correct backend URLs.


Scripts/download-assets.sh
```diff
@@ -73,7 +73,7 @@ else
     fi 
 
     echo "Cloning assets from ${REPO_URL}"
-    git clone --depth 1 ${REPO_URL} ${CONFIGURATION_LOCATION}
+    git clone --depth 1 -b feature/custom-backend --single-branch ${REPO_URL} ${CONFIGURATION_LOCATION}
```
vytis: Needs to be reverted before merging

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -234,12 +293,27 @@ class LandingViewController: AuthenticationStepViewController {
     }
 
     // MARK: - Adaptivity Events
+    
+    private var logoHidden: Bool {
```
typfel: For consistently this could be moved into method `updateLogoVisibility()`

WireCommonComponents/BackendEnvironment+Shared.swift
```diff
@@ -20,10 +20,17 @@ import Foundation
 import WireTransport
 
 extension BackendEnvironment {
-    public static let shared: BackendEnvironment = {
-        let type = EnvironmentType(userDefaults: .standard)
+    
+    public static let backendSwitchNotification = Notification.Name("backendEnvironmentSwitchNotification")
+    
+    public static var shared: BackendEnvironment = {
```
typfel: I understand it was already like this but couldn't the `environment` be property on the `SessionManager` and the UI would need do the bookkeeping of saving and restoring it.

One singleton less :)

vytis: Yeah good point, not sure exactly why we have it in UI actually

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
vytis: Well the problem is that we use it in `viewDidLayoutSubviews()`. We hide the logo on small devices and need to prevent it reappearing after switching to custom backend

WireCommonComponents/BackendEnvironment+Shared.swift
vytis: To move it to SE we would also need to change lots of initialzers and inject `Bundle.backendBundle` instead of `Environment`, because in SE we don't know anything about this backend bundle. I would keep it as is for the time being, it does not cause any issues at the moment.

typfel: Ok

fastlane/Fastfile
```diff
@@ -6,6 +6,8 @@ platform :ios do
     desc "Fetch dependencies"
     lane :prepare do |options|
         build_type = options[:build_type]
+        # Delete configurations in case we switch ti different branch or repo
```
alexaubry: Typo

```suggestion
        # Delete configurations in case we switch to different branch or repo
```

PR #3374, opened 2019-04-23
Fix size of Wire logo in share extension

alexaubry:
### Issues

After the icons refactor, the size of the Wire logo on top of the share extension was broken.

### Solutions

Fix the size by downscaling the image.

### DIFF
--- a/Wire-iOS Share Extension/ShareExtensionViewController.swift
+++ b/Wire-iOS Share Extension/ShareExtensionViewController.swift
@@ -132,10 +132,11 @@ class ShareExtensionViewController: SLComposeServiceViewController {
     }
 
     private func setupNavigationBar() {
+        let iconSize = CGSize(width: 32, height: 26.3)
         guard let item = navigationController?.navigationBar.items?.first else { return }
         item.rightBarButtonItem?.action = #selector(appendPostTapped)
         item.rightBarButtonItem?.title = "share_extension.send_button.title".localized
-        item.titleView = UIImageView(image: WireStyleKit.imageOfLogo(color: .black))
+        item.titleView = UIImageView(image: WireStyleKit.imageOfLogo(color: .black).downscaling(to: iconSize))
     }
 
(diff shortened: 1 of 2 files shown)


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -132,10 +132,11 @@ class ShareExtensionViewController: SLComposeServiceViewController {
     }
 
     private func setupNavigationBar() {
+        let iconSize = CGSize(width: 32, height: 26.3)
         guard let item = navigationController?.navigationBar.items?.first else { return }
         item.rightBarButtonItem?.action = #selector(appendPostTapped)
         item.rightBarButtonItem?.title = "share_extension.send_button.title".localized
-        item.titleView = UIImageView(image: WireStyleKit.imageOfLogo(color: .black))
+        item.titleView = UIImageView(image: WireStyleKit.imageOfLogo(color: .black).downscaling(to: iconSize))
```
typfel: Why couldn't we achieve the same effect with `contentMode = .scaleAspectFit`?

alexaubry: I'll try

alexaubry: `imageOfLogo` returns a 272pt by 224pt image, so it still looks odd to fit, because it will take the entire height of the navigation bar. I guess we could also add a height constraint, if that's better than downscaling the image.

typfel: Didn't you add a way to construct icons of custom sizes? I don't have strong opinion on what's better constraint/downscaling.

alexaubry: This is a special case, it is generated differently in the style kit so it's actually not a `StyleKitIcon`. I could always speak with Eike later to make it a proper icon.

PR #3377, opened 2019-04-24
Disabled team creation

vytis:
### Issues

There is no need for creating teams on mobile anymore.

### Solutions

To keep the layout the same we just make the button transparent for now, later will be removing the code related to this feature.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -855,7 +855,7 @@
 // Settings - top level
 "self.settings.create_team.title" = "Create a team";
 "self.settings.manage_team.title" = "Manage Team";
-"self.settings.add_team_or_account.title" = "New Team or Account";
+"self.settings.add_team_or_account.title" = "Add Account";
```
typfel: So I guess we need update localisation for other languages too?

PR #3383, opened 2019-04-30
Fix scroll position when opening conversation with many unread messages

typfel:
### Issues

When opening a conversation with more unread messages than the size of the message window we would lose the scroll position.

### Causes

When opening a conversation with unread message the following things were happening:

1. Load initial message window of size 30
2. Load message window which includes the first unread message
3. Scroll to the first unread message
4. We change properties of some messages when they appear on screen and save (link attachments)
5. The fetch results controller monitors the save notifications and updates its `fetchObjects` property
6. We discover that there's new messages in the message window and perform a table view update

The problem is caused by the the last table view update which is happening because the `NSFetchResultsController` is not very strict about the `fetchLimit` and might include more messages if they are modified after the initial fetch.

### Solutions

Only work on the `fetchObjects` within the `fetchLimit` if the message window is detached from the bottom.

### Todo
- [x] Handle new incoming messages

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationTableViewDataSource.swift
@@ -85,7 +85,10 @@ final class ConversationTableViewDataSource: NSObject {
     }
     
     @objc public var messages: [ZMConversationMessage] {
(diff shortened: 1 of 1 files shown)


typfel:
This solution doesn't currently work with new incoming messages which are effectively ignored.

PR #3384, opened 2019-04-30
Feature/swift5

billypchan:
## What's new in this PR?

upgrade the project for Swift 5

### framework
- [x] https://github.com/wireapp/wire-ios-shared-resources/pull/35
- [x] https://github.com/wireapp/ios-snapshot-test-case/pull/3
- [x] https://github.com/wireapp/DifferenceKit/pull/1
- [x] https://github.com/wireapp/Down/pull/14
- [x] https://github.com/wireapp/Cartography/pull/4
- [x]  wireapp/wire-ios-ziphy
- [x] wireapp/wire-ios-sync-engine
- [x] wireapp/wire-ios-share-engine
- [x] wireapp/wire-ios-canvas

### Swift 5 related issues
- [x] fix 13 warnings in common components target
- [x] fix ~~234~~ Swift5 related warnings in Wire-iOS target
- [x] fix warnings in tests


Wire-iOS Tests/Utils/VariantsBuilder.swift
```diff
@@ -43,6 +43,10 @@ struct WritableKeyPathApplicator<Type>: Hashable {
     var hashValue: Int {
```
alexaubry: If we implement `hash(into:)` we probably also need to remove `hashValue`.

Wire-iOS/Sources/UserInterface/Conversation/Mentions/ZMUser+Mention.swift
```diff
@@ -32,6 +32,10 @@ class HashBox<Type: NSObjectProtocol>: Hashable, Equatable {
     var hashValue: Int {
```
alexaubry: Same here.

Wire-iOS Tests/Utils/VariantsBuilder.swift
typfel: @billypchan I think we have similar case in https://github.com/wireapp/wire-ios-cryptobox/pull/30 in `GenericHash`

PR #3386, opened 2019-05-02
Improve initial position when a loading conversation

typfel:
### Issues

When loading a conversation with unread messages we would first load and display the last messages and then reload the messages again around first unread message. This is both unnecessary performance wise and looks bad.

### Causes

We didn't used to reload message when scrolling to the first unread message but after some re-factoring that's now the case.

### Solutions

Before displaying the table view load either the last messages or messages near the first unread message.

NOTE: this needs to be done in `viewDidLayoutSubviews` because only then is bounds set which makes it possible to scroll the table view.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/UpsideDownTableView+Scrolling.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/UpsideDownTableView+Scrolling.swift
@@ -29,7 +29,7 @@ extension UpsideDownTableView {
     func scroll(toIndex indexToShow: Int, completion: ((UIView)->())? = .none) {
         guard numberOfSections > 0 else { return }
         
-        let rowIndex = numberOfCells(inSection: indexToShow) - 1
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/UpsideDownTableView+Scrolling.swift
```diff
@@ -19,19 +19,33 @@
...
     @objc(scrollToBottomAnimated:)
     func scrollToBottom(animated: Bool) {
         // upside-down tableview's bottom is rightside-up tableview's top
         super.scrollToTop(animated: animated)
     }
+    
+    func scroll(toIndex indexToShow: Int, completion: ((UIView)->())? = .none) {
+        guard numberOfSections > 0 else { return }
+        
+        let rowIndex = numberOfCells(inSection: indexToShow) - 1
```
billypchan:
```suggestion
        guard let rowIndex = numberOfCells(inSection: indexToShow) - 1,  rowIndex >= 0 else { return }
```

PR #3387, opened 2019-05-02
Improve performance when opening conversations with unread messages

typfel:
### Issues

Opening a conversation with many unread text messages would take a long time.

### Causes

The first time a text message is displayed we will fetch any potential link attachments. When opening a conversation with many unread messages this results in a `enqueueChanges`, which results in save call for each method call. This in combination with the fetch results controller observing each save can take seconds to complete.

### Solution 1

~Let sync engine perform a delayed save so that we will only perform a single save at the end.~

### Solution 2

Add new method `enqueueDelayedChanges` to perform changes with a delayed save.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1020


typfel:
Performance needs more investigation, not as good as solution 1.

PR #3390, opened 2019-05-03
Snapshot test for  typing indication and refactoring

billypchan:
## What's new in this PR?

Add snapshot test for typing indicator.
Refactoring the code about typing indicator reveal. Do not reveal it animated when viewDidLoad.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+TypingIndicator.swift
```diff
@@ -0,0 +1,37 @@
...
+
+extension ConversationInputBarViewController {
+
+    @objc
+    func updateTypingIndicatorVisibility(animated: Bool) {
+        guard let typingIndicatorView = typingIndicatorView else { return }
+
+        let count = typingUsers?.count ?? 0
+
+        if let typingUsers = typingUsers as? Set<ZMUser>, count > 0 {
```
typfel: Let's annotate `typingUsers` with the correct type in `ConversationInputBarViewController+Private.h`.

billypchan: It is declared as '@property (nonatomic, nullable) NSSet *typingUsers;' in the header.

typfel: Right I mean let's add the type annotation like this:

`@property (nonatomic, nullable) NSSet<ZMUser *> *typingUsers`

PR #3395, opened 2019-05-07
Conversation screen - open profile when tapped on the avatar

billypchan:
## What's new in this PR?

Allow user to tap on an avatar in the conversation to see the profile of a user.


billypchan:
@alexaubry I checked the desktop app do not recognize tap gesture on the name label. We can collect the feedbacks from users and the UX team.

PR #3396, opened 2019-05-08
Fix warning for deprecated encodedOffset property

billypchan:
### Issues

`String.Index.encodedOffset` is deprecated after Swift 4.2

### Solutions

Create a string's whole range with `utf8.count`


Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
```diff
@@ -18,6 +18,12 @@
 
 import Foundation
 
+extension String {
+    var wholeRangeInUTF8: NSRange {
+        return NSRange(location: 0, length: utf8.count)
```
alexaubry: To get the whole range, just use `NSRange(startIndex ..< endIndex, in: self)`. 

This abstracts away the implementation details, if Apple ever move away from UTF8. Besides, NSString uses UTF16, so using UTF8 count is incorrect.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
```diff
@@ -28,9 +34,9 @@ import Foundation
     let searchQueryMatchRange: NSRange
 
     init?(text: String?, cursorPosition: Int) {
-        guard let text = text else { return nil }
-        let wholeRange = NSRange(location: 0, length: text.endIndex.encodedOffset)
-        let matches = mentionRegex.matches(in: text, range: wholeRange)
+        guard let text = text, text.count > 0 else { return nil }
```
alexaubry: Using `isEmpty` is better than count, because counting the number is O(n). Also, why do we need to check if it’s not empty? Doesn’t NSRegularExpression cover that?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
billypchan: I tried `NSRange(startIndex ..< endIndex, in: self)` and also `NSRangeFromString()` and create a NSRange from Range but they can not pass the cases fro Emojis. May be we can rewrite the `NSRegularExpression.matches` before to accept `Range` instead of `NSRange`.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
billypchan: Maybe we can introduce SwiftLint for this kind of fixes.

billypchan:
@alexaubry Let's wait for the @vytis's comment about this.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
alexaubry: @billypchan The reason the tests fail is because UTF8 and UTF16 encoding do not produce the same number of bytes, so that means the UTF8 range will have more bytes than the UTF16 range, which is what NSString needs. You can use this tool to check the count needed for the tests: https://ethproductions.github.io/bytes/?e=utf-16

vytis:
The main problem with this PR is that `NSString` and corresponding `NSRange` is using UTF16 encoding. However from Swift 5 `String` can use different encodings under the hood. When using from pure Swift it's UTF8 but can be UTF16 when working with strings coming from Objective-C. One important thing that neither `Range` nor `NSRange` work in bytes, so in this situation we should not care what is the length of the string in bytes.

To cut the long story short it should be:

```swift
extension String {
    var wholeRange: NSRange {
        return NSRange(location: 0, length: endIndex.utf16Offset(in: self))
    }
}
```

Screenshot 2019-05-20 at 11 48 31

Wire-iOS/Sources/UserInterface/Conversation/InputBar/MentionsHandler.swift
```diff
@@ -18,6 +18,12 @@
 
 import Foundation
 
+extension String {
+    var wholeRange: NSRange {
+        return NSRange(location: 0, length: endIndex.utf16Offset(in: self))
```
alexaubry: Are we sure we want to do it like this? The approach with passing the range and the string is probably the most encoding-agnostic way, and lets Apple take care of the underlying implementation.

```suggestion
        return NSRange(startIndex ..< endIndex, in: self)
```

vytis: There is more info on [SE-241](https://github.com/apple/swift-evolution/blob/master/proposals/0241-string-index-explicit-encoding-offset.md), but the main idea in this case is that we need to make sure we use UTF16 indexes when constructing  `NSRange`, because this is the encoding that `NSString` is using. Your proposed change would be fine if we would be creating a `Range` and using only Swift APIs, but in this case we are passing it to `NSRegularExpression` which is an Objective-C class working with `NSString`.

vytis: I misread the proposed change a bit and it seems indeed that it's a cleaner API that basically does the same under the hood - https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSRange.swift#L257-L261

PR #3398, opened 2019-05-10
bump cartography

billypchan:
## What's new in this PR?

bump cartography fro Swift 5 support

### TODO
- [x] check the cartography release is valid
- [x] check CI can not fetch new commits of Cartography issue


vytis:
@billypchan seems to be working now. The problem was that our fork of Cartography had tags that were the same (e.g. `3.0.6`) but pointing to different commits. Azure Pipelines has Carthage cache pre-populated with most popular dependencies including Cartography. There is a bug in Carthage that makes `carthage bootstrap` fail in the situation when we have conflicting tags (https://github.com/Carthage/Carthage/issues/2786). 
The fix was simple - remove all the custom tags from our fork and use only those that are in the upstream repository.

PR #3400, opened 2019-05-13
ZIOS-11404 - Indicate legal hold in conversation title & conversation details

nicorsm:
## What's new in this PR?

Added legal hold icon to title view and conversation detail.
To be adapted with Jacob's implementation at https://github.com/wireapp/wire-ios/compare/feature/legalhold-details#diff-3ffdd914b9f464d5d9572f6428f6ac94R181 and needs fixes + tests.


typfel:
I took the decision in data model to call legal hold `legalhold` (all lower case ) when using it in properties and variable names.

In my eyes it looks better but I'm open for discussion, but whatever we pick I think we should be consistent.

nicorsm:
Personally, I see better `LegalHold` since they're two different words, but it is indifferent to me.

typfel:
Yes, I see the argument that they are two different words. I also don't have strong opinion on this, let's sleep on it and then we can decide :)

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -200,6 +200,7 @@
 "list.archived_conversations_close" = "Close archive";
 
 "conversation.voiceover.verified" = "verified";
+"conversation.voiceover.legalhold" = "conversation is under legal hold";
```
alexaubry: Accessibility labels shouldn't be sentences. We should probably do something like Label: "Legal Hold", Value: "Active"

Wire-iOS/Sources/UserInterface/Availability/IconStringsBuilder.swift
```diff
@@ -25,15 +25,21 @@ public class IconStringsBuilder {
     // - and, obviously, a color
     
     static func iconString(with icon: NSTextAttachment?, title: String, interactive: Bool, color: UIColor) -> NSAttributedString {
+        return iconString(with: icon == nil ? [] : [icon!], title: title, interactive: interactive, color: color)
+    }
+    
+    static func iconString(with icons: [NSTextAttachment], title: String, interactive: Bool, color: UIColor) -> NSAttributedString {
```
alexaubry: Can we check how it behaves in RTL mode? (use "RTL pseudo-language" as the language in the scheme options). I'm worried that hard-coding the order with spaces might not respect the semantics of this scenario (or other languages, e.g. Chinese that doesn't have spaces). Maybe we can also research a better way to combine string lists?

Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
```diff
@@ -41,20 +41,31 @@ class ConversationTitleView: TitleView {
...
                         title: conversation.displayName.localizedUppercase,
                         interactive: self.interactive && conversation.relatedConnectionState != .sent)
         
+        var accessibilityLabel = conversation.displayName.localizedUppercase
+        
         if conversation.securityLevel == .secure {
-            self.accessibilityLabel = conversation.displayName.localizedUppercase + ", " + "conversation.voiceover.verified".localized
-        } else {
-            self.accessibilityLabel = conversation.displayName.localizedUppercase
+            accessibilityLabel += ", " + "conversation.voiceover.verified".localized
```
alexaubry: Same here with RTL support.

Wire-iOS/Sources/UserInterface/Availability/IconStringsBuilder.swift
nicorsm: It doesn't change between LTR and RTL. I've added some controls to diversify the UI, so the title view is mirrored in RTL languages. About spaces, I tried with Chinese and it works as well. Anyway, I switched to a padding element that is added as a blank `NSTextAttachment`. Changes are coming!

Wire-iOS/Resources/Base.lproj/Localizable.strings
nicorsm: I understand... we have two values here that could be `active` or `disabled`: `verified` and `legalhold`.
- shall we report them in accessibility if they're on both states, or just `active`? (`disabled` state is not visible)
- since they're two values, they don't fit in one `accessibilityLabel` and one `accessibilityValue`. Shall I compose the `accessibilityLabel` with the format you specified?
Thx :)

typfel:
@alexaubry are you happy with this PR now?

Wire-iOS/Resources/Base.lproj/Localizable.strings
alexaubry: 1. We should just report that it's active.
2. It's better to use the a11y value separately, because the label describes what it is (usually a noun or nominal group) and value gives more information about the state or the content. You mentioned it was not possible to do that?

Wire-iOS/Sources/UserInterface/Conversation Options/UINavigationController+LegalHoldButton.swift
```diff
@@ -0,0 +1,36 @@
...
+    @objc func legalHoldItem() -> UIBarButtonItem {
+        let item = UIBarButtonItem(icon: .legalholdactive, target: self, action: #selector(presentLegalholdDetails))
+        item.accessibilityIdentifier = "legalhold"
+        item.accessibilityLabel = "legal hold details" //TODO localize
+        item.tintColor = .vividRed
+        return item
+    }
+    
+    @objc func presentLegalholdDetails() {
+        //let viewController = LegalholdDetailsViewController(conversation: conversation)
```
typfel: This can be implemented after merging `feature/legalhold` into this PR.

PR #3405, opened 2019-05-14
ZIOS-11422: Make the conversation not secure when dismissing the alert

alexaubry:
### Issues

When a conversation becomes unverified after discovering a client when sending a message, we show an alert to the user asking them whether we want to send the message anyway.

If they cancel, the conversation remains degraded, and the next time the user sends a message, we show the alert again.

### Causes

We do not decrease the security of the conversation when the user cancels the alert.

### Solutions

Mark the conversation as not secure when cancelling the alert, so that we don't ask the user again. That way, we align the behavior with calls and the upcoming legal hold feature.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-data-model/pull/722
- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1023


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -762,9 +762,11 @@ - (void)presentConversationDegradedActionSheetControllerForUsers:(NSSet<ZMUser *
     UIAlertController *controller = [UIAlertController controllerForUnknownClientsForUsers:users completion:^(ConversationDegradedResult result) {
         switch (result) {
             case ConversationDegradedResultCancel:
+                [self.conversation makeNotSecure];
                 [self.conversation doNotResendMessagesThatCausedDegradation];
```
typfel: I'd prefer to mark the conversation as not secure inside `doNotResendMessagesThatCausedDegradation`, that way there no possibility for the UI to mess up.

Ideally `makeNotSecure` would not even be exposed to the UI but I think it's used by calling in the voice channel atm.

typfel:
Calling degradation has special handling which also needs to be updated https://github.com/wireapp/wire-ios-sync-engine/blob/d8a7328cc515782fe5fb8a147b0b883f8a6693ee/Source/Calling/VoiceChannelV3.swift#L148

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
alexaubry: I'll close this PR then.

PR #3408, opened 2019-05-14
feature: Inform user that they are no longer under legal hold

billypchan:
## What's new in this PR?

Create an alert to Inform user that they are no longer under legal hold.
Create the snapshot test for the alert.
Clean up the helper method for creating an alert with ok button.

### TODO

- [x] receive signal for "When the admin turns off legal hold for you" and display the alert


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -159,6 +161,15 @@ final class AppRootViewController: UIViewController {
                                                            application: UIApplication.shared)
                 
             sessionManager.urlHandler.delegate = self
+
+            /// setup UserChangeInfo observer after sessionManager is created since ZMUserSession.shared() depends on it.
+            self.setupUserChangeInfoObserver()
```
typfel: The `AppRootViewController` lives outside the user session life cycle, it's created once at launch. So don't think this will work in a multiple account scenario or if the user is logged out when the app launches.

I'd suggest moving this logic to the `ZClientViewController` which is the root "logged in" - view controller.

billypchan:
@typfel updated as your suggestion.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+ZMUserObserver.swift
```diff
@@ -0,0 +1,39 @@
+
```
typfel: Accidental extra space

Wire-iOS/Sources/UserInterface/Helpers/UIViewController+Alert.swift
```diff
@@ -18,24 +18,79 @@
...
+extension UIAlertController {
+        
+    /// Create an alert with a OK button
+    ///
+    /// - Parameters:
+    ///   - title: title of the alert
+    ///   - message: message of the alert
+    ///   - okActionHandler: a nullable closure for the OK button
+    /// - Returns: the alert presented
+    static func ok(title: String,
```
typfel: This function name is not very descriptive. How about something more verbose like:

`alertWithOKButton(title: String, message: String, actionHandler: ((UIAlertAction) -> Void)?)`

Wire-iOS/Sources/UserInterface/Helpers/UIViewController+Alert.swift
```diff
@@ -18,24 +18,79 @@
...
-        let okAction = UIAlertAction(title: "general.ok".localized, style: .cancel, handler: okActionHandler)
+        let okAction =  UIAlertAction.ok(style: .cancel, handler: okActionHandler)
         alert.addAction(okAction)
 
-        present(alert, animated: true, completion: nil)
+        return alert
     }
+
+    //MARK: - legal hold
+    static func legalHoldDeactivated(okActionHandler: ((UIAlertAction) -> Void)? = nil) -> UIAlertController {
```
typfel: Why expose a `okActionHandler` which is never used.

billypchan: It was added for any actions after the button is clicked and it will be removed.

PR #3411, opened 2019-05-15
Feature/swift snapshot

billypchan:
## What's new in this PR?

Introduce swift snapshot testing. 

Compare both snapshot frameworks:

|  | swift snapshot | ios snapshot test case |
| --- | ------------- | ------------- |
| subclassing | just subclass XCTestCase | need to subclass FBSnapshotTestCase |
| create/update reference image | auto created when first run/set `record = true`  | need to set `recordMode = true` |
| verifying `UIViewController` | can verify directly  | only verifies `UIView` |
| snapshot artifacts | stored only the failed images (discussion: https://github.com/pointfreeco/swift-snapshot-testing/pull/220)  | stored the failed images/ref. images/diff. images |
| development | actively developing  | no major update recently  |

### TODO
- [x] remove file extension in the snapshot path
- [x] fix snapshot reference image path on CI server.
- [x] test for fail case on CI server
- [x] document the benefit of swift snapshot testing
- [x] create an example for UIView snapshot
- [x] create an example for UIAlertController(and other UIView subclasses) snapshot
- [ ] Extract business logic form `CoreDataSnapshotTestCase` and create a new Class to handle it.
### Dependencies

- [x] fork swift-snapshot-testing


Wire-iOS Tests/VersionInfoViewControllerSnapshotTests.swift
```diff
@@ -36,6 +45,9 @@ final class VersionInfoViewControllerSnapshotTests: ZMSnapshotTestCase {
     }
 
     func testForInitState(){
-        verify(view: sut.view)
+
+        let failure = verifySnapshot(matching: sut, as: .image, snapshotDirectory: snapshotDirectory)
```
vytis: It will be pretty annoying to add the snapshot directory parameter to all cases where we have `verify()` now. Where does the library expect the snapshots to be located? Maybe it's better to move/rename them to have the same structure?

billypchan: @vytis the custom snapshot directory has to be provided in the method parameter. I will refactor this part as a `XCTestCase` extension.

Wire-iOS Tests/SwiftSnapshotTesting/XCTestCase+SnapshotTesting.swift
```diff
@@ -0,0 +1,44 @@
...
+        return path
+    }
+
+    func verify(matching value: UIViewController,
+                file: StaticString = #file,
+                testName: String = #function,
+                line: UInt = #line) {
+
+        let failure = verifySnapshot(matching: value,
+                                     as: .image,
```
vytis: Maybe add this as a parameter defaulting to `.image`, so that in future we can use other strategies if needed

vytis: Since it seems to be more complicated than expected - maybe it's better to leave it as `.image` in there and don't add the new parameter. If we want to adopt some other types of snapshots we would use the new API. It is actually a better outcome because we have added this method only to be backwards compatible with old snapshot library API.

billypchan: I added a work-a-round to stimulate "defaulting to `.image`", there would be a finite number of classes support image snapshoting so it would be too hard to maintain.

Cartfile.private
```diff
@@ -1,2 +1,3 @@
 github "wireapp/ios-snapshot-test-case" "4.0.0-xcode_10_2"
 github "wireapp/wire-ios-testing" ~> 19.0
+github "pointfreeco/swift-snapshot-testing" ~> 1.5
```
alexaubry: If we decide to go with this solution, we will need to fork it into the wireapp org before merging the PR.

Wire-iOS Tests/UIAlertViewController+CompanyLoginTests.swift
```diff
@@ -0,0 +1,42 @@
...
+        }, completion: {_ in })
+    }
+
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+
+    func testForAlert(){
+        verify(matching: sut)
```
alexaubry: Might be a personal preference, but I think we should indicate what we are matching it to. What are your thoughts about specifying that we are matching an image in the arguments, e.g. `verify(matchingImage: sut)`?

Cartfile.private
billypchan: It will be forked after this is approved. Thanks.

Wire-iOS Tests/UIAlertViewController+CompanyLoginTests.swift
billypchan: This is one of the interfaces of the new `verify` method which is equivalent to ` verify(matching: sut, as: .image)`

related discussions:
https://stackoverflow.com/questions/56256602/generic-function-parameter-default-value

vytis:
We have a bit of a problem with repository size increasing quite a bit because of snapshots. The problem is that git just treats PNGs as binary blobs and when they change it stores both copies of the file. It also applies to renames - for git it's basically delete one and add another. I think filenames for the new snapshot files will be slightly different than old ones? Maybe we should think what to do to avoid putting snapshots in current repository to avoid increasing the size indefinitely... Any ideas @alexaubry @typfel  @nicorsm ?

billypchan:
@vytis We can create another PR for the splitting repository task.

PR #3416, opened 2019-05-17
Snapslot of UserImageViewContainer

billypchan:
## What's new in this PR?

Create snapshots of `UserImageViewContainer` for PR https://github.com/wireapp/wire-ios/pull/3414.
A new protocol `ProfileImageFetchable` is created to solve the issue that `MockUser` cannot override `UserType` extension function `fetchProfileImage` since it is not declared in `UserType`. (I tried move it to the protocol definition but it is not possible due to Swift-only parameter types.)

Replace `ZMUserSession` with `ZMUserSessionInterface` in several places in order to allow injecting `MockZMUserSession`

### TODO
- [x] clean up `static func image(inTestBundleNamed name: String, bundle: Bundle) -> UIImage`


Wire-iOS Tests/UserImageViewContainerSnapshotTests.swift
```diff
@@ -0,0 +1,58 @@
...
+        mockUser.profileImage = image(inTestBundleNamed: "unsplash_matterhorn.jpg")
+    }
+    
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+
+    func testForNoUserImageWithoutSession(){
+        sut = UserImageViewContainer(size: .big, maxSize: maxSize, yOffset: -8, userSession:nil)
```
vytis: What's the reason for this offset? I also see the snapshot has some empty space around the avatar, do we need it?

billypchan: This snapshot is simulating the appearance inside CallAccessoryViewController so I applied same size settings:

https://github.com/wireapp/wire-ios/blob/c0987fce4ef3d49c054e8c5521f25890f33d4e52/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallAccessoryViewController.swift#L29

vytis: In this case we are testing the view itself, so I don't see a reason for this offset - it's just confusing to see some constants without any extra info.

billypchan:
@vytis the snapshots are updated for your suggestion.

PR #3422, opened 2019-05-21
Feature: indicate legal hold in conversation list

billypchan:
## What's new in this PR?

Add legal hold indication in `ConversationListTopBarViewController`.

### TODO
- [ ] present legal hold info screen when the indicator is tapped.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -118,7 +137,12 @@ class ConversationListTopBarViewController: UIViewController {
         guard let middleView = topBar?.middleView as? UILabel else { return }
         middleView.text = ZMUser.selfUser().name
     }
-    
+
+    @objc
+    func presentLegalHoldInfo() {
+        ///TODO: present legalhold screen
```
typfel: This can already be done by presenting the `LegalHoldDetailsViewController`

Wire-iOS Tests/ConversationList/ConversationListTopBarViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,60 @@
...
+final class ConversationListTopBarViewControllerSnapshotTests: ZMSnapshotTestCase {
+    
+    var sut: ConversationListTopBarViewController!
+    var mockAccount: Account!
+    var mockSelfUser: MockUser!
+
+    override func setUp() {
+        super.setUp()
+        mockAccount = Account.mockAccount(imageData: mockImageData)
+        mockSelfUser = MockUser.firstMockUser()
```
alexaubry: Could we add a black background here? It's hard to read the contents of the snapshot otherwise.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -20,19 +20,21 @@ import UIKit
...
     private var account: Account
+    private let selfUser: UserType
     
     var topBar: TopBar? {
         return view as? TopBar
     }
     
     
-    init(account: Account) {
+    init(account: Account, selfUser: UserType = ZMUser.selfUser()) {
```
alexaubry: Having the self user as a default parameter kind of defeats the purpose of dependency injection here. Maybe we could just pass it at call site?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -87,7 +94,19 @@ class ConversationListTopBarViewController: UIViewController {
             return titleLabel
         }
     }
-        
+
+    func createLegalHoldView() -> UIView {
+        let imageView = UIImageView(frame: .zero)
```
alexaubry: You probably don't need to pass in the frame here.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
typfel: Right it doesn't support the user case right now, let's integrate that separately.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -87,7 +94,19 @@ class ConversationListTopBarViewController: UIViewController {
...
+
+        imageView.setIcon(.legalholdactive, size: .tiny, color: .vividRed)
+        imageView.isUserInteractionEnabled = true
```
alexaubry: If the image view is user interaction enabled, we need to set it up for voiceover properly (label, traits), cf what we did for the indicator in the navigation bar of the conversation.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
typfel: I think that depends on how it's used. The default parameter makes it's easier to use in the code  but still allows it be replaced by a mock object in the tests.

alexaubry: Sure, I just mean it's easy to forget that we have this parameter and try to use the self user in the tests by accident. But both are fine I guess.

typfel: Right, it's a trade off. I suspect this intializer is called in only one place in the production code so the value of the  default parameter is limited.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
billypchan: Right, I will check some other similar tappable indicator for voiceover support.

PR #3430, opened 2019-05-22
ZIOS-11408: Ask users to confirm legal hold before sending message

alexaubry:
### Issues

We want to warn the user before they send a message if legal hold was discovered in the process.

### Solutions

We adopt the new API to detect privacy changes and unity the flow with security level degradation and legal hold status.

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/726
- [ ] https://github.com/wireapp/wire-ios-request-strategy/pull/144
- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/1026
- [ ] https://github.com/wireapp/wire-ios-share-engine/pull/110


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -717,10 +717,13 @@
...
 "meta.degraded.degradation_reason_message.plural" = "%@ started using new devices.";
+"meta.degraded.degradation_reason_message.someone" = "Someone started using a new device.";
 "meta.degraded.dialog_message" = "Do you still want to send your message?";
-"meta.degraded.show_device_button" = "Show device";
-"meta.degraded.send_anyway_button" = "Send anyway";
+"meta.degraded.verify_devices_button" = "Verify Devices…";
+"meta.degraded.send_anyway_button" = "Send Anyway";
 "meta.degraded.cancel_sending_button" = "Cancel";
+"meta.legalhold.info_button" = "What Is Legal Hold?";
+"meta.legalhold.send_alert_title" = "The conversation is now subject to legal hold.";
```
typfel: Shouldn't the copy be "The conversation is now subject to legal hold. Do you still want to send your message?"

alexaubry: "Do you still want to send your message?" is the message of the alert, like we do for the degradation (under `meta.degraded.dialog_message`)

alexaubry: I think in the spec it doesn’t specify that we need to change the title, but just add the verify devices option, but I can check again.

typfel: Ok, missed that there was a title _and_ a message

PR #3437, opened 2019-05-24
Fix: large size GIF may freeze the app

billypchan:
### Issues

When a user receives a large (>10MB) animated GIF image, the app may freeze due to high CPU usage.

### Causes

The line ` FLAnimatedImage(animatedGIFData: imageData)` consumes CPU usage.

### Solutions

Replace with `SwiftyGif` for better performance and catches exceptions when creating an image object from data.

### TODO
- [x] replace other places using `FLAnimatedImage`
- [ ] create tests for performance
- [x] retire `MediaAsset` & `MediaAssetView`


billypchan:
Fixed non-GIF image handling in my SwiftyGif fork and will restart the PR.

PR #3442, opened 2019-05-28
ZIOS-11427: Add test for json file content

nicorsm:
## What's new in this PR?

Added tests for static URL loaded from the configuration bundle.

### DIFF
--- a/Wire-iOS Tests/URL+WireTests.swift
+++ b/Wire-iOS Tests/URL+WireTests.swift
@@ -26,8 +26,9 @@ class URL_WireTests: XCTestCase {
     override func setUp() {
         super.setUp()
         let bundle = Bundle.backendBundle
-        EnvironmentType.production.save(in: .standard)
-        be = BackendEnvironment(userDefaults: .standard, configurationBundle: bundle)
+        let defaults = UserDefaults(suiteName: "URLWireTests")!
+        EnvironmentType.production.save(in: defaults)
+        be = BackendEnvironment(userDefaults: defaults, configurationBundle: bundle)
     }
     
     override func tearDown() {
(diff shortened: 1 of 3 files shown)


Wire-iOS Tests/URL+WireTests.swift
```diff
@@ -0,0 +1,60 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class URL_WireTests: XCTestCase {
+
+    let be = BackendEnvironment.shared
```
vytis: Maybe it would be safer to explicitly load production environment, since it's the only one we really care about.

Wire-iOS Tests/URL+WireTests.swift
```diff
@@ -0,0 +1,60 @@
...
+    
+    func testThatWebsiteURLsAreLoadedCorrectly() {
+        XCTAssertNotEqual(be.websiteURL.absoluteString, "")
+        XCTAssertEqual(URL.wr_website, be.websiteURL)
```
vytis: If website URL will be set to "random.org" this and other tests will pass. It's better to maybe check full URLs manually instead.

Wire-iOS Tests/URL+WireTests.swift
```diff
@@ -0,0 +1,74 @@
...
+@testable import Wire
+
+class URL_WireTests: XCTestCase {
+    
+    var be: BackendEnvironment!
+    
+    override func setUp() {
+        super.setUp()
+        let bundle = Bundle.backendBundle
+        EnvironmentType.production.save(in: .standard)
```
vytis: Just to be safe - better to create your own user defaults and pass it in instead of using `.standard`. This way we would avoid any issues with other tests being influenced somehow.

PR #3444, opened 2019-05-29
Fix: Double status indicators is shown on input bar

billypchan:
### Issues

When a team member is available, the placeholder on the input bar has two dot icons.

### Causes

~~Unknown, using a `NSTextAttachment` with no image seems to be the cause of the issue.~~
Side effect of calling `NSAttributedString.applying(transform: textTransform)` with padding `NSTextAttachment`.

### Solutions

Use a space instead of an empty attachment, which works beter for dynamic font size.


typfel:
Looks like this padding change is breaking other snapshot tests.

billypchan:
@typfel The snapshots are restored to the version before https://github.com/wireapp/wire-ios/pull/3435#pullrequestreview-241747133.

PR #3446, opened 2019-05-31
Fix: group options in cell are not responsive to tapping

billypchan:
### Issues

group options collection view cells are not responsive to tapping.

### Causes

Unknown, may be same as https://github.com/wireapp/wire-ios/pull/3403

### Solutions

Implement empty `didSelectItemAt` method in `GroupDetailsSectionController`, which is the parent class of several cell classes.
Change `GroupOptionsSectionController`'s `GroupOptionsSectionControllerDelegate` property to `unowned` to indicated that the delegate is not optional and it is assigned when `init`.

### discussion
Actually `GroupDetailsSectionController` is an abstract base class of several cell classes. It would be better to rewrite it as a protocol (need some time since it has some default implementations).


Wire-iOS/Sources/UserInterface/GroupDetails/Sections/GroupOptionsSectionController.swift
```diff
@@ -52,7 +52,7 @@ class GroupOptionsSectionController: GroupDetailsSectionController {
 
     // MARK: - Properties
 
-    private weak var delegate: GroupOptionsSectionControllerDelegate?
+    private unowned let delegate: GroupOptionsSectionControllerDelegate
```
typfel: Even though it's nice that gets rid of the optional and it's fairly safe since the delegate is a required parameter to the initialiser, it's still no guarantee that the delegate will always be around since it's can be deallocated at any time.

PR #3447, opened 2019-05-31
Fix: convert GroupDetailsSectionController to a protocol

billypchan:
## What's new in this PR?

Follow-up of https://github.com/wireapp/wire-ios/pull/3446. Refactor `GroupDetailsSectionController` into a protocol to prevent creating an abstract class with default implementations. When the class conforms new `GroupDetailsSectionController` type and want to use the original default method implementations, access the new helper methods in protocol's extension instead.


billypchan:
@typfel this PR is not a must-do to fix the issue but suggest an alternative for similar issues.

typfel:
@billypchan since this is not strictly necessary, let's close this?

PR #3451, opened 2019-06-04
Fix: "You started using *This device*" button doesn't work 

billypchan:
### Issues

Link text in the system message is not tappable in release build. (It is tappable in debug build)

### Causes

Same as https://github.com/wireapp/wire-ios/pull/3403

### Solutions

Add an empty implementation of `textView(_ textView: UITextView, shouldInteractWith url: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool` in `ConversationIconBasedCell` and let the children of `ConversationIconBasedCell` overrides it.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationIconBasedCell.swift
```diff
@@ -166,3 +166,10 @@ class ConversationIconBasedCell: UIView, UITextViewDelegate {
     }
 
 }
+
+extension ConversationIconBasedCell: UITextViewDelegate {
+    public func textView(_ textView: UITextView, shouldInteractWith url: URL, in characterRange: NSRange, interaction: UITextItemInteraction) -> Bool {
+        //Fixes Swift 5.0 release build child class overrode method not called bug
```
typfel: typo:

```suggestion
        // Fixes Swift 5.0 release build child class overridden method not called bug
```

PR #3463, opened 2019-06-12
Feature: legalhold activation dialog

billypchan:
## What's new in this PR?

Refactors `RequestPasswordViewController` to `RequestPasswordController` to  prevent subclassing `UIAlertController` as it should not be subclassed(https://developer.apple.com/documentation/uikit/uialertcontroller#//apple_ref/doc/uid/TP40014538-CH1-SW2).

Create `RequestPasswordContext` for the legal hold request context.

## TODO
- [x] receive signal for "When the admin turns on legal hold for you" and display the alert
- [x] process the user's choice
- [x] SSO case: disable password input
- [ ] reappear every time the app is put back into the foreground


Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -0,0 +1,113 @@
...
+        alertController.addAction(okAction)
+        alertController.addAction(cancelAction)
+    }
+    
+    @objc
+    func passwordTextFieldChanged(_ textField: UITextField) {
+        guard let passwordField = alertController.textFields?[0] else { return }
+
+        ///TODO: update with password requirement
+        okAction.isEnabled = (passwordField.text ?? "").count > 6
```
alexaubry: The only requirement for entering an existing password is that it's not empty.

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1655,3 +1655,7 @@
 "share_extension.voiceover.conversation_secure" = "All clients verified.";
 "share_extension.voiceover.conversation_secure_with_ignored" = "Not all clients verified.";
 
+// Legal hold - alert
+"legalhold_request.alert.title" = "Legal Hold Requested";
+"legalhold_request.alert.detail" = "All future messages will be recorded by the device with fingerprint:\n%@\nThis includes deleted, edited, and timed messages in all conversations.";
```
alexaubry: I think we need to have two newlines around the password.

```suggestion
"legalhold_request.alert.detail" = "All future messages will be recorded by the device with fingerprint:\n\n%@\n\nThis includes deleted, edited, and timed messages in all conversations.";
```

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
```diff
@@ -20,12 +20,13 @@
 import Foundation
 
 extension NSData {
-    public func attributedFingerprint(attributes: [NSAttributedString.Key : AnyObject], boldAttributes: [NSAttributedString.Key : AnyObject], uppercase: Bool = false) -> NSAttributedString? {
-        
+
+    /// return a lower case and space between every btye string of the given data
+    var fingerprintString: String {
         let strings: [String] = self.mapBytes { (char: UInt8) -> (String) in
```
alexaubry: Isn't there a simpler way to do this?

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -0,0 +1,113 @@
...
+            [weak self, unowned alertController] (action: UIAlertAction) -> Void in
+            if let passwordField = alertController.textFields?[0] {
+                let password = passwordField.text ?? ""
+                self?.callback(.success(password))
+            }
+        }
+        
+        let cancelAction = UIAlertAction(title: cancelTitle, style: .cancel) {
+            [weak self, unowned alertController] (action: UIAlertAction) -> Void in
+            self?.callback(.failure(NSError(domain: "\(type(of: alertController))", code: 0, userInfo: [NSLocalizedDescriptionKey: "User cancelled input"])))
```
alexaubry: Code could be `NSUserCancelledError`. I'm not sure if we need a user info either, since it's never gonna be shown.

```suggestion
            self?.callback(.failure(NSError(domain: "\(type(of: alertController))", code: NSUserCancelledError, userInfo: nil)))
```

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+ZMUserObserver.swift
```diff
@@ -19,9 +19,15 @@ import Foundation
...
-            !ZMUser.selfUser().isUnderLegalHold {
-            presentLegalHoldDeactivatedAlert()
+        if changeInfo.legalHoldStatusChanged {
+
+            if ZMUser.selfUser().hasLegalHoldRequest {
+                presentLegalHoldActivatedAlert(){ (password) in
+                    ///TODO: activation request
+                }
+            } else if !ZMUser.selfUser().isUnderLegalHold {
+                presentLegalHoldDeactivatedAlert()
```
alexaubry: Legal hold status might also change when user clients are added or removed. We need to make sure legal hold was enabled previously before showing this alert.

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
billypchan: @alexaubry it is copied from the original code and maybe it was a requirement.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+ZMUserObserver.swift
billypchan: I will update data model for this case. Thanks!

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
billypchan: We can merge this logic with `NSString.fingerprintStringWithSpaces`.

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
typfel: I can we can drop the length requirement. Generally it makes little sense do password validation when creating a new password since these rules change over time.

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
billypchan: fixed in https://github.com/wireapp/wire-ios/pull/3464

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -0,0 +1,102 @@
...
+            cancelTitle = "general.accept".localized
+        }
+
+        alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
+
+        switch context {
+        case .removeDevice,
+             .legalHold(_, true):
+            alertController.addTextField {(textField: UITextField) -> Void in
+                textField.placeholder = "self.settings.account_details.remove_device.password".localized
```
alexaubry: Here we need to add an observer to disable the button OK if the text is empty.

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -0,0 +1,102 @@
...
+                textField.isSecureTextEntry = true
```
alexaubry: Can we also set the `textContentType` to `.password`?

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -99,4 +106,11 @@ final class RequestPasswordController {
         alertController.addAction(okAction)
         alertController.addAction(cancelAction)
     }
+
+    @objc
+    func passwordTextFieldChanged(_ textField: UITextField) {
+        guard let passwordField = alertController.textFields?[0] else { return }
+
+        okAction.isEnabled = passwordField.text?.count > 0
```
alexaubry: okAction also needs to be disabled by default when you initialize it.

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
```diff
@@ -0,0 +1,87 @@
...
+        }
+        
+        let attributedRemoteIdentifier = fingerprintString.fingerprintString(attributes: attributes, boldAttributes: boldAttributes)
+        
+        let paragraphStyle = NSMutableParagraphStyle()
+        paragraphStyle.lineSpacing = 10
+        return attributedRemoteIdentifier! && [.paragraphStyle: paragraphStyle]
+    }
+    
+    public func mapBytes<T: Any, E: Any>(callback: (E) -> (T)) -> [T] {
```
typfel: Didn't we delete this? Did you have a merge conflict?

alexaubry:
@billypchan Please redo the snapshots for RequestPasswordControllerSnapshotTests, since we disabled the button.

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
billypchan: Git did not say it is conflicting for the renamed file. Thanks!

Wire-iOS/Sources/Helpers/syncengine/UserClient+RemoveTask.swift
```diff
@@ -62,6 +68,7 @@ private class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
     private var strongReference: ClientRemovalObserver? = nil
     let userClientToDelete: UserClient
     let controller: UIViewController
+    private var requestPasswordController: RequestPasswordController?
```
typfel: Is really necessary to keep a reference to `RequestPasswordController`?

billypchan: to make `textField` in `RequestPasswordController` reponse to events, a reference to `RequestPasswordController` is necessary.

typfel: I don't understand why it's necessary `requestPasswordController` is not used anywhere as far as I can see.

alexaubry: Following up on what Jacob says, I agree it doesn't make sense to init and retain a request password controller. Ideally, `RequestPasswordController` should just have a static method that takes all the parameters from what is now the initializer, and returns the UIAlertController to present. 

We can use an NSNotification to observe the text changes, which means that we actually don't need to use the targets. You can see what we did on the CompanyLoginController to set up the text field observing with the token and notification center.

It could look something like:

```swift
class RequestPasswordController {
    enum Context { ... }

    static func makePasswordRequestAlert(for context: Context, completionHandler: @escaping (String?) -> Void) -> UIAlertController {
        // set up the alert in a self-contained way with NotificationCenter
    }

}
```

billypchan: I refactored reference keywords and it works correctly now.

PR #3464, opened 2019-06-12
Clean up Fingerprint Data to spaced string methods

billypchan:
## What's new in this PR?

Commented by @alexaubry in https://github.com/wireapp/wire-ios/pull/3463#discussion_r292808013, unify the Fingerprint Data to spaced string methods with existing `NSString.fingerprintStringWithSpaces()` method.


Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
```diff
@@ -1,63 +1,60 @@
...
+//
 
 
 import Foundation
 
-extension NSData {
+extension Data {
+    /// return a lower case and space between every btye string of the given data
+    var fingerprintString: String? {
+        let string = String(data: self, encoding: .utf8)
```
alexaubry: You can use the non nullable decoder here:

```suggestion
        let string = String(decoding: self, as: UTF8.self)
```

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
```diff
@@ -1,63 +1,60 @@
...
+
         let paragraphStyle = NSMutableParagraphStyle()
         paragraphStyle.lineSpacing = 10
         return attributedRemoteIdentifier! && [.paragraphStyle: paragraphStyle]
     }
-    
+}
+
+extension NSData {
     public func mapBytes<T: Any, E: Any>(callback: (E) -> (T)) -> [T] {
```
typfel: Is `mapBytes` still in used anywhere else?

billypchan: removed!

Wire-iOS/Sources/Helpers/NSData+Fingerprint.swift
```diff
@@ -1,79 +1,44 @@
...
 // along with this program. If not, see http://www.gnu.org/licenses/.
-// 
+//
 
 
 import Foundation
 
-extension NSData {
+extension Data {
+    /// return a lower case and space between every btye string of the given data
```
typfel: Typo

```suggestion
    /// return a lower case and space between every byte string of the given data
```

PR #3468, opened 2019-06-14
Use mock core data with swift snapshot testing

billypchan:
## What's new in this PR?

This PR takes `CallInfoRootViewControllerSnapshotTests`'s `testOneToOneIncomingAudioRinging` as an example for using mock core data in new swift snapshot testing framework.

A class `MockCoreData` is created, which is extracted from `CoreDataSnapshotTestCase` to allow doing snapshot testing without subclassing `XCTestCase`


Wire-iOS Tests/Utils/MockCoreData.swift
```diff
@@ -0,0 +1,250 @@
...
+//
+
+import WireTesting
+import XCTest
+@testable import Wire
+
+
+/// This class provides a `NSManagedObjectContext` in order to test views with real data instead
+/// of mock objects.
+final class MockCoreData {
```
typfel: Since this class provides a way to to use real core data objects instead of mock objects I think the name `MockCoreData` is a bit misleading. How about `CoreDataFixture`

Wire-iOS Tests/Utils/MockCoreData.swift
```diff
@@ -0,0 +1,250 @@
...
+//
+
+import WireTesting
+import XCTest
+@testable import Wire
+
+
+/// This class provides a `NSManagedObjectContext` in order to test views with real data instead
+/// of mock objects.
+final class CoreDataFixture {
```
typfel: We should also rename the file to `CoreDataFixture.swift`

PR #3471, opened 2019-06-17
ZIOS-11478: Implement second device alert

alexaubry:
### Issues

- We want to notify users on all devices when legal hold is disabled
- We want to notify users on their other devices that legal hold was accepted

### Solutions

- Implement the alert for legal hold on second device

### Bugs Fixed

We also fix a bug where the conversation list header wouldn't observe changes, and where it would show the wrong shape for team users after logging in.

### To-Do

- [x] Show the pending request alert when coming to the foreground
- [x] ACK legal hold status when the user taps the disabled alert
- [x] Fix bug where the "legal hold is disabled" alert is shown after the user accepts legal hold


Wire-iOS/Sources/Managers/LegalHoldDisclosureController.swift
```diff
@@ -0,0 +1,224 @@
...
+class LegalHoldDisclosureController: NSObject, ZMUserObserver {
+
+    enum DisclosureState {
+        /// No legal hold status is being disclosed.
+        case none
+
+        /// The user is being warned about a pending legal hold alert.
+        case warningAboutPendingRequest(UIAlertController)
+
+        ///
```
typfel: Incomplete comment?

alexaubry: Added the missing comments, thanks

typfel:
Since you deleted `UIViewController+LegalHoldAlertSnapshotTests.swift` shouldn't there also be some associated snapshots to delete?

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -45,6 +45,7 @@ final class AppRootViewController: UIViewController {
     fileprivate let mediaManagerLoader = MediaManagerLoader()
 
     var authenticationCoordinator: AuthenticationCoordinator?
+    var legalHoldDisclosureController: LegalHoldDisclosureController?
```
typfel: Why does the `LegalHoldDisclosureController` live in the `AppRootViewController` and not in the `ZClientViewController`? Wouldn't the need for `resetLegalHoldDisclosureControllerIfNeeded()` go away if was created & destroyed together with the `ZClientViewController`?

alexaubry: That's a good point, thanks.

alexaubry: Moved to the client view controller. Does it look good to you now?

billypchan:
snapshot test fails due to title bar is overflown. Maybe caused by the addition of top right indicator?

Screenshot 2019-06-20 at 10 38 21

billypchan:
Another snapshot failure is related to mockusers name, Tarja is the last user in `people-01.json` and James is the first one. 

Screenshot 2019-06-20 at 10 49 57

alexaubry:
@billypchan I couldn't check the snapshots before now because I was away, but thanks for the input!

typfel:
@billypchan are you ok with this now?

PR #3477, opened 2019-06-19
Fix: enter password alert not showing after entered a wrong password

billypchan:
### Issues

In setting screen, enter password alert not showing after entered a wrong password. (Wrong password alert appears after `remove device` cell is tapped)

### Causes

Multiple instance of `ClientRemovalObserver` exists due to they are not released after created when `remove device` is button is tapped everytime. And they are not release since self-retained and captured in closures.

### Solutions

Refactor and make ViewControllers retain `ClientRemovalObserver`. Refactor closures not to retain `ClientRemovalObserver`.

### DIFF
--- a/Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
+++ b/Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
@@ -100,15 +100,14 @@ final class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
         if !passwordIsNecessaryForDelete {
             controller.requestPassword { [weak self] newCredentials in
                 guard let emailCredentials = newCredentials,
-                    emailCredentials.password?.isEmpty == false,
-                    let weakSelf = self else {
+                    emailCredentials.password?.isEmpty == false else {
                     self?.endRemoval(result: ClientRemovalUIError.noPasswordProvided)
                     return
                 }
-                weakSelf.credentials = emailCredentials
-                weakSelf.startRemoval()
+                self?.credentials = emailCredentials
+                self?.startRemoval()
 
-                weakSelf.passwordIsNecessaryForDelete = true
+                self?.passwordIsNecessaryForDelete = true
             }
         } else {
             controller.presentAlertWithOKButton(message: "self.settings.account_details.remove_device.password.error".localized)
--- a/Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
+++ b/Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
@@ -72,10 +72,6 @@ final class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
         observerToken = ZMUserSession.shared()?.add(self)
     }
 
-    deinit {
-        observerToken = nil
-    }
-    
     func startRemoval() {
         controller.showLoadingView = true
         ZMUserSession.shared()?.delete(userClientToDelete, with: credentials)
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
```diff
@@ -64,11 +55,9 @@ enum ClientRemovalUIError: Error {
...
-
-    private var strongReference: ClientRemovalObserver? = nil
-    let userClientToDelete: UserClient
-    let controller: UIViewController
+final class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
+    var userClientToDelete: UserClient
+    unowned let controller: UIViewController
     let completion: ((Error?)->())?
     var credentials: ZMEmailCredentials?
     private var passwordIsNecessaryForDelete: Bool = false
```
typfel: What purpose does this property have? I don't see it being set to anything else than `true`

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
```diff
@@ -111,31 +98,26 @@ private class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
         controller.showLoadingView = false
 
         if !passwordIsNecessaryForDelete {
-            controller.requestPassword { newCredentials in
+            controller.requestPassword { [weak self] newCredentials in
                 guard let emailCredentials = newCredentials,
-                    emailCredentials.password?.isEmpty == false else {
-                    self.endRemoval(result: ClientRemovalUIError.noPasswordProvided)
+                    emailCredentials.password?.isEmpty == false,
+                    let weakSelf = self else {
```
typfel: The name `weakSelf` is misleading since we are actually capturing a strong reference to self here.

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
billypchan: It is used as a flag to indicate the password should show for the first time, seems not related to SSO.

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
```diff
@@ -111,31 +98,26 @@ private class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
...
+                                               with: weakSelf.credentials)
+                weakSelf.controller.showLoadingView = true
+
+                weakSelf.passwordIsNecessaryForDelete = true
             }
-            passwordIsNecessaryForDelete = true
-        }
-        else {
-            controller.displayError(NSLocalizedString("self.settings.account_details.remove_device.password.error", comment: ""))
+        } else {///TODO: not showing if self is not top VC?
```
typfel: This would never happens since we kill the observer if the user navigates away?

billypchan: yes, but it should not occurs since the UI is blocked by the loading spinner.

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
typfel: Ah ok

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
```diff
@@ -111,31 +98,26 @@ private class ClientRemovalObserver: NSObject, ZMClientUpdateObserver {
...
+                    emailCredentials.password?.isEmpty == false,
+                    let weakSelf = self else {
+                    self?.endRemoval(result: ClientRemovalUIError.noPasswordProvided)
                     return
                 }
-                self.credentials = emailCredentials
-                ZMUserSession.shared()?.delete(self.userClientToDelete, with: self.credentials)
-                self.controller.showLoadingView = true
+                weakSelf.credentials = emailCredentials
+                ZMUserSession.shared()?.delete(weakSelf.userClientToDelete,
```
typfel: The next two lines could be replaced with `weakSelf.startRemoval()`?

Wire-iOS/Sources/Helpers/syncengine/ClientRemovalObserver.swift
billypchan: `self` is weakened in the capture list above.

PR #3481, opened 2019-06-20
Fix: Chinese text paragraphs line spacings are not consistent

billypchan:
### Issues

Wire-iOS Tests_ReferenceImages_64_Wire_iOS_Tests MultiParagraphSnapshotTests_testThatLineHeightOfListIsConsistent_Chinese_414@2x_Before_73de87976a61a706eff61102fb103643b593411f

In the above snapshot, the line spacing of Chinese paragraph is not consistent, the first paragraph's line spacings are too close.

### Causes

Seems like related to the text rendering of textView has an issue when some components missing font attribute. ~~(the space between list number and text has no font attributes)~~ When the line break of each paragraph has no font attribute, the line spacing is inconsistent.

### Solutions

After trial and error, I found that adding font attribute to ~~all components~~ line break component of an attributed string fixes the issue.

I will raise an issue in the Down's page instead of patching it in this project.

testThatLineHeightOfListIsConsistent_Chinese_414@2x

Above is the snapshot after fixing.

### Dependency
- [x] https://github.com/wireapp/Down/pull/15


Cartfile
```diff
@@ -6,5 +6,5 @@ github "wireapp/FormatterKit" "1.8.1-swift3.0.2"
 github "wireapp/wire-ios-canvas" ~> 10.0
 github "wireapp/HockeySDK-iOS" ~> 5.1.4
 github "wireapp/FLAnimatedImage" "1.0.12-wire"
-github "wireapp/Down" "v2.1.2"
+github "wireapp/Down" "fix/ZIOS-11365-Chinese-paragraph"
```
alexaubry: Don't forget releasing Down

PR #3491, opened 2019-06-25
Fix: app crashes when the quoted image has zero width

billypchan:
### Issues

App crashes when the quoted image has zero width.

### Causes

The `multiplier` of the ratio constraint is `nan` in this case.

### Solutions

set `multiplier` to 1 when the image's width is zero
remove unnecessary height constraint.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/MessageContent/ImageMessageContentView.swift
```diff
@@ -63,10 +59,10 @@ class ImageContentView: UIView {
     private func updateAspectRatio(for resource: PreviewableImageResource) {
         let contentSize = resource.contentSize
         imageAspectConstraint.apply(imageView.removeConstraint)
-        imageAspectConstraint = imageView.heightAnchor.constraint(equalTo: imageView.widthAnchor, multiplier: contentSize.height / contentSize.width)
+        imageAspectConstraint = imageView.heightAnchor.constraint(equalTo: imageView.widthAnchor, multiplier: contentSize.width == 0 ? 1 : (contentSize.height / contentSize.width))
```
nicorsm: I would place the calculation of the multiplier on a constant, just to shorten a little bit this line 🙂

billypchan: I checked the doc (https://developer.apple.com/documentation/uikit/nslayoutdimension/1500934-constraint) and this question(https://stackoverflow.com/questions/19593641/can-i-change-multiplier-property-for-nslayoutconstraint) and found `constant` here is `The offset constant for this relationship.`.
The only way to update the `multiplier` of a constraint is recreation.

nicorsm: Sorry, what I wrote was misleading 😅I mean moving calculations to a `let` constant

billypchan: `contentSize` varies in run time so the `multiplier` is not a constant.

alexaubry: I think Nicola meant that it would be better to move the calculation of the multiplier to a constant right before the call, so that the line is shorter, which I agree would be more legible.

alexaubry:
```suggestion
        let imageAspectMultiplier = contentSize.width == 0 ? 1 : (contentSize.height / contentSize.width)
        imageAspectConstraint = imageView.heightAnchor.constraint(equalTo: imageView.widthAnchor, multiplier: imageAspectMultiplier)
```

PR #3493, opened 2019-06-25
Fix: group conversation with only 1 other person appears as 1 to 1

billypchan:
### Issues

The group conversation on the conversation list has a round shape, instead of square shape for groups.

### Causes

`Mode` of `ConversationAvatarView` does not consider the conversation type

### Solutions

Init `Mode` with conversation type and return `mode = .four` when conversation type is `.group` and number of users >=1.

Refactor `ConversationAvatarView`'s interface. Hide `users` and `conversation` properties to prevent confusion and create a new `func configure(context: Context)` method for both connection request and started conversation cases.


Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -134,41 +135,25 @@ final public class ConversationAvatarView: UIView {
...
-                    $0.container.isOpaque = false
-                    $0.container.backgroundColor = UIColor(white: 0, alpha: 0.24)
-                }
-
-                $0.allowsInitials = mode.showInitials
-                $0.shape = mode.shape
-                index = index + 1
-            }
-            self.setNeedsLayout()
+            mode = Mode(users: users)
```
typfel: Don't we lose the `converationType` information here?

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -134,41 +135,25 @@ final public class ConversationAvatarView: UIView {
...
                 return
             }
-            
+
             let stableRandomParticipants = conversation.stableRandomParticipants.filter { !$0.isSelfUser }
 
             self.accessibilityLabel = "Avatar for \(self.conversation?.displayName ?? "")"
             self.users = stableRandomParticipants
+
+            ///override the mode
```
typfel:
```suggestion
            // override the mode
```

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
billypchan: `converationType` is optional, it can not nil for the pending connection case.

typfel: But don't we call this setter when the participants of a group changes?

billypchan: It would be clearer that do not expose both `users` and `conversation` of `ConversationAvatarView` and create a `context` to handle pending cases.

typfel: Yep, exposing both is problematic. What do you mean by a `context`?

billypchan: I reduced the public properties and no need to create a `Context` now.

typfel: Ok, but that's no merged yet? Because `converation` and `users` are still public.

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -131,44 +130,36 @@ extension Mode {
...
-                    $0.container.isOpaque = false
-                    $0.container.backgroundColor = UIColor(white: 0, alpha: 0.24)
-                }
-
-                $0.allowsInitials = mode.showInitials
-                $0.shape = mode.shape
-                index = index + 1
-            }
-            self.setNeedsLayout()
+    func configure(context: Context) {
```
billypchan: @typfel all public properties are hidden and only expose this method.

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -131,44 +130,36 @@ extension Mode {
 }
 
 final public class ConversationAvatarView: UIView {
+    enum Context {
+        case connect(users: [ZMUser])
```
typfel: I don't think you can connect to multiple people at once, maybe this can be an a single `ZMUser`?

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -131,44 +130,36 @@ extension Mode {
...
+        case converse(conversation: ZMConversation)
```
typfel:
```suggestion
        case conversation(conversation: ZMConversation)
```

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListItemView+Update.swift
```diff
@@ -28,13 +28,19 @@ extension ConversationListItemView {
...
-    
+
+
+    /// configure without a conversation, i.e. when displaying a pending user
+    ///
+    /// - Parameters:
+    ///   - title: title of the cell
+    ///   - subtitle: subtitle of the cell
+    ///   - users: the pending user(s)
     @objc func configure(with title: NSAttributedString?, subtitle: NSAttributedString?, users: [ZMUser]) {
```
typfel: I'm thinking this configure might be redundant. We actually know that a conversation is for a pending user by looking at the `conversationType == .connection`. Seems like it should be possible to have only one configure method taking a conversation.

That would make the `Context` enum unnecessary since we can compute everything based on the conversation.

billypchan: This method only called by `ConnectRequestsCell` and it is no longer used for displaying a connection request.

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
billypchan: When displaying multiple incoming connecting requests in one cell, it is possible have multiple users. It looks more make sense to drop the `Context` and accepts `[ZMConversation]` for configure method.

typfel: Ah ok I see

PR #3495, opened 2019-06-26
ZIOS-9914: Conversation List Accessibility Improvements

alexaubry:
### Issues

We didn't fully support accessibility features on the conversation list. 

### Solutions

- Fix the grouping of elements on the conversation cell
- Improve data read by voiceover to include the status subtitle and availability
- Improve gesture support and labels in archived screen
- Simplify layout (not directly related to a11y but helped with switching the containers)
- Improve accessibility in start UI

### To-Do 

- [ ] Add custom actions to the cell to avoid swiping 
- [ ] What about the join call / play|pause button?


vytis:
It is expected that snapshots changed slightly?

alexaubry:
@vytis Yes, because of the stack view somehow the vertical arrangement seems to be a bit different.

PR #3500, opened 2019-06-27
Fix: Add a way to see the verification status of a device in the device list

billypchan:
## What's new in this PR?

In `ClientTableViewCell`, Set 'isAccessibilityElement' of 'verifiedLabel' to true to make it accessible. 
In `UserClientCell`, Set 'isAccessibilityElement' of 'verifiedIconView' to true to make it accessible and set accessibility label for both verified/unverified state.


typfel:
The `ClientTableViewCell`only used when displaying your own clients in settings. Victor was asking for an identifier on the `UserClientCell`.

I'd suggest updating the identifier for the shield icon depending on the verification status.

billypchan:
@typfel applied the fix to `UserClientCell`

Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
```diff
@@ -25,9 +25,18 @@ class UserClientCell: SeparatorCollectionViewCell {
...
-    private let verifiedIconView = UIImageView()
-    private let unverifiedIconView = UIImageView()
-    
+    private let verifiedIconView: UIImageView = {
+        let imageView = UIImageView()
+
+        imageView.image = WireStyleKit.imageOfShieldverified
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        imageView.contentMode = .center
+        imageView.accessibilityIdentifier = "img.shield"
```
alexaubry: If you're gonna set the ID later depending on the status, you can remove it here.

```suggestion
```

Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
```diff
@@ -117,6 +121,8 @@ class UserClientCell: SeparatorCollectionViewCell {
         let boldAttributes: [NSAttributedString.Key: AnyObject] = [NSAttributedString.Key.font: boldFingerprintFont.monospaced()]
         
         verifiedIconView.image = client.verified ? WireStyleKit.imageOfShieldverified : WireStyleKit.imageOfShieldnotverified
+        verifiedIconView.accessibilityLabel = client.verified ? "verified": "unverified"
```
alexaubry: The accessibility label is what is gonna be spoken by VoiceOver, you need to set the identifier with the automation locators, and the label with a localized string.

```suggestion
        verifiedIconView.accessibilityIdentifier = client.verified ? "verified" : "unverified"
        verifiedIconView.accessibilityLabel = client.verified ? "device.verified".localized : "device.not_verified".localized
```

Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
```diff
@@ -25,9 +25,18 @@ class UserClientCell: SeparatorCollectionViewCell {
     
     private let deviceTypeIconView = UIImageView()
     private let accessoryIconView = UIImageView()
-    private let verifiedIconView = UIImageView()
-    private let unverifiedIconView = UIImageView()
-    
+    private let verifiedIconView: UIImageView = {
```
typfel: Let's not do unrelated re-factoring. IMO it makes the code hard to read if we use different styles within the same class.

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientTableViewCell.swift
```diff
@@ -28,7 +28,11 @@ class ClientTableViewCell: UITableViewCell {
     let labelLabel = UILabel(frame: CGRect.zero)
     let activationLabel = UILabel(frame: CGRect.zero)
     let fingerprintLabel = UILabel(frame: CGRect.zero)
-    let verifiedLabel = UILabel(frame: CGRect.zero)
+    let verifiedLabel: UILabel = {
```
typfel: Also here let's not change the style.

Wire-iOS/Sources/UserInterface/UserProfile/Devices/UserClientCell.swift
```diff
@@ -43,15 +42,16 @@ class UserClientCell: SeparatorCollectionViewCell {
...
         deviceTypeIconView.contentMode = .center
-        
+
         verifiedIconView.image = WireStyleKit.imageOfShieldverified
         verifiedIconView.translatesAutoresizingMaskIntoConstraints = false
         verifiedIconView.contentMode = .center
-        verifiedIconView.accessibilityIdentifier = "img.shield"
-        
+        verifiedIconView.isAccessibilityElement = true
+        verifiedIconView.accessibilityIdentifier = "device_cell.verifiedShield"
```
typfel: This could also be removed since it's guaranteed to be overwritten in the configure method.

PR #3502, opened 2019-07-01
Fix: Missing send key on iPad when option is toggled

billypchan:
### Issues

When toggling "send button" option in the Setting screen, the keyboard is not updated immediately on iPad.

### Causes

On iPhone devices, since the setting screen covers the whole screen and the conversation screen calls `viewWillAppear` after setting screen is dismissed. On iPad devices, the setting screen is a popover and `viewWillAppear` would not be called and the keyboard would not refresh.

### Solutions

Post a notification when `send button` option is toggled and `InputBar` observes this notification and updates the keyboard return key type.


Wire-iOS/Sources/Components/Settings+Account.swift
```diff
@@ -69,4 +69,13 @@ extension Settings {
...
     }
+
+    @objc
+    func notifySendButtonEnabling() {
+        NotificationCenter.default.post(name: .sendButtonEnabling, object: self, userInfo: nil)
+    }
+}
+
+extension Notification.Name {
+    static let sendButtonEnabling = Notification.Name("SendButtonEnabling")
```
typfel: Let make the connection to the setting more clear by giving it the same name + `Changed`:

`disableSendButtonChanged`

Let's also move it to the `Settings` class since that's where it's used.

typfel: Ah I see now that it's already in the Accounts class :) So just the renaming then.

billypchan: We should clean up the notification names in `Setting.m` in the future.

Wire-iOS/Sources/Components/Settings+Account.swift
```diff
@@ -69,4 +69,13 @@ extension Settings {
         self.setValue(conversationURI.absoluteString, for: UserDefaultLastViewedConversation, in: account)
         self.defaults().synchronize()
     }
+
+    @objc
+    func notifySendButtonEnabling() {
```
typfel: Let's also rename this to `notifyDisableSendButtonChanged`

PR #3506, opened 2019-07-02
Fix: legal hold activatation alert not presented if there is alreay an presented view controller

billypchan:
### Issues

When the app's topmost view controller is presented modally, the legal hold alert does not show.

### Causes

The alert is presented from a covered view controller.

### Solutions

Present the view controller from the topmost view controller.
Improve the `UIViewControler.presentTopmost` method so that it can present the alert correctly.


Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
```diff
@@ -86,7 +81,12 @@ public extension UIApplication {
...
         return visibleWindow.last
     }
-    
+
+
+    /// Get the top most view controller
+    ///
+    /// - Parameter onlyFullScreen: if false, also search for all kinds of presented view controller
+    /// - Returns: the top most view controller 
     func wr_topmostController(onlyFullScreen: Bool = true) -> UIViewController? {
```
typfel: Since we removed `wr_topmostViewController()` can rename this one to `wr_topmostViewController(onlyFullScreen:)`? I think make sense to be more specific here.

PR #3508, opened 2019-07-02
Fix: Add accessibilityIdentifier to UserClientCell

billypchan:
## What's new in this PR?

~~Group the accessibility info of `UserClientCell`'s subviews to the cell.~~
Add `accessibilityIdentifier` to `UserClientCell`.


typfel:
I think now we are abusing the accessibility that's displayed to the user for automation purposes. I'd say we don't need display everything at the top level here but let's check with @VictorWissink when he's back.

billypchan:
I had some discussion with the QA team after we redesign the accessibility of the conversation list cell(https://github.com/wireapp/wire-ios/pull/3495) and still improving that. QA team encountered some issues related to missing info after putting all info in the cell level and thinking about make the subviews accessible.

PR #3509, opened 2019-07-02
ZIOS-11526: Legal hold device should be the first device in the device list

nicorsm:
### Issues

Legal hold device was not on the top of the list, because we were checking for the `type` instead of the `deviceClass`.

### Solutions

I've changed the conditions in the `sortedByRelevance()` extension method, plus adjusting the logic there in order to get the desired behaviour.


Wire-iOS/Sources/Helpers/UserClient+Fingerprint.swift
```diff
@@ -30,8 +30,10 @@ extension Sequence where Element: UserClientType {
     
     func sortedByRelevance() -> [UserClientType] {
         return sorted { (lhs, rhs) -> Bool in
-            if lhs.type == .legalHold {
+            if lhs.deviceClass == .legalHold && rhs.deviceClass != .legalHold {
```
typfel: What's the additional logic here solving? Was it not enough to change from `type` to `deviceClass`?

nicorsm: Yes.. yesterday evening seemed not to work but tried right now and it works. Late work's fault 😁

PR #3511, opened 2019-07-03
ZIOS-11523: App crashes after "log out" is tapped on "Too Many Devices" screen

nicorsm:
### Issues

App was crashing when tapping logout button in the "Too Many Devices" screen during login flow because of a fatal error.

### Solutions

Since the user is authenticated successfully in this phase, I'm checking if there is an active account to log out before throwing the fatal error.

### DIFF
--- a/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
+++ b/Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
@@ -391,7 +391,8 @@ extension AuthenticationCoordinator {
     /// Signs the current user out with a warning.
     private func signOut(warn: Bool) {
         if warn {
-            let signOutAction = AuthenticationCoordinatorAlertAction(title: "general.ok".localized, coordinatorActions: [.showLoadingView, .signOut(warn: false)], style: .destructive)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -391,16 +391,24 @@ extension AuthenticationCoordinator {
...
                                                             message: "self.settings.account_details.log_out.alert.message".localized,
                                                             actions: [.cancel, signOutAction])
 
             presentAlert(for: alertModel)
         } else {
+            
             guard let accountId = unauthenticatedSession.accountId,
                   let unauthenticatedAccount = sessionManager.accountManager.account(with: accountId) else {
+                    
+                if sessionManager.accountManager.accounts.count == 1 {
```
typfel: Why don't have an `unauthenticatedAccount` at this point?  

When get the error about `too-many-clients` I think it should end up calling `SessionManager.clientRegistrationDidFail`, which should in theory create an unauthenticated session?

typfel: I think it would be worth looking into this because the current solution doesn't work with multiple accounts.

PR #3513, opened 2019-07-03
Fix: reduce build time warnings

billypchan:
## What's new in this PR?

Reduce build time warnings caused by "Cannot find protocol definition" and duplicated implementation of methods.

The number of warnings is reduced from ~220 to 154.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+StartUI.m
```diff
@@ -27,9 +27,6 @@
 
 @import MobileCoreServices;
 
-@interface ConversationListViewController (StartUI)
```
typfel: This is the extension that's implemented in this file, we should not remove it. We could move to to `ConversationListViewController+StartUI.h`though and delete the unnamed extension that's currently there.

billypchan: There is also a definition in `ConversationListViewController+Internal.h` for `@interface ConversationListViewController (StartUI) <StartUIDelegate>` and I will move it to `ConversationListViewController+StartUI.h`

We have 4 headers, 11 swift extensions and 1 objc category for one view controller class which is too much and should do some cleanup job.

typfel: Yep, definitely but let's sit down together and plan how we should clean this up before we start doing the bigger clean up.

PR #3522, opened 2019-07-05
Fix: Unable to login when logged out from phone only account

billypchan:
## What's new in this PR?

Fix the issue that when the session is expired and login with the phone number not working issue.

### DIFF
--- a/Wire-iOS/Sources/Authentication/Event Handlers/Input/AuthenticationReauthenticateInputHandler.swift
+++ b/Wire-iOS/Sources/Authentication/Event Handlers/Input/AuthenticationReauthenticateInputHandler.swift
@@ -18,11 +18,13 @@
 
 import Foundation
 
+private let zmLog = ZMSLog(tag: "AuthenticationReauthenticateInputHandler")
+
 /**
  * Handles input in the reauthentication phase.
  */
 
-class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
+final class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
 
     weak var statusProvider: AuthenticationStatusProvider?
 
@@ -44,6 +46,8 @@ class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
             // If we get `String`, start the phone login flow
             let request = AuthenticationLoginRequest.phoneNumber(phoneInput)
             return [.startLoginFlow(request)]
+        } else {
+            zmLog.safePublic("Unable to handle context: \(context)")
         }
 
         // Do not handle other cases.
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Authentication/Event Handlers/Input/AuthenticationReauthenticateInputHandler.swift
```diff
@@ -39,10 +41,13 @@ class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
             // If we get `(String, String)`, start the email flow
             let request = AuthenticationLoginRequest.email(address: email, password: password)
             return [.startLoginFlow(request)]
-        } else if let phoneInput = context as? String {
+        } else if let phoneNumber = context as? PhoneNumber {
+            let phoneInput = phoneNumber.fullNumber
             // If we get `String`, start the phone login flow
```
typfel: This comment could be removed now since we expect the type `PhoneNumber` now.

Wire-iOS/Sources/Authentication/Event Handlers/Input/AuthenticationReauthenticateInputHandler.swift
```diff
@@ -39,10 +41,13 @@ class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
             // If we get `(String, String)`, start the email flow
             let request = AuthenticationLoginRequest.email(address: email, password: password)
             return [.startLoginFlow(request)]
-        } else if let phoneInput = context as? String {
+        } else if let phoneNumber = context as? PhoneNumber {
+            let phoneInput = phoneNumber.fullNumber
```
typfel: Why not just forward `phoneNumber.fullNumber ` directly, no need for a separate variable.

Wire-iOS/Sources/Authentication/Event Handlers/Input/AuthenticationReauthenticateInputHandler.swift
```diff
@@ -39,10 +41,13 @@ class AuthenticationReauthenticateInputHandler: AuthenticationEventHandler {
...
             // If we get `String`, start the phone login flow
             let request = AuthenticationLoginRequest.phoneNumber(phoneInput)
             return [.startLoginFlow(request)]
+        } else {
+            zmLog.safePublic("Unable to handle context: \(context)")
```
typfel: 👍

typfel: But we should not log the context to the public

billypchan: However safePublic logging does not support `Any`, I will change it to logging context's type.

PR #3524, opened 2019-07-08
Fix: service user avatar appeared as group conversation grid

billypchan:
### Issues

After https://github.com/wireapp/wire-ios/pull/3493, a conversation with a service user only appears as a group conversation.

### Causes

All group conversations are displayed in `.four` Mode no matter the type of `users`

### Solutions

Refactor the logic, when there is only one user in a group conversation, display it in `.one` mode.
Add tests for the `Mode` enum.


Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -107,13 +107,24 @@ extension Mode {
     ///   - conversationType: when conversationType is nil, it is a incoming connection request
     ///   - users: number of users involved in the conversation
     fileprivate init(conversationType: ZMConversationType? = nil, users: [UserType]) {
-        switch (conversationType, users.count) {
-        case (.group?, 1...):
+        let isServiceUser: Bool
```
typfel: Why not move this inside the `case .group?, 1:`

like

```
self = users[0].isServiceUser ?  .one(serviceUser: true) : .four
```

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -107,13 +107,24 @@ extension Mode {
...
+        if users.count > 0 {
+            isServiceUser = users[0].isServiceUser
+        } else {
+            isServiceUser = false
+        }
+
+        switch (conversationType, users.count, isServiceUser) {
+        case (.group?, 1, true),
+             (.oneOnOne?, _, _),
```
typfel: A service user can never be inside `oneOnOne` or `.connection` so we can remove these cases.

Wire-iOS Tests/ConversationAvatarViewTests.swift
```diff
@@ -35,6 +35,84 @@ extension Array where Element: ZMUser {
     }
 }
 
+protocol CoreDataFixed {
```
typfel: This will saves us some typing but I think we should discuss what we want to achieve with this protocol. I'd prefer if we do this in separate PR.

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
```diff
@@ -171,7 +182,7 @@ final public class ConversationAvatarView: UIView {
         }
     }
     
-    private var mode: Mode = .one(serviceUser: false) {
+    var mode: Mode = .one(serviceUser: false) {
```
typfel: Let's make the setter private?

Wire-iOS/Sources/UserInterface/Components/Views/ConversationAvatarView.swift
billypchan: The logic is refactored and the if case is moved inside the switch case.

Wire-iOS Tests/ConversationAvatarViewTests.swift
billypchan: remove the test for now

PR #3526, opened 2019-07-09
CoreDataFixed protocol

billypchan:
## What's new in this PR?

Create `CoreDataFixed` protocol to simplify accessing properties of `CoreDataFixture`.  This is part of migrating `CoreDataSnapshotTestCase` to new `SnapshotTesting`.


Wire-iOS Tests/SwiftSnapshotTesting/CoreDataFixture.swift
```diff
@@ -248,3 +248,27 @@ extension CoreDataFixture {
     }
 }
 
+
+protocol CoreDataFixed {
```
typfel: I propose a name more closely linked to the `CoreDataFixture` like `CoreDataFixtureTestHelper`.

PR #3528, opened 2019-07-10
ZIOS-11562: Force app lock and make time configurable from build config

nicorsm:
## What's new in this PR?

Implemented custom timer on app lock and possibility to force it by a custom `applock.json` file in the configuration folder. TODO: tests.

### DIFF
--- a/Wire-iOS Tests/AppLockTests.swift
+++ b/Wire-iOS Tests/AppLockTests.swift
@@ -1,0 +1,110 @@
+//
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+@testable import WireCommonComponents
+
+class AppLockTests: XCTestCase {
+
+    let decoder = JSONDecoder()
+    
+    func testThatForcedAppLockDoesntAffectSettings() {
+
+        //given
+        let fileData = "{\"forceAppLock\":true,\"timeout\":900}".data(using: .utf8)!
+        AppLockRules.shared = try! decoder.decode(AppLockRules.self, from: fileData)
+        
+        //when
+        XCTAssertTrue(AppLockRules.shared.forceAppLock)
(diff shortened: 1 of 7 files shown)


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -6825,6 +6828,7 @@
 			isa = PBXResourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				7CC0FDA022D62B3F00B67B9D /* applock.json in Resources */,
```
typfel: This file doesn't seem to be committed

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Options.swift
```diff
@@ -186,6 +186,7 @@ extension SettingsCellDescriptorFactory {
         
         if context.canEvaluatePolicy(LAPolicy.deviceOwnerAuthentication, error: &error) {
             let lockApp = SettingsPropertyToggleCellDescriptor(settingsProperty: self.settingsPropertyFactory.property(.lockApp))
+            lockApp.settingsProperty.enabled = !AppLockRules.shared.forceAppLock
```
typfel: Won't this always make it enabled if `forceAppLock` is false?

Wire-iOS.xcodeproj/project.pbxproj
nicorsm: Because it's inside the Configuration folder...

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Options.swift
typfel: Ah never mind, now realized you are disabling the control itself.

Wire-iOS.xcodeproj/project.pbxproj
typfel: Ah I see

Wire-iOS Tests/AppLockTests.swift
```diff
@@ -0,0 +1,110 @@
...
+
+class AppLockTests: XCTestCase {
+
+    let decoder = JSONDecoder()
+    
+    func testThatForcedAppLockDoesntAffectSettings() {
+
+        //given
+        let fileData = "{\"forceAppLock\":true,\"timeout\":900}".data(using: .utf8)!
+        AppLockRules.shared = try! decoder.decode(AppLockRules.self, from: fileData)
```
typfel: We could create `AppLockRules` using the default initialiser instead of decoding from JSON and add a separate tests which tests the JSON decoding.

typfel:
All these singletons and static method involved with the app lock feature aren't making this easier to test. I release it was already like this and now is not the time re-factor.

But maybe we could exposing the `AppLockRules` through the `AppLock` in order to only have one singleton.

typfel:
I guess it also needs PR on https://github.com/wireapp/wire-ios-build-configuration or there another reason why the tests fail?

nicorsm:
@typfel exactly, I'll make PR on build config.

PR #3531, opened 2019-07-12
ZIOS-11563: Added jailbreak detection

nicorsm:
## What's new in this PR?

Added function to detect jailbroken devices through several methods.


nicorsm:
Moved to https://github.com/wireapp/wire-ios-sync-engine/pull/1058

PR #3533, opened 2019-07-12
Observe team image in TeamImageView 

typfel:
## What's new in this PR?

Observe the team in `TeamImageView` and update the image when it changes.

### Dependency

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1056
- [x] https://github.com/wireapp/wire-ios-data-model/pull/760


Wire-iOS/Sources/UserInterface/SelfProfile/AccountView.swift
```diff
@@ -453,19 +459,44 @@ extension PersonalAccountView {
...
+        } else {
+            return nil
+        }
+    }
+    
+}
+
+fileprivate extension Account {
+    
+    var teamImageViewContent: TeamImageView.Content? {
```
billypchan: We can merge this method with above `TeamType.teamImageViewContent` as a method of `TeamImageView.Content`

PR #3534, opened 2019-07-15
Jailbreak detection UI implementation

nicorsm:
## What's new in this PR?

Added UI implementation for jailbreak detection. Alert is handled by an instance of `JailbrokenDeviceViewController` (derivated from `BlockerViewController`, the former `BlacklistViewController`). Copy is NOT final.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1058
- [x] https://github.com/wireapp/wire-ios/pull/3529 (update from develop)


Wire-iOS/Sources/AppStateController.swift
```diff
@@ -193,6 +198,12 @@ extension AppStateController : SessionManagerDelegate {
         }
     }
     
+    func sessionManagerDidBlacklistJailbrokenDevice() {
+        guard !UIDevice.isSimulator else { return }
```
typfel: We could solve this by adding a `#if TARGET_OS_SIMULATOR` which always return false inside the jailbreak detection.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -193,6 +198,12 @@ extension AppStateController : SessionManagerDelegate {
         }
     }
     
+    func sessionManagerDidBlacklistJailbrokenDevice() {
```
typfel: Maybe move it below the `sessionManagerDidBlacklistCurrentVersion()`

Wire-iOS/Sources/AppState.swift
```diff
@@ -24,6 +24,7 @@ enum AppState : Equatable {
     case authenticated(completedRegistration: Bool)
     case unauthenticated(error : NSError?)
     case blacklisted
```
typfel: We could also add flag to this enum instead of introducing a new case. Like `case blacklisted(jailbroken: Bool)`

Wire-iOS/Sources/AppStateController.swift
nicorsm: Tried right now but doesn't seems to work...

typfel: It doesn't work how? Can you show me the code.

nicorsm: I tried with `#if TARGET_OS_SIMULATOR` but it doesn't work. Switching to `#if targetEnvironment(simulator)` instead seems to work...

nicorsm: (look at https://github.com/wireapp/wire-ios-sync-engine/pull/1058/files )

typfel: Ah yeah sorry my bad, `#if TARGET_OS_SIMULATOR` is the objective-c syntax

billypchan:
@nicorsm both target and merging branch need to be synchronized with the develop branch.

Wire-iOS/Sources/UserInterface/JailbrokenDeviceViewController.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+
+class JailbrokenDeviceViewController: BlockerViewController {
+
+    override func showAlert() {
```
billypchan: Can we use extension instead of extending `BlockerViewController`?

Wire-iOS/Sources/UserInterface/BlockerViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+
+import Foundation
+
+class BlockerViewController : LaunchImageViewController {
+    
+    var applicationDidBecomeActiveToken : NSObjectProtocol? = nil
+    
+    deinit {
+        if let token = applicationDidBecomeActiveToken {
+            NotificationCenter.default.removeObserver(token)
```
billypchan: it is not necessary after iOS9

Wire-iOS/Sources/UserInterface/BlockerViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+        applicationDidBecomeActiveToken = NotificationCenter.default.addObserver(forName: UIApplication.didBecomeActiveNotification, object: nil, queue: nil) { [weak self](_) in
+            self?.showAlert()
+        }
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        showAlert()
+    }
+    
+    func showAlert() {
```
billypchan: I think `BlockerViewController` with different contexts is better than creating 2 subclasses for different use cases.

nicorsm: I'd prefer to create two subclasses that may be expanded where needed, but for the simplicity of this class I can also use a single one.

PR #3541, opened 2019-07-18
Fix: allow self user checking the legal hold details

billypchan:
### Issues

legal hold red dot not tappable in a 1-1 conversation which the other user is removed

### Causes

If the other has no 1-1 conversation, the `LegalHoldDetailsViewController` does not initialize. But in this case, we should also consider the current conversation.

### Solutions

Pass the conversation for which presents `LegalHoldDetailsViewController` and use it as subject conversation if there is no 1-1 conversation.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListTopBarViewController.swift
```diff
@@ -178,7 +178,7 @@ final class ConversationListTopBarViewController: UIViewController {
 
     @objc
     func presentLegalHoldInfo() {
-        LegalHoldDetailsViewController.present(in: self, user: ZMUser.selfUser())
+        LegalHoldDetailsViewController.present(in: self, user: ZMUser.selfUser(), from: nil)
```
nicorsm: We could set `conversation: ZMConversation?` to `nil` by default in order to avoid adding `from: nil)` to the method call.

PR #3555, opened 2019-07-29
Fix: Share extension crashes when sharing a specific link with a large image

billypchan:
### Issues

Share extension crashes when sharing a web page with a very large image (withd > 5000px)

### Causes

When assigning the large image to `UIImageView`, app crashes due to excessing memory limit allowed by the system (~ 120MB).

### Solutions

Resize the image to fill the preview view if necessary.

### Todo
- [ ] move image resize functions to utilities to speed up the build time.


Wire-iOS/Sources/Components/AudioPlayer/AudioTrackPlayer.m
```diff
@@ -27,6 +27,7 @@
 @import WireUtilities;
 @import WireSyncEngine;
 #import "KeyValueObserver.h"
+#import "Wire-Swift.h"
```
nicorsm: Is this needed here?

billypchan: It is needed since one of a image resize method is translated to Swift.

nicorsm: Ah ok I see, it's used for the Now Playing artwork 👍

PR #3562, opened 2019-08-01
Fix: share extension crash due to missing file

billypchan:
## What's new in this PR?

Fix share extension crash due to missing `session_manager.json` file


marcoconti83:
🎉

PR #3568, opened 2019-08-07
Fix: no applock displayed after restarting the app

billypchan:
### Issues

When `forceAppLock` is enabled, no applock displayed after restarting the app.

### Causes

If the `timeout` is not reached, applock is not shown.

### Solutions

When the app launches, skip the time checking and show app lock screen if `forceAppLock` is enabled.


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -177,6 +183,23 @@ extension AppLockViewController {
     }
     
     @objc func applicationDidBecomeActive() {
-        self.showUnlockIfNeeded()
+        /// if this is the first time became active, i.e. when app launch, show the look screen if forceAppLock is true
+        if !AppLockViewController.becameActive, AppLock.rules.forceAppLock {
```
typfel: Reading the spec I don't think this should behave differently when `forceAppLock` is on/off. It should always ask of authentication on a app launch.

typfel:
What about clearing on the `lastUnlockedDate` on `didFinishLaunchingNotification`? That would allow us to keep the existing logic.

PR #3570, opened 2019-08-07
Fix: no applock displayed after restarting the app

billypchan:
### Issues

When `forceAppLock` is enabled, no applock displayed after restarting the app.

### Causes

If the `timeout` is not reached, applock is not shown.

### Solutions

~~When the app launches, set `lastUnlockedDate` to invalid value and app lock screen will be shown.~~ Set `lastUnlockedDate` to an invalid value so that app lock always appear when app launches.

Remove `lockAppLastDate` from `SettingsPropertyFactory` and not save the value in the keychain.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
@@ -43,6 +43,9 @@ extension Notification.Name {
         return shared.dimContents
     }
 
+     /// flag to identify the app is just launched. the value is false until applicationDidBecomeActive is called
+    static var becameActive: Bool = false
+
     convenience init() {
         self.init(nibName:nil, bundle:nil)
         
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -177,6 +180,13 @@ extension AppLockViewController {
     }
     
     @objc func applicationDidBecomeActive() {
-        self.showUnlockIfNeeded()
+        if !AppLockViewController.becameActive {
+            AppLockViewController.becameActive = true
+
+            /// invlidate lastUnlockedDate, app lock will show since the timeout is reached
+            AppLock.lastUnlockedDate = Date.init(timeIntervalSince1970: 0)
```
typfel: If the app was just launched`lastUnlockedDate` would be already be set to `Date.init(timeIntervalSince1970: 0)` since it's not persisted between launches?

billypchan: removed the unnecessary checkings

PR #3577, opened 2019-08-09
Allow override configuration_path and hockey_app_id in fastfile 

billypchan:
## What's new in this PR?

To allow submitting a build to Hockey app with current configuration sets with variations, e.g. change timeout setting and force app lock policy, and send it under different hockey_app_id. New options are added to `fastfile`.

- add `configuration_path` option to `prepare` lane
- add `hockey_app_id` option to `upload_hockey` lane

If the options are not provided, the default value will be applied.


typfel:
Can you explain the use case of overriding `hockey_app_id`? It's based on the `build_type` AFAIK so when do need a build_type which doesn't go to the matching hockey id?

Similarly what's the use case of for overriding `configuration_path`?

billypchan:
reopen in
https://github.com/wireapp/wire-ios/pull/3579

PR #3580, opened 2019-08-14
fastfile & download-assets.sh: config repo url and branch options

billypchan:
## What's new in this PR?

Allow setting config repo URL and branch options when running`download-assets.sh`
Update `fastfile` to support passing new parameter to `download-assets.sh` via `setup.sh`

### DIFF
--- a/Scripts/download-assets.sh
+++ b/Scripts/download-assets.sh
@@ -32,7 +32,7 @@ OVERRIDES_DIR=
 
 usage()
 {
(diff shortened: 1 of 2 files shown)


Scripts/download-assets.sh
```diff
@@ -50,6 +52,18 @@ while [ "$1" != "" ]; do
         -h | --help )           usage
                                 exit
                                 ;;
+        -c)                     shift
```
typfel: Can't you move let this fall through at line 40 to avoid duplicating the code?

typfel: This was not that easy to do?

billypchan: The `shift` is moved before `case` and duplicated `shift`s are removed.

PR #3593, opened 2019-08-21
Fix: NetworkConditionHelper supports multiple radio access technology

billypchan:
### Issues

After iOS12, iOS can return multiple network service technologies instead of single `currentRadioAccessTechnology`. (`NetworkConditionHelper` is used for checking the quality of service before starting a video call)

### Solutions
Refactor `NetworkConditionHelper` to check for `serviceCurrentRadioAccessTechnology` dictionary instead and choose the best quality type in it.


Wire-iOS/Sources/Components/NetworkConditionHelper.swift
```diff
@@ -0,0 +1,91 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import CoreTelephony
+
+enum NetworkQualityType: Int, Comparable {
+    case unknown = 0
+    case type2G
```
typfel: Let's lose the `type` prefix?

billypchan: The caller uses `typeXX` prefix and Swift not allow an enum name starting with a number.

PR #3595, opened 2019-08-21
Fix: Country supports multiple cellular providers

billypchan:
### Issues

After iOS12, iOS can return multiple `serviceSubscriberCellularProviders` instead of single `subscriberCellularProvider`. The `Country` class use it to create an instance with the country code of the carrier.

### Solutions

Refactor `Country`'s `countryFromDevice` method to adapt `serviceSubscriberCellularProviders`


Wire-iOS/Sources/Authentication/Helpers/CountryCode/Country+Extension.swift
```diff
@@ -0,0 +1,46 @@
...
+    ///
+    /// - Returns: a Country object
+    @objc
+    class func countryFromDevice() -> Country? {
+        let networkInfo = CTTelephonyNetworkInfo()
+
+        let carrier: CTCarrier?
+        if #available(iOS 12, *) {
+            /// Get the carrier from first cellular provider
+            carrier = networkInfo.serviceSubscriberCellularProviders?.values.first
```
typfel: Since this is dictionary we are now getting a random provider. I suggest that we return the first carrier which also has a ` isoCountryCode`.

Wire-iOS/Sources/Authentication/Helpers/CountryCode/Country.m
```diff
@@ -72,7 +53,7 @@ + (nullable instancetype)countryWithISO:(NSString *)ISO e164:(NSNumber *)e164
     return country;
 }
 
-+ (nullable instancetype)countryWithISO:(NSString *)ISO
++ (nullable instancetype)countryWithISO:(NSString * __nullable)ISO
```
typfel: It's enough to declare nullability in the header. We can remove this and not have to worry about keeping this in sync.

billypchan: Compile complains for `Nullability specifier 'nonnull' conflicts with existing specifier '_Nullable'` if not included nullability marker. I will refactor this method in Swift.

typfel: It probably complains because `nullable` in the return value. You need remove all nullability annotations.

typfel: Ah but I see now that the entire file full nullability annotations.

billypchan: refactor to Swift and the nullability is much clearer :)

Wire-iOS/Sources/Authentication/Helpers/CountryCode/Country+Extension.swift
```diff
@@ -0,0 +1,60 @@
...
+            return Country.countryWithISO(iso: isoCountryCode)
+        } else {
+            return Country.countryWithISO(iso: NSLocale.current.regionCode?.lowercased())
+        }
+    }
+
+    private class func countryWithISO(iso: String?) -> Country? {
+        guard let iso = iso else { return nil }
+
+        for country in allCountries() {
```
billypchan: @typfel `countryWithISO` is actually searching a `Country` from a static array from `allCountries`, currently Swift `convince init` mush call a `init` method and does not suit this case.

typfel: OK 👍

PR #3597, opened 2019-08-21
fix: deprecated warnings

billypchan:
## What's new in this PR?

Fix warnings regarding deprecations.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Constraint.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Constraint.swift
@@ -27,7 +27,12 @@ extension KeyboardAvoidingViewController {
         topEdgeConstraint = constraints[.top]
         topEdgeConstraint?.constant = topInset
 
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController+Constraint.swift
```diff
@@ -27,7 +27,12 @@ extension KeyboardAvoidingViewController {
         topEdgeConstraint = constraints[.top]
         topEdgeConstraint?.constant = topInset
 
-        bottomEdgeConstraint = viewController.bottomLayoutGuide.bottomAnchor.constraint(equalTo: bottomLayoutGuide.bottomAnchor, constant: 0)
+        if #available(iOS 11.0, *) {
```
typfel: Can't we use one of the helper methods in `UIViewController+SafeArea.swift`?

typfel:
Why the `Don't merge` label?

billypchan:
I would write more comment about this PR later.

PR #3614, opened 2019-08-27
Feature: support sharing all file type

billypchan:
## What's new in this PR?

Support share all file types via Safari or File app (with `UTI = public.fileurl`).
The filename is extracted from the file URL and prefilled in the text editor.

### DIFF
--- a/Wire-iOS Share Extension/Info.plist
+++ b/Wire-iOS Share Extension/Info.plist
@@ -32,8 +32,18 @@
 		<dict>
 			<key>NSExtensionActivationRule</key>
 			<string>
-				TRUEPREDICATE
-			</string>
+                SUBQUERY(extensionItems, $extensionItem, SUBQUERY($extensionItem.attachments, $attachment, ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.plain-text").@count &gt;= 1).@count &gt;= 1
+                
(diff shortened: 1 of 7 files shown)


Wire-iOS Share Extension/Info.plist
```diff
@@ -34,7 +34,7 @@
 			<string>
                 SUBQUERY(extensionItems, $extensionItem, SUBQUERY($extensionItem.attachments, $attachment, ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO "public.plain-text").@count &gt;= 1).@count &gt;= 1
                 
-                OR SUBQUERY(extensionItems, $extensionItem, SUBQUERY($extensionItem.attachments, $attachment, SUBQUERY($attachment.registeredTypeIdentifiers, $uti, $uti UTI-CONFORMS-TO "public.url" AND NOT $uti UTI-CONFORMS-TO "public.file-url").@count &gt;= 1).@count &gt;= 1).@count &gt;= 1
```
typfel: Does the remaining `OR  SUBQUERY` make any sense when we remove the `AND NOT` or will this this always match now?

billypchan: `public.file-url` is a subtype of `public.url` and it was excluded.

```
/*
 *  kUTTypeURL
 *
 *    The bytes of a URL
 *    (OSType 'url ')
 *
 *    UTI: public.url
 *    conforms to: public.data
 *
 *
 *  kUTTypeFileURL
 *
 *    The text of a "file:" URL 
 *    (OSType 'furl')
 *
 *    UTI: public.file-url
 *    conforms to: public.url
 *
 */
```

the 13 `SUBQUERY` in `NSExtensionActivationRule` code seems can be refactored in a simpler way. But I would need to find out the original requirement.

typfel: I looked up the original commit which added these rules and didn't contain hints of what the requirements were.

PR #3616, opened 2019-08-28
Retire IS_IPAD_PORTRAIT_LAYOUT macro

billypchan:
## What's new in this PR?

Retire IS_IPAD_PORTRAIT_LAYOUT macro.


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+rightView.swift
```diff
@@ -30,3 +30,17 @@ extension SplitViewController {
...
     }
 }
+
+extension SplitViewController {
+
+    @objc(updateLayoutSizeForTraitCollection:size:)
+    func updateLayoutSize(for traitCollection: UITraitCollection?, size: CGSize) {
+        if traitCollection?.horizontalSizeClass == .compact {
+            layoutSize = .compact
+        } else if isIPadRegular(), UIApplication.shared.statusBarOrientation.isPortrait {
```
typfel: Would be nice to remove the dependency on `UIApplication.shared.statusBarOrientation.isPortrait` and `device.userInterfaceIdiom == .pad`. Couldn't we achieve the behaviour by checking if `size.height > size.width`?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+ConversationListContentDelegate.swift
```diff
@@ -0,0 +1,41 @@
...
+
+import Foundation
+
+extension ConversationListViewController: ConversationListContentDelegate {
+    public func conversationList(_ controller: ConversationListContentController!, didSelect conversation: ZMConversation!, focusOnView focus: Bool) {
+        setSelectedConversation(conversation)
+    }
+
+    public func conversationList(_ controller: ConversationListContentController!, willSelectIndexPathAfterSelectionDeleted conv: IndexPath!) {
+        if isIPadRegular() && UIApplication.shared.statusBarOrientation.isPortrait {
```
typfel: Maybe this decision could be made one level up by the `ZClientVIewController` which could then query `SplitViewController.layoutSize == .regularPortrait`

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+rightView.swift
typfel: I'd say that `isIPadRegular()` is just as bad as the macro and we try to remove the usage of it.

billypchan: The `isIPadRegular` is a method with a default parameter for injecting a mock `UIDevice` for testing. The problem of using a macro is we can not use it in Swift and we can not inject orientation and size class for testing.

typfel: I see, but I'm not proposing to go back to macros but instead to rely on the trait collection and size. To modify the trait collection and size in a test?

billypchan: `size.height > size.width` does not related to layout modes. E.g. iPad pro has 3 split modes in landscape orientations and the 1/3 mode match the case `size.height > size.width` and we still want to show the list.

billypchan: Remove the unused size parameter.

typfel: OK, let's keep like this for now then but how about using the existing `isIPadRegularPortrait()` method.

PR #3617, opened 2019-08-29
Move methods from ConversationListViewController to ZClientViewController

billypchan:
### Issues
The methods `showDataUsagePermissionDial` and `showAvailabilityBehaviourChangeAlertIfNeeded` in
`ConversationListViewController` are not related to the class

### Solutions

Put the methods in `ZClientViewController` and clean up unnecessary properties `isComingFromRegistration` & `needToShowDataUsagePermissionDialog`.
Refactor viewDidAppear to Swift.


Wire-iOS Tests/ZClientViewControllerSnapshotTests.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    override func tearDown() {
+        sut = nil
+        coreDataFixture = nil
+        
+        super.tearDown()
+    }
+
+    func testForShowDataUsagePermissionDialogIfNeeded() {
```
typfel: I don't think this test doesn't needs to be a snapshot test. This seems like the perfect opportunity to for a unit test on a view model.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -21,6 +21,16 @@ import Foundation
...
 extension ZClientViewController {
 
+    // MARK: - Setup methods
+
+    @objc
+    func setupConversationListViewController(account: Account) {
+        conversationListViewController = ConversationListViewController()
+        conversationListViewController.account = account
+
+        needToShowDataUsagePermissionDialog = false
```
typfel: This seems unrelated to this method can't it be moved to the `init` just like `self.dataUsagePermissionDialogDisplayed = false`

Wire-iOS Tests/ZClientViewControllerSnapshotTests.swift
billypchan: refactored to test the alert is created or not for the different stages.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+PushPermissions.swift
```diff
@@ -52,7 +52,12 @@ extension ConversationListViewController {
         contentContainer.alpha = 1.0
     }
 
-    @objc func showPushPermissionDeniedDialogIfNeeded() {
+    private var isComingFromRegistration: Bool {
+        return ZClientViewController.shared()?.isComingFromRegistration ?? false
+    }
+
+    @objc
+    func showPushPermissionDeniedDialogIfNeeded() {
```
typfel: I guess this dialogue is also not directly related to the conversation list. Can of course be re-factored in a separate PR.

billypchan: This does not add a alert but a `PermissionDeniedViewController` to `ConversationListViewController` so it should stays here.

PR #3624, opened 2019-09-02
Fix: ConversationListViewController - translate overrode methods and observer methods to Swift

billypchan:
## What's new in this PR?

Translate overrode methods and these observer methods to Swift:
```
ZMConversationListObserver
ZMInitialSyncCompletionObserver
```

### TODO
- [ ] this delegate is should be placed in ViewModel. I will move it later.


Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+ViewLifeCycle.swift
```diff
@@ -39,4 +55,47 @@ extension ConversationListViewController {
...
+
+        return .lightContent
+    }
+
+    override open func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
+        coordinator.animate(alongsideTransition: { context in
+            // we reload on rotation to make sure that the list cells lay themselves out correctly for the new
+            // orientation
+            self.listContentController.reload()
+        }) { context in
```
typfel: Should this second block stay?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+ViewLifeCycle.swift
```diff
@@ -39,4 +55,47 @@ extension ConversationListViewController {
...
+        }
+
+        super.viewWillTransition(to: size, with: coordinator)
+    }
+
+    override open var shouldAutorotate: Bool {
+        return true
+    }
+
+    override open var definesPresentationContext: Bool {
```
typfel: Hmm can't we just assign `definesPresentationContext = true` where we construct the `ConversationListViewController`?

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+Internal.h
```diff
@@ -27,8 +27,16 @@
 @property (nonatomic, nullable) ConversationActionController *actionsController;
 @property (nonatomic) BOOL viewDidAppearCalled;
 
+@property (readwrite, nonatomic, nonnull) UIView *contentContainer;
+
+@property (nonatomic, nonnull) id userObserverToken;
```
typfel: These tokens aren't assigned until `viewDidLoad` so it would be safer to make them `nullable`.

billypchan: These tokens are set only I think `nonnull` is acceptable. I will check them still necessary if I do not keep a reference to the observers.

typfel: On the other hand If they are only set there's also no inconvenience if they nullable. The safety I'm talking about, is when someone in the future writes codes that reads them.

Wire-iOS/Sources/UserInterface/ConversationList/ConversationListViewController+ViewLifeCycle.swift
billypchan: removed!

PR #3631, opened 2019-09-04
Feature: deleting a conversation

billypchan:
## What's new in this PR?

Add delete group action item if needed and display a confirm alert before deleting.

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -692,7 +692,7 @@
 "meta.menu.leave" = "Leave Group…";
 "meta.menu.archive" = "Archive";
 "meta.menu.unarchive" = "Unarchive";
-"meta.menu.delete" = "Delete conversation";
+"meta.menu.delete" = "Delete Group";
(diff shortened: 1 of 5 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -766,6 +766,11 @@
 "profile.connection_request_dialog.button_cancel" = "Ignore";
 "profile.connection_request_dialog.button_connect" = "Connect";
 
+// Delete group
+"conversation.delete_request_dialog.title" = "Delete group conversation?";
+"conversation.delete_request_dialog.message" = "This will delete the whole group from everyone's list and clear all the content on all member's devices. You would not able to restore any content from this group, and all group members will get notified of this action.";
+"conversation.delete_request_error_dialog.title" = "Error occurs when deleting a group conversation";
```
typfel: So we will need to talk to @johnxnguyen about the correct copy here.

typfel: But we can do that in a separate PR

billypchan: I think the alert message can be shortened.

PR #3632, opened 2019-09-04
Fix: retire IS_IPAD_LANDSCAPE_LAYOUT macro

billypchan:
## What's new in this PR?

Retire IS_IPAD_LANDSCAPE_LAYOUT macro and clean up `isConversationViewVisible` and `isConversationListVisible`


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+leftView.swift
```diff
@@ -19,6 +19,10 @@
 import Foundation
 
 extension SplitViewController {
+    @objc
+    var isConversationViewVisible: Bool {
+        return (isIPadRegular() && UIApplication.shared.statusBarOrientation.isLandscape) || !isLeftViewControllerRevealed
```
typfel: Can't this be: `return layoutSize == .regularLanscape ||  isLeftViewControllerRevealed`

`UIApplication.shared.statusBarOrientation` is a deprecated method I think we should limit our use of it.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -267,4 +267,13 @@ extension ZClientViewController {
         conversationListViewController.select(conversation)
     }
 
+    @objc
+    var isConversationViewVisible: Bool {
+        return splitViewController.isConversationViewVisible
+    }
+
+    var isConversationListVisible: Bool {
+        return (isIPadRegular() && UIApplication.shared.statusBarOrientation.isLandscape) || (splitViewController.isLeftViewControllerRevealed && conversationListViewController.presentedViewController == nil)
```
typfel: Seems like this could also be simplified to:

`return splitViewController.layoutSize == .regularLanscape || (splitViewController.isLeftViewControllerRevealed && conversationListViewController.presentedViewController == nil)`

Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController+leftView.swift
billypchan: `UIApplication.shared.statusBarOrientation` is still the only way to get the interface orientation.

PR #3633, opened 2019-09-05
Fix: iPad 12inch - 2/3 split mode has wider margin than the compact mode

billypchan:
### Issues

On iPad 12-inch's 2/3 split mode, the right conversation view has wider margins than the compact mode. The right conversation view takes 368 pt and the margins take 96pt on both sides, which is ~50% of the view's width.

### Causes

We use the same margin size for regular mode but not considering the view's width.

### Solutions

When the view's width is less than 400pt, use compact mode's margin sizes.
Refactor `UIView.conversationLayoutMargins` and create `UITraitEnvironment .conversationHorizontalMargins` which get the `horizontalSizeClass` from the view instead of from `UIApplication.shared.keyWindow?.traitCollection`.

### TODO
- [ ] retire `UIView.conversationLayoutMargins` and replace with `UITraitEnvironment .conversationHorizontalMargins`


Wire-iOS/Sources/Helpers/UIView+LayoutMargins.swift
```diff
@@ -18,29 +18,53 @@
...
+extension UITraitEnvironment {
+    var conversationHorizontalMargins: HorizontalMargins {
+        guard traitCollection.horizontalSizeClass == .regular else {
+            return HorizontalMargins(userInterfaceSizeClass: .compact)
+        }
+
+        let userInterfaceSizeClass: UIUserInterfaceSizeClass
+
+        /// on iPad 9.7 inch 2/3 mode, right view's width is  396pt, use the compact mode's narrower margin
+        if let viewWidth = ZClientViewController.shared()?.splitViewController.rightView.frame.width,
```
typfel: I would prefer not to rely on accessing a singleton when calculating this. Can't we pass  n the view size as parameter to this method?

billypchan: I would use the windows' width instead.

PR #3637, opened 2019-09-06
Fix: iOS 13 navigation bar titles

nicorsm:
### Issues

Navigation bar items are using the default font size and weight on iOS 13.0 (e.g. "FOR COMPANIES").

IMG_0646

### Causes

This is probably caused by the new `UINavigationBarAppearance` object introduced in iOS 13.0: https://developer.apple.com/documentation/uikit/uinavigationbarappearance

### Solutions

I've added a possible fix by setting the `titleTextAttributes` also to the new `standardAppearance` property. **This needs to be tested on iOS 13.0**, so let's wait for the switch to Xcode 11.


billypchan:
no longer needed after UI update.

PR #3640, opened 2019-09-06
Fix: retire ipad checking macros

billypchan:
## What's new in this PR?

Translated the code using `IS_IPAD_FULLSCREEN` to Swift.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListCell+Constraint.swift
```diff
@@ -19,6 +19,25 @@
 import Foundation
 
 extension ConversationListCell {
```
typfel: How about renaming this file to just `ConversationListCell.swift` since it's now just a place to put the swift implementation.

typfel:
If you haven't already please test that verify that the affected screens/interactions still work.

PR #3646, opened 2019-09-09
Fix: extract code form ConversationListViewController to ViewModel

billypchan:


Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+            DispatchQueue.main.async {
+                if pushesDisabled,
+                    let weakSelf = self {
+                    weakSelf.viewController.observeApplicationDidBecomeActive()
+
+                    Settings.shared().lastPushAlertDate = Date()
+
+                    weakSelf.viewController.showPermissionDeniedViewController()
+
+                    weakSelf.viewController.concealContentContainer()
```
typfel: I'd suggest calling `concealContentContainer` from inside `showPermissionDeniedViewController()` and remove it from the `ConversationListContainerViewModelDelegate` protocol.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+
+        if !pushAlertHappenedMoreThan1DayBefore {
+            return
+        }
+
+        UNUserNotificationCenter.current().checkPushesDisabled({ [weak self] pushesDisabled in
+            DispatchQueue.main.async {
+                if pushesDisabled,
+                    let weakSelf = self {
+                    weakSelf.viewController.observeApplicationDidBecomeActive()
```
typfel: Can't this be called from inside `showPermissionDeniedViewController()` and be removed from the view model protocol?

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+            return
+        }
+
+        if AutomationHelper.sharedHelper.skipFirstLoginAlerts || viewController.hasUsernameTakeoverViewController {
+            return
+        }
+
+        let pushAlertHappenedMoreThan1DayBefore: Bool = Settings.shared().pushAlertHappenedMoreThan1DayBefore
+
+        if !pushAlertHappenedMoreThan1DayBefore {
```
typfel: This can simplified with a guard statement and and done without a local variable.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    private var isComingFromRegistration: Bool {
+        return ZClientViewController.shared()?.isComingFromRegistration ?? false
+    }
+
+    func showPushPermissionDeniedDialogIfNeeded() {
+        // We only want to present the notification takeover when the user already has a handle
+        // and is not coming from the registration flow (where we alreday ask for permissions).
+        if isComingFromRegistration || nil == ZMUser.selfUser().handle {
```
typfel: Can be a guard statement

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+            return
+        }
+
+        if AutomationHelper.sharedHelper.skipFirstLoginAlerts || viewController.hasUsernameTakeoverViewController {
```
typfel: Can be a guard statement

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+                animated: Bool = false,
+                completion: (() -> ())? = nil) {
+        selectedConversation = conversation
+
+        viewController.dismissPeoplePicker(with: { [weak self] in
+            self?.viewController.selectOnListContentController(self?.selectedConversation, scrollTo: message, focusOnView: focus, animated: animated, completion: completion)
+        })
+    }
+
+    func removeUserProfileObserver() {
```
typfel: Let's move these methods to `ConversationListViewControllerViewModel+Handles.swift` where they are used.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+Handles.swift
```diff
@@ -0,0 +1,159 @@
...
+        case .keepSuggestion(let suggested):
+            setSuggested(handle: suggested)
+        case .learnMore:
+            if let viewController = viewController as? UIViewController {
+                URL.wr_usernameLearnMore.openInApp(above: viewController)
+            }
+        }
+    }
+
+    func createConversation(withUsers users: Set<ZMUser>?, name: String?, allowGuests: Bool, enableReceipts: Bool) {
```
typfel: Let's move the conversation creation methods to a separate file since they are not related to the user handle.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed!

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+typealias CompletionHandler = (Bool) -> Void
+
+protocol ConversationListContainerViewModelDelegate: class {
+    init(selfUser: SelfUserType, viewModel: ConversationListViewController.ViewModel)
+
+    func updateBottomBarSeparatorVisibility(with controller: ConversationListContentController)
+    func dismissPeoplePicker(with block: @escaping Completion)
+    func scrollViewDidScroll(scrollView: UIScrollView!)
+
+    func setState(_ state: ConversationListState,
```
typfel: I'd argue that putting every argument on separate line is not more readable in this case.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+typealias Completion = ()->()
+typealias CompletionHandler = (Bool) -> Void
```
typfel: I'd prefer to not a have generic `CompletionHandler` `typealias` which takes a boolean argument since it will be unclear what that boolean actually means.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+    func selectOnListContentController(_ conversation: ZMConversation!, scrollTo message: ZMConversationMessage?, focusOnView focus: Bool, animated: Bool, completion: (() -> Void)?) -> Bool
+
+    var hasUsernameTakeoverViewController: Bool { get }
+}
+
+extension ConversationListViewController: ConversationListContainerViewModelDelegate {}
+
+extension ConversationListViewController {
+    final class ViewModel: NSObject {
+        weak var viewController: ConversationListContainerViewModelDelegate! {
```
typfel: Declare as optional?

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+                updateNoConversationVisibility()
+                updateArchiveButtonVisibility()
+                showPushPermissionDeniedDialogIfNeeded()
+            }
+        }
+
+        let account: Account
+        var selectedConversation: ZMConversation?
+
+        weak var userProfile: UserProfile? = ZMUserSession.shared()?.userProfile
```
typfel: Why does `userProfile` need to be weak?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController+ConversationListCellDelegate.swift
```diff
@@ -1,4 +1,4 @@
-//
+
```
typfel: Space character

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -26,7 +26,9 @@ extension ZClientViewController {
     ///TODO: caller to Swift
     @objc
     func setupConversationListViewController(account: Account, selfUser: UserType) {
-        conversationListViewController = ConversationListViewController(account: account, selfUser: selfUser as! SelfUserType)
+        let viewModel = ConversationListViewController.ViewModel(account: account)
+        conversationListViewController = ConversationListViewController(selfUser: selfUser as! SelfUserType, viewModel: viewModel)
+        viewModel.viewController = conversationListViewController
```
typfel: Can't we do this inside the `ConversationListViewController` init?

billypchan: The separated creation of VM and VC makes testing easier, i.e. we can use a mock VC instead of the `ConversationListViewController`.
We can create a coordinator to manage the setup of the VM and VC.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+import Foundation
+
+typealias Completion = ()->()
+typealias CompletionHandler = (Bool) -> Void
+
+protocol ConversationListContainerViewModelDelegate: class {
+    init(viewModel: ConversationListViewController.ViewModel)
+
+    func updateBottomBarSeparatorVisibility(with controller: ConversationListContentController)
+    func dismissPeoplePicker(with block: @escaping Completion)
```
typfel: I think we can remove this method since we can achieve the same thing with `setState(_ state:, ..)`.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
typfel: One does not exclude the other. We can assign it the `ConversationListViewController` init and allow for the property to be assign separately at the same time.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed!

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+Handles.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: We have same `typealias` named `ImageSaveCompletion` and `ToggleHandler` already. I would suggest renaming it as `BoolResultHandler`

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
typfel: Having a shared completion handler type alias is not helpful since to make the code understandable we need a meaningful name on the `Bool` variable. I think we should create a separate completion handler for each use case.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: The `weak` reference is set since this class is created: https://github.com/wireapp/wire-ios/pull/439

`userProfile` is used for `requestSettingHandle` and it can be replace as a property.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
```diff
@@ -0,0 +1,207 @@
...
+        var connectionRequestsObserverToken: Any?
+
+        var actionsController: ConversationActionController?
+
+        init(account: Account) {
+            self.account = account
+        }
+
+        deinit {
+            removeUserProfileObserver()
```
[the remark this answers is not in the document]
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController+ConversationListCellDelegate.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel.swift
billypchan: fixed

billypchan: fixed

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
```diff
@@ -39,9 +39,9 @@ extension ConversationListViewController.ViewModel: StartUIDelegate {
     }
 
     func startUI(_ startUI: StartUIViewController!, didSelect conversation: ZMConversation!) {
-        viewController?.dismissPeoplePicker(with: {
+        viewController?.setState(.conversationList, animated:true) {
```
typfel: It's redundant to set the state here since `ZClientViewController.shared()?.select(conversation, focusOnView: true, animated: true)` will end up in the view model which also sets the state.

billypchan: nice catch and I also fixed in `ProfilePresenter`

typfel:
Let's merge this after we've created the `release/3.44`

Wire-iOS/Sources/Authentication/Interface/Helpers/UIAlertController+Newsletter.swift
```diff
@@ -72,8 +72,8 @@ extension UIAlertController {
         #endif
     }
 
-    @objc static func showNewsletterSubscriptionDialogIfNeeded(presentViewController: UIViewController,
-                                                         completionHandler: @escaping (Bool) -> Void) {
+    static func showNewsletterSubscriptionDialogIfNeeded(presentViewController: UIViewController,
+                                                         completionHandler: @escaping BoolResultHandler) {
```
typfel: This still doesn't tell me what the boolean result means. I would expect a completion handler like this:
 `(_ subscribe: Bool) -> Void `

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
typfel: I still think this makes sense to do. I'm not proposing to create the view model inside the `ConversationListViewController` just that it configures itself as the delegate on the injected view model.

PR #3651, opened 2019-09-10
[Fix] Delete group action should not be visible from the conversation list

typfel:
### Issues

- `Delete group` action should not be visible from the conversation list
- `Notifications...` action should not be visible from the group details menu

### Solutions

Introduce separate lists of actions for list and group details.


billypchan:
snapshot of `testForActionMenu` needs to be updated for the contexts.

PR #3655, opened 2019-09-11
Add a separate you localization case for missed calls in groups

typfel:
### Issues

In German the `you` case of a sentence needs to be treated differently than the third person case.

#### Third person:
Verpasster Anruf von **John**

#### Second person (Currently):
Verpasster Anruf von **Du**

#### Second person (Expected):
Verpasster Anruf von **Dir**

### Solutions

Add separate localization string for the second person case.


billypchan:
@typfel I would suggest directly editing the DE localization files to see the result. Then we can suggest the translators for this case.

PR #3656, opened 2019-09-11
Fix: Added validation step for phone during signup

nicorsm:
### Issues

Phone numbers inserted with blank space were sent to the server as not validated.

### Causes

This is because we are currently validating the number, but the instance sent to the server is not validated, so it still sends the number with blank space.

### Solutions

I'm validating the number before the submission.


typfel:
I'd like to understand why this stopped working with the validator changes. I would expect us to validate the number again here https://github.com/wireapp/wire-ios-sync-engine/blob/92a74cc3b8e1e0ad183aadd797b24df0213e6d5e/Source/UnauthenticatedSession/UnauthenticatedSession%2BLogin.swift#L60

Maybe that doesn't work anymore?

nicorsm:
@typfel I checked on the dev build before the release of changes for the validators, and it was not working as well... it isn't a regression 😁

typfel:
I see.

PR #3666, opened 2019-09-12
Fix: snapshot test for "Everything archived" label

billypchan:
## What's new in this PR?

Inject `ConversationListType` when creating `ConversationListViewController.ViewModel` for mocking `ZMConversationList`.

Creating a snapshot test for https://github.com/wireapp/wire-ios/pull/3652


typfel:
I think we should re-think the `ZMConversationList` interface it's really not very suitable for testing.

PR #3669, opened 2019-09-13
Fix: clean up unsed Swift methods

billypchan:
## What's new in this PR?

Clean up methods without callers.


Wire-iOS/Sources/UserInterface/Helpers/UITableViewCell+Reuse.swift
```diff
@@ -48,9 +48,3 @@ extension UICollectionView {
         return dequeueReusableCell(withReuseIdentifier: T.zm_reuseIdentifier, for: indexPath) as! T
     }
 }
-
-extension UITableView {
```
typfel: It makes sense to keep this.

PR #3672, opened 2019-09-13
Fix: select country codes plist from file transfer doesnt trigger transfer

billypchan:
## What's new in this PR?

Simplify the file URL creation logic to fix the bundle file can not be copied issue.
Bump version to 3.45


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+FilesPopover.swift
```diff
@@ -60,16 +60,14 @@ extension ConversationInputBarViewController {
...
-                    let sourceLocation = Bundle(for: type(of: self)).url(forResource: "CountryCodes", withExtension: "plist") else { return }
+                    let sourceLocation = Bundle.main.url(forResource: "CountryCodes", withExtension: "plist") else { return }
 
-
-                let destLocationString = URL(fileURLWithPath: basePath).appendingPathComponent(sourceLocation.lastPathComponent).absoluteString
-                let destLocation = URL(fileURLWithPath: destLocationString)
+                let destLocation = URL(fileURLWithPath: basePath).appendingPathComponent(sourceLocation.lastPathComponent)
 
                 do {
                     try FileManager.default.copyItem(at: sourceLocation, to: destLocation)
```
typfel: If we don't care about the error we can remove the do block and make it a `try?`

PR #3673, opened 2019-09-16
Fix: Zero-bytes attachments seen as always loading files

nicorsm:
### Issues

Zero-bytes attachments are always rendered as loading files (with three dots loading view).

### Solutions

I've added some clauses in order to render it as normal files attachments.


typfel:
Why does a zero size file have the `.uploading` transferState?

PR #3675, opened 2019-09-17
Fix: When clearing content of group conversation and searching for the name, then tapping on it, does not open the conversation

billypchan:
### Issues

When clearing the content of group conversation and searching for the name, then tapping on it, does not open the conversation.

### Causes

The selected conversation is still archived and not found in the conversation list. 

### Solutions
Create a new method `ZMConversation.revealClearedOrArchived` which is extracted from `ConversationListViewModel.selectItem`. Call `revealClearedOrArchived` when the user tapped on a search item to make the conversation not archived before opening it.
Remove animation of minimizing overlay of the call controller.

### Note
In this PR I refactored the method to access `archivedConversations` and `clearedConversations`.
The original way is getting them with `ZMConversationList` static Helper method. The better way would be changing them to properties of `ZMUserSession`. I will move this change to data model.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ZMConversation+RevealArchived.swift
```diff
@@ -0,0 +1,48 @@
...
+extension ZMConversation {
+
+    @discardableResult
+    func revealClearedOrArchived(userSession: ZMUserSessionInterface? = ZMUserSession.shared(),
+                                 completionHandler: Completion?) -> Bool {
+        var containedInOtherLists = false
+
+        guard let userSession = userSession else { return containedInOtherLists }
+
+        if userSession.archivedConversations.contains(self) {
```
typfel: It seems unnecessary to check if the archivedConversations contains the conversation since `isArchived = true` by definition means it's part of that list.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+ArchivedListViewControllerDelegate.swift
```diff
@@ -23,14 +23,12 @@ extension ConversationListViewController.ViewModel: ArchivedListViewControllerDe
...
     }
 
-    func archivedListViewController(_ controller: ArchivedListViewController, didSelectConversation conversation: ZMConversation) {
-        ZMUserSession.shared()?.enqueueChanges({
-            conversation.isArchived = false
-        }, completionHandler: { [weak self] in
-
+    func archivedListViewController(_ controller: ArchivedListViewController,
+                                    didSelectConversation conversation: ZMConversation) {
+        conversation.revealClearedOrArchived() { [weak self] in
```
typfel: I would try simplify this to

```
UserSession.shared.enqueueChanges {
    conversation.isArchived = false
}, competionHandler: {
       self?.viewController?.setState(.conversationList, animated: true, completion:{
	         self?.viewController?.selectOnListContentController(conversation, scrollTo: nil, focusOnView: true, animated: true, completion: nil)
	   })
}
```

typfel:
The animation changes to dismissing call seems unrelated. Let's do that in separate PR.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+StartUIDelegate.swift
```diff
@@ -41,7 +41,9 @@ extension ConversationListViewController.ViewModel: StartUIDelegate {
     }
 
     func startUI(_ startUI: StartUIViewController, didSelect conversation: ZMConversation) {
-        ZClientViewController.shared()?.select(conversation, focusOnView: true, animated: true)
+        conversation.revealClearedOrArchived() {
```
typfel: Same thing here let's set `conversation.isArchived = false`

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/Model/ZMConversation+RevealArchived.swift
typfel: Given that it's safe to set `conversation.isArchived = false` I would remove this method.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ViewModel/ConversationListViewControllerViewModel+ArchivedListViewControllerDelegate.swift
typfel: But isn't the unarchiving already covered by `select(itemToSelect: Any?)`?

billypchan: It would be too late if unarchive in `select(itemToSelect: Any?)`. There was a guard to prevent conversation list not to select a conversation which does not on the list so we have to unarchive it first.

typfel: Ah ok, I'll see if it's possible to unify it.

PR #3685, opened 2019-09-25
reduce non critical warnings

billypchan:
## What's new in this PR?

Reduce non-critical warnings to reduce the number of lines of build log and force on newly created warnings.

### Todo:
- [ ] rewrite objc c code using Swift protocol  as Swift class extension


Wire-iOS/Sources/Helpers/FormattedText.swift
```diff
@@ -41,7 +41,8 @@ enum FormattedText {
      */
 
     static func list(from array: [String]) -> String {
-        #warning("TODO iOS 13: Use the Foundation list formatter.")
+        ///TODO:
+//        #warning("TODO iOS 13: Use the Foundation list formatter.")
```
typfel: If we don't want warning for this let's rewrite as:
```
// TODO iOS 13: Use the Foundation list formatter.
```

PR #3696, opened 2019-10-02
Fix: newly created group doesnt open and show up in recent list view

billypchan:
### Issues

After a user creates a new group, the conversation is not opened

### Causes

The new created conversation is not on the conversation list. The conversation screen is dismissed after selected a nil conversation.

### Solutions

When `ConversationDirectoryChangeInfo` contains a `groups` in the list, update the conversation list for non folder enabled mode.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
@@ -664,14 +664,14 @@ extension ConversationListViewModel: ConversationDirectoryObserver {
             reload()
         } else {
             for updatedList in changeInfo.updatedLists {
-                if let kind = self.kind(conversationListType: updatedList) {
+                if let kind = self.kind(of: updatedList) {
                     updateForConversationType(kind: kind)
                 }
             }
         }
     }
 
-    private func kind(conversationListType: ConversationListType) -> Section.Kind? {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -669,15 +669,15 @@ extension ConversationListViewModel: ConversationDirectoryObserver {
                 case .unarchived:
                     kind = .conversations
                 case .contacts:
-                    kind = .contacts
+                    kind = folderEnabled ? .contacts : .conversations
                 case .pending:
                     kind = .contactRequests
                 case .groups:
-                    kind = .group
+                    kind = folderEnabled ? .group : .conversations
```
typfel: This will be easy to forget to update in the future. Can't we just ignore the changes if `folderEnabled = false`?

Maybe add something like after the switch 

```
if folderEnabled && !kind.isOneOf([.conversations, .pending]) {
    continue
}
```

billypchan: This part maps  a`ConversationListType` to a `Kind`. I will extract this part and make it easier to understand.

typfel: It's also unnecessary to update the list if `.groups` and `.contacts` changes when folders are not displayed.

PR #3704, opened 2019-10-07
Fix: update conversation selection behaviour

billypchan:
## What's new in this PR?

Expend the collapsed folder when one of the items is selected.

### DIFF
--- a/Wire-iOS Tests/ConversationList/ConversationListViewModel/ConversationListViewModelTests.swift
+++ b/Wire-iOS Tests/ConversationList/ConversationListViewModel/ConversationListViewModelTests.swift
@@ -228,19 +228,14 @@ final class ConversationListViewModelTests: XCTestCase {
 
     // MARK: - state
     func testThatSectionIsExpendedAfterSelected() {
+        ///GIVEN
         sut.folderEnabled = true
-
         let mockConversation = ZMConversation()
-
         fillDummyConversations(mockConversation: mockConversation)
-
-        ///WHEN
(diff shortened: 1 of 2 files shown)


Wire-iOS Tests/ConversationList/ConversationListViewModel/ConversationListViewModelTests.swift
```diff
@@ -227,6 +227,26 @@ final class ConversationListViewModelTests: XCTestCase {
...
     // MARK: - state
+    func testThatSectionIsExpendedAfterSelected() {
+        sut.folderEnabled = true
+
+        let mockConversation = ZMConversation()
+
+        fillDummyConversations(mockConversation: mockConversation)
+
+        ///WHEN
+        sut.setCollapsed(sectionIndex: Int(sectionGroups), collapsed: true)///todo
```
typfel: A test should preferably only have one `WHEN` following:
```
GIVEN
WHEN
THEN
```

I would move this code to the `GIVEN` section.

Wire-iOS Tests/ConversationList/ConversationListViewModel/ConversationListViewModelTests.swift
```diff
@@ -227,6 +227,26 @@ final class ConversationListViewModelTests: XCTestCase {
...
+
+        ///THEN
+        XCTAssert(sut.collapsed(at: Int(sectionGroups)))
```
typfel: Similar here I'd move this into the given section or remove the assertion if we already have another test which test that `setCollapsed` works.

PR #3705, opened 2019-10-07
Fix: do not show favorite item for archived conversation

billypchan:
## What's new in this PR?

Do not show favorite item for archived conversation.
Removed duplicated snapshot tests.


typfel:
Hmm I'm not sure the tests are duplicated since we do not show the same menu items in the list and in the group details.

billypchan:
> Hmm I'm not sure the tests are duplicated since we do not show the same menu items in the list and in the group details.

Yes, I found the items are different, I will undo it.

PR #3706, opened 2019-10-07
Feature: folder list and creation

nicorsm:
## What's new in this PR?

PoC for folder list (through `FolderPickerViewController`) and creation (through `FolderCreationController`).

PR #3707, opened 2019-10-08
Feature: create ConversationInterface

billypchan:
## What's new in this PR?

Add moving to / removing from folder action item to action sheet.
Created `ConversationInterface` as an interface of `ZMConversation` to mock its property easily. 
Update Swift snapshot helper method to snapshot a `UIAlertViewController` with `actionsheet` style.

Add test `testForActionMenu_removeFolder` to test new Swift `MockConversation`. 

### TODO
- [x] wait for https://github.com/wireapp/wire-ios/pull/3706 for the "move to folder" action
- [ ] retire objc `MockConversation` (will keep both for now.)
- [ ] The classes in `ConversationInterface` should be replaced with interface, e.g. `UserType` instead of `ZMUser`


typfel:
Can we do the conversation interface and the snapshot testing in separate PR? We can do it properly later by exposing a ConversationType from data model.

Wire-iOS Tests/ConversationList/Container/ConversationActionControllerSnapshotTests.swift
```diff
@@ -0,0 +1,87 @@
...
+
+    var isReadOnly: Bool = false
+
+    var isFavorite: Bool = false
+
+    var mutedMessageTypes: MutedMessageTypes = .none
+
+    var activeParticipants: Set<ZMUser> = []
+
+    var folderName: String?
```
billypchan: TODO: use `folder` property after data model is updated

PR #3711, opened 2019-10-09
Fix: create a way for qa to see what conversation list is active

billypchan:
## What's new in this PR?

Set `ConversationListContentController.collectionView.accessibilityValue` to ~~the set of expended folder's raw value.~~ show the folder enabled state.


typfel:
I don't see how this solves the problem QA is asking for. As I understood it they want know if the the user have the folder list or the recent list selected.

PR #3712, opened 2019-10-09
Feature: support folders in the conversation list

typfel:
## What's new in this PR?

Add support for displaying custom folders in the conversation list.

## Notes

Since the `Section.Kind` now contains a associated value with the introduction of `.folder(LabelType)` it can no longer confirm to `Codable`, it also conceptually doesn't work since we don't want to persist the `LabelType`. To still be able to persist the collapse state the `Section.Kind` now exposes an `identifier` String instead.


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -208,13 +243,12 @@ final class ConversationListViewModel: NSObject {
...
-
-    /// get section's name in raw value
+    /// Get section's title
     ///
     /// - Parameter sectionIndex: section index of the collection view
-    /// - Returns: raw value of the section's type
+    /// - Returns: localizated title
     func sectionName(of sectionIndex: Int) -> String? {
-        return kind(of: sectionIndex)?.rawValue
+        return kind(of: sectionIndex)?.title
```
billypchan: We should return a non-localized value here since it is used for a cell's accessibility identifier.

PR #3720, opened 2019-10-11
Fix: Update folder copy 

billypchan:
Update text copy


typfel:
@billypchan  snapshots tests failed

PR #3722, opened 2019-10-11
Fix: iPad - app crash when sharing an image from full screen viewer

billypchan:
### Issues

App crashes when sharing an image form image viewer on iPad.

### Causes

The share popover is not configured. 

### Solutions

Config the popover before presenting


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController+MessageActionResponder.swift
```diff
@@ -37,8 +40,15 @@ extension FullscreenImageViewController {
 extension FullscreenImageViewController: MessageActionResponder {
     public func perform(action: MessageAction, for message: ZMConversationMessage!, view: UIView) {
         switch action {
-        case .forward,
-             .showInConversation,
+        case .forward:
+            if isIPadRegular() {
```
typfel: Why add this logic here when it's seemingly also covered later in:

```
///Do not dismiss Modal for forward since share VC is present in a popover
         let shouldDismissModal = action != .delete && action != .copy &&
             !(action == .forward && isIPadRegular())
```

billypchan: clean up as suggested.

PR #3728, opened 2019-10-14
convert PassthroughWindow to Swift

billypchan:
## What's new in this PR?

convert `PassthroughWindow` to Swift


Wire-iOS/Sources/UserInterface/Components/Views/PassthroughWindow.swift
```diff
@@ -1,42 +1,41 @@
...
-// 
+//
 
+import UIKit
+class PassthroughWindow: UIWindow {
 
-#import "PassthroughWindow.h"
-
+    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
+        for v in subviews {
```
typfel: Let's not use one letter variable names and write `view` instead of `v`.

PR #3732, opened 2019-10-15
fix: Longpress on external link formatted with markdown - 2 things are opened simultaneously

billypchan:
### Issues

On iOS 13.1, both menu and "Visit link" alert appear when long pressed a link.

### Causes

same as https://github.com/wireapp/wire-ios/pull/3714

### Solutions

Show the alert only if when the tap gesture has `.ended` state


Wire-iOS/Sources/UserInterface/Components/Views/LinkInteractionTextView.swift
```diff
@@ -100,33 +100,45 @@ extension LinkInteractionTextView: UITextViewDelegate {
...
-                // data detector links should be handle by the system
-                return dataDetectedURLSchemes.contains(URL.scheme ?? "") || !(interactionDelegate?.textView(self, open: URL) ?? false)
+                needFixForRepeatedGesture = false
             } else if #available(iOS 13.0, *) {
+                needFixForRepeatedGesture = true
+            } else {
+                needFixForRepeatedGesture = false
+            }
+
+            let closure: () -> Bool = {
```
typfel: How about a more descriptive name for this closure, something like `performLinkInteraction`

Wire-iOS/Sources/UserInterface/Components/Views/LinkInteractionTextView.swift
```diff
@@ -100,33 +100,45 @@ extension LinkInteractionTextView: UITextViewDelegate {
...
         switch interaction {
         case .invokeDefaultAction:
             
             guard !UIMenuController.shared.isMenuVisible else {
                 return false // Don't open link/show alert if menu controller is visible
             }
 
-            // if alert shown, link opening is handled in alert actions
-            if showAlertIfNeeded(for: URL, in: characterRange) { return false }
+            let needFixForRepeatedGesture: Bool
```
typfel: 👍

PR #3733, opened 2019-10-15
Use difference kit to update conversation list

billypchan:
## What's new in this PR?

Use DifferentKit to update collection view.

### todo
- [ ] retire ZMChangedIndexes/ZMOrderedSetState https://github.com/wireapp/wire-ios-data-model/pull/803


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -185,6 +186,34 @@ final class ConversationListViewModel: NSObject {
     // Local copies of the lists.
     private var sections: [Section] = []
 
+    typealias DiffKitSection = ArraySection<Int, SectionItem>
+
+    /// make items has different hash in different sections
+    struct SectionItem: Hashable, Differentiable {
```
typfel: Can't make Section conform to `DifferentiableSection` and `ZMConversation` to `Differentiable` instead of introducing a new container?

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -511,16 +540,8 @@ final class ConversationListViewModel: NSObject {
...
-    private func changedIndexes(oldConversationList: [AnyHashable],
-                                newConversationList: [AnyHashable]) -> ZMChangedIndexes? {
-        let startState = ZMOrderedSetState(orderedSet: NSOrderedSet(array: oldConversationList))
-        let endState = ZMOrderedSetState(orderedSet: NSOrderedSet(array: newConversationList))
-        let updatedState = ZMOrderedSetState(orderedSet: [])
-        
-        return ZMChangedIndexes(start: startState, end: endState, updatedState: updatedState, moveType: ZMSetChangeMoveType.uiCollectionView)
-    }
-    
+    ///TODO: retire ZMChangedIndexes/ZMOrderedSetState?
```
typfel: Yes!

Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
billypchan: The reason of creating this new container is the same `ZMConversation` object may exists in different section, e.g. in favourite and group folder. If we do not consider `section` number as part of the  `Differentiable`, it may cause wrong result when calculating change set.

typfel: I think DifferenceKit should support an element existing two separate sections.

typfel: "Supports both linear and sectioned collection even if contains duplicates"

billypchan: @typfel confirmed the `SectionItem` is necessary to make `StagedChangeset` calculation correct. For example, when we want to collapse a folder, without this container the changeset would be:
```
   /// StagedChangeset result: move a favourite item to people folder, weired moving animation will start
        elementMoved: [
            (source: [element: 0, section: 1], target: [element: 0, section: 3])
        ] 
```

with the `SectionItem` which provides different hash for the same `ZMConversation` object in different folders, it can give a correct result:
```
       /// we just want the favorite item to be removed when collapsed.
        elementDeleted: [
            [element: 0, section: 1]
        ]
```

typfel: Hmm want experiment a bit with this well well but I can imagine that using an `Int` as the identifier for a section is problematic.

`ArraySection<Int, SectionItem>`

For example:

```
(section: 1, elements: [a,b,c])
(section: 2, elements: [d,e,f])
(section: 3, elements: [g,h,i])
```

If section 2 goes away it will look like all the elements in in section 3 moved to section 2 and section 3 disappeared.

```
(section: 1, elements: [a,b,c])
(section: 2, elements: [g,h,i])
```

--- 

Generally I would like to to not create an intermediate structure just for calculating the diff. I would much prefer adapt our existing structure`[Section]` to implement the `DifferentiableSection` and `Differentiable`.

Also note that the `Section` already have an identifier we can use .

billypchan:
continue in https://github.com/wireapp/wire-ios/pull/3742

PR #3736, opened 2019-10-17
fix: Missing accessibility ID for obfuscated elements(audio, video, location)

billypchan:
### Issues

When a message cell is obfuscated, the accessibility identifier is not updated

### Causes

It was defined in `ObfuscationView` but now it is not accessible. 

### Solutions

Define `accessibilityIdentifier` in `ConversationMessageCellDescription`. 
In `ConversationMessageCellTableViewAdapter`, instead of one-time setting `accessibilityLabel` and `accessibilityIdentifier`, override them to always getting the updated value.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Rich Content/ConversationImageMessageCell.swift
```diff
@@ -152,7 +152,10 @@ class ConversationImageMessageCellDescription: ConversationMessageCellDescriptio
     let supportsActions: Bool = true
     let containsHighlightableContent: Bool = true
     
-    let accessibilityIdentifier: String? = "ImageCell"
+    var accessibilityIdentifier: String? {
+        return configuration.isObfuscated ? "Obfuscated image message" : "ImageCell"
```
typfel: To keep things consistent how about naming it `ObfuscatedImageCell` same applied for the other cell as well.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Rich Content/ConversationImageMessageCell.swift
```diff
@@ -30,7 +30,7 @@ class ConversationImageMessageCell: UIView, ConversationMessageCell {
     
     private var containerView = UIView()
     private var imageResourceView = ImageResourceView()
-    private let obfuscationView = ObfuscationView(icon: .photo)
+    private lazy var obfuscationView = ObfuscationView(icon: .photo)
```
typfel: Why change this property to be lazy? Seems unrelated to the PR.

PR #3739, opened 2019-10-17
fix: hidden notificationsWindow without using makeKey

billypchan:
### Issues

In https://github.com/wireapp/wire-ios/pull/3730, we found that the other window in front of main window overrides status bar style and hidden.

### Causes

The notification windows and call windows always exist and cover the main window

### Solutions

Set empty windows hidden instead of calling `makeKey` to prevent it overrides the status bar.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
@@ -26,20 +26,16 @@ extension Notification.Name {
     static let appUnlocked = Notification.Name("AppUnlocked")
 }
 
-@objcMembers final class AppLockViewController: UIViewController {
+final class AppLockViewController: UIViewController {
     fileprivate var lockView: AppLockView!
     fileprivate var localAuthenticationCancelled: Bool = false
     fileprivate var localAuthenticationNeeded: Bool = true
 
     fileprivate var dimContents: Bool = false {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -26,20 +26,16 @@ extension Notification.Name {
...
-            self.view.isHidden = !self.dimContents
-                        
-            if dimContents {
-                AppDelegate.shared().notificationsWindow?.makeKey()
-            } else {
-                AppDelegate.shared().window.makeKey()
-            }
+            view.isHidden = !dimContents
+
+            AppDelegate.shared().notificationsWindow?.isHidden = !dimContents
```
typfel: Can't you access the window from the view instead? `view.window?.isHidden  = !dimContents`? 

Isn't unnecessary to hide/unhide the view when we now do it on the window level.

What about he `CallWindow` don't we a have a similar problem there?

billypchan: The `CallWindow` has another issue and will be fixed in separated PR.

typfel:
Have you tested this on real device? Does it still work if you need enter your pin code with the keyboard?

billypchan:
Yes, tested with device and works

PR #3741, opened 2019-10-18
App Center Support

nicorsm:
### Issues

Hockey is saying goodbye, so it's time to move to App Center.

### Solutions

1. Replaced Hockey APIs with respective counterparts on App Center
2. Replaced old Hockey references with new App Center

### TODO

Enable App Center on extensions after this gets solved https://github.com/microsoft/appcenter-sdk-apple/issues/1819

PR #3748, opened 2019-10-22
Fix: retire STRINGIZE

billypchan:
### Issues

`STRINGIZE` preprocessor macro is not available in Swift.

### Solutions

Add `Wire SSO URL Scheme` key to plist file and access it from `Bundle` extension.


Wire-iOS/Wire-Info.plist
```diff
@@ -172,5 +172,7 @@
 	</array>
 	<key>WireGroupId</key>
 	<string>${WIRE_BUNDLE_ID}</string>
+	<key>Wire SSO URL Scheme</key>
+	<string>$(WIRE_SSO_URL_SCHEME)</string>
```
nicorsm: I was thinking about moving App Center ID there too. Furthermore, App Center requires to insert the App Center ID inside this file, so we can definitively get rid out of the preprocessor files.

billypchan: @nicorsm All ID should be stored in a private repo. `WIRE_SSO_URL_SCHEME` is refer to a value in an external `.xcconfig` file.

nicorsm: Yeah sure, I meant the reference to the ID, like you did with `WIRE_SSO_URL_SCHEME`. It's temporarily described as `appcenter-${APP_SECRET}` here https://github.com/wireapp/wire-ios/pull/3741/files

PR #3777, opened 2019-11-05
Updated UI for the badges if their values are more than 99

KaterinaNikolaenko:
## What's new in this PR?

Changed badge label value in the ConversationListHeaderView if badge number is > 99.


billypchan:
@KaterinaNikolaenko the snapshot should be updated for this change.

PR #3782, opened 2019-11-07
Fix avs crash

David-Henner:
### Issues

Joining a group call from a second device crashes 

### Solutions

Use CallParticipant instead of UUID & update the logic responsible for updating the participants & participants states


Wire-iOS Tests/CallHapticsControllerTests.swift
```diff
@@ -31,20 +31,30 @@ fileprivate class MockCallHapticsGenerator: CallHapticsGeneratorType {
     }
 }
 
-final class CallHapticsControllerTests: XCTestCase {
+final class CallHapticsControllerTests: ZMSnapshotTestCase {
```
billypchan: These are no snapshot tests below, any reason for changing this?

David-Henner: Nice catch!
I've used it to get easy access to the managed object context, and forgot to update it in the end

johnxnguyen: Discussed 1:1. For now it's ok to leave it like this. We made a ticket to refactor the test classes to provide a context for normal unit tests.

billypchan: You may use `CoreDataFixture` class which extracted the methods in `ZMSnapshotTestCase `

PR #3790, opened 2019-11-11
Fix: crash in UITextView replace method

KaterinaWire:
Fix: crash in UITextView replace method

### Issues
Crash when replacing text in TextView.

### Causes
Replacing text with a given range that >  string length.

### Solutions
Add a range length check.

### DIFF
--- a/Wire-iOS Tests/UITextView+ReplaceTests.swift
+++ b/Wire-iOS Tests/UITextView+ReplaceTests.swift
@@ -44,7 +44,7 @@ class UITextView_ReplaceTests: XCTestCase {
         sut.replace(range, withAttributedText: attributedText!)
         
         // THEN
-        XCTAssertEqual(sut.text, text, "Should have been a success")
+        XCTAssertEqual(sut.text, text)
     }
     
     func testRangeIsInsideTheReplacementString() {
@@ -58,7 +58,7 @@ class UITextView_ReplaceTests: XCTestCase {
         sut.replace(range, withAttributedText: attributedText!)
         
         // THEN
-        XCTAssertEqual(sut.text, "12345678🐶5678🐶", "Should have been a success")
+        XCTAssertEqual(sut.text, "12345678🐶5678🐶")
     }
     
     func testGivenRangeIsOutsideTheWholeRange() {
@@ -72,6 +72,6 @@ class UITextView_ReplaceTests: XCTestCase {
         sut.replace(range, withAttributedText: attributedText!)
         
         // THEN
-        XCTAssertEqual(sut.text, text, "Should have been a success")
+        XCTAssertEqual(sut.text, text)
     }
 }
--- a/Wire-iOS Tests/UITextView+ReplaceTests.swift
+++ b/Wire-iOS Tests/UITextView+ReplaceTests.swift
@@ -1,0 +1,63 @@
+//
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class UITextView_ReplaceTests: XCTestCase {
+
+    var sut: UITextView!
+    
+    override func setUp() {
+        super.setUp()
+        sut = UITextView()
+    }
+
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+    
+    func testRangeIsOutsideTheReplacementString() {
+        // GIVEN
+        let text = "12345678"
+        sut.text = text
+        let attributedText = sut.attributedText
+        let range = NSRange(location: 0, length: 24)
+        
+        // WHEN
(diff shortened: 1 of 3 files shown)


billypchan:
It would be better to name the branch as `fix/ZIOS-11793-App-crash` to indicate it is a fix.

Wire-iOS/Sources/UserInterface/Components/Views/UITextView+Replace.swift
```diff
@@ -20,6 +20,9 @@ import Foundation
 
 extension UITextView {
     func replace(_ range: NSRange, withAttributedText replacement: NSAttributedString) {
+        guard range.length <= self.text.count else {
```
johnxnguyen: The `length` of an `NSString` is not always the same as the `count` of a `String` (see [here](https://www.logcg.com/en/archives/3253.html)). It's a little "gotcha" that has caused a lot of range issues in the past 😄 .

Also, I would expect that the only valid value of `range` is a range that is entirely contained by the `self.attributedText.wholeRange`.

KaterinaWire: I've changed this part. Now I use `NSMaxRange` to find the end of the ranges. So, if the end of the given `range` is outside of `self.attributedText.wholeRange`, it will not replace the string.

billypchan:
I would suggest the title of this PR "Fix: crash in UITextView replace method"

Wire-iOS Tests/UITextView+ReplaceTests.swift
```diff
@@ -0,0 +1,63 @@
...
+        sut = nil
+        super.tearDown()
+    }
+    
+    func testRangeIsOutsideTheReplacementString() {
+        // GIVEN
+        let text = "12345678"
+        sut.text = text
+        let attributedText = sut.attributedText
+        let range = NSRange(location: 0, length: 24)
```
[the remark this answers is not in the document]
KaterinaWire: I added this case to another method `testGivenRangeIsOutsideTheWholeRange()` (NSRange(location: 4, length: 6)).

Wire-iOS Tests/UITextView+ReplaceTests.swift
```diff
@@ -0,0 +1,63 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class UITextView_ReplaceTests: XCTestCase {
```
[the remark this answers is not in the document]
KaterinaWire: Done

Wire-iOS Tests/UITextView+ReplaceTests.swift
```diff
@@ -0,0 +1,77 @@
...
+        let text = "12345678🐶"
+        sut.text = text
+        let attributedText = sut.attributedText
+        let range = NSRange(location: 0, length: 4)
+        
+        // WHEN
+        sut.replace(range, withAttributedText: attributedText!)
+        
+        // THEN
+        XCTAssertEqual(sut.text, "12345678🐶5678🐶", "Should have been a success")
```
johnxnguyen: I don't think the assertion message here is necessary.

Wire-iOS Tests/UITextView+ReplaceTests.swift
```diff
@@ -0,0 +1,77 @@
...
+        let text = "12345678"
+        sut.text = text
+        let attributedText = sut.attributedText
+        let range = NSRange(location: 4, length: 6)
+        
+        // WHEN
+        sut.replace(range, withAttributedText: attributedText!)
+        
+        // THEN
+        XCTAssertEqual(sut.text, text, "Should have been a success")
```
johnxnguyen: I don't think the assertion message here is necessary.

Wire-iOS/Sources/UserInterface/Components/Views/UITextView+Replace.swift
```diff
@@ -20,6 +20,9 @@ import Foundation
 
 extension UITextView {
     func replace(_ range: NSRange, withAttributedText replacement: NSAttributedString) {
+        guard NSMaxRange(self.attributedText.wholeRange) >= NSMaxRange(range)  else {
```
johnxnguyen: `NSMaxRange(range)` can be replaced with `range.upperBound`, which (I think) reads better.

johnxnguyen: Also, I think it would be safer to check the lower bounds too, like:

```swift
guard 
  range.lowerBound >= attributedText.wholeRange.lowerBound,
  range.upperBound <= attributedText.wholeRange.upperBound
  else { return }
```

Wire-iOS Tests/UITextView+ReplaceTests.swift
KaterinaWire: The assertion message has been deleted.

Wire-iOS Tests/UITextView+ReplaceTests.swift
KaterinaWire: The assertion message has been deleted.

Wire-iOS/Sources/UserInterface/Components/Views/UITextView+Replace.swift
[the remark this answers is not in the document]
KaterinaWire: Is it still relevant?

Wire-iOS/Sources/UserInterface/Components/Views/UITextView+Replace.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Components/Views/UITextView+Replace.swift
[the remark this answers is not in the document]
billypchan: It is fine now.

PR #3795, opened 2019-11-13
Feature/add selfUser to the participants in the conversation

KaterinaWire:
## What's new in this PR?

Add ZMUser.selfUser() to the participants list.


billypchan:
The merge target should be `feature/conversation-roles` instead of `develop`

PR #3796, opened 2019-11-14
[Fix] Permission alerts not showing

David-Henner:
### Issues

For the case of microphone permission:

GIVEN microphone permissions are revoked
WHEN user accepts an incoming call
THEN call overlay is dismissed without asking user to grant microphone permissions

### Causes

The logic responsible for displaying the alert asking the user to grant permissions failed to make the `notificationsWindow` visible

### Solutions

- Make `notificationsWindow` visible when displaying permissions alert
- Hide `notificationsWindow` upon completion of the alert

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
+++ b/Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
@@ -21,6 +21,7 @@
 #import "AppDelegate.h"
 #import "UIAlertController+Wire.h"
 #import "UIResponder+FirstResponder.h"
+#import "Wire-Swift.h"
 
 @import Photos;
 #import <AVFoundation/AVFoundation.h>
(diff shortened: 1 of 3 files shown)


johnxnguyen:
Do you think we also need to adjust the logic for the other permission alerts?

billypchan:
I would recommend showing the alert on the main window instead of `notificationsWindow`.  In this PR https://github.com/wireapp/wire-ios/pull/3739 we make `notificationsWindow` not always visible. The `NotificationWindow` is used for showing `AppLock` screen only and the permission alert should not cover the dialog appear on that.

Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
```diff
@@ -134,14 +135,35 @@ + (void)wr_warnAboutMicrophonePermission
 {
     UIAlertController *noMicrophoneAlert = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"voice.alert.microphone_warning.title", nil)
```
billypchan: You may use factory method `UIAlertController.alertWithOKButton` to save some boilplate code.

David-Henner: Done in c84837f

David-Henner:
@billypchan Updated based on your comments

billypchan:
@David-Henner We should fix similar issue for other alerts.

David-Henner:
@billypchan Right,
I wasn't sure if that was in the scope.
Are we fine to do that in the same PR?

PR #3800, opened 2019-11-17
Fix: convert Settings+ColorScheme to Swift

billypchan:
## What's new in this PR?

Convert `Settings+ColorScheme` helper to Swift.
Refactor `Settings.shared` to non-nil property to reduce unwrapping.


Wire-iOS/Settings+ColorScheme.swift
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @objc
+    var defaults: UserDefaults {
+        return UserDefaults.standard
+    }
+
+    var colorScheme: SettingsColorScheme? {
+        get {
+            guard let string = defaults.string(forKey: UserDefaultColorScheme) else { return nil }
```
David-Henner: Just a thought, would like to know your opinion:
Since the default `colorSchemeVariant` seems to be `.light` 
Could we set it in the `userDefaults` if it happens to be `nil` ?
Rather than doing `colorScheme.variant = Settings.shared().colorScheme?.colorSchemeVariant ?? .light` in the `AppRootViewController`

Wire-iOS/Sources/Managers/ColorSchemeController.swift
```diff
@@ -45,14 +45,9 @@ class ColorSchemeController: NSObject {
...
-        let colorScheme = ColorScheme.default
-        switch Settings.shared().colorScheme {
-        case .light:
-            colorScheme.variant = .light
-        case .dark:
-            colorScheme.variant = .dark
-        }
+    @objc
+    func settingsColorSchemeDidChange(notification: Notification?) {
+        ColorScheme.default.variant = Settings.shared().colorScheme?.colorSchemeVariant ?? .light
```
David-Henner: Referring to my previous comment, if implemented, we could do 
`ColorScheme.default.variant = Settings.shared().colorScheme.colorSchemeVariant` and thus avoid duplicating the `?? .light` logic

Wire-iOS/Settings+ColorScheme.swift
billypchan: Nice catch, I will charge this to non optional

PR #3812, opened 2019-11-19
Convert ProfileViewController to MVVM

billypchan:
## What's new in this PR?

Convert `ProfileViewController` to MVVM pattern.
Convert callers of `ProfileViewController` to Swift.


Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
```diff
@@ -207,18 +207,27 @@ extension GroupDetailsViewController {
...
     }
     
+    func suggestedBackButtonTitle(for controller: ProfileViewController?) -> String? {
+        //no-op
+        return nil
+    }
+    
+    func profileViewController(_ controller: ProfileViewController?, wantsToCreateConversationWithName name: String?, users: Set<ZMUser>) {
+        //no-op
+    }
```
David-Henner: Is it possible to avoid duplication of those methods (also defined in `MessageDetailsContentViewController`, `ZClientViewController`...) with a default protocol implementation ?

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewController.swift
```diff
@@ -0,0 +1,534 @@
...
+        presentAlert(controller, targetView: targetView)
+    }
+    
+    // MARK: Delete Contents
+    
+    private func presentDeleteConfirmationPrompt(from targetView: UIView) {
+        guard let conversation = viewModel.conversation else { return }
+        
+        let controller = UIAlertController(title: ClearContentResult.title, message: nil, preferredStyle: .actionSheet)
+        ClearContentResult.options(for: conversation) .map { $0.action(viewModel.handleDeleteResult) }.forEach(controller.addAction)
```
David-Henner: typo space before `.map` ?

Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
billypchan: originally `ProfileViewControllerDelegate` has three optional methods.  Since Swift does not allow optional protocol methods I would leave it like this. 
The better solution would be split `ProfileViewControllerDelegate` into 3 protocols and just conform to the needed protocols.

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController.m
```diff
@@ -380,11 +379,7 @@ - (UIViewController *)participantsController
         case ZMConversationTypeOneOnOne:
         case ZMConversationTypeConnection:
         {
-            viewController = [UserDetailViewControllerFactory createUserDetailViewControllerWithUser:self.conversation.firstActiveParticipantOtherThanSelf
-                                          conversation:self.conversation
-                         profileViewControllerDelegate:self
-                               viewControllerDismisser:self];
-
+            viewController = [self createUserDetailViewControllerWithUser];
```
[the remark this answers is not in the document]
billypchan: The reason of this change is one of the original parameter is changed to Swidt only.
I will shorten it to `createUserDetailViewController`

Wire-iOS/Sources/UserInterface/GroupDetails/GroupDetailsViewController.swift
David-Henner: Wouldn't this work ?

```
protocol ProtocolName {
   func nonOptionalFunc()
   func optionalFunc()
}

extension ProtocolName {
   func optionalFunc() {
      // default implementation
   }
}
```

billypchan: I would perfer split this protocol into 2 since only one class implements `suggestedBackButtonTitle`

PR #3816, opened 2019-11-21
Feature: Change the accessibilityIdentifier for UserCell

KaterinaWire:
## What's new in this PR?

Change the accessibilityIdentifier for the UserCell according to which section (admins/members) it's in.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
@@ -27,7 +27,7 @@ extension UIImageView {
     }
 }
 
-class UserCell: SeparatorCollectionViewCell {
+class UserCell: SeparatorCollectionViewCell, SectionListCellType {
 
     var hidesSubtitle: Bool = false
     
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
```diff
@@ -63,6 +63,19 @@ class UserCell: SeparatorCollectionViewCell {
...
+        let prefix: String
+        
+        if let sectionName = sectionName {
+            prefix = "\(sectionName) - "
+        } else {
+            prefix = ""
+        }
+        
+        return prefix + "participants.section.participants.cell"
+    }
```
billypchan: Instead of copying the default implementation of `SectionListCellType`, we can create one new member property say `cellIdentifier` in `SectionListCellType` and create the `identifier` with `return prefix + cellIdentifier`

Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ParticipantsSectionController.swift
```diff
@@ -68,6 +68,17 @@ private struct ParticipantsSectionViewModel {
...
+    var accessibilityTitle: String {
+        switch teamRole {
+        case .member:
+            return "Members"
+        case .admin:
+            return "Admins"
+        default:
+            return ""
+        }
+    }
```
billypchan: It would be better to refector this as an extension property of `TeamRole` and we can resue it.

Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
KaterinaWire: Updated

Wire-iOS/Sources/UserInterface/GroupDetails/Sections/ParticipantsSectionController.swift
KaterinaWire: I created a new extension for `TeamRole`

PR #3817, opened 2019-11-21
Feature: Rename section titles

KaterinaWire:
## What's new in this PR?

Rename section titles: "Conversation Admins/Members" to "Group Admins/Members".


billypchan:
Some snapshots need to be updated?

PR #3818, opened 2019-11-21
Feature/status propagation

David-Henner:
## What's new in this PR?

Hide the other users status if team size is bigger than the optimal limit


Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
-github "microsoft/appcenter-sdk-apple" "2.4.0"
+github "microsoft/appcenter-sdk-apple" "2.5.1"
```
billypchan: I would suggest not to update it for this release.

Wire-iOS Tests/ZMUserAvailabilityTests.swift
```diff
@@ -0,0 +1,113 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+
+class ZMUserAvailabilityTests: XCTestCase {
```
billypchan: `ZM` prefix is no longer necessary

billypchan:
```suggestion
final class UserAvailabilityTests: XCTestCase {
```

David-Henner: It's testing `ZMUser+Availability`, I think it's easier to look up if there are any tests for the extension by keeping the naming consistent.
Which in this case would mean also updating the `ZMUser+Availability` to `User+Availability`.
Would you agree to this change ?

Wire-iOS/Sources/Helpers/UserType+Availability.swift
```diff
@@ -0,0 +1,26 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension UserType {
+    var shouldHideAvailability: Bool {
```
billypchan: This should be moved to Data Model and we can move the test to there also

Wire-iOS/Sources/Helpers/ZMUser+Availability.swift
```diff
@@ -0,0 +1,38 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireDataModel
+
+extension ZMUser {
+    var shouldHideAvailability: Bool {
```
billypchan: This can be also moved to Data Model, we can do it after release

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -261,7 +261,7 @@ class ProfileHeaderViewController: UIViewController, Themeable {
     }
     
     private func updateAvailabilityVisibility() {
-        availabilityTitleViewController.view?.isHidden = options.contains(.hideAvailability) || !options.contains(.allowEditingAvailability) && user.availability == .none
+        availabilityTitleViewController.view?.isHidden = options.contains(.hideAvailability) || !options.contains(.allowEditingAvailability) && user.availability == .none || user.shouldHideAvailability
```
billypchan:
```suggestion
        let isHidden: Bool
        if user.shouldHideAvailability {
          isHidden = true
        } else {
          isHidden =  availabilityTitleViewController.view?.isHidden = options.contains(.hideAvailability) || !options.contains(.allowEditingAvailability) && user.availability == .none
        }
        
        availabilityTitleViewController.view?.isHidden = isHidden
```

billypchan: I would break it down to reduce the run time and easier to understand.

billypchan:
@David-Henner we can do the refactoring in another PR.

PR #3820, opened 2019-11-22
Feature: UI toggle conversation admin on/off in participant profile

billypchan:
## What's new in this PR?

Add a switch to toggle group admin setting. A new class `IconToggleSubtitleCell` is created.
Refactor `ProfileViewControllerTests` & `ProfileDetailsViewControllerTests` with new snapshot framewrok.

### TODO
- [ ] self user can manage his/her admin role?
- [ ] the default group admin setting should be on


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -716,6 +716,8 @@
 "participants.section.admins.footer" = "There are no admins.";
 "participants.section.members.footer" = "There are no members.";
 
+"profile.profile.group_admin_options.title" = "Group admin";
+
```
KaterinaWire: I added such string to the previous PR, so it might be a conflict.

Wire-iOS/Sources/UserInterface/UserProfile/Details/Cells/IconToggleSubtitleCell.swift
```diff
@@ -0,0 +1,118 @@
...
+    func configure(with configuration: CellConfiguration, variant: ColorSchemeVariant) {
+        guard case let .iconToggle(title,
+                                   subtitle,
+                                   identifier,
+                                   titleIdentifier,
+                                   icon,
+                                   color,
+                                   get,
+                                   set) = configuration else { preconditionFailure() }
+        let mainColor = color ?? UIColor.from(scheme: .textForeground, variant: variant)
```
KaterinaWire: Can we get mainColor as variant.mainColor()?

Wire-iOS/Resources/Base.lproj/Localizable.strings
billypchan: The title of the toggle cell and status label should be seperated in two cases since we may change one of them in the future.

Wire-iOS/Sources/UserInterface/Conversation Options/IconActionCell.swift
```diff
@@ -69,11 +69,16 @@ final class IconActionCell: UITableViewCell, CellConfigurationConfigurable {
...
         label.text = title
         separator.backgroundColor = UIColor.from(scheme: .cellSeparator, variant: variant)
     }
 }
 
+extension ColorSchemeVariant{
+    func mainColor(color: UIColor?) -> UIColor {
+        return color ?? UIColor.from(scheme: .textForeground, variant: self)
+    }
+}
```
KaterinaWire: Would it be better to put this extension in a separate file and move it to Sources -> Helpers?

billypchan: moved to `ColorScheme.swift`

PR #3821, opened 2019-11-22
Feature: Add Admin label to participant's profile

KaterinaWire:
## What's new in this PR?

Add Admin label and icon to participant's profile.


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -835,6 +835,7 @@
 // User Details
 "profile.details.title" = "Details";
 "profile.details.guest" = "Guest";
+"profile.details.group_admin" = "GROUP ADMIN";
```
billypchan: We can keep this in lower case and make it upper case in the code.

Wire-iOS/Sources/UserInterface/Components/Views/GroupRoleLabelIndicator.swift
```diff
@@ -0,0 +1,71 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class GroupRoleLabelIndicator: UIStackView, Themeable {
```
billypchan: Do we have snapshot for this class?

Wire-iOS/Resources/Base.lproj/Localizable.strings
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/Components/Views/GroupRoleLabelIndicator.swift
KaterinaWire: I added a new test for this label

billypchan:
I would recommend refactoring exist `GuestLabelIndicator` to `LabelIndicator` and init with `icon` and `title` parameter instead of creating a new `GroupRoleLabelIndicator` with similar code.

PR #3826, opened 2019-11-25
Fix/after entering exact name of user the other section is still visible

KaterinaWire:
### Issues
The participant is shown inside own section, but the other section is also still visible with the text "there are no admins." or "there are no members."

### Solutions
Hide section if there are not participants.


billypchan:
`GroupParticipantsDetailViewControllerTests` failed

KaterinaWire:
> `GroupParticipantsDetailViewControllerTests` failed

Done

PR #3827, opened 2019-11-25
[Refactor] Move user availability extension to wire-ios-data-model

David-Henner:
## What's new in this PR?

Moved the shouldHideAvailability property implementation & tests from wire-ios to wire-ios-data-model


David-Henner:
Pending on the dependencies to be updated to use `data-model` `196.0` or higher

PR #3832, opened 2019-11-26
Fix/admin label text color in dark mode

KaterinaWire:
### Issues
The `admin label` is clipped when text size is bigger than normal.

### Solutions
Admin label and icon were moved to `containerView` instead from the `stackView`.


billypchan:
18 tests failed (testDefaultOptions_NoAvailability_Edit, ...)

PR #3834, opened 2019-11-27
Feature/profile screen update a participant's name

KaterinaWire:
## What's new in this PR?

The verified shield was moved right to the userName.
The available status was moved left to the userName.


Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -275,7 +275,7 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
             isHidden = options.contains(.hideAvailability) || !options.contains(.allowEditingAvailability) && user.availability == .none
         }
         
-        availabilityTitleViewController.view?.isHidden = isHidden
+        availabilityTitleViewController.view?.isHidden = true//isHidden
```
billypchan: `availabilityTitleViewController.view` is always hidden?

KaterinaWire: According to the new design spec, we should move the availability to the left of the userName. I left this line of comments because I was not sure about the implementation.

Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileTitleView.swift
```diff
@@ -23,6 +23,7 @@ import Cartography
     
     let verifiedImageView = UIImageView(image: WireStyleKit.imageOfShieldverified)
     let titleLabel = UILabel()
+    var variant: ColorSchemeVariant
```
billypchan: this should be put`public func configure(with user: UserType)`

billypchan:
We should update or create snapshot test for this change.

Wire-iOS/Sources/UserInterface/UserProfile/Views/ProfileTitleView.swift
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
KaterinaWire: It was wrong. I added a new state to check if accessibility should be hidden.
According to the new design spec, we should move the availability to the left of the userName.

billypchan:
The font of the title changed, please verify this with design spec.

KaterinaWire:
> The font of the title changed, please verify this with design spec.

The font of the user name?

KaterinaWire:
> The font of the title changed, please verify this with design spec.

I changed the font as it was before

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -269,7 +269,7 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
     
     private func updateAvailabilityVisibility() {
         let isHidden: Bool
-        if user.shouldHideAvailability {
+        if user.shouldHideAvailability || (conversation != nil) {
```
billypchan: I would recommend creating a context( and a view model) for this VC to handle this complicated logics.

Wire-iOS Tests/ProfileDetailsViewControllerTests.swift
```diff
@@ -20,7 +20,7 @@ import SnapshotTesting
 import XCTest
 @testable import Wire
 
-final class ProfileDetailsViewControllerTests: XCTestCase {
+final class ProfileDetailsViewControllerTests: ZMSnapshotTestCase {
```
billypchan:
```suggestion
final class ProfileDetailsViewControllerTests: XCTestCase {
```

Wire-iOS Tests/DeviceManagement/ProfileViewControllerTests.swift
```diff
@@ -20,7 +20,7 @@ import SnapshotTesting
 import XCTest
 @testable import Wire
 
-final class ProfileViewControllerTests: XCTestCase {
+final class ProfileViewControllerTests: ZMSnapshotTestCase {
```
billypchan:
```suggestion
final class ProfileViewControllerTests: XCTestCase {
```

billypchan:
The snapshots changed for the accent color should be restored.

PR #3844, opened 2019-12-02
Feature: refactoring guest label indicator group role label indicator

KaterinaWire:
## What's new in this PR?

Create a LabelIndicator view and  init method with an icon, title and accessibilityIdentifier  parameters instead of  GuestLabelIndicator and GroupRoleLabelIndicator.


Wire-iOS/Sources/UserInterface/Components/Views/LabelIndicator.swift
```diff
@@ -18,43 +18,45 @@
...
+final class LabelIndicator: UIView {
+    
     private let variant: ColorSchemeVariant
-    private let groupRoleIcon = UIImageView()
+    private let indicatorIcon = UIImageView()
     private let titleLabel = UILabel()
     private let containerView = UIView()
     
-    init() {
+    init(icon: StyleKitIcon, title: String, accessibilityIdentifier: String) {
```
billypchan: We can create a context for `LabelIndicator` having value like `.groupRole` and `.guest`

Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -219,7 +219,6 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
     
     func applyColorScheme(_ variant: ColorSchemeVariant) {
         availabilityTitleViewController.availabilityTitleView?.colorSchemeVariant = variant
-        guestIndicator.colorSchemeVariant = variant
```
billypchan: still needed for variant update?

KaterinaWire: I set in init method `self.variant = ColorScheme.default.variant`

billypchan: This method would be called after color scheme is update, which may happens after `init` is called

KaterinaWire: Add `colorSchemeVariant` to the `LabelIndicator` view.

PR #3848, opened 2019-12-04
Feature: Update UI for the user profile: add status to change group role 

KaterinaWire:
## What's new in this PR?

Add a new state for the `ProfileDetailsViewController` and the method for the `ProfileDetailsContentControllerDelegate protocol` to track the current group role in the UI.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
@@ -169,7 +169,7 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
         guestIndicatorStack.alignment = .center
         
         updateGuestIndicator()
-        updategroupRoleIndicator()
+        updateGroupRoleIndicator()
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -235,14 +235,18 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
...
         if let _ = conversation {
             groupRoleIndicator.isHidden = !user.isAdminGroup
         } else {
             groupRoleIndicator.isHidden = true
         }
     }
     
+    func switchGroupRoleIndicatorToAdmin(_ isAdminRole: Bool) {
+        groupRoleIndicator.isHidden = !isAdminRole
+    }
```
billypchan: We can refactor this to `var isAdminRole` and handle the UI update in the `set` block

KaterinaWire: Fixed

PR #3850, opened 2019-12-04
Update permissions strings

David-Henner:
### Issues

Permissions strings used for camera and microphone were not consistent between the post-install permissions alerts and our "custom" permissions alerts

### Solution

- Used the localised `NS...UsageDescription` from `InfoPlist.strings` instead of the strings defined in `Localizable.strings`


Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
```diff
@@ -103,7 +103,7 @@ + (void)wr_warnAboutCameraPermissionWithCompletion:(dispatch_block_t)completion
     }
     
     UIAlertController *noVideoAlert = [UIAlertController alertWithOKButtonWithTitle:NSLocalizedString(@"voice.alert.camera_warning.title", nil)
-                                                                            message:NSLocalizedString(@"voice.alert.camera_warning.explanation", nil)
+                                                                            message:NSLocalizedStringFromTable(@"NSCameraUsageDescription", @"InfoPlist", nil)
```
billypchan: in `Wire-Info.plist` there is also a copy of `NSCameraUsageDescription` & `NSMicrophoneUsageDescription`, should we also update that copy?

Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
```diff
@@ -103,7 +103,7 @@ + (void)wr_warnAboutCameraPermissionWithCompletion:(dispatch_block_t)completion
     }
     
     UIAlertController *noVideoAlert = [UIAlertController alertWithOKButtonWithTitle:NSLocalizedString(@"voice.alert.camera_warning.title", nil)
-                                                                            message:NSLocalizedString(@"voice.alert.camera_warning.explanation", nil)
```
billypchan: If it is not needed we can remove these strings fro `Localizable.strings(Base)` file and it will be updated in crowdin.

billypchan:
Do we need to check similar cases for photo and contact permission also?

David-Henner:
This is what it looks like, I don't know if it's necessary to have the same text in those cases. I think it's okay to leave it as it is. What do you think @billypchan ?

### Photos

<div>


</div>

### Contacts

<div>


</div>

Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
David-Henner: Done in f466892

Wire-iOS/Sources/Helpers/UIApplication+Permissions.m
David-Henner: Done in e39df36

marcoconti83:
Waiting for next release before merging this

typfel:
@David-Henner  I guess we can merge this now?

PR #3861, opened 2019-12-16
fix: use ZMConversation.defaultAdminRoleName as role name checker

billypchan:
## What's new in this PR?

Use `ZMConversation.defaultAdminRoleName` provided from data-model to identify a role name is an admin or not.


billypchan:
notice: 35 tests failed, e.g. `testForOptionsForNonTeamUser `

PR #3864, opened 2019-12-17
create role helpers

billypchan:
## What's new in this PR?

Create helpers to access a role of a particiant in a conversation.

### DIFF
--- a/Wire-iOS/Sources/Helpers/UserType+ZMUser.swift
+++ b/Wire-iOS/Sources/Helpers/UserType+ZMUser.swift
@@ -41,6 +41,7 @@ extension UserType {
     }
 
     
+    ///TODO: mv to data model
     func isAdminGroup(conversation: ZMConversation?) -> Bool {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/Helpers/UserType+ZMUser.swift
```diff
@@ -45,4 +45,14 @@ extension UserType {
...
+
+    ///TODO: mv to data model
+    func participantRole(in conversation: ZMConversation?) -> ParticipantRole? {
+        return zmUser?.participantRoles.first(where: { $0.conversation == conversation })
+    }
+
+    ///TODO: mv to data model
+    func role(in conversation: ZMConversation?) -> Role? {
+        return participantRole(in: conversation)?.role
+    }
```
johnxnguyen: Why not move these to DM already?

billypchan: I will move them after approval, there are also some other methods should be moved.

PR #3869, opened 2019-12-18
fix: pass searchDirectory to ContactsDataSource.search

billypchan:
## What's new in this PR?

For an unknown reason `ContactsDataSource.search` can not compile on CI sevser due to `searchDirectory` member property is not seen by this Swift extension. Pass it by parameter to try to solve it.

### TODO:
- [ ] convert ContactsDataSource to Swift


marcoconti83:
Which Xcode version is the server using?

billypchan:
@marcoconti83 it is 10.2.1. I should change it to 10.3 to match Azure's version?

PR #3870, opened 2019-12-18
Fix: Hide a groupAdminStatus for the wirelessUser

KaterinaWire:
## What's new in this PR?

Hide a groupAdminStatus if the user is a  wireless guest.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
@@ -152,7 +152,7 @@ final class ProfileDetailsContentController: NSObject,
             ///Do not show group admin toggle for self user or requesting connection user
             var items: [ProfileDetailsContentController.Content] = []
             if let conversation = conversation {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -152,7 +152,7 @@ final class ProfileDetailsContentController: NSObject,
             ///Do not show group admin toggle for self user or requesting connection user
             var items: [ProfileDetailsContentController.Content] = []
             if let conversation = conversation {
-                if (viewer.zmUser?.canModifyOtherMember(in: conversation) ?? false) {
+                if (viewer.zmUser?.canModifyOtherMember(in: conversation) ?? false) && !user.isWirelessUser {
```
johnxnguyen: It would be nice to express these conditions with readable names, such as

```swift
let viewerCanChangeOtherRoles = viewer.zmUser?.canModifyOtherMember(in: conversation) ?? false
let userCanHaveRoleChanged = !user.isWirelessUser
```

PR #3881, opened 2019-12-19
Fix: Hide a "Group admin" label in the Profile screen

KaterinaWire:
## What's new in this PR?

Hide a "Group admin" label in the Profile screen if it's not a group conversation.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
@@ -257,7 +257,7 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
     }
 
     private func updateGroupRoleIndicator() {
-        groupRoleIndicator.isHidden = !(conversation.map(user.isAdminGroup) ?? false)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/SelfProfile/ProfileHeaderViewController.swift
```diff
@@ -257,7 +257,7 @@ final class ProfileHeaderViewController: UIViewController, Themeable {
     }
 
     private func updateGroupRoleIndicator() {
-        groupRoleIndicator.isHidden = !(conversation.map(user.isAdminGroup) ?? false)
+        groupRoleIndicator.isHidden = !(conversation.map(user.isAdminGroup) ?? false) || (conversation?.conversationType != .group)
```
billypchan:
```suggestion
        let groupRoleIndicatorHidden: Bool
        switch conversation?.conversationType {
        case .group?:
            groupRoleIndicatorHidden = !(conversation.map(user.isAdminGroup) ?? false)
        default:
            groupRoleIndicatorHidden = true
        }
        groupRoleIndicator.isHidden = groupRoleIndicatorHidden

```

billypchan: use switch is clearer to read

PR #3882, opened 2019-12-19
bump data model

billypchan:


billypchan:
need to change the partner - guest room creation logic in DM

PR #3890, opened 2019-12-20
restore disabled StartedConversationCellTests

billypchan:
### Issues

When running `StartedConversationCellTests` on the CI server, it crashes randomly and the tests can not pass.

### Causes

We were trying to test different cell appearances by changing a self user's role, which may affect the results of other tests since the clean up changed core data is missing. 

### Solutions

Instead of changing `selfUser`'s role, inject a `SwiftMockUser` and change its properties. 
Create a new `MockSystemMessage` which conforms `SystemMessageNewConversationProperties` and inject it to `ConversationMessageSectionController`

### Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-data-model/pull/863


billypchan:
TODO: refactor failed tests with mock self user injection. Do not write to core data in the tests.

PR #3904, opened 2019-12-30
Fix: Add a conversation change observer to a cell

KaterinaWire:
### Issues
No response when tapping badge item (Reply/Share/Delete) .

### Causes
The conversationDidChange event reloads all tableView.

### Solutions
Add an observer of the conversation change in the cell.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
@@ -60,9 +60,3 @@ extension ConversationContentViewController {
         updateTableViewHeaderView()
     }
 }
-
-extension ConversationContentViewController: ZMConversationObserver {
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController+Scrolling.swift
```diff
@@ -63,8 +63,6 @@ extension ConversationContentViewController {
 
 extension ConversationContentViewController: ZMConversationObserver {
     public func conversationDidChange(_ note: ConversationChangeInfo) {
-        guard note.createdRemotelyChanged else { return }
-        
-        dataSource?.loadMessages(forceRecalculate: true)
+        guard note.createdRemotelyChanged else { return }        
     }
 }
```
billypchan: This can be removed (and the observer also)

KaterinaWire: Removed

PR #3905, opened 2019-12-30
convert ShareContactsViewController to Swift

billypchan:
## What's new in this PR?

convert `ShareContactsViewController` and its callers to Swift


billypchan:
continue in https://github.com/wireapp/wire-ios/pull/3906

PR #3906, opened 2020-01-01
convert ShareContactsViewController & PermissionDeniedViewController to Swift

billypchan:
## What's new in this PR?

convert `ShareContactsViewController` & `PermissionDeniedViewController` to Swift


Wire-iOS/Sources/Permissions/PermissionDeniedViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+    
+    private var initialConstraintsCreated = false
+    private let heroLabel: UILabel = UILabel.createHeroLabel()
+    private var settingsButton: Button!
+    private var laterButton: UIButton!
+    private let backgroundBlurView: UIVisualEffectView = UIVisualEffectView.createBackgroundBlurView()
+
+    private var monochromeStyle = false
+
+    class func addressBookAccessDeniedViewController(withMonochromeStyle monochromeStyle: Bool) -> PermissionDeniedViewController {
```
typfel: Pretty sure this monochrome style is not used anymore. Did you check if this ever called with `monochromeStyle == true`?

Wire-iOS/Sources/Permissions/PermissionDeniedViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        
+        updateViewConstraints()
+    }
+    
+    @available(*, unavailable)
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func viewDidLoad() {
```
typfel: Can be deleted

Wire-iOS/Sources/Permissions/PermissionDeniedViewController.swift
billypchan: Yes, it is always false now.

Wire-iOS/Sources/Permissions/PermissionDeniedViewController.swift
```diff
@@ -0,0 +1,178 @@
...
+        vc.heroLabel.attributedText = attributedText
+        
+        vc.settingsButton.setTitle("registration.address_book_access_denied.settings_button.title".localized.uppercased(), for: .normal)
+        
+        vc.laterButton.setTitle("registration.address_book_access_denied.maybe_later_button.title".localized.uppercased(), for: .normal)
+
+        return vc
+    }
+    
+    class func push() -> PermissionDeniedViewController {
```
typfel: This name is a bit out of context now, I see where it comes from but it doesn't make much sense anymore. It looks like this class method is creating a default instance of the `PermissionDeniedViewController`. Maybe we can call it something like `default`

billypchan: restored to original objc name `pushDeniedViewController`

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+ShareContactsViewControllerDelegate.swift
```diff
@@ -1,4 +1,4 @@
-//
+
```
typfel: Extra space added

PR #3908, opened 2020-01-02
Fix: Restored scrollToFirstHighlightedUser and fixed tests.

KaterinaWire:
### Issues
Test crashes.

### Causes
Before implementing a conversation roles feature we had only 1 section for the GroupParticipantsDetailViewController. 

### Solutions
Calculate the section to which the selected user belongs.


Wire-iOS/Sources/UserInterface/GroupDetails/GroupParticipantsDetail/GroupParticipantsDetailViewModel.swift
```diff
@@ -42,6 +42,11 @@ final class GroupParticipantsDetailViewModel: NSObject, SearchHeaderViewControll
         }
     }
     
+    var sectionOfFirstSelectedParticipant: Int? {
+        guard let first = selectedParticipants.first as? ZMUser else { return nil }
```
typfel: How about renaming `first` to `user` ?

Also `sectionOfFirstSelectedParticipant` is quite similar to `indexOfFirstSelectedParticipant `. Do you think it would make sense to combine them into a `var indexPathOfFirstSelectedParticipant: IndexPath? `

KaterinaWire: Yes, I agree. 
I fixed it.

PR #3911, opened 2020-01-03
Fix: app center does not track for crashes

billypchan:
### Issues

 app center does not track for crashes

### Causes

`setTrackingEnabled` should be called after services started.
The dummy AppCenter ID overrides the AppCenter ID in other `.xcconfig` files.

### Solutions

Correct the setup.
Clean up duplicated frameworks existed in both wire compound and wire-ios targets.


Wire-iOS/Sources/AppDelegate+AppCenter.swift
```diff
@@ -43,18 +44,20 @@ extension AppDelegate {
         
         let appCenterTrackingEnabled = !TrackingManager.shared.disableCrashAndAnalyticsSharing
```
marcoconti83: Where does `TrackingManager.shared.disableCrashAndAnalyticsSharing` come from? Is this what is set when the user is asked about sharing crashes and other data?

billypchan: It is from the user setting

marcoconti83:
@pwnsdx will review this too

Wire-iOS/Resources/Configuration/Project-Base.xcconfig
```diff
@@ -62,8 +62,7 @@ ENABLE_STRICT_OBJC_MSGSEND = YES
 
 // LLVM compiler - Preprocessing
 
-APP_CENTER_APP_ID_KEY=dummyAppCenterID
```
billypchan: It causes the crashes are not tracked since it override `APP_CENTER_APP_ID_KEY` defined in other xcconfig files.

Wire-iOS/Sources/AppDelegate+AppCenter.swift
```diff
@@ -43,17 +44,20 @@ extension AppDelegate {
...
         
         if appCenterTrackingEnabled {
+            MSCrashes.setDelegate(self)
+            
             MSAppCenter.start()
+            
+            MSAppCenter.setLogLevel(.verbose)
+            
+            // This method must only be used after Services have been started.
+            MSAppCenter.setTrackingEnabled(appCenterTrackingEnabled)
```
billypchan: @pwnsdx Here is the change related to security. I moved `MSAppCenter.setTrackingEnabled` after `MSAppCenter.start` to fix the installation.

PR #3914, opened 2020-01-03
[Feature] Custom app lock

David-Henner:
## What's new in this PR?

Improved App Lock security with an option to use the account password instead of device PIN in case Biometrics fail or aren't enabled.

### Solutions

- Added new `AppLock` Rule `useBiometricsOrAccountPassword` to activate/deactivate using the account password
- Updated `AppLock` logic to handle both cases 
- Updated presentation logic with VIPER architecture
- Updated request password controller for App Lock
- Updated unit tests

### Dependencies

https://github.com/wireapp/wire-ios-sync-engine/pull/1135


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -51,121 +49,58 @@ final class AppLockViewController: UIViewController {
...
-
-    @objc func applicationWillResignActive() {
-        if AppLock.isActive {
-            self.dimContents = true
-        }
+// MARK: - AppLockManagerDelegate
+extension AppLockViewController: AppLockUserInterface {
+    func presentRequestPasswordController(with message: String, callback: @escaping RequestPasswordController.Callback) {
+        let passwordController = RequestPasswordController(context: .unlock(message.localized), callback: callback)
+        self.passwordController = passwordController
```
johnxnguyen: Why do we need to store a reference to the password controller?

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -0,0 +1,99 @@
...
+extension AppLockInteractor: AppLockInteractorInput {
+    var isLockTimeoutReached: Bool {
+        let lastAuthDate = appLock.lastUnlockedDate
+        
+        // The app was authenticated at least N seconds ago
+        let timeSinceAuth = -lastAuthDate.timeIntervalSinceNow
+        if timeSinceAuth >= 0 && timeSinceAuth < Double(appLock.rules.appLockTimeout) {
+            return false
+        }
+        return true
```
johnxnguyen: This could be simplified with a range, for example:

```swift
let isWithinTimeoutWindow = (0..<appLock.rules.appLockTimeout).contains(timeSinceAuth)
return !isWithinTimeoutWindow
```

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
David-Henner: It seems to be deallocated if there is no reference to it

johnxnguyen: Because the `passwordController.alertController` is what is being presented?

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol AppLockInteractorInput: class {
+    var isLockTimeoutReached: Bool { get }
+    func evaluateAuthentication(description: String)
```
johnxnguyen: It seems like this method is only called in one place. Perhaps you don't need to pass the description as an argument.

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
David-Henner: Yes, `passwordController` wouldn't be retained, which means `passwordController.alertController` neither. In which case the alert disappear instantly

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
David-Henner: The reason behind it is that I'm keeping the strings in one place in a struct in the presenter file.  
```
private struct AuthenticationMessageKey {
    static let accountPassword = "self.settings.privacy_security.lock_password.description.unlock"
    static let wrongPassword = "self.settings.privacy_security.lock_password.description.wrong_password"
    static let wrongOfflinePassword = "self.settings.privacy_security.lock_password.description.wrong_offline_password"
    static let deviceAuthentication = "self.settings.privacy_security.lock_app.description"
}
```

David-Henner: I'll then pass the `deviceAuthentication` string to `evaluateAuthentication(description:)`

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
johnxnguyen: Would it make sense to for `RequestPasswordController` to inherit `UIAlertController`, then you could just present the password controller instead?

David-Henner: I don't see a reason against it

johnxnguyen: Otherwise I would leave a comment stating that a reference should be kept, otherwise someone might remove it thinking it's not needed (like me ahah).

David-Henner: To keep the scope of the PR I'd leave a comment like you suggested and mark the controller for improvement (maybe for the upcoming service run)

David-Henner: Since the controller is used in other parts of the app

Wire-iOS/Sources/UserInterface/Helpers/RequestPasswordController.swift
```diff
@@ -25,6 +25,7 @@ final class RequestPasswordController {
     enum RequestPasswordContext {
         case removeDevice
         case logout
+        case unlock(String)
```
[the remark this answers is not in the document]
David-Henner: done here ab2dc35

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
David-Henner: done here ab2dc35

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -0,0 +1,83 @@
...
+
+class AppLockInteractor {
+    weak var output: AppLockInteractorOutput?
+    
+    // For tests
+    var appLock: AppLock.Type = AppLock.self
+    var dispatchQueue: DispatchQueue = DispatchQueue.main
+    
+    private var isObserverSetup: Bool = false
+    private var moc: NSManagedObjectContext? {
```
typfel: Doesn't seem to be used anymore

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class AppLockInteractor {
+    weak var output: AppLockInteractorOutput?
+    
+    // For tests
+    var appLock: AppLock.Type = AppLock.self
+    var dispatchQueue: DispatchQueue = DispatchQueue.main
+    
+    private var isObserverSetup: Bool = false
```
typfel: Does this still serve a purpose?

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1136,6 +1136,9 @@
 "self.settings.privacy_security.lock_app.subtitle.none" = "Unlock by entering your passcode.";
 
 "self.settings.privacy_security.lock_app.description" = "Unlock Wire";
+"self.settings.privacy_security.lock_password.description.unlock" = "Unlock Wire with account password.";
```
typfel: @johnxnguyen @David-Henner are these strings the final ones?

johnxnguyen: They look fine to me. @David-Henner do you want to run them by Astrid?

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
David-Henner: removed (881ddca)

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
David-Henner: removed (881ddca)

Wire-iOS/Resources/Base.lproj/Localizable.strings
David-Henner: Yes, would be wise

PR #3916, opened 2020-01-04
Fix: convert InviteContactsViewController

billypchan:


Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewControllerDelegate.swift
```diff
@@ -0,0 +1,60 @@
...
+}
+
+@objc
+protocol SearchUserType {
+    var user: ZMUser? { get }
+}
+
+extension ZMSearchUser: SearchUserType { }
+
+typealias SearchUser = UserType & SearchUserType
```
johnxnguyen: Why do we need this?

billypchan: For mocking a search user in snapshot testings.

typfel: But only one method has been updated to use the the `SearchUser` type and all other methods still use the `ZMSearchUser`. Besides that I don't think we should introduce the `SearchUserType` protocol since it exposes the `ZMUser` to the UI which are currently trying to remove with the current `UserType` re-factorings.

I would propose to review the methods taking a `ZMSearchUser` and see if it's possible to replace it with `UserType`. I suspect that in many cases the retrieval of the underlying `ZMUser` only necessary due to missing methods on the `UserType` or `ZMSearchUser` which now exist.

billypchan: refactor the delegate method to pass a `UserType` instead of `SearchUser`

Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+        tableView.sectionIndexBackgroundColor = .clear
+        tableView.sectionIndexColor = .accent()
+        
+        bottomContainerSeparatorView.backgroundColor = UIColor.from(scheme: .separator, variant: .dark)
+        bottomContainerView.backgroundColor = UIColor.from(scheme: .searchBarBackground, variant: .dark)
+        
+        titleLabel.textColor = UIColor.from(scheme: .textForeground, variant: .dark)
+    }
+    
+    private func invite(user: ZMSearchUser, from view: UIView) {
```
johnxnguyen: Is it possible to use `UserType` here?

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewControllerDelegate.swift
```diff
@@ -0,0 +1,51 @@
...
+    func contactsViewControllerDidNotShareContacts(_ controller: ContactsViewController)
+    
+    func contactsViewControllerDidConfirmSelection(_ controller: ContactsViewController)
+}
+
+@objc
+protocol ContactsViewControllerContentDelegate: NSObjectProtocol {
+    
+    
+    func contactsViewController(_ controller: ContactsViewController, shouldDisplayActionButtonFor user: ZMSearchUser) -> Bool
```
johnxnguyen: Is it possible to use `UserType` here?

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController+ContactsDataSourceDelegate.swift
```diff
@@ -50,7 +50,7 @@ extension ContactsViewController: ContactsDataSourceDelegate {
         cell.actionButton.isHidden = actionButtonHidden(user: user)
 
         if !cell.actionButton.isHidden,
-            let index = contentDelegate?.contactsViewController(self, actionButtonTitleIndexFor: user),
+            let index = contentDelegate?.contactsViewController(self, actionButtonTitleIndexFor: user.user, isIgnored: user.user?.isIgnored ?? false),
```
typfel: So we are missing `isIgnored` on the `UserType` let's add it then.

billypchan: I am fixing the crash when testing, objc allow casting `MockUser` to `ZMSearchUer` and getting its `isIgnored` property. After converted to Swift it is not allowed.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewControllerDelegate.swift
billypchan: removed user parameter since it is not used

Wire-iOS/Sources/UserInterface/InviteContacts/InviteContactsViewController.swift
billypchan: `ZMSearchUser.contact` is needed in callee `inviteUserOrOpenConversation` so it is not possible.

johnxnguyen: Perhaps this is a case for a `SearchUserType’. @typfel thoughts?

typfel: Yes, if we stick to the rule of not exposing concrete types I think we would need a `SearchUserType`. We should define it in data model / sync engine though. 

But this PR is about converting an existing class to Swift. I think it makes sense to not start re-factoring the types at the same time. I would propose to leave it as a `ZMSearchUser` in this PR and introduce the `SearchUserType` in a separate PR.

PR #3917, opened 2020-01-06
[Fix] UserType conformance

johnxnguyen:
### Issues

Some functionality involving `UserType` (for example the new roles feature) is hard (if not impossible) to test.

### Causes

The `UserType` protocol leaks the concrete `ZMUser` type (through the `zmUser` property) such that certain functionality uses methods not exposed by the `UserType` protocol. Therefore any attempt to mock the user type is fruitless because the concrete `ZMUser` type is always used in the production code. In general, accessing this `zmUser` property is bad practice, because it defeats the purpose of the `UserType` abstraction.

### Solutions

The `UserType` protocol was exerted in https://github.com/wireapp/wire-ios-data-model/pull/869 to be able to remove `zmUser`. This PR contains

- updates to `UserType` conformance
- general clean up and removal of redundant code
- unboxing of `ZMUser` from `UserType` where appropriate


Wire-iOS/Sources/UserInterface/ConnectRequests/UserConnectionView.swift
```diff
@@ -105,7 +105,7 @@ public final class UserConnectionView: UIView, Copyable {
     private var correlationLabelText: NSAttributedString? {
         return type(of: self).correlationFormatter.correlationText(
             for: user,
-            addressBookName: user.zmUser?.addressBookEntry?.cachedName
+            addressBookName: (user as? ZMUser)?.addressBookEntry?.cachedName
```
typfel: So this is still leaking the concrete type

Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+SearchResults.swift
```diff
@@ -42,14 +42,19 @@ extension StartUIViewController {
 }
 
 extension StartUIViewController: SearchResultsViewControllerDelegate {
+
+    private func unboxedUser(from user: UserType) -> ZMUser? {
+      return (user as? ZMUser) ?? (user as? ZMSearchUser)?.user
```
billypchan: `ZMSearchUser` has its own unbox logic in `isConnected`, `isBlocked` getters, should we just pass the `UserType` object without unboxing?

Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -289,7 +291,7 @@ final class ProfileDetailsContentController: NSObject,
         guard
             let role = newParticipantRole,
             let session = ZMUserSession.shared(),
-            let user = user.zmUser
+            let user = (user as? ZMUser) ?? (user as? ZMSearchUser)?.user
             else { return }
         
         conversation?.updateRole(of: user, to: role, session: session) { (result) in
```
billypchan: It should be better to change the parameter type to `UserType` to save the unboxing code.

johnxnguyen: Here we are about to make the request to the backend to update the role for the user. To use the `UserType` here we'll have to introduce a few more changes, which at the time seemed not necessary but now it might be worth doing for the sake of completion.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -65,15 +65,15 @@ final class ProfileViewControllerViewModel: NSObject {
...
 
     var hasLegalHoldItem: Bool {
         return bareUser.isUnderLegalHold || conversation?.isUnderLegalHold == true
     }
     
     var showVerifiedShield: Bool {
-        if let user = bareUser.zmUser {
+        if let user = fullUser {
             let showShield = user.trusted() &&
                 !user.clients.isEmpty &&
```
billypchan: we can move `trusted` property to `UserType`
`clients` can be replaced with `allClients` in `UserType`

Wire-iOS/Sources/UserInterface/ConnectRequests/UserConnectionView.swift
johnxnguyen: I found that in some cases it made sense to unbox the user as to use the `UserType` would require more changes in other classes. But looking at this now, it looks like we can just add the address book entry to the user type 👍

Wire-iOS/Sources/UserInterface/StartUI/StartUI/StartUIViewController+SearchResults.swift
johnxnguyen: The unboxing here is not for the `isConnected` and `isBlocked` properties, but for passing the `ZMUser` to the delegate (which then uses it to do more stuff on the SE side).

johnxnguyen:
Regarding all the comments: there is more work to do for ensuring the proper use of `UserType` which is out of scope for this PR. I will address this separately and merge this if approved.

PR #3920, opened 2020-01-06
converted ZClientViewController.mediaPlaybackManager callers and ConversationContentViewControllerDelegate to Swift

billypchan:
## What's new in this PR?

converted `ZClientViewController.mediaPlaybackManager` and `ConversationContentViewControllerDelegate` callers to Swift


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ConversationContentViewControllerDelegate.swift
```diff
@@ -0,0 +1,135 @@
...
+            NotificationCenter.default.addObserver(self, selector: #selector(menuDidHide(_:)), name: UIMenuController.didHideMenuNotification, object: nil)
+            
+            return false
+        }
+        
+        return true
+    }
+    
+    func conversationContentViewController(_ contentViewController: ConversationContentViewController, performImageSaveAnimation snapshotView: UIView?, sourceRect: CGRect) {
+        if let snapshotView = snapshotView {
```
typfel: Could unwrap the snapshot view or return here:

`guard let snapshotView = snapshotView else { return }`

typfel: Or change the method not to accept an optional `UIView`

billypchan: I checked that the caller may pass a nil snapshotView so I make it optional.

typfel: Right but the method does nothing if the view is nil we could just as well not allow it.

PR #3934, opened 2020-01-14
[Feature] Add launch flag for disabling the interactive dismissal of keyboard

typfel:
## What's new in this PR?

Adds the launch flag:

`--disable-interactive-keyboard-dismissal`

Which disables the interactive keyboard dismissal of the conversation input bar keyboard. This is necessary since the automation doesn't see the buttons underneath the UITextView `accessoryView` when running on iOS 13.

### DIFF
--- a/WireCommonComponents/AutomationHelper.swift
+++ b/WireCommonComponents/AutomationHelper.swift
@@ -69,7 +69,7 @@ final public class AutomationEmailCredentials: NSObject {
     /// Whether we should disable the call quality survey.
     public let disableCallQualitySurvey: Bool
     
-    /// Wheater we should disable dismissing the conversation input bar keyboard by dragging it downwards.
+    /// Whether we should disable dismissing the conversation input bar keyboard by dragging it downwards.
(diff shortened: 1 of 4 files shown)


WireCommonComponents/AutomationHelper.swift
```diff
@@ -37,49 +37,52 @@ final public class AutomationEmailCredentials: NSObject {
...
+    public let disableAutocorrection : Bool
     
     /// Whether address book upload is enabled on simulator
-    @objc public let uploadAddressbookOnSimulator : Bool
+    public let uploadAddressbookOnSimulator : Bool
 
     /// Whether we should disable the call quality survey.
     public let disableCallQualitySurvey: Bool
     
+    /// Wheater we should disable dismissing the conversation input bar keyboard by dragging it downwards.
```
johnxnguyen: Typo.

```suggestion
    /// Whether we should disable dismissing the conversation input bar keyboard by dragging it downwards.
```

PR #3937, opened 2020-01-14
[Fix] AppLock not showed on launch

David-Henner:
### Issues

AppLock wasn't showing up after app launch

### Causes

We relied on `isUserSessionActive` to decide if we need to show the appLock.
However, on app launch, the userSession is instantiated after checking `isUserSessionActive`.
This leads to the condition evaluating to `false` instead of `true`, therefore the appLock will not show.

### Solutions

Rely on appState transitions instead of userSession.

### DIFF
--- a/Wire-iOS/Sources/AppStateController.swift
+++ b/Wire-iOS/Sources/AppStateController.swift
@@ -122,12 +122,20 @@ final class AppStateController : NSObject {
             zmLog.debug("transitioning to app state: \(newAppState)")
             lastAppState = appState
             appState = newAppState
+            
             delegate?.appStateController(transitionedTo: appState) {
                 completion?()
             }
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/AppStateController.swift
```diff
@@ -246,3 +254,8 @@ extension AppStateController : AuthenticationCoordinatorDelegate {
     }
     
 }
+
+extension AppStateController {
```
typfel: If you do the extension on `Notification.Name` instead you can use . syntax when subscribing.

David-Henner: What about the userInfo key? Any suggestion on how to make it short as well?

PR #3946, opened 2020-01-17
Fix: Tap connect button on incoming request does not dismiss the screen

billypchan:
### Issues

Tap connect button on incoming request does not dismiss the screen

### Causes

`UIViewController` has a `select` method, which conflicts with `ZClientViewController`'s `select` method without argument label.

### Solutions

Name the label as `conversation`.
~~I also restore the property type `var connectionRequests` from `[ZMConversation]` to `ZMConversationList?` to prevent casting.~~


Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.swift
```diff
@@ -136,7 +137,7 @@ final class ConnectRequestsViewController: UIViewController, UITableViewDataSour
         }
         
         cell.ignoreBlock = { [weak self] in
-            guard self?.connectionRequests.isEmpty == true else { return }
+            guard self?.connectionRequests?.count == 0 else { return } ///TODO: copy below
```
typfel: Is the TODO completed?

billypchan: fixed

PR #3948, opened 2020-01-17
Fix: convert AppDelegate

billypchan:
## What's new in this PR?

Convert `AppDelegate` to Swift


Wire-iOS/Sources/AppDelegate+URL.swift
```diff
@@ -19,11 +19,11 @@
...
 
 extension AppDelegate {
-    @objc
-    func open(url: URL, options: [UIApplication.OpenURLOptionsKey : AnyObject]) -> Bool {
+
+    func open(url: URL, options: [UIApplication.OpenURLOptionsKey : Any]) -> Bool {
         guard let urlHandler = sessionManager?.urlHandler else { return false }
 
-        return urlHandler.openURL(url, options: options)
+        return urlHandler.openURL(url, options: options as! [UIApplication.OpenURLOptionsKey : AnyObject])
```
typfel: Do the force unwrapping in the guard?

Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -18,9 +18,228 @@
...
+    func application(_ application: UIApplication, didReceiveRemoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        zmLog.info("application:didReceiveRemoteNotification:fetchCompletionHandler: notification: \(userInfo)")
+        
+        launchType = (application.applicationState == .inactive || application.applicationState == .background) ? .push : .direct
+    }
+    
+    func application(_ application: UIApplication, performFetchWithCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        zmLog.info("application:performFetchWithCompletionHandler:")
+        
+        rootViewController?.performWhenAuthenticated(){
```
typfel:
```suggestion
        rootViewController?.performWhenAuthenticated() {
```

Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -18,9 +18,228 @@
...
+            ZMUserSession.shared()?.application(application, performFetchWithCompletionHandler: completionHandler)
+        }
+    }
+    
+    func application(_ application: UIApplication, handleEventsForBackgroundURLSession identifier: String, completionHandler: @escaping () -> Void) {
+        zmLog.info("application:handleEventsForBackgroundURLSession:completionHandler: session identifier: \(identifier)")
+        
+        rootViewController?.performWhenAuthenticated(){
```
typfel:
```suggestion
        rootViewController?.performWhenAuthenticated() {
```

Wire-iOS/Sources/AppDelegate+URL.swift
typfel: But I guess the real fix is to change the urlHandler API in sync engine to match the type.

billypchan: removed the !

typfel: How can that be allowed? `Any` is more generic than `AnyObject`

billypchan: Correct, but when I try to cast it the compiler say:
```
Conditional cast from '[UIApplication.OpenURLOptionsKey : Any]' to '[UIApplication.OpenURLOptionsKey : AnyObject]' always succeeds
```

PR #3953, opened 2020-01-17
converted InvisibleInputAccessoryView

billypchan:
converted InvisibleInputAccessoryView and delegate to Swift


Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBar.swift
```diff
@@ -194,7 +194,7 @@ final class InputBar: UIView {
     
     override public var bounds: CGRect {
         didSet {
-            invisibleInputAccessoryView?.intrinsicContentSize = CGSize(width: UIView.noIntrinsicMetric, height: bounds.height)
+            invisibleInputAccessoryView?.contentSize = CGSize(width: UIView.noIntrinsicMetric, height: bounds.height)
```
johnxnguyen: Shouldn't this be `intrinsicConentSize`?

billypchan: `UIView` has a get only `intrinsicContentSize`, I change this to `overriddenIntrinsicContentSize` to replace original `_intrinsicContentSize`

PR #3958, opened 2020-01-20
Fix: Connection request screen is too tall on iPhone 11

billypchan:
### Issues

On iPhone with notch, the connect request cell is too high and covered by the notch.

### Causes

We used `tableView.bounds.size.height` for the cell height, which is not considering the safe area.

### Solutions

Use `safeAreaLayoutGuideOrFallback.layoutFrame.size.height` instead.


Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.swift
```diff
@@ -109,15 +108,15 @@ final class ConnectRequestsViewController: UIViewController, UITableViewDataSour
         
         let inset: CGFloat = connectionRequests.count > 1 ? 48 : 0
         
-        return max(0, tableView.bounds.size.height - inset)
+        return max(0, view.safeAreaHeight - inset)
```
typfel: I would prefer that we don't introduce `safeAreaHeight` and instead call `safeAreaLayoutGuideOrFallback.layoutFrame.size.height` directly, since it makes it more clear that we are relying on fallback behaviour.

Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestCell.swift
```diff
@@ -19,6 +19,7 @@
 import UIKit
 
 final class ConnectRequestCell: UITableViewCell {
+    static var identifier = "ConnectRequestCell"
```
typfel: We could use the method from `UITableViewCell+Reuse`

billypchan: fixed

PR #3963, opened 2020-01-21
[Fix] Make links accessible to automation on iOS 13

typfel:
### Issues

When running on iOS 13 the "this device" link, included in a system message, is no longer exposed to automation/accessibility.

### Causes

By trial and error I discovered that `isEditable` must now be true for links to exposed as accessibility elements in iOS 13.

### Solutions

Remove code which sets `isEditable` to `false`. The textview is still not editable since we override `canBecomeFirstResponder` and the gesture recogniser.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/WebLinkTextView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/WebLinkTextView.swift
@@ -38,8 +38,6 @@ import UIKit
     }
 
     private func setup() {
-        isSelectable = true
-        isEditable = false
         isScrollEnabled = false
         bounces = false
         backgroundColor = UIColor.clear


Wire-iOS/Sources/UserInterface/Components/Views/WebLinkTextView.swift
```diff
@@ -38,8 +38,6 @@ import UIKit
     }
 
     private func setup() {
-        isSelectable = true
```
typfel: Deleted this as well since `isSelectable` is true by default.

Wire-iOS/Sources/UserInterface/Components/Views/WebLinkTextView.swift
```diff
@@ -38,8 +38,6 @@ import UIKit
...
-        isEditable = false
```
billypchan: This `UITextView` subclass becomes editable but it should not have any action when the user tap on it. Should we make it for testing only?

typfel: As mention in the PR description, the textview is still not editable since we prevent it by other means.

typfel: I would not make this change testing only since it's also a fix for users who rely on the accessibility functionality.

PR #3964, opened 2020-01-21
[Refactor] Hide UserType in Typing Classes

johnxnguyen:
## What's new in this PR?

This PR is part of an ongoing effort to all public references that leak concrete user objects. See the architecture issues for more info: https://github.com/wireapp/ios-architecture/issues/89

It contains adjustments for changes introduced in https://github.com/wireapp/wire-ios-sync-engine/pull/1147

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1147


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -349,8 +349,9 @@ final class TypingMatcher: ConversationStatusMatcher {
     
     func description(with status: ConversationStatus, conversation: ZMConversation) -> NSAttributedString? {
         let statusString: NSAttributedString
-        if status.isGroup, let typingUsers = conversation.typingUsers() {
-            let typingUsersString = typingUsers.compactMap { $0 as? ZMUser }.map { $0.displayName(in: conversation) }.joined(separator: ", ")
+        if status.isGroup {
+            let typingUsers = conversation.typingUsers.compactMap { $0 as? ZMUser }
+            let typingUsersString = typingUsers.map { $0.displayName(in: conversation) }.joined(separator: ", ")
```
johnxnguyen: `displayName(in:)` is a `ZMUser` method, which uses a display name generator that requires `ZMUser` (because of `NSManagedObjectID`. I would have also refactored this, however I understand that the display name generator will be retired soon.

PR #3965, opened 2020-01-21
Fix/app crash conversation view controller+profile view controller delegate

KaterinaWire:
## What's new in this PR?

- Fix app crash (we still can't reproduce this crash).
- Bump the data model version.


johnxnguyen:
I'm not sure I see the relevance of the `PopupIconButtonView` changes to the crash. Is it necessary in this PR?

Cartfile.resolved
```diff
@@ -16,7 +16,7 @@ github "wireapp/swift-protobuf" "1.5.0"
 github "wireapp/swift-snapshot-testing" "1.6.0"
 github "wireapp/wire-ios-canvas" "10.0.0"
 github "wireapp/wire-ios-cryptobox" "20.1.0"
-github "wireapp/wire-ios-data-model" "204.0.2"
+github "wireapp/wire-ios-data-model" "204.0.3"
```
johnxnguyen: What is included in this bump?

KaterinaWire:
> I'm not sure I see the relevance of the `PopupIconButtonView` changes to the crash. Is it necessary in this PR?

@johnxnguyen  No, I can move it to another PR.

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ProfileViewControllerDelegate.swift
```diff
@@ -21,7 +21,7 @@ import Foundation
 extension ConversationViewController {
     @objc
     func createUserDetailViewController() -> UIViewController {
-        guard let user = conversation.firstActiveParticipantOtherThanSelf else {
+        guard let user = (conversation.firstActiveParticipantOtherThanSelf ?? conversation.connectedUser) else {
```
johnxnguyen: Could you perhaps elaborate what this change is doing?

Cartfile.resolved
KaterinaWire: Fix. After refactoring, I didn't make appendSelfConversationWithLastReadOfConversation method as a public.

johnxnguyen: Is it relevant for this PR though?

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ProfileViewControllerDelegate.swift
KaterinaWire: Crash logs show this line of code as a cause of crashes. We did such check in another VC (ConversationContentViewController):
```if conversation.conversationType == .connection {
            otherParticipant = conversation.firstActiveParticipantOtherThanSelf ?? conversation.connectedUser
        } else {
            otherParticipant = conversation.firstActiveParticipantOtherThanSelf
        }
```
Since we can't reproduce this crash now, I decided to try to fix it. If it helps, we can refactor this logic and put into a data model (firstActiveParticipantOtherThanSelf exists now only in the UI project)

Cartfile.resolved
KaterinaWire: No, but we get crashes in develop branch now. So it's important.

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ProfileViewControllerDelegate.swift
johnxnguyen: Looking in the source code, this method `createUserDetailViewController()` is being called when the conversation type is either the self conversation, one to one, or connection. `firstActiveParticipant` tries to find the first local participant that isn't the self user. But if this is called on the self conversation, there wouldn't be any other participants, which would lead to the `fatal()`. Perhaps that's the source of the crash?

Cartfile.resolved
johnxnguyen: I believe It's good to limit scopes of PR's to contain only relevant changes with some documentation, so that we can refer to the PRs at a later date and also revert them if needed. It can be hard to track down changes hiding in other PRs.

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+ProfileViewControllerDelegate.swift
KaterinaWire: Yes, it might be the self conversation or connection. This fix will help if it's a connection.

PR #3968, opened 2020-01-22
[Refactor] Hide UserType in Conversation Classes

KaterinaWire:
## What's new in this PR?

This PR is part of an ongoing effort to all public references that leak concrete user objects. See the architecture issues for more info: wireapp/ios-architecture#89

It contains adjustments for changes introduced in  wireapp/wire-ios-sync-engine#1150

Needs releases with:

- [ ]  wireapp/wire-ios-sync-engine#1150


KaterinaWire:
> Is it possible to change the argument of `addOrShowError(participants:)` to accept `UserType`s as well? And further up the chain?

@johnxnguyen Done

Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
```diff
@@ -75,7 +75,7 @@ extension ZMConversation {
                                 switch result {
                                 case .success:
                                     if user.isServiceUser {
-                                        Analytics.shared().tagDidRemoveService(user)
+                                        Analytics.shared().tagDidRemoveService(user as! ServiceUser)
```
johnxnguyen: I suggest doing an `if let` to cast as `ServiceUser` so as to avoid a potential crash.

KaterinaWire: Fixed

PR #3974, opened 2020-01-23
[Feature] Linear onboarding welcome screen

David-Henner:
## What's new in this PR?

Welcome screen updated according to the new design specs


Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -257,38 +256,67 @@ class LandingViewController: AuthenticationStepViewController {
...
-        contentStack.translatesAutoresizingMaskIntoConstraints = false
-
+        topStack.translatesAutoresizingMaskIntoConstraints = false
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+        buttonStackView.translatesAutoresizingMaskIntoConstraints = false
+        enterpriseLoginButton.translatesAutoresizingMaskIntoConstraints = false
+        messageLabel.translatesAutoresizingMaskIntoConstraints = false
+        
+        var widthConstraint: NSLayoutConstraint
+        if traitCollection.horizontalSizeClass == .regular {
```
typfel: The `horizontalSizeClass` can change at runtime, when using the app in split screen mode for example. Wouldn't it be possible to achieve the same thing with maximum width constraint plus leading trailing insets?

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -257,38 +256,67 @@ class LandingViewController: AuthenticationStepViewController {
...
+            widthConstraint = contentView.widthAnchor.constraint(equalToConstant: 375)
```
billypchan: We should consider the device with width = 320, e.g. iPhone 4 inch?

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -257,38 +256,67 @@ class LandingViewController: AuthenticationStepViewController {
...
 
-        view.addSubview(contentStack)
+        view.addSubview(contentView)
     }
 
     private func createConstraints() {
-        contentStack.translatesAutoresizingMaskIntoConstraints = false
-
+        topStack.translatesAutoresizingMaskIntoConstraints = false
+        contentView.translatesAutoresizingMaskIntoConstraints = false
```
johnxnguyen: You could wrap this in an array and use a for each.

David-Henner: done here 4625d54

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
David-Henner: done here 4625d54

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
David-Henner: should be taken into account here 4625d54

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -85,34 +88,53 @@ class LandingViewController: AuthenticationStepViewController {
...
-        button.accessibilityIdentifier = "CreateAccountButton"
-        button.addTapTarget(self, action: #selector(LandingViewController.createAccountButtonTapped(_:)))
-        button.setContentCompressionResistancePriority(.required, for: .vertical)
-        button.setContentCompressionResistancePriority(.required, for: .horizontal)
-        button.setContentHuggingPriority(.required, for: .vertical)
-
+    
+    let enterpriseLoginButton: Button = {
+        let button = Button(style: .fullMonochrome, variant: .light)
+        button.setBackgroundImageColor(UIColor.from(scheme: .secondaryAction), for: .normal)
```
typfel: I'd suggest to create a new color for this in the ColorScheme.swift or use the named color directly here if it's a one off that's shouldn't be repeated in other places.

Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -257,43 +245,73 @@ class LandingViewController: AuthenticationStepViewController {
...
-
+        disableAutoresizingMaskTranslation(for: [
+            topStack,
+            contentView,
+            buttonStackView,
+            enterpriseLoginButton,
+            messageLabel
+        ])
+        
+        let widthConstraint = contentView.widthAnchor.constraint(equalToConstant: 375)
```
billypchan: The snapshot of different screen size should be created with `verifyInAllIPhoneSizes`

PR #3976, opened 2020-01-24
[Refactor] Hide Concrete Users in Calling Classes

johnxnguyen:
## What's new in this PR?

This PR is part of an ongoing effort to remove all concrete user objects from the UI project. See the architecture issues for more info: https://github.com/wireapp/ios-architecture/issues/89.

This PR contains changes to match changes in https://github.com/wireapp/wire-ios-sync-engine/pull/1154.


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -37,7 +37,7 @@ fileprivate extension VoiceChannel {
         
         switch state {
         case .incoming(video: false, shouldRing: true, degraded: _):
-            return initiator.map { .avatar($0) } ?? .none
+            return (initiator as? ZMUser).map { .avatar($0) } ?? .none
```
johnxnguyen: @typfel I decided to keep `ZMUser` in the CallInfoViewControllerAccessoryType` because of the `Equatable` issues we talked about regarding `UserType`.

I'm not sure if opaque types can help us here, because the accessory type is an enum and the avatar case needs to hold a reference to an equatable user.

typfel: Make sense.

Yep doesn't look opaque types would help here https://github.com/apple/swift-evolution/blob/master/proposals/0244-opaque-result-types.md

I'd propose we refactor the calling implementation later.

PR #3983, opened 2020-01-24
[Fix] Remove all whitespaces at the end of a search string

KaterinaWire:
### Issues
Extra space at the end is a cause that the user disappears in the search UI.

### Causes
We use the original search string (with spaces) to search.

### Solutions
Remove whitespaces.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
@@ -250,7 +250,7 @@ extension UIViewController {
         var options = options
         options.updateForSelfUserTeamRole(selfUser: ZMUser.selfUser())
 
-        let request = SearchRequest(query: query, searchOptions: options, team: ZMUser.selfUser().team)
+        let request = SearchRequest(query: query.trim(), searchOptions: options, team: ZMUser.selfUser().team)
         if let task = searchDirectory?.perform(request) {
             task.onResult({ [weak self] in self?.handleSearchResult(result: $0, isCompleted: $1)})
             task.start()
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -261,7 +261,8 @@ extension UIViewController {
 
     @objc
     public func searchForUsers(withQuery query: String) {
-        self.performSearch(query: query, options: [.conversations, .contacts, .teamMembers, .directory])
+        let trimmedQuery = query.trimmingCharacters(in: .whitespaces)
```
billypchan: we can group this trimming code as a string's property in extension to prevent repeated code.

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -261,7 +261,8 @@ extension UIViewController {
...
+        self.performSearch(query: trimmedQuery, options: [.conversations, .contacts, .teamMembers, .directory])
```
typfel: There are `searchForLocalUsers()` and `searchForServices()` also call performSearch(). We would catch them all if we do the trimming inside `performSearch()`.

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
KaterinaWire: Done

PR #3987, opened 2020-01-27
fix: crash when dismiss search screen

billypchan:
Fix a potential crash when the use dismisses the search screen and the keyboard is visible caused by `SearchResultsView` may be freed after before the animation starts.  (keyboardFrameDidChange may be called during the dismissal)
Convert `UIView+Zeta` and its callers to Swift.


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -24,6 +24,50 @@ extension ConversationInputBarViewController {
     func postImage(_ image: MediaAsset) {
         sendController.sendMessage(withImageData: image.data())
     }
+    
+    ///TODO: chnage to didSet after ConversationInputBarViewController is converted to Swift
```
johnxnguyen: This is a TODO for later?

billypchan: Yes, I leave here and it should be refactored after this class is converted to Swift

PR #3989, opened 2020-01-27
[Fix] Change the display of the verified shield.

KaterinaWire:
### Issues
Verified shield shows up without having verified own devices.

### Causes
Do not check own devices.

### Solutions
Check for own verified devices for selfUser().


Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
```diff
@@ -50,6 +50,16 @@ class UserCell: SeparatorCollectionViewCell, SectionListCellType {
     
     weak var user: UserType? = nil
     
+    private var showVerifiedShield: Bool {
```
johnxnguyen: I suggest to name this `shouldShowVerifiedShield` because it's not an action but rather a query.

typfel:
Since this bug comes from the fact that rules for when we show the verified badge have changed, I would propose to update `ZMUser.trusted` in data model to take your own devices into account.

Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
KaterinaWire: Done

KaterinaWire:
> Since this bug comes from the fact that rules for when we show the verified badge have changed, I would propose to update `ZMUser.trusted()` in data model to take your own devices into account.

@typfel Yes, it makes sense.

typfel:
Changing the definition of  `trusted()` will have unwanted side effects on  `ConversationChangeInfo.usersThatCausedConversationToDegrade`. I still don't think we should have this business logic in the UI though. How about adding a new `verified: Bool` property on the `UserType` which also takes the self user into account.

typfel:
BTW: a very similar method logic can also be found in `ProfileViewControllerViewModel.showVerifiedShield`

KaterinaWire:
@typfel  I like the idea of ​​adding `verified: Bool` property to the `UserType `, since we are going to use  `UserType` instead of `ZMUser` in the UI.

typfel:
> @typfel I like the idea of ​​adding `verified: Bool` property to the `UserType `, since we are going to use `UserType` instead of `ZMUser` in the UI.

Good! Then let's take approach.

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -72,16 +72,11 @@ final class ProfileViewControllerViewModel: NSObject {
...
     
-    var showVerifiedShield: Bool {
-        if let user = fullUser {
-            let showShield = user.trusted() &&
-                !user.clients.isEmpty &&
-                context != .deviceList &&
-                ZMUser.selfUser().trusted()
-            return showShield
+    var shouldShowVerifiedShield: Bool {
+        guard let user = fullUser else {
```
typfel: No need to unwrap to the `fullUser` since `isVerified` exists on the `UserType`

typfel: Unrelated to this PR but I think there's no need for the `fullUser` anymore in this class (ping @johnxnguyen )

KaterinaWire: fullUser is optional, that's why I unwrapped it.
Otherwise, I need to do (fullUser?.isVerified ?? false)

typfel: use the `bareUser`, fullUser is just casting it to ZMUser.

KaterinaWire: Fixed

Wire-iOS Tests/CallInfoConfigurationTests.swift
```diff
@@ -99,7 +99,8 @@ class CallInfoConfigurationTests: XCTestCase {
         let mockVoiceChannel = MockVoiceChannel(conversation: mockConversation)
         let fixture = CallInfoTestFixture(otherUser: mockConversation.connectedUser!)
         
-        ((mockConversation.sortedActiveParticipants.last as Any) as? MockUser)?.untrusted = true
+        ((mockConversation.sortedActiveParticipants.first as Any) as? MockUser)?.isTrusted = true
+        ((mockConversation.sortedActiveParticipants.last as Any) as? MockUser)?.isTrusted = false
```
KaterinaWire: There are 2 participants in MockConversation: self user and otherUser. The previous property `untrusted` was false by default. As a result, selfUser was trusted (untrusted == false) and otherUser (mockConversation.sortedActiveParticipants.last) was set as untrusted.
Now we have only one property `isTrusted` and by default it's false.

Wire-iOS Tests/CallInfoConfigurationTests.swift
```diff
@@ -116,7 +117,8 @@ class CallInfoConfigurationTests: XCTestCase {
         let mockVoiceChannel = MockVoiceChannel(conversation: mockConversation)
         let fixture = CallInfoTestFixture(otherUser: mockConversation.connectedUser!)
         
-        ((mockConversation.sortedActiveParticipants.last as Any) as? MockUser)?.untrusted = true
+        ((mockConversation.sortedActiveParticipants.first as Any) as? MockUser)?.isTrusted = true
+        ((mockConversation.sortedActiveParticipants.first as Any) as? MockUser)?.isTrusted = false
```
johnxnguyen: Is one of these supposed to be fore the `last`?

KaterinaWire: Yes

Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
johnxnguyen: @typfel looks like the `fullUser` is still used in several places in this file, but we'll clean it up when we start digging into the UI more.

PR #3999, opened 2020-01-28
retire DeveloperMenuState

billypchan:
## What's new in this PR?

Retire `DeveloperMenuState` and read the config with `Bundle`.


typfel:
The usage  of `enableDeveloperMenu` has lost its meaning since it's now used control all sorts of debug related controls. 

I would propose to rename to to `developerModeEnabled`.

WireCommonComponents/AppCenter.swift
```diff
@@ -57,6 +57,9 @@ public extension Bundle {
         return Bundle.appMainBundle.infoForKey("UseAppCenter") == "1"
     }
 
+    static var enableDeveloperMenu: Bool {
```
typfel: I think `AppCenter.swift` is not the right place for this property.

billypchan: moved to a new file under Wire iOS target

Wire-iOS/Sources/Components/Settings.swift
```diff
@@ -1,32 +1,30 @@
...
-#else
-    return NO;
-#endif
+extension Settings {
+    @objc
+    func startLogging() {
+        #if targetEnvironment(simulator)
+        loadEnabledLogs()
+        #else
+        ZMSLog.startRecording(withIsInternal: Bundle.enableDeveloperMenu)
```
typfel:
```suggestion
        ZMSLog.startRecording(withIsInternal: Bundle.developerModeEnabled)
```

?

PR #4001, opened 2020-01-28
set status bar style and hidden with child for statusbar style/hidden method

billypchan:
## What's new in this PR?

Continue https://github.com/wireapp/wire-ios/pull/3730

Retire outdated`UIApplication.setStatusBarStyle/Hidden` method.
Change `UIViewControllerBasedStatusBarAppearance` to `YES` to let every view controller has its own status bar setting.
Define `childForStatusBarStyle/childForStatusBarHidden` for container view controllers.

Update snapshots which should have status bar spacings.

### TODO
- [x] fix the issue that when an alert is shown on convo list the status bar is darkened issue.
- [x] update snapshot with status bar spacing


Wire-iOS/Sources/Authentication/Interface/ViewControllers/AuthenticationCredentialsViewController.swift
```diff
@@ -144,6 +144,10 @@ final class AuthenticationCredentialsViewController: AuthenticationStepControlle
     override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
         return wr_supportedInterfaceOrientations
     }
+    
+    override var prefersStatusBarHidden: Bool {
+        return false
```
typfel: Isn't this the default, or why do we override this?

Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
```diff
@@ -51,10 +51,7 @@ final class ClearBackgroundNavigationController: UINavigationController {
         }
     }
     
-    override var preferredStatusBarStyle: UIStatusBarStyle {
-        return topViewController?.preferredStatusBarStyle ?? .lightContent
-    }
-    
+    ///TODO: child?
```
typfel: Since ClearBackgroundNavigationController is a subclass of UINavigationController, which already is a container view controller, it will already handle this?

Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
```diff
@@ -105,3 +70,12 @@ public extension UIApplication {
     }
 }
 
+extension UINavigationController {
```
typfel: Is this necessary? UINavigationController doesn't have sensible default for this?

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -312,8 +316,7 @@ final class AppRootViewController: UIViewController {
                     viewController.didMove(toParent: self)
                     previousViewController.removeFromParent()
                     self.visibleViewController = viewController
-                    UIApplication.shared.wr_updateStatusBarForCurrentControllerAnimated(true)
-                    completionHandler?()
+                        completionHandler?()
```
typfel: Incorrect Indentation?

typfel:
You've some testing to verify the status bar still behaves like it should on a device?

Wire-iOS/Sources/Authentication/Interface/ViewControllers/AuthenticationCredentialsViewController.swift
billypchan: remove all unnecessary override

Wire-iOS/Sources/UserInterface/Components/Controllers/ClearBackgroundNavigationController.swift
billypchan: checked that it is handled

typfel:
What causes the snapshots to change? Setting `UIViewControllerBasedStatusBarAppearance` to `YES`?

Wire-iOS/Sources/UserInterface/Helpers/UIApplication+StatusBar.swift
billypchan: checked that this override is necesary. Related discussion:
https://stackoverflow.com/questions/41026514/preferred-status-bar-style-of-view-controller-is-ignored-when-in-navigation-cont

billypchan:
@typfel I believe It is related the navigation bar in the snapshots.

PR #4014, opened 2020-01-31
convert TokenFieldDelegate

billypchan:
## What's new in this PR?

convert TokenFieldDelegate to Swift

### DIFF
--- a/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift
+++ b/Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift
@@ -32,18 +32,8 @@ protocol Shareable {
     func previewView() -> UIView?
 }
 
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/Components/ShareViewController/ShareViewController.swift
```diff
@@ -241,4 +231,18 @@ final class ShareViewController<D: ShareDestination, S: Shareable>: UIViewContro
     @objc func keyboardFrameDidChange(notification: Notification) {
         updatePopoverFrame()
     }
+
```
typfel: Let's keep the `// MARK: - TokenFieldDelegate` ?

billypchan: fixed

PR #4016, opened 2020-01-31
Retire KeyValueObserver

billypchan:
Retire `KeyValueObserver` and replace with Swift `observe`


Wire-iOS/Sources/Managers/MediaPlaybackManager.swift
```diff
@@ -115,16 +117,18 @@ final class MediaPlaybackManager: NSObject, AVSMedia {
     
     // MARK: - Active Media Player State
     private func startObservingMediaPlayerChanges() {
-        titleObserver = KeyValueObserver.observe(activeMediaPlayer as Any, keyPath: "title", target: self, selector: #selector(activeMediaPlayerTitleChanged(_:)), options: [.initial, .new])
+        setObserver(activeMediaPlayer as? NSObject & MediaPlayer)
```
typfel: Why can't call `observe` here directly like you do the other cases?

billypchan: It is a work-a-round for compiler error. If I directly write the code like ` activeMediaPlayer.observe(...)`, since activeMediaPlayer is declared as a `MediaPlayer` abstract type and not necessarily a `NSObject`, compiles fails due to it does not know `activeMediaPlayer` has the method `observe`.

typfel: Ok

typfel: It doesn't work to change the  property declaration to:
`dynamic private(set) weak var activeMediaPlayer: (MediaPlayer & NSObject)?`

billypchan: Tested changing the `activeMediaPlayer` type but it still gives error:
`Generic parameter 'Value' could not be inferred`

typfel: Ok

PR #4018, opened 2020-01-31
Convert image utilities

billypchan:
Convert `UIImage+ImageUtilities` to Swift

### Discussion
- [ ] should these methods be moved to wire-utilities?


Wire-iOS/Sources/Helpers/UIImage+ImageUtilities.swift
```diff
@@ -32,4 +30,124 @@ extension UIImage {
...
+
+        return image
+    }
+
+    class func deviceOptimizedImage(from imageData: Data) -> UIImage? {
+        return UIImage(from: imageData, withMaxSize: UIScreen.main.nativeBounds.size.height)
+    }
+
+    convenience init?(from imageData: Data, withMaxSize maxSize: CGFloat) {
+        guard let source: CGImageSource = CGImageSourceCreateWithData(imageData as CFData, nil),
```
typfel: We are no longer releasing the `source` and `scaledImage`. Is that not necessary in Swift?

Wire-iOS/Sources/Helpers/UIImage+ImageUtilities.swift
```diff
@@ -32,4 +30,124 @@ extension UIImage {
...
+
+        if let height = properties[kCGImagePropertyPixelHeight] as? CGFloat,
+            let width = properties[kCGImagePropertyPixelWidth] as? CGFloat {
+            return CGSize(width: width, height: height)
+        }
+
+        return .zero
+    }
+
+    convenience init?(from imageData: Data, withShorterSideLength shorterSideLength: CGFloat) {
```
typfel: Same question here about releasing resources.

Wire-iOS/Sources/Helpers/UIImage+ImageUtilities.swift
```diff
@@ -32,4 +30,124 @@ extension UIImage {
...
+            longSideLength = shorterSideLength * (size.height / size.width)
+        }
+
+        guard let scaledImage = CGImageSourceCreateThumbnailAtIndex(source, 0, UIImage.thumbnailOptions(withMaxSize: longSideLength)) else { return nil }
+
+        ///TODO: read screen scale
+        self.init(cgImage: scaledImage, scale: 2.0, orientation: .up)
+    }
+
+    class func image(color: UIColor, andSize size: CGSize) -> UIImage? {
```
typfel: This could also be a `convenience init`, that would keep the old method signature.

Wire-iOS/Sources/Helpers/UIImage+ImageUtilities.swift
billypchan: Yes, the compiler warning message said it is no need to call release method.

Wire-iOS/Sources/Helpers/UIImage+ImageUtilities.swift
billypchan: It is not allowed in Swift since `init` is not called.

typfel: It's allowed if you end with:

`self = image` instead of `return image`

billypchan: `self` is immutable, the work-a-round is:
`self.init(cgImage: (image?.cgImage)!)`

typfel: Ah I see, the `self =` is only works with structs

PR #4021, opened 2020-02-03
Convert self device update alert

billypchan:
Convert `UIAlertController+NewSelfClients` to Swift

## TODO
- [ ] snapshot test for the alert


Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController+NewDeviceAlert.swift
```diff
@@ -92,3 +92,33 @@ extension SelfProfileViewController {
...
 }
+
+extension UIAlertController {
+    convenience init(forNewSelfClients clients: Set<UserClient>) {
+        var deviceNamesAndDates: [String] = []
+        
+        for userClient in clients {
+            let deviceName: String?
+            
+            if userClient.model?.isEmpty == false {
```
typfel: This could be an `if let`, then you could declare  `deviceName` as non-optional.

billypchan: fixed

PR #4022, opened 2020-02-03
Snapshot self device alert

billypchan:
Refactor `SelfProfileViewControllerSnapshotTests` with Swift snapshot framework
Add snapshot test for new self clients alert.

### Discussion:
- [ ] `readReceiptsEnabledChangedRemotely` should be added to `UserType`?


Wire-iOS Tests/DeviceManagement/ZMSnapshotTestCase+MockUserClient.swift
```diff
@@ -18,12 +18,12 @@
...
 
-extension ZMSnapshotTestCase {
+extension NSManagedObjectContext {
     func mockUserClient(fingerprintString: String = "102030405060708090102030405060708090102030405060708090") -> UserClient! {
-        let client = UserClient.insertNewObject(in: uiMOC)
+        let client = UserClient.insertNewObject(in: self)
         client.remoteIdentifier = "102030405060708090"
 
-        client.user = ZMUser.insertNewObject(in: uiMOC)
+        client.user = ZMUser.insertNewObject(in: self)
```
johnxnguyen: There exists a `UserClientType` abstraction defined in DataModel. It would be nice if we use it to mock objects so we don't have to have CoreData in our tests.

billypchan: replaced with mockUserClient in SelfProfileViewControllerSnapshotTests, will retire this method in separated PR.

Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -18,14 +18,21 @@
 
 import UIKit
 
+///TODO: add readReceiptsEnabledChangedRemotely to UserType
+protocol EditableUserReadReceiptsStatus {
+    var readReceiptsEnabledChangedRemotely: Bool { get }
+}
```
johnxnguyen: I'm not sure that it's a good idea to add this to `UserType`. The `UserType` shouldn't expose properties used primarily for business logic.

Wire-iOS Tests/DeviceManagement/ZMSnapshotTestCase+MockUserClient.swift
johnxnguyen: So the plan is to keep using this method for now, and remove it in another PR?

Wire-iOS Tests/Mocks/SwiftMockUser.swift
```diff
@@ -43,7 +43,13 @@ final class MockSelfUser: SwiftMockUser, SelfLegalHoldSubject {
 
 /// a new simple Mock User without objc
 class SwiftMockUser: NSObject, UserType {
-    var name: String? = nil
+    var name: String? = nil {
```
johnxnguyen: https://github.com/wireapp/wire-ios/pull/4066 replaces the `SwiftMockUser` with `MockUserType`. Perhaps we should wait for it to be merged first and update changes here after?

PR #4026, opened 2020-02-03
[Fix] landing page backend title

David-Henner:
### Issues

The welcome message on the landing page uses the title from the backend configuration on Staging and Production, where it should only be used for custom backends

### Solutions

Use custom title only for custom backend


Wire-iOS/Sources/Authentication/Landing/LandingViewController.swift
```diff
@@ -383,11 +383,13 @@ class LandingViewController: AuthenticationStepViewController {
         }
     }
     
-    private func updateCustomBackendLabel() {
+    private func updateCustomBackendLabels() {
         switch BackendEnvironment.shared.environmentType.value {
         case .production, .staging:
             customBackendStack.isHidden = true
+            messageLabel.text = "landing.welcome_message".localized(args: "Wire")
```
typfel: We shouldn't hardcode this since we will need adapt it for white label builds. Maybe we could use `WIRE_PRODUCT_NAME`

PR #4027, opened 2020-02-04
Fix: Alert not shown

billypchan:
Fix for the issue that alert is not shown on `notificationWindow`. Affect alerts: warnAboutNoInternetConnection, save image and remove user


Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
```diff
@@ -103,11 +103,11 @@ extension ZMConversation {
     func warnAboutNoInternetConnection() -> Bool {
         if AppDelegate.isOffline {
             let internetConnectionAlert = UIAlertController(title: "voice.network_error.title".localized, message: "voice.network_error.body".localized, cancelButtonTitle: "general.ok".localized)
-            AppDelegate.shared.notificationsWindow?.rootViewController?.present(internetConnectionAlert, animated: true)
+            AppDelegate.shared.window?.rootViewController?.present(internetConnectionAlert, animated: true)
```
typfel: Weren't we presenting it on the notification window because the calling UI is displayed in the notification window and the alert wouldn't be visible when in a call.

billypchan: This alert also presented when the user press the calling button in conversation screen. We should test in both cases.

billypchan: Refactor the code and showing the alert in `notificationsWindow`

Wire-iOS/Sources/Helpers/syncengine/ZMConversation+Calling.swift
```diff
@@ -101,13 +101,19 @@ extension ZMConversation {
...
-            AppDelegate.shared.notificationsWindow?.rootViewController?.present(internetConnectionAlert, animated: true)
-            return true
-        } else {
+        guard AppDelegate.isOffline else {
             return false
         }
+        
+        let internetConnectionAlert = UIAlertController.alertWithOKButton(title: "voice.network_error.title".localized,
+                                                                          message: "voice.network_error.body".localized) { _ in
+            AppDelegate.shared.notificationsWindow?.isHidden = true
```
typfel: Wouldn't this potentially hide a call which displays an alert or that's not a scenario which can happen?

billypchan: This alert is shown when `startVideoCall` or `startAudioCall` is called. The call screen is not presented at the point.

typfel: But in that case isn't it unnecessary to present the alert on the notification window in first place? I was under the assumption that we presented an alert while the calling UI was on screen.

billypchan: Showing the alert on `notificationWindow`:
Simulator Screen Shot - iPhone Xʀ - 2020-02-05 at 14 26 08

Showing the alert in main window would be just fine since it is the fallback case and  call window would not show.

billypchan: @typfel Restore the code and show the alert on main window

PR #4034, opened 2020-02-05
convert country, CountryCodeResultsTableViewController and WIRESTAN macro

billypchan:
## What's new in this PR?

Convert `Country` and its callers to Swift.
Convert objc flag `WIRESTAN` macro as a Swift flag in `OTHER_SWIFT_FLAGS`, all the functions using this macro to Swift.
Convert `CountryCodeResultsTableViewController`


Wire-iOS/Sources/Authentication/Helpers/CountryCode/Country.swift
```diff
@@ -0,0 +1,196 @@
...
+@objc
+final class Country: NSObject {
+    let iso: String
+
+    let e164: UInt
+
+    class var defaultCountry: Country {
+
+        #if WIRESTAN
+        if UserDefaults.standard.string(forKey: "ZMBackendEnvironmentType") == "staging" {
```
typfel: This code seems a bit outdated since we don't store the `ZMBackendEnvironmentType` in .standard anymore. A better way to check this would be: `BackendEnvironment.shared.environmentType == .staging`

Wire-iOS/Sources/Authentication/Helpers/CountryCode/CountryCodeBaseTableViewController.m
```diff
@@ -18,49 +18,13 @@
 
 
 #import "CountryCodeBaseTableViewController.h"
-#import "Country.h"
+#import "Wire-Swift.h"
```
typfel: Can't we delete this file as well? It doesn't contain any code...

billypchan: `CountryCodeTableViewController` extends this class and it is written in objc, I can removed this class later.

Wire-iOS/Sources/Authentication/Helpers/CountryCode/Country.swift
billypchan: fixed

typfel:
Did you try it out in the phone registration?

billypchan:
Tested in both new account registration and edit phone number screen.

PR #4035, opened 2020-02-05
[Refactor] setup of Backend Environment on launch

David-Henner:
### Issues

- Custom backend environments are overridden by default backend environment on launch

### Causes

1. The `BackendEnvironment` implementation relies on `UserDefaults` to retrieve the persisted environment type.
2.  `setupBackendEnvironment` method modifies `UserDefaults` directly after launch to override the environment type.

If the goal is to override the environment type, it works.
However, if we have a custom backend, it will be overridden on launch, which isn't wanted.

### Solutions

- Replace the key used in the launch arguments to one that is specific to its use case
- Add interface to `AutomationHelper` to retrieve and persist the backend type override
- Add interface to `BackendEnvironment` extension to set backend type from a string describing the type


Cartfile.resolved
```diff
@@ -25,6 +25,6 @@ github "wireapp/wire-ios-share-engine" "168.0.0"
 github "wireapp/wire-ios-sync-engine" "297.1.0"
 github "wireapp/wire-ios-system" "28.1.0"
 github "wireapp/wire-ios-testing" "20.1.0"
-github "wireapp/wire-ios-transport" "66.1.1"
+github "wireapp/wire-ios-transport" "66.2"
```
billypchan: `Cartfile`'s data-model minor version is also updated?

David-Henner: data-model isn't in the `Cartfile`

billypchan: I was thinking some conflict occurs in the compounds.

David-Henner: Makes sense, that's the reason for the build failures. I wrote `66.2` instead of `66.2.0`

WireCommonComponents/AutomationHelper+BackendEnvironment.swift
```diff
@@ -0,0 +1,36 @@
...
+extension AutomationHelper {
+    private static let backendEnvironmentTypeOverrideKey = "BackendEnvironmentTypeOverrideKey"
+    
+    public func backendEnvironmentTypeOverride() -> String? {
+        return UserDefaults.standard.string(forKey: AutomationHelper.backendEnvironmentTypeOverrideKey)
+    }
+    
+    public func persistBackendTypeOverrideIfNeeded(with type: String?) {
+        guard shouldPersistBackendType else { return }
+        UserDefaults.standard.set(type, forKey: AutomationHelper.backendEnvironmentTypeOverrideKey)
```
typfel: The override won't work in the share extension since it's saved in `.standard`

PR #4039, opened 2020-02-06
[Fix] hide phone login if custom backend

David-Henner:
## What's new in this PR?

Hide phone login tab if we're on custom backend


Wire-iOS/Sources/Authentication/Interface/ViewControllers/AuthenticationCredentialsViewController.swift
```diff
@@ -148,6 +148,8 @@ final class AuthenticationCredentialsViewController: AuthenticationStepControlle
     func configure(with featureProvider: AuthenticationFeatureProvider) {
         if case .reauthentication? = flowType {
             tabBar.isHidden = prefilledCredentials != nil
+        } else if case .custom(url: _) = BackendEnvironment.shared.environmentType.value {
```
johnxnguyen: I think you can omit the `(url: _)` part of the pattern if you don't use it.

PR #4043, opened 2020-02-07
[Fix] Don't assume list of call participant doesn't contain any duplicates

typfel:
### Issues

The app sometimes crashes in group calls.

### Causes

The `CallHapticsController` tries to generate a dictionary with the call participants using the call participant itself as the key, this crashes if a call participant exists twice in the list.

### Solutions

Don't assume that call participant list doesn't contain any duplicates and employ a uniquing strategy.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
@@ -86,9 +86,7 @@ final class CallHapticsController {
     }
     
     private func createVideoStateMap(using participants: [CallParticipant]) -> [CallParticipant : Bool] {
-        return Dictionary(uniqueKeysWithValues: participants.map {
-            ($0, $0.state.isSendingVideo)
-        })
+        return Dictionary(participants.map { ($0, $0.state.isSendingVideo) }, uniquingKeysWith: { (first, _) in first })
     }
 }
 


Wire-iOS/Sources/UserInterface/Calling/Haptic Feedback/CallHapticsController.swift
```diff
@@ -86,9 +86,7 @@ final class CallHapticsController {
     }
     
     private func createVideoStateMap(using participants: [CallParticipant]) -> [CallParticipant : Bool] {
-        return Dictionary(uniqueKeysWithValues: participants.map {
-            ($0, $0.state.isSendingVideo)
-        })
+        return Dictionary(participants.map { ($0, $0.state.isSendingVideo) }, uniquingKeysWith: { (first, _) in first })
```
johnxnguyen: So we're only using the participant's user id as the key?

typfel: We use user id and client id:
https://github.com/wireapp/wire-ios-sync-engine/blob/be18314915de9acba8cccba257f5c9b0407b02c8/Source/Calling/CallState.swift#L54

typfel: At the root I'd say this is a AVS bug because they give us the list call participants.

PR #4047, opened 2020-02-07
[Refactor] Inviting address book contacts

johnxnguyen:
## What's new in this PR?

This PR heavily refactors the address book contact feature available only for consumer users. It contains:

- Objective C to Swift conversions
- Merging of `InviteContactsViewController` and `ContactsViewController`
- Large grain simplifications
- Deletion of a lot of dead code
- Fixes some recent regressions (see Issues)

### Simplifications / Dead Code Clean Up

- There was some code relating to connecting to an address book contact. This seems to be left over from a distant age and was not being used, since public directory Wire accounts aren't included in the list of contacts.
- There was lots of boiler plate code for selecting and deselecting contacts. This wasn't being used and could be removed.

### Regressions

1. Going to the invite contacts screen while address book access is denied shows a blank screen.
2. Trying to invite an address book contact when no email or sms client installed just pops the view controller and does not show the alert.
3. Trying to invite an address book contact even with email or sms client just pops the view controller and does not show a modal overlay.

### Causes

1. We do actually try to display a screen asking for access permissions, but it was being hidden immediately after it was being shown. The reason is because some view setup code was in an initializer, which was hiding a view that was being unhidden in `viewDidLoad`. The assumption was that `viewDidLoad` would always be called after `init` was finished, but it's not the case.
2. We do actually try to show the alert, but we are also popping the view controller and so the alert doesn't get retained. We want to pop the view controller so that when opening a conversation transitions smoothly. However we don't need to pop the view controller when trying to invite someone.
3. Same as 2.

### Solutions

1. Do all view setup in the `viewDidLoad` method.
2. Refactored handling of button actions, which leads to different handling of inviting and opening conversations actions.
3. Same as 2.


Wire-iOS/Sources/UserInterface/ContactsUI/ContactsDataSource.swift
```diff
@@ -0,0 +1,155 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@objc protocol ContactsDataSourceDelegate: class {
```
typfel: Is still necessary to expose these protocols and  classes to Objective-C?

Wire-iOS Tests/ContactsCellSnapshotTests.swift
```diff
@@ -22,15 +22,10 @@ import XCTest
 final class ContactsCellSnapshotTests: ZMSnapshotTestCase {
 
     var sut: ContactsCell!
-    let buttonTitles = ["contacts_ui.action_button.open".localized,
-                        "contacts_ui.action_button.invite".localized,
-                        "connection_request.send_button_title".localized]
```
typfel: So we can delete this entry from the `.strings`file?

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
```diff
@@ -18,21 +18,206 @@
 
 import Foundation
 
-extension ContactsViewController {
-    @objc
-    func updateActionButtonTitles() {
-        actionButtonTitles = contentDelegate?.actionButtonTitles(for: self)
+class ContactsViewController: UIViewController {
```
billypchan:
```suggestion
final class ContactsViewController: UIViewController {
```

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
```diff
@@ -18,21 +18,206 @@
...
+    weak var delegate: ContactsViewControllerDelegate?
+
+    var bottomContainerView = UIView()
+    var bottomContainerSeparatorView = UIView()
+    var noContactsLabel = UILabel()
+    var searchHeaderViewController = SearchHeaderViewController(userSelection: .init(), variant: .dark)
+    var separatorView = UIView()
+    var tableView = UITableView()
+    var inviteOthersButton = Button(style: .empty, variant: ColorScheme.default.variant)
+    var emptyResultsLabel = UILabel()
```
billypchan: If they are not nil-ed in the code, we can change them to `let`

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsDataSource.swift
johnxnguyen: Ah yeah, forgot to remove those annotations.

Wire-iOS Tests/ContactsCellSnapshotTests.swift
johnxnguyen: We still use the strings, they're just located inside the cell now instead of being injected into them.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
johnxnguyen: Done.

Wire-iOS/Sources/UserInterface/ContactsUI/ContactsViewController.swift
johnxnguyen: Done

PR #4049, opened 2020-02-07
Convert TokenizedTextView

billypchan:
Convert `TokenizedTextView` and its callers to Swift


Wire-iOS/Sources/Components/TokenField/TokenField+DefaultAppearance.swift
```diff
@@ -19,6 +19,38 @@
 import UIKit
 
 extension TokenField {
+
+    @objc
+    func setupSubviews() {
```
typfel: Why are these TokenFields necessary? Is it not possible to convert `TokenizedTextView` without touching the `TokenField`

billypchan: The conversion here is for assigning a non objc `tokenizedTextViewDelegate`. The original objc code cast `delegate` (UITextViewDelegate) to `TokenizedTextViewDelegate` and I separated them here.

typfel: Ok 👍

PR #4050, opened 2020-02-07
convert TextView

billypchan:
## What's new in this PR?

Convert `TextView` to Swift


Wire-iOS/Sources/UserInterface/Components/TextView/TextView.swift
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    // MARK: - Copy/Pasting
+    override func paste(_ sender: Any?) {
+        let pasteboard = UIPasteboard.general
+        zmLog.debug("types available: \(pasteboard.types)")
+
+        if pasteboard.hasImages,
+            let image = UIPasteboard.general.mediaAsset() {
+            (delegate as? TextViewProtocol)?.textView(self, hasImageToPaste: image)
```
johnxnguyen: I don't think it's a good idea to couple the `TextViewProtocol` to the `UITextViewDelegate` because they're not codependent. We could take this opportunity to separate them (and perhaps find a better name than `TextViewProtocol`).

billypchan: fixed

johnxnguyen:
@billypchan looks like a merge went wrong, you've got unrelated commits.

Wire-iOS/Sources/UserInterface/Components/TextView/TextView.swift
```diff
@@ -0,0 +1,210 @@
...
+    func textView(_ textView: UITextView, hasImageToPaste image: MediaAsset)
+    func textView(_ textView: UITextView, firstResponderChanged resigned: Bool)
+}
+
+// Inspired by https://github.com/samsoffes/sstoolkit/blob/master/SSToolkit/SSTextView.m
+// and by http://derpturkey.com/placeholder-in-uitextview/
+@objc
+class TextView: UITextView {
+
+    weak var textViewDelegate: TextViewProtocol?
```
johnxnguyen: I find this a bit confusing and I think we could use a better name. At least something like `InformalTextViewDelegate`.

billypchan: renamed.

PR #4054, opened 2020-02-10
[Feature] Automatic SSO flow 

David-Henner:
## What's new in this PR?

Start the SSO flow automatically upon user clicking on `Login with SSO` button.
- Fetch SSO Settings from BE
- Fallback on manual SSO flow if no SSO code 

Note: Did some refactoring to make the flow more understandable

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
+++ b/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
@@ -243,11 +243,14 @@ extension CompanyLoginController {
     /// Fetches SSO code and starts flow automatically if code is returned on completion
     /// Otherwise, falls back on prompting user for SSO code
     func startAutomaticSSOFlow() {
-        SessionManager.shared?.activeUnauthenticatedSession.fetchSSOSettings { [displayCompanyLoginPrompt, attemptLoginWithSSOCode] result in
+        delegate?.controller(self, showLoadingView: true)
+        SessionManager.shared?.activeUnauthenticatedSession.fetchSSOSettings { [weak self] result in
+            guard let `self` = self else { return }
+            self.delegate?.controller(self, showLoadingView: false)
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -292,12 +290,35 @@ import WireCommonComponents
                 return
             }
             BackendEnvironment.shared = backendEnvironment
-            notifyDelegate()
+            self.delegate?.controllerDidUpdateBackendEnvironment(self)
```
johnxnguyen: This call was previously run in on the main thread. Is it still the case or is it not necessary?

David-Henner: The completion block the call is defined in is running on the main thread already

Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -247,15 +235,34 @@ import WireCommonComponents
...
+extension CompanyLoginController {
     
+    /// Fetches SSO code and starts flow automatically if code is returned on completion
+    /// Otherwise, falls back on prompting user for SSO code
+    func startAutomaticSSOFlow() {
+        delegate?.controller(self, showLoadingView: true)
+        SessionManager.shared?.activeUnauthenticatedSession.fetchSSOSettings { [weak self] result in
+            guard let `self` = self else { return }
+            self.delegate?.controller(self, showLoadingView: false)
+            guard result.error == nil, let ssoCode = result.value?.ssoCode else {
```
typfel: Maybe a switch would be easier to read when acting on a `Result`.

```
switch result {
 case .success(value), let ssoCode = value.ssoCode:
  self.attemptLoginWithSSOCode(ssoCode)
default:
  self.displayCompanyLoginPrompt(ssoOnly: true)
}
```

PR #4067, opened 2020-02-13
Convert NSString+Fingerprint

billypchan:
Convert `NSString+Fingerprint` to Swift


Wire-iOS/Sources/UserInterface/Helpers/String+Fingerprint.swift
```diff
@@ -0,0 +1,49 @@
...
+    func split(every: Int) -> [String] {
+        var result = [String]()
+
+        for i in stride(from: 0, to: count, by: every) {
+            let start = index(startIndex, offsetBy: i)
+            let end = index(start, offsetBy: every, limitedBy: endIndex) ?? endIndex
+            result.append(String(self[start..<end]))
+        }
+
+        return result
```
johnxnguyen: Could you use `map` instead?

Wire-iOS Tests/StringFingerprintTests.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    func testThatFingerprintSplitsProperlyFor4() {
+        // given
+        let testStrings = ["abc", "mfngsdnfgljsfgjdns", "!!@#!@#!@#AASDF", ""]
+
+        let resultStrings = ["abc", "mfng sdnf gljs fgjd ns", "!!@# !@#! @#AA SDF", ""]
+
+        for i in 0..<testStrings.count {
+            // when
+
```
johnxnguyen: Please clean up the vertical whitespace.

Wire-iOS Tests/String+FingerprintTests.swift
```diff
@@ -0,0 +1,48 @@
...
+@testable import Wire
+
+final class String_FingerprintTests: XCTestCase {
+    func testFingerprintAttributes() {
+        let regularAttributes = [
+                                 NSAttributedString.Key.font: UIFont.systemFont(ofSize: UIFont.systemFontSize)
+                                 ]
+        let boldAttributes = [
+                              NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: UIFont.systemFontSize)
+                              ]
```
johnxnguyen: Please clean up the vertical whitespace.

Wire-iOS Tests/String+FingerprintTests.swift
```diff
@@ -0,0 +1,48 @@
...
+        let attributedString = fingerprintString.fingerprintString(attributes: regularAttributes, boldAttributes: boldAttributes)
+
+        var bold = true
+        attributedString.enumerateAttributes(in: NSRange(location: 0, length: attributedString.length), options: [], using: { attrs, range, stop in
+            let stringInRange = ((attributedString.string as NSString?)?.substring(with: range))?.trimmingCharacters(in: CharacterSet.whitespaces)
+            if (stringInRange?.count ?? 0) == 0 {
+                return
+            }
+            XCTAssertEqual(attrs as? [NSAttributedString.Key: UIFont], bold ? boldAttributes : regularAttributes)
```
johnxnguyen: I think this could be simplified by avoiding to enumerate attributes and instead checking attributes over ranges directly. For example

```swift
stride(from: 0, to: 20, by: 6).forEach {
  let boldRange = NSRange(location: $0, length: 2)
  // Assert attributes
}

stride(from: 3, to: 20, by: 6).forEach {
  let regularRange = NSRange(location: $0, length: 2)
  // Assert attributes
}
```

Wire-iOS/Sources/UserInterface/Helpers/String+Fingerprint.swift
billypchan: fixed

Wire-iOS Tests/String+FingerprintTests.swift
```diff
@@ -0,0 +1,51 @@
...
+        let boldAttributes = [NSAttributedString.Key.font: UIFont.boldSystemFont(ofSize: UIFont.systemFontSize)]
+
+        // WHEN
+        let attributedString = fingerprintString.fingerprintString(attributes: regularAttributes, boldAttributes: boldAttributes)
+
+        // THEN
+        stride(from: 0, to: 20, by: 6).forEach {
+            var boldRange = NSRange(location: $0, length: 2)
+
+            let attrs = attributedString.attributes(at: $0, effectiveRange: &boldRange)
```
johnxnguyen: The effective range will store the range of the attributes that it finds, rather than returning the attributes in the passed in range.

What you want to do is create the attributed substring with the `boldRange` and then ask for the attributes over the whole substring. Like...

```swift
let substring = attributedString.attributedSubstring(from: boldRange)
let attrs = substring.attributes(at: boldRange.location, effectiveRange: nil)
...
```

Wire-iOS Tests/String+FingerprintTests.swift
```diff
@@ -0,0 +1,51 @@
...
+            XCTAssertEqual(attrs as? [NSAttributedString.Key: UIFont], boldAttributes)
+        }
+
+        stride(from: 3, to: 20, by: 6).forEach {
+            var regularRange = NSRange(location: $0, length: 2)
+            let attrs = attributedString.attributes(at: $0, effectiveRange: &regularRange)
```
johnxnguyen: Same as above

Wire-iOS/Sources/UserInterface/Helpers/String+Fingerprint.swift
```diff
@@ -17,16 +17,11 @@
...
-            let end = index(start, offsetBy: every, limitedBy: endIndex) ?? endIndex
-            result.append(String(self[start..<end]))
-        }
-
-        return result
+    
+    func split(every: Int) -> Array<String> {
+        return stride(from: 0, to: count, by: every).map {
+            String(Array(Array(self)[$0..<min($0 + every, count)]))
+            }
```
johnxnguyen: This is a bit more confusing. I meant refactoring using map to avoid creating a temporary array and appending results to it. Like

```swift
stride(from: 0, to: count, by: every).map {
  let start = index(startIndex, offsetBy: $0)
  let end = ....
  return String(self[start..<end])
}
```

billypchan: refactoring number range to Index range.

PR #4068, opened 2020-02-13
Convert NSString URL helper

billypchan:
Refactor NSString URL helper to URL and String extension. 
Move the new extension to Wire-iOS target.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/ArticleView.swift
```diff
@@ -127,10 +127,10 @@ final class ArticleView: UIView {
     }
     
     private func formatURL(_ URL: Foundation.URL) -> NSAttributedString {
-        let urlWithoutScheme = URL.absoluteString.removingURLScheme(URL.scheme!)
-        let displayString = urlWithoutScheme.removingPrefixWWW().removingTrailingForwardSlash()
+        let urlWithoutScheme = URL.urlWithoutScheme
```
typfel: Would it make sense to use URLComponents to modify the URL instead? https://developer.apple.com/documentation/foundation/urlcomponents

billypchan: I checked the API and found that there is no method to remove the leading scheme & "://" and "www.", the custom remover is needed.

typfel: Ok

PR #4072, opened 2020-02-13
Convert AVSLogObserver

billypchan:
## What's new in this PR?

Convert AVSLogObserver

### dependency
- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/1167


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -85,7 +84,7 @@ final class AppRootViewController: UIViewController {
     override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
         appStateController = AppStateController()
         fileBackupExcluder = FileBackupExcluder()
-        avsLogObserver = AVSLogObserver()
+        SessionManager.startAVSLogging()
```
typfel: Why not start and stop the logging inside `SessionManager`.

billypchan: I will move it to `SessionManager` init method. `stopAVSLogging` is unnesscessary and I will remove it.

billypchan: @typfel After testing, I would prefer calling `SessionManager.startAVSLogging()` when `AppRootViewController` is init. An `SessionManager` is created in `AppRootViewController.launch` method and some logs may lose.

The first messages logged in current version:
`"AVS inited with flags=0x0 [avs 5.4.199 (x86_64/darwin)]\n"`

If we call `SessionManager.startAVSLogging()` in `SessionManager.init`, the first log message is:
`"AVSMediaManager: setUiStartsAudio: no\n"`

typfel: OK

PR #4076, opened 2020-02-14
[Refactor] Getting the self user

johnxnguyen:
## What's new in this PR?

New additions in https://github.com/wireapp/wire-ios-sync-engine/pull/1166 provide us a new way to get a reference to the current self user.

This PR contains changes to integrate this machinery by configuring the `SelfUser` helper class. 

### Issues

Currently, when we invoke `ZMUser.selfUser()`, a `ZMUser` instance is dynamically returned depending on whether we're running tests or not. This is confusing and can lead to several mock "self users" being in scope, such as one created locally that is passed around and the one created automatically when `ZMUser.selfUser()` is invoked somewhere in the code being tested. Since we never really know when `ZMUser.selfUser()` is being called in which tests, it can lead to unexpected results.

### Solutions

Admittedly, having an easy way to get the self user is extremely helpful. Currently we do this with `ZMUser.selfUser()`, but this PR puts in place `SelfUser.current` which we will migrate to.

`SelfUser` is configured with a `SelfUserProvider` as soon as it becomes available. This is done in the `AppRootViewController` when we transition to an authenticated state by setting `SelfUser.provider`.

However, this is only done automatically when running the production code. If we're running tests, then we need to set it explicitly, otherwise a call to `SelfUser.current` will intentionally crash. This will help the developer discover when they need to mock the self user and provide it. It is likely that the test class that needs to provide the self user will conform to `SelfUserProvider` and call `SelfUser.provider = self` in the test setup.

### Needs Releases With 

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1166
- [x] https://github.com/wireapp/wire-ios-share-engine/pull/149

### Notes

Adoption of this new system will be completed in a separate PR.


johnxnguyen:
@typfel I realize that we should probably set `SelfUser.provider = nil` when we no longer have any active user sessions, but I'm unsure which of the app states this corresponds to. All except `authenticated`?

typfel:
> @typfel I realize that we should probably set `SelfUser.provider = nil` when we no longer have any active user sessions, but I'm unsure which of the app states this corresponds to. All except `authenticated`?

Yes

Wire-iOS Tests/Mocks/MockUser+SelfUserFromSession.swift
```diff
@@ -19,7 +19,7 @@
 import Foundation
 @testable import Wire
 
-extension MockUser: SelfUserProvider {
+extension MockUser: SelfUserProviderUI {
```
johnxnguyen: This is a temporary fix to avoid name classes with `SelfUserProvider` in SyncEngine. It will be removed entirely in the next PR.

Wire-iOS Tests/TestingAppDelegate.swift
```diff
@@ -0,0 +1,35 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+@testable import Wire
+
+/// The App Delegate to use for the test target.
+
+@objc(TestingAppDelegate)
+final class TestingAppDelegate: AppDelegate {
```
billypchan: Is it possible to create a new Test delegate instead of extending AppDelegate?

johnxnguyen: We could, if we're sure that it won't break tests. I can investigate it in the next PR.

PR #4077, opened 2020-02-14
[Fix] Name and avatar is not aligned correctly on iPad

KaterinaWire:
### Issues
Name and avatar is not aligned correctly on iPad.

### Causes
We use a smaller margin (as for iPhone) to display name and avatar on iPad.

### Solutions
Remove an extra check.


KaterinaWire:
> Did you tested on iPhone?
@billypchan 
Yes

PR #4079, opened 2020-02-17
[Refactor] Update URL action handling after the API changes in SessionManager

typfel:
## What's new in this PR?

After the URL action handling was refactored in https://github.com/wireapp/wire-ios-sync-engine/pull/1168 we need to update UI.

- Implement the `URLActionDelegate` which replaces `SessionManagerURLHandlerDelegate`
- The API for connecting to a bot changed
- the `openURL()` method is now throws and error if the URL is malformed

### Dependencies

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1168


Wire-iOS/Sources/AppRootViewController+URLActionDelegate.swift
```diff
@@ -0,0 +1,97 @@
...
+            alert.addAction(cancelAction)
+            
+            self.present(alert, animated: true, completion: nil)
+            
+        case .accessBackend(configurationURL: let configurationURL):
+            let alert = UIAlertController(title: "url_action.switch_backend.title".localized,
+                                          message: "url_action.switch_backend.message".localized(args: configurationURL.absoluteString),
+                                          preferredStyle: .alert)
+            let agreeAction = UIAlertAction(title: "general.ok".localized,
+                                            style: .default) { _ in
```
johnxnguyen: Putting `title` and `style` would make trailing block align much closer to the gutter, making it more readable.

PR #4083, opened 2020-02-17
fix: prevent SectionCollectionViewController out of bound cell access

billypchan:
## What's new in this PR?

guard for `SectionCollectionViewController` out of bound section access.


Wire-iOS/Sources/UserInterface/Components/Controllers/SectionCollectionViewController.swift
```diff
@@ -95,18 +95,30 @@ extension SectionCollectionViewController: UICollectionViewDataSource {
...
     
     func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        guard visibleSections.indices.contains(section) else { return 0 }
+
         return visibleSections[section].collectionView(collectionView, numberOfItemsInSection: 0)
     }
     
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard visibleSections.indices.contains(indexPath.section) else {
+            fatal("Unknown section")
```
johnxnguyen: Perhaps it's worth reporting the section index too, it might give some insight.

Wire-iOS/Sources/UserInterface/Components/Controllers/SectionCollectionViewController.swift
```diff
@@ -95,18 +95,30 @@ extension SectionCollectionViewController: UICollectionViewDataSource {
...
+        }
+
         return visibleSections[indexPath.section].collectionView(collectionView, cellForItemAt:indexPath)
     }
     
     func collectionView(_ collectionView: UICollectionView, viewForSupplementaryElementOfKind kind: String, at indexPath: IndexPath) -> UICollectionReusableView {
+        guard visibleSections.indices.contains(indexPath.section) else {
+            fatal("Unknown section")
```
johnxnguyen: Also here.

Wire-iOS/Sources/UserInterface/Components/Controllers/SectionCollectionViewController.swift
billypchan: The issue should be fixed since the guard is added in `numberOfItemsInSection`. But it is fine to see the issue is fixed or not.

PR #4084, opened 2020-02-17
fix: app crash when accessing MediaAsset.data

billypchan:
## What's new in this PR?

Change `MediaAsset.data` to optional and unwrapp it before access.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/UIPasteboard+MediaAsset.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/UIPasteboard+MediaAsset.swift
@@ -47,9 +47,10 @@ extension UIPasteboard {
         return nil
     }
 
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Components/UIPasteboard+MediaAsset.swift
```diff
@@ -47,9 +47,10 @@ extension UIPasteboard {
         return nil
     }
 
-    @objc func setMediaAsset(_ image: MediaAsset?) {
-        guard let image = image else { return }
+    func setMediaAsset(_ image: MediaAsset?) {
+        guard let image = image,
+              let data = image.data() else { return }
```
johnxnguyen: You could use optional chaining here: `let data = image?.data()`

PR #4085, opened 2020-02-17
refactoring TypingIndicatorViewTests

billypchan:
## What's new in this PR?

Refactoring TypingIndicatorViewTests with new snapshot framework.
Add tests for setHidden method.


Wire-iOS Tests/TypingIndicatorViewTests.swift
```diff
@@ -18,16 +18,16 @@
...
-class TypingIndicatorViewTests: ZMSnapshotTestCase {
+final class TypingIndicatorViewSnapshotTests: XCTestCase {
 
     var sut: TypingIndicatorView!
     
     override func setUp() {
         super.setUp()
         sut = TypingIndicatorView()
         sut.translatesAutoresizingMaskIntoConstraints = false
-        sut.layer.speed = 0
```
typfel: This set to `0` to turn off the typing animation which was making the tests unreliable, that's not necessary anymore?

billypchan: Tested with Swift snapshot framework and it is not longer needed.

PR #4087, opened 2020-02-18
Convert ZoomTransition

billypchan:
Convert `ZoomTransition` to Swift. Refactor the shared code from `VerticalTransition`.


Wire-iOS/Sources/UserInterface/Components/Transitions/ZoomTransition.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+        
+        toView.layoutIfNeeded()
+        
+        fromView.alpha = 1
+        fromView.layer.needsDisplayOnBoundsChange = false
+
+        if reversed {
+            
+            UIView.wr_animate(easing: .easeInExpo, duration: 0.35, animations: {               fromView.alpha = 0
```
KaterinaWire: `fromView.alpha = 0` should be put to the next line

billypchan: fixed

PR #4104, opened 2020-02-20
Replace objc spinner extensions with SpinnerCapable protocol 

billypchan:
## What's new in this PR?

Current the loading spinner is created in `UIViewController (LoadingView)` extension and it uses `objc_get/setAssociatedObject` to store the spinner view. To replace this, `SpinnerCapable` protocol is created. 

Example use:

For any `UIViewController` want to show a spinner, conforms `SpinnerCapable `:
```
final class MockLoadingSpinnerViewController: UIViewController, SpinnerCapable {
   var dismissSpinner: Completion?
   :
   :
}
```

Show the spinner:

```
viewController.isSpinnerVisible = true
```

Dismiss the spinner:

```
viewController.isSpinnerVisible = false
```

Check the spinner is presented:

```
if viewController.isSpinnerVisible { ... }
```

Set the spinner subtitle:
```
viewController.showSpinner(title: "RESTORING…")
```


Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,102 @@
...
+        get {
+            return loadingSpinnerView.spinnerSubtitleView.subtitle
+        }
+        
+        set {
+            loadingSpinnerView.spinnerSubtitleView.subtitle = newValue
+        }
+    }
+
+    var showSpinner: Bool {
```
marcomaddy1982: could be renamed in `var isLoading: Bool?`

billypchan: It is a synonym of `isLoading`. I will rename it after PoC is approved.

johnxnguyen:
Do you think it's possible to avoid the conforming view controller to have to declare and retain a reference to the spinner view? 

I wonder if this could be achieved by having the Spinner protocol define a default method to create a spinner view controller, present it and return a closure (containing logic to dismiss the spinner), which could be retained by the initial view controller and called when it wants.

billypchan:
@johnxnguyen presenting a new ViewController would have some problems, some current ViewController using spinner is not full-screen, e.g. Camera keyboard VC. Adding a spinner view as a subview can also allow multiple visible VC having spinners at the same time.

marcomaddy1982:
I'm agree with @johnxnguyen, I guess it is not necessary have a reference of the spinner in the view controller. I would move to his proposed implementation using a protocol.

johnxnguyen:
> @johnxnguyen presenting a new ViewController would have some problems, some current ViewController using spinner is not full-screen

What about a child view controller?

Also, is the spinner supposed to block the touches to the underlying view controller?

johnxnguyen:
In fact, you could even keep the spinner as a `UIView`, you would just return the closure to tear down the view. For example: 

```swift
protocol SpinnerCapable {
 
 func presentSpinner() -> () -> Void

}

extension SpinnerCapable where Self: UIViewController {

  func presentSpinner() -> () -> Void {
    // Starts animating when it appears, stops when it disappears
    let spinnerView = SpinnerView()
    
    view.addSubview(spinnerView)
    
    // Setup constraints    
    ...

    return { spinnerView.removeFromSuperview() }
  }
}

class ViewController: UIViewController, SpinnerCapable {

  func doSomething() {
    let dimissSpinner = presentSpinner()
    someHeavyComputation(onComplete: dismissSpinner) 
  }

}

```

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+        
+        get {
+            return dismissSpinner != nil
+        }
+    }
+}
+
+typealias SpinnerViewController = UIViewController & SpinnerCapable
+typealias SpinnerCompletion = ((Completion?) -> Void)
```
johnxnguyen, line 42: I'm not sure I see the need to pass in a completion handler as an argument. Couldn't the user simply just call `dismissSpinner()` and then whatever else they want to do after that?

Also perhaps a better name for `SpinnerCompletion` is something that describes what the closure will do, such as `SpinnerDismisser`.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,99 @@
...
+//
+
+import Foundation
+
+protocol SpinnerCapable: class {
+    var dismissSpinner: SpinnerCompletion? { get set }
+}
+
+extension SpinnerCapable where Self: UIViewController {
+    var showSpinner: Bool {
```
johnxnguyen, line 26: `showSpinner` is imperative language which should be reserved for methods and functions. I suggest to use descriptive language, such as `isSpinnerVisible` or `isSpinnerHidden`.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,99 @@
...
+        set {
+            if newValue {
```
johnxnguyen, line 28: You should guard check that there doesn't already exist a spinner before starting a new one. Since starting a second spinner would replace `dismissSpinner`, you would never be able to dismiss previous spinners.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+}
+
+typealias SpinnerViewController = UIViewController & SpinnerCapable
+typealias SpinnerCompletion = ((Completion?) -> Void)
+
+extension UIViewController {
+    
+    func presentSpinner(title: String? = nil) -> SpinnerCompletion {
```
johnxnguyen, line 46: This extension makes *every* view controller capable of presenting a spinner. Do you mean to extended `SpinnerCapable where Self == UIViewController` ?

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+        
+        get {
+            return dismissSpinner != nil
+        }
+    }
+}
+
+typealias SpinnerViewController = UIViewController & SpinnerCapable
```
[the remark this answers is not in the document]
billypchan: `SpinnerViewController`is referenced in several place, I think it is still saves some brackets, how about rename it like `SpinnerCapableViewController` ?

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
billypchan: The `Completeion` closure is for snapshot testing after spinner dismissed, I will check is it still needed.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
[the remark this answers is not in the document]
johnxnguyen: Yes that's a good solution.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
billypchan: It will be named to `showLoadingView` later after objc code is removed

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
billypchan: It is removed.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
billypchan: fixed.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
johnxnguyen: Why don't we make this private and provide a computed property `isSpinnerVisible`, since there is the method `showSpinner(title:)`

billypchan: fixed

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,101 @@
...
+
+extension SpinnerCapable where Self: UIViewController {
+    func showSpinner(title: String) {
+        dismissSpinner = presentSpinner(title: title)
+    }
+
+    var isSpinnerVisible: Bool {
+        set {
+            if newValue {
+                //do not show doubled spinner
```
typfel, line 34:
```suggestion
                // do not show double spinners
```

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,101 @@
...
+
+        return {
+            spinnerView.removeFromSuperview()
+        }
+    }
+
+    fileprivate func createSpinner(title: String? = nil) -> LoadingSpinnerView {
+        let loadingSpinnerView = LoadingSpinnerView()
+        loadingSpinnerView.backgroundColor = UIColor(white: 0, alpha: 0.5)
+
```
typfel, line 72:
```suggestion
```

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -0,0 +1,101 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+typealias SpinnerCapableViewController = UIViewController & SpinnerCapable
+
+protocol SpinnerCapable: class {
+    var dismissSpinner: Completion? { get set }
```
typfel, line 23: How about having a more specific type for the completion closure, like you wrote in the description I find `SpinnerCompletion` more readable.

billypchan: `SpinnerCompletion` is replaced a simple `Completion` (() -> Void) since it is no longer needed.

typfel: Yes, I understand it works with `Completion` but for readability I still think it makes sense to add a `typealias SpinnerCompletion = Completion`.

billypchan: added the typealias and renames the parameters.

PR #4111, opened 2020-02-21
Remove wrapInNavigationController objc interfaces

billypchan:
Remove `wrapInNavigationController` objc interfaces and convert their callers to Swift.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DefaultNavigationBar.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DefaultNavigationBar.swift
@@ -67,16 +67,6 @@ import UIKit
 
 extension UIViewController {
 
-    @objc
-    func wrapInNavigationController(_ navigationControllerClass: UINavigationController.Type) -> UINavigationController {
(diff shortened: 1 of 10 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/DefaultNavigationBar.swift
```diff
@@ -67,16 +67,6 @@ import UIKit
...
-    func wrapInNavigationController(_ navigationControllerClass: UINavigationController.Type) -> UINavigationController {
-        return wrapInNavigationController(navigationControllerClass: navigationControllerClass, navigationBarClass: DefaultNavigationBar.self)
-    }
-
-    @objc
-    func wrapInNavigationController() -> UINavigationController {
-        return wrapInNavigationController(navigationControllerClass: RotationAwareNavigationController.self, navigationBarClass: DefaultNavigationBar.self)
-    }
-    
     func wrapInNavigationController(navigationControllerClass: UINavigationController.Type = RotationAwareNavigationController.self,
```
johnxnguyen: It would be nice to hide the first argument label for the callers, since we see "navigationController" three times in a single line.

PR #4112, opened 2020-02-21
convert MediaAsset callers to Swift

billypchan:
## What's new in this PR?

Convert `MediaAsset` protocol callers to Swift to prepare for converting the protocol to Swift


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.swift
```diff
@@ -40,4 +41,31 @@ extension FullscreenImageViewController {
...
+        let imageData = message.imageMessageData?.imageData
+        
+        DispatchQueue.global(qos: .default).async(execute: { [weak self] in
+            
+            let mediaAsset: MediaAsset
+            
+            if imageIsAnimatedGIF == true {
+                mediaAsset = FLAnimatedImage(animatedGIFData: imageData)
+            } else if let imageData = imageData,
+                   let image = UIImage(data: imageData) {
```
johnxnguyen: You could clean up this with `if let image = imageData.map(UIImage.init)`

billypchan: fixed

PR #4116, opened 2020-02-24
[Feature] Change the placeholder in the input bar

KaterinaWire:
## What's new in this PR?

If the contact has a set availability status and a long name, the placeholder in the input bar can be broken for WEB app. It was decided for all platforms to show "This person is" + availability status.


Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
```diff
@@ -44,7 +44,7 @@
                 guard availability != .none && !user.shouldHideAvailability else { //Should use the default placeholder string
                     return nil
                 }
-                title = "availability.\(availability.canonicalName).placeholder".localized(args: user.name ?? "").localizedUppercase
+                title = "availability.\(availability.canonicalName).placeholder".localized(args: "availability.default.user_name.title".localized).localizedUppercase
```
typfel: I'd suggest placing the "This person" in all the existing placer holder strings, like this:
`"availability.available.placeholder" = "This person is available";`

The reason is that it's problematic to assume that we can concatenate the two strings "This person" and "is available" in all languages.

Wire-iOS/Sources/UserInterface/Availability/AvailabilityTitleView.swift
```diff
@@ -120,7 +120,7 @@ final class AvailabilityTitleView: TitleView, Themeable, ZMUserObserver {
         super.configure(icon: icon, title: title, interactive: isInteractive, showInteractiveIcon: showInteractiveIcon)
         
         accessibilityLabel = options.contains(.allowSettingStatus) ? "availability.accessibility_label.change_status".localized : "availability.accessibility_label.status".localized
-        accessibilityValue = availability.localizedName
+        accessibilityValue = availability.localizedName + " " + (user.name ?? "")
```
typfel: Why is this necessary now? You were modifying the placeholder in the input bar and this will affect the navigation title.

`accessibilityValue` is user facing for people who rely on voice over.

KaterinaWire: Removed

PR #4133, opened 2020-02-26
Convert IconButton

billypchan:
## What's new in this PR?

Convert `IconButton` to Swift

### TODO
- [x] split attributable string helper into 2 files and only left the localization path in common component, to prevent adding many import lines in main project.


Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+import UIKit
+import WireCommonComponents
+
+enum IconButtonStyle {
+    case `default`
+    case circular
+    case navigation
+}
+
+struct IconDefinition: Equatable {
```
marcomaddy1982: would be better `IconConfiguration` or `IconInfo`?

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+    let iconType: StyleKitIcon
```
marcomaddy1982: I guess we could name this just `let type: StyleKitIcon`, the prefix `icon` seems to be a bit redundant

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+    let iconSize: CGFloat
```
marcomaddy1982: Same of previous comment. The prefix `icon` seems to be a bit redundant.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+    let renderingMode: UIImage.RenderingMode
+}
+
+///TODO: remove public after ButtonWithLargerHitArea is converted to Swift
+public class IconButton: ButtonWithLargerHitArea {
+    var circular = false {
```
marcomaddy1982: IMHO I would rename it in `var isCircular: Bool = false`

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+
+    var borderWidth: CGFloat = 0.5 {
+        didSet {
+            updateCircular()
+        }
+    }
+
+    var roundCorners = false {
```
marcomaddy1982: IMHO I would rename it in `var hasRoundCorners = false`.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+    init() {
+        super.init(frame: .zero)
+
+        hitAreaPadding = CGSize(width: 20, height: 20)
+    }
+
+    convenience init(style: IconButtonStyle, variant: ColorSchemeVariant = ColorScheme.default.variant) {
+        self.init()
+
+        setIconColor(UIColor.from(scheme: .iconNormal, variant: variant), for: .normal)
```
marcomaddy1982: IMHO I would avoid to have so long init method. I suggest to separate these statements in methods with a more descriptive naming.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+    private var iconDefinitionsByState: [UIControl.State: IconDefinition] = [:]
+    private var priorState: UIControl.State?
+
+    init() {
+        super.init(frame: .zero)
+
+        hitAreaPadding = CGSize(width: 20, height: 20)
+    }
+
+    convenience init(style: IconButtonStyle, variant: ColorSchemeVariant = ColorScheme.default.variant) {
```
marcomaddy1982: IMHO I would avoid to have so long init method. I suggest to separate the next statements in methods with a more descriptive naming.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+}
+
+struct IconDefinition: Equatable {
+    let iconType: StyleKitIcon
+    let iconSize: CGFloat
+    let renderingMode: UIImage.RenderingMode
+}
+
+///TODO: remove public after ButtonWithLargerHitArea is converted to Swift
+public class IconButton: ButtonWithLargerHitArea {
```
marcomaddy1982: General comment on the class: I would reorganise the methods in a more readable way: you could it do in two ways
* group private and public methods and use more descriptive pragma mark
or
* separate related methods in extension.

Wire-iOS/Sources/UserInterface/ConnectRequests/OutgoingConnectionBottomBar.swift
```diff
@@ -55,7 +55,7 @@ import Cartography
         cancelButton.setTitle("profile.cancel_connection_button_title".localized(uppercased: true), for: .normal)
         cancelButton.titleLabel?.font = FontSpec(.small, .light).font!
         cancelButton.setTitleColor(UIColor.from(scheme: .textForeground), for: .normal)
-        cancelButton.titleImageSpacing = 24
+        cancelButton.setTitleImageSpacing(24)
```
marcomaddy1982: IMHO I would avoid to use this magic number. you could create a static constant for example.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+
+    /// set icon to a new icon or no icon
+    ///
+    /// - Parameters:
+    ///   - iconType: the StyleKitIcontype
+    ///   - iconSize: StyleKitIcon.Size
+    ///   - state: UIControl state
+    ///   - renderingMode: Default rendering mode is AlwaysTemplate
+    ///   - force: force update
+    func setIcon(_ iconType: StyleKitIcon?,
```
marcomaddy1982: IMHO this method should be clean up a bit.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+        guard let iconType = iconType else {
+            removeIcon(for: state)
+            return
+        }
+
+        let newIcon = IconDefinition(iconType: iconType, iconSize: iconSize, renderingMode: renderingMode)
+
+        let currentIcon = iconDefinitionsByState[state]
+
+        if !force, let currentIcon = currentIcon, currentIcon == newIcon {
```
marcomaddy1982: You could use a `guard` statement here instead of an `if` with a `return` and probably you could create a unique bigger `guard` grouping  it with the initial `guard` statement.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+
+        setImage(image.withRenderingMode(renderingMode), for: state)
+    }
+
+    func removeIcon(for state: UIControl.State) {
+        iconDefinitionsByState[state] = nil
+        setImage(nil, for: state)
+    }
+
+    func setIconColor(_ color: UIColor?, for state: UIControl.State) {
```
marcomaddy1982: IMHO this method should be clean up a bit.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
```diff
@@ -0,0 +1,310 @@
...
+        tintColor = iconColor(for: state)
+    }
+
+    func updateCircularCornerRadius() {
+        guard circular else { return }
+
+        /// Create a circular mask. It would also mask subviews.
+
+        let radius: CGFloat = bounds.size.height / 2
+        let maskPath = UIBezierPath(roundedRect: bounds, byRoundingCorners: .allCorners, cornerRadii: CGSize(width: radius, height: radius))
```
marcomaddy1982: General comment for indentation:  I would indent all the `UIBezierPath` going in a new line for each init params.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: I followed the original naming and I would perfer keeping it.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: fixed.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: fixed

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: I would prefer to follow orginal naming since slibing class `Button` also has same naming.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: fixed

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: I will clean up these after `ButtonWithLargerHitArea` is converted to Swift.

Wire-iOS/Sources/UserInterface/ConnectRequests/OutgoingConnectionBottomBar.swift
billypchan: I will clean up these in another PR. Thanks!

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: There are quite a lot of callers and I will clean them up in separate PR.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: The first guard handles the nil case of `iconType` and I would prefer seperated them.

Wire-iOS/Sources/Components/Buttons/IconButton.swift
billypchan: fixed

PR #4134, opened 2020-02-27
[Fix] Display the full name in it's entirety on the self profile page

KaterinaWire:
## What's new in this PR?
To unify the behaviour with android, we should display the full name in it's entirety on the self profile page, spanning over a maximum of three lines.


billypchan:
Test `testTestForAUserWithALongName` failed. We should run this test with `verifyAllIPhoneSizes` to see if this changes fits on smallest screen.

PR #4135, opened 2020-02-27
Move AttributedStringOperators from common component to app target

billypchan:
### Issues

In https://github.com/wireapp/wire-ios/pull/4133, we found that the bridging header imports `WireCommonComponents` via `IconButton.h`. If we remove `IconButton.h`, a lot of source files have to import `WireCommonComponents`.

### Causes

A lot of files needs attributed String helper in `AttributedStringOperators`.

### Solutions

Move `AttributedStringOperators` to project target. Leave `String.localized` method in `WireCommonComponents`.


Wire-iOS Tests/Mocks/MockUser+PointOfView.swift
```diff
@@ -17,13 +17,10 @@
...
 
 extension MockUser {
-    @objc var pov: PointOfView {
-        if MockUser.mockSelf() == self {
-            return .secondPerson
-        }
-
-        return .none
+    var pov: PointOfView {
+        return MockUser.mockSelf() == self ?.secondPerson : .none
```
johnxnguyen: This functionality is present in `UserType+Helpers.swift`. It should be fine to delete this.

billypchan: file removed.

PR #4141, opened 2020-02-28
[Fix] Connection request modal not showing on iPad

typfel:
### Issues

The connection request modal is not presented on iPad

### Causes

The custom transition animation is never completed because we guard against `transitionContext.fromView` being nil, and return early. When presenting using the `.formSheet` presentationStyle the `fromView` is not set when presenting and `toView` is not set when dismissing.

### Solutions

Don't assume `toView` and `fromView` is always available.


billypchan:
@typfel I will apply the similar fix to other `Transition`.

PR #4143, opened 2020-02-28
Refactored LinkOpeningOption storedPreference with associatedtype

billypchan:
## What's new in this PR?

Refactored duplicate implementation of `storedPreference` with associatedtype in `LinkOpeningOption`

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/LinkOpening/LinkOpener.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/LinkOpening/LinkOpener.swift
@@ -70,14 +70,14 @@ protocol LinkOpeningOption {
     var displayString: String { get }
     static var availableOptions: [Self] { get }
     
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/Settings/LinkOpening/LinkOpener.swift
```diff
@@ -63,17 +63,29 @@ extension NSURL {
 }
 
 protocol LinkOpeningOption {
-
+    associatedtype E: RawRepresentable where E.RawValue == Int
```
typfel: Let's use a more descriptive name then `E`, how about `ApplicationOption`?

billypchan: Rename to `ApplicationOptionEnum` to indicate it is a Enum with rawValue.

PR #4158, opened 2020-03-05
Feature: SpinnerButton

billypchan:
## What's new in this PR?

Create `SpinnerButton` which has a spinner on the button with customized disable state color scheme.


Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,82 @@
...
+            
+            showSpinner ? spinner.startAnimation() : spinner.stopAnimation()
+        }
+    }
+    
+    override init() {
+        super.init()
+        
+        /// multi line support
+        let inset: CGFloat = 10
```
marcomaddy1982: It would be nice place this statements in a separate method and not directly in the 'init()'

billypchan: @marcomaddy1982 fixed, and the UI size constant is still not finalised and will be updated later.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        progressSpinner.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            progressSpinner.centerYAnchor.constraint(equalTo: centerYAnchor),
+            progressSpinner.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -SpinnerButton.iconInset)])
+
+        return progressSpinner
+    }()
+    
+    var showSpinner: Bool = false {
```
marcomaddy1982: I guess the 'var showSpinner: Bool = false' is a bit redundant, you could override 'var isHidden: Bool = false'. Basically they have the same scope.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,83 @@
...
+final class SpinnerButton: Button {
+    ///TODO: get these from design
+    private static let iconSize = StyleKitIcon.Size.tiny.rawValue
+    private static let iconInset: CGFloat = 10
+    private static let textInset: CGFloat = 5
+    private static let inset: CGFloat = 10
+
+    private lazy var spinner: ProgressSpinner = {
+        let progressSpinner = ProgressSpinner()
+
```
marcomaddy1982: I guess you could remove one line here (I don-t know how much importance you give to style comment).

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,83 @@
...
+    private static let iconInset: CGFloat = 10
+    private static let textInset: CGFloat = 5
+    private static let inset: CGFloat = 10
+
+    private lazy var spinner: ProgressSpinner = {
+        let progressSpinner = ProgressSpinner()
+        
+
+        // the spinner covers the text with alpha BG
+        progressSpinner.backgroundColor = variant == .light ? UIColor(white: 1, alpha: 0.8) : UIColor(white: 0, alpha: 0.8) ///TODO: update from design
```
marcomaddy1982: IMHO it is more readable if the ternary operator is indented as:
progressSpinner.backgroundColor = variant == .light 
            ? UIColor(white: 1, alpha: 0.8)
            : UIColor(white: 0, alpha: 0.8)

Wire-iOS/Sources/Components/Buttons/Button.swift
```diff
@@ -53,7 +53,7 @@ class Button: ButtonWithLargerHitArea {
             updateStyle(variant: variant)
         }
     }
-    private var variant: ColorSchemeVariant = ColorScheme.default.variant
```
marcomaddy1982: Instead to remove the private access modifier you could create a public set method for it. This is just a thought. :-)

Wire-iOS/Sources/Components/Buttons/Button.swift
```diff
@@ -108,6 +102,17 @@ class Button: ButtonWithLargerHitArea {
             setBorderColor(UIColor(white: 1.0, alpha: 0.16), for: .highlighted)
         }
     }
+    
+    func updateEmptyStyle() {
```
marcomaddy1982: this should be private

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
marcomaddy1982: Basically something like this:

`override var isHidden: Bool {
        didSet {
            isEnabled = !isHidden
            isHidden
                ? spinner.stopAnimation()
                : spinner.startAnimation()
        }
    }`

marcomaddy1982: And probably `isEnabled = !isHidden` should be removed cause if the spinner is hidden it could not be interested so doesn't make sense disable it.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,83 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+
+/// A button with spinner at the trailing side. Title text is non trancated.
+final class SpinnerButton: Button {
+    ///TODO: get these from design
```
marcomaddy1982: A good best practice could be place this UI constant in a separate extension place on top of the class definition, something like this:

```suggestion
    extension SpinnerButton {
    ///TODO: get these from design
    private static let iconSize = StyleKitIcon.Size.tiny.rawValue
    private static let iconInset: CGFloat = 10
    private static let textInset: CGFloat = 5
    private static let inset: CGFloat = 10
}

/// A button with spinner at the trailing side. Title text is non trancated.
final class SpinnerButton: Button { ... }
```

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
```diff
@@ -0,0 +1,83 @@
...
+            
+            showSpinner ? spinner.startAnimation() : spinner.stopAnimation()
+        }
+    }
+    
+    override init() {
+        super.init()
+        
+        /// multi line support
+        if let titleLabel = titleLabel {
```
marcomaddy1982: IMHO you could place these statement in a separated `private` method called something like `configureTitleLabel`.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
billypchan: the `showSpinner` property only controls the visibility of the spinner but not the button. When the spinner is shown, the button should be disabled.

Wire-iOS/Sources/Components/Buttons/Button.swift
billypchan: It is overridden by `SpinnerButton` so its has to be internal.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
billypchan: I will move them after the design spec is finalized.

Wire-iOS/Sources/Components/Buttons/SpinnerButton.swift
marcomaddy1982: ok now it is clear, what do you think to rename `var showSpinner: Bool` in `var isLoading: Bool`?

PR #4160, opened 2020-03-05
[Fix] restore conversation after unblocking and user

marcomaddy1982:
### Issues

Issue 1: Tapping on the "Unblock" action of the sheet in the profile another unnecessary action sheet (having the same "Cancel" and "Unblock" action is shown)

Issue 2: When an user unblock another one, the unblocked user is not shown in the conversation list after closing the search screen

### Causes

After unblocking an user the 'ZMUserSession enqueued changes' were not saved.

### Solutions

Issue 1: show the confirmation action sheet just when we need to block the suer
Issue 2: Save the  'ZMUserSession enqueued changes' after unblocking an user.

### Dependencies

Jira Ticket: https://wearezeta.atlassian.net/browse/ZIOS-12451


Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -174,10 +184,14 @@ final class ProfileViewControllerViewModel: NSObject {
...
     func transitionToListAndEnqueue(leftViewControllerRevealed: Bool = true, _ block: @escaping () -> Void) {
         ZClientViewController.shared?.transitionToList(animated: true,
-                                                         leftViewControllerRevealed: leftViewControllerRevealed) {
-                                                            ZMUserSession.shared()?.enqueueChanges(block)
+                                                       leftViewControllerRevealed: leftViewControllerRevealed) {
+                                                        self.saveEnqueueChanges(block)
         }
     }
+    
+    func saveEnqueueChanges(_ block: @escaping () -> Void) {
```
typfel: I would drop the `save` prefix since `enqueue` is already matching the action we do.

PR #4161, opened 2020-03-05
[Refactor] Replace Set<ZMUser> with UserSet

johnxnguyen:
## What's new in this PR?

This PR is part of an ongoing effort to remove references to concrete user objects. See https://github.com/wireapp/ios-architecture/issues/89 for more details.

---

### Issues

There are several cases where we wish to hold a group of users in a `Set`, and this is currently achieved with `Set<ZMUser>`. Unfortunately, we can't declare a type of `Set<UserType>` because `UserType` is a protocol, and therefore can't conform to `Hashable` as required by `Set`. 

### Solutions

There already existed a solution to this, however it was only used in a single place. The idea is to wrap the opaque instance of `UserType` inside a concrete `HashBox`. Since `UserType` extends `NSObjectProtocol`, we can use the `isEqual(to:)` method and `hashValue` to provide `Hashable` conformance to `HashBox`.

This allows us to define a `Set<HashBox<UserType>>`. However, boxing and unboxing the underlying user is tedious, so I've provided a `UserSet` type that conforms to `Collection` and `SetAlgebra`. It takes care of all the boxing and unboxing, so it can be used just like a `Set<UserType>`.

This PR implements this new type and integrates it so that all occurrences of `Set<ZMUser>` have been replaced with `UserSet`

### Dependencies

DataModel: `[ZMUser]` argument types have been replaced with `[UserType]`, allowing us to use the `UserSet` right up to the framework boundary.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/908


Wire-iOS/Sources/Helpers/UserSet.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+
+import Foundation
+
+/// A set of instances conforming to `UserType`.
+
+struct UserSet: Collection, SetAlgebra {
+
+    typealias Storage = Set<HashBox<UserType>>
+    typealias Element = UserType
```
marcomaddy1982: typealias Element = UserType seems to be never used.

johnxnguyen: It's a requirement of `Collection`, and although it can be inferred I've chosen to state it explicitly for clarity.

Wire-iOS/Sources/Helpers/UserSet.swift
```diff
@@ -0,0 +1,117 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// A set of instances conforming to `UserType`.
+
+struct UserSet: Collection, SetAlgebra {
```
marcomaddy1982: You have already implemented the necessary init method, but for completion you could add this protocol ExpressibleByArrayLiteral to the UserSet

johnxnguyen: `ExpressibleByArrayLiteral` is inherited by `SetAlgebra`, but I could also explicitly state it too.

marcomaddy1982: No at this point it doesn't make sense. Sorry I didn't noticed.

PR #4162, opened 2020-03-05
[Fix] missing cryptobox logs on public builds

typfel:
### Issues

Cryptobox logs are not activated for public builds

### Causes

We don't activate any log tags for public builds

### Solutions

Activate the default log tags for all builds.

### Notes

In order to make easier to test crypto box error logs I've added a debug action for corrupting a session on the device details screen.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
@@ -42,7 +42,7 @@ final class ProfileClientViewController: UIViewController {
     var fromConversation: Bool = false
 
     /// Used for debugging purposes, disabled in public builds
-    var deleteDeviceButton: ButtonWithLargerHitArea?
+    var debugMenuButton: ButtonWithLargerHitArea?
 
     var showBackButton: Bool = true {
         didSet {
@@ -109,7 +109,7 @@ final class ProfileClientViewController: UIViewController {
         self.setupVerifiedToggle()
         self.setupVerifiedToggleLabel()
         self.setupResetButton()
-        self.setupDeleteButton()
+        self.setupDebugMenuButton()
         self.createConstraints()
         self.updateFingerprintLabel()
     }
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
```diff
@@ -309,11 +309,11 @@ final class ProfileClientViewController: UIViewController {
...
-        if let deleteDeviceButton = self.deleteDeviceButton {
-            constrain(contentView, descriptionTextView, deleteDeviceButton) { contentView, reviewInvitationTextView, deleteDeviceButton in
-                deleteDeviceButton.right == contentView.right
-                deleteDeviceButton.left == contentView.left
-                deleteDeviceButton.top == reviewInvitationTextView.bottom + 10
+        if let debugMenuButton = self.debugMenuButton {
+            constrain(contentView, descriptionTextView, debugMenuButton) { contentView, reviewInvitationTextView, debugMenuButton in
+                debugMenuButton.right == contentView.right
+                debugMenuButton.left == contentView.left
+                debugMenuButton.top == reviewInvitationTextView.bottom + 10
```
marcomaddy1982: What is that magic number 10?

Wire-iOS/Sources/UserInterface/UserProfile/Devices/ProfileClientViewController.swift
```diff
@@ -365,6 +383,19 @@ final class ProfileClientViewController: UIViewController {
         }
         self.presentingViewController?.dismiss(animated: true, completion: .none)
     }
+    
+    @objc private func onCorruptSessionTapped() {
+        let sync = self.userClient.managedObjectContext!.zm_sync!
+        let selfClientID = ZMUser.selfUser()?.selfClient()?.objectID
+        sync.performGroupedBlockAndWait {
+            let client = try! sync.existingObject(with: self.userClient.objectID) as! UserClient
```
marcomaddy1982: I know that it is just for debug purpose but is it really necessary the force casting?

PR #4164, opened 2020-03-09
[Chore] Don't use deprecated Github API

johnxnguyen:
### Issues

Using the `access_token` query parameter in GitHub API endpoints is deprecated and will be removed July 1st 2020.

### Solutions

Use the Authorization HTTP header instead, via `curl -u <username:token> <endpoint>`

### Notes

I also added echoed an error message when either the access token or username isn't set and return early.

### DIFF
--- a/Scripts/download-avs.sh
+++ b/Scripts/download-avs.sh
@@ -30,10 +30,13 @@ AVS_FRAMEWORK_NAME="avs.framework"
 # CREDENTIALS
 ##################################
 # prepare credentials if needed
-GITHUB_USERNAME="`git config user.email`"
 
-if [[ -n "${GITHUB_ACCESS_TOKEN}" ]] && [[ -n "${GITHUB_USERNAME}" ]]; then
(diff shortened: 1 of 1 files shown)


typfel:
I guess the script now aborts if there's are no credentials available.

johnxnguyen:
> I guess the script now aborts if there's are no credentials available.

Yeah I realized that we don't actually want that logic, I'll revert it. But in this case, we actually expect there to be credentials so I think it's failing because of the `git config user.name` call. I'm not really familiar with how the CI works and am a little confused about which environment the script is running on.

typfel:
I don't know what effect this has on our build machines but we could also rely on getting the credentials from the `credential-osxkeychain` helper

```
git credential-osxkeychain get
host=github.com
protocol=https
<press enter>

password=xxxxx
username=typfel
```

Then we could get rid of the requirement of exposing the github token through the environment variable `GITHUB_ACCESS_TOKEN`.

Hmm but maybe it's better to do this as separate PR and not break the build machines.

PR #4171, opened 2020-03-10
[Refactor]: Profile Image Download

marcomaddy1982:
### Issues

Too much protocol dependencies on the users objects (`ZMUser` and `ZMSearchUser`) to fetch the profile image.

### Solutions

I tought that would be a good practice to delete the `ProfileImageFetchable` protocol and move its the logic in an `UserType` extension with default implementation called `UserType+ProfileImage`.
Moreover I tried to clean up the profile image fetching logic separating it in smaller methods.


Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
```diff
@@ -0,0 +1,143 @@
...
+                                 completion: completion)
+            return
+        }
+        
+        completion(cachedImage, true)
+    }
+    
+    // MARK: ImageSize Helper
+    
+    private func findProfileImageSize(for sizeLimit: Int?) -> ProfileImageSize {
```
typfel: the `find` prefix makes me think we are searching for a `ProfileImageSize` (and we might not find it), which is not the case.

IMO I would drop the `find` prefix.

Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
```diff
@@ -0,0 +1,143 @@
...
+    
+    private func findCachedImage(imageCache: ImageCache<UIImage>,
+                                 cacheKey: String) -> UIImage? {
+        guard let cachedImage = imageCache.cache.object(forKey: cacheKey as NSString) else {
+            return nil
+        }
+        return cachedImage
+    }
+    
+    private func findCachedImageKey(for imageSize: ProfileImageSize,
```
typfel: I also here I would drop the `find` prefix.

Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
```diff
@@ -0,0 +1,143 @@
...
+        if let sizeLimit = sizeLimit {
+            derivedKey = "\(derivedKey)_\(sizeLimit)"
+        }
+        
+        return derivedKey
+    }
+    
+    // MARK: Preview Image Helper
+    
+    private func findPreviewImage(for imageSize: ProfileImageSize) {
```
typfel: I would expect `findPreviewImage` to return me a preview image but it's actually requesting either the preview or complete profile image to be downloaded.

I suggest renaming this method to something along the lines for `downloadProfileImage()`

marcomaddy1982: I was thinking to something like `requestPreviewImage`? basically like the two methods in the switch.

Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
```diff
@@ -0,0 +1,143 @@
...
+        case .preview:
+            requestPreviewProfileImage()
+        default:
+            requestCompleteProfileImage()
+        }
+    }
+    
+    // MARK: Dowload Image Helper
+    
+    private func downloadImageProfile(for imageSize: ProfileImageSize,
```
typfel: This name is also a bit misleading since this method is not downloading the profile image. It's pre-processing (desaturating & downscaling) an already downloaded image.

marcomaddy1982: something like `processDownloadedProdileImage`?

Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
typfel: But the two methods are called `requestPreviewProfileImage()` and `requestCompleteProfileImage()`,  wouldn't `requestProfileImage()` be more fitting?

marcomaddy1982: Yes, as you previously mentioned it is better `downloadProfileImage`.

Wire-iOS/Sources/Helpers/syncengine/UserType+ProfileImage.swift
```diff
@@ -0,0 +1,143 @@
...
+                                                 sizeLimit: sizeLimit,
+                                                 isDesaturated: isDesaturated) else {
+            completion(nil, false)
+            return
+        }
+        
+        guard let cachedImage = cachedImage(imageCache: imageCache,
+                                            cacheKey: cacheKey) else {
+            downloadProfileImage(for: imageSize)
+            processDownloadedProdileImage(for: imageSize,
```
typfel:
```suggestion
            processDownloadedProfileImage(for: imageSize,
```

PR #4172, opened 2020-03-10
[Chore] Add test for verifying that log files are included in debug reports

typfel:
## What's new in this PR?

Add test for verifying that log files are included in debug reports


Wire-iOS Tests/LoggingTest.swift
```diff
@@ -0,0 +1,30 @@
...
+
+import XCTest
+import WireSystem
+
+class LoggingTest: XCTestCase {
+
+    func testThatCurrentLogIsNotEmpty() {
+        // Assert that some logs have been written to the current log file, which is the file
+        // which gets attached to debug reports.
+        XCTAssertNotNil(ZMSLog.currentLog)
```
johnxnguyen: How does this assert that logs have been written to the file?

typfel: It doesn't :) 

I'll  add an additional assert that the file is not empty.

johnxnguyen: Is that coming in this PR or another?

typfel: No added it now.

PR #4173, opened 2020-03-11
[Fix] Hide verified shield if it's a connection

KaterinaWire:
### Issues
The verified shield is shown on the connection request page, overlapping the "Cancel" label/button.

### Solutions
Check if the conversation is a connection.

### Dependencies
Jira Ticket: https://wearezeta.atlassian.net/browse/ZIOS-12573

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
@@ -305,8 +305,10 @@ final class ConversationSystemMessageCellDescription {
             return [AnyConversationMessageCellDescription(timerCell)]
 
         case .conversationIsSecure:
-            let shieldCell = ConversationVerifiedSystemMessageSectionDescription()
-            return [AnyConversationMessageCellDescription(shieldCell)]
+            if conversation.conversationType != .connection {
+                let shieldCell = ConversationVerifiedSystemMessageSectionDescription()
+                return [AnyConversationMessageCellDescription(shieldCell)]
+            }
 
         case .decryptionFailed:
             let decryptionCell = ConversationCannotDecryptSystemMessageCellDescription(message: message, data: systemMessageData, sender: sender, remoteIdentityChanged: false)
--- a/Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
@@ -20,7 +20,13 @@
 import UIKit
 
 final class ConversationTitleView: TitleView {
-    var conversation: ZMConversation
+    private var shouldShowVerifiedShield: Bool = false
+    public var conversation: ZMConversation {
+        didSet {
+            shouldShowVerifiedShield = conversation.securityLevel == .secure && conversation.conversationType != .connection
+        }
+    }
+
     var interactive: Bool = true
     
     @objc init(conversation: ZMConversation, interactive: Bool = true) {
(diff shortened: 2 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -305,8 +305,10 @@ final class ConversationSystemMessageCellDescription {
             return [AnyConversationMessageCellDescription(timerCell)]
 
         case .conversationIsSecure:
-            let shieldCell = ConversationVerifiedSystemMessageSectionDescription()
-            return [AnyConversationMessageCellDescription(shieldCell)]
+            if conversation.conversationType != .connection {
```
johnxnguyen: You could move this check into the case pattern, like:

```swift
case .conversationIsSecure where conversation.conversationType != .connection:
```

Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
```diff
@@ -46,7 +46,7 @@ final class ConversationTitleView: TitleView {
             attachments.append(.legalHold())
         }
         
-        if conversation.securityLevel == .secure {
+        if conversation.securityLevel == .secure && conversation.conversationType != .connection {
```
johnxnguyen: It would be nice to move this into an private extension for the conversation, like:

```swift
private extension ZMConversation {
  
  var shouldShowVerifiedShield: Bool { ... }

}
```

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
KaterinaWire: I can't because then I will “jump” to the default section.

johnxnguyen: Ah right, makes sense 👍

Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
KaterinaWire: I can also add `didSet` to the `conversation` in this class.
```
private var shouldShowVerifiedShield: Bool = false
    public var conversation: ZMConversation {
        didSet {
            shouldShowVerifiedShield = conversation.securityLevel == .secure && conversation.conversationType != .connection
        }
    }
```
Wouldn't it better?

johnxnguyen: The downside to adding a stored property is that we are adding mutable state, which should be avoided whenever possible. Comparing that to a computed property in an extension, you don't need to worry about that validity of state since we read the value directly from the conversation.

KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Conversation/ConversationTitleView.swift
```diff
@@ -96,3 +96,9 @@ extension NSTextAttachment {
     }
 }
 
+private extension ZMConversation {
+    var shouldShowVerifiedShield: Bool {
+        return self.securityLevel == .secure &&
```
typfel: I think it's weird that conversation of type `.connection` has securityLevel set to `.secure` in the first place. I would consider just not checking if we can avoid setting it to `.secure` in the data model.

PR #4184, opened 2020-03-13
Convert TokenField

billypchan:
## What's new in this PR?

Convert `TokenField` to Swift.

### tested cases
1. press backspace to remove a token
2. select a token
3. clear the tokenField (by cancel button)
4. insert text between tokens (result: inserted at the end of textView)


Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+        }
+    }
+    
+    private(set) var userDidConfirmInput = false
+    
+    private var accessoryButtonTopMargin: NSLayoutConstraint!
+    private var accessoryButtonRightMargin: NSLayoutConstraint!
+    
+    private var toLabel: UILabel = UILabel()
+    ///TODO:
```
johnxnguyen: Todo?

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+    private var toLabelLeftMargin: NSLayoutConstraint!
+    private var toLabelTopMargin: NSLayoutConstraint!
+    private var currentTokens: [Token] = [] ///TODO: public getter
```
johnxnguyen: Todo?

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+    private var textAttributes: [NSAttributedString.Key : Any] {
+        var attributes: [NSAttributedString.Key : Any] = [:]
+        
+        let inputParagraphStyle = NSMutableParagraphStyle()
+        inputParagraphStyle.lineSpacing = lineSpacing
+        attributes[.paragraphStyle] = inputParagraphStyle
+        
+            attributes[.font] = font
+            attributes[.foregroundColor] = textColor
+
```
johnxnguyen: The indentation here is off.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+        super.resignFirstResponder()
+        return textView.resignFirstResponder()
+    }
+
+    // MARK: - Interface
+    var tokens: [Token] {
+        return currentTokens
+    }
+    
+    func addToken(forTitle title: String, representedObject object: AnyObject) { ///TODO: compare any
```
johnxnguyen: Todo?

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+            // Calling -insertText: forces textView to update its contentSize, while other public methods do not.
+            // Broken contentSize leads to broken scrolling to bottom of input field.
+            textView.insertText("")
+            
+            delegate?.tokenField(self, changedFilterTextTo: "")
+            
+            invalidateIntrinsicContentSize()
+            
+            // Move the cursor to the end of the input field
+            textView.selectedRange = NSRange(location: textView.text.count, length: 0)
```
johnxnguyen: Beware that `String`s `count` is different from `NSString`s `length`. `NSString` counts characters based on utf16, so you could do the same with string: `textView.text.uft16.count`.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+            setNeedsLayout()
+            updateLayout()
+            scrollToBottomOfInputField()
+        } else {
+            textView.attributedText = collapsedString
+            invalidateIntrinsicContentSize()
+        }
+    }
+
+    func updateMaxTitleWidth(for token: Token?) {
```
johnxnguyen: It'd be good to make the token non optional, since the method has no other purpose than to work on the token.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+            tokenMaxSizeWidth -= toLabel.frame.size.width + (hasAccessoryButton ? accessoryButton.frame.size.width : 0.0) + tokenOffset
+        } else if currentTokens.count == 1 {
+            tokenMaxSizeWidth -= hasAccessoryButton ? accessoryButton.frame.size.width : 0.0
+        }
+        token?.maxTitleWidth = tokenMaxSizeWidth
+    }
+    
+    // searches by isEqual:
+    func token(forRepresentedObject object: AnyObject?) -> Token? {
+        return currentTokens.first(where:{ $0.representedObject as? NSObject == object as? NSObject})
```
johnxnguyen: Why do we have to work with erased types? It would be better to use generics.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -20,14 +20,520 @@ import UIKit
...
+        
+        return NSAttributedString(string: collapsedText, attributes: textAttributes)
+    }
+    
+    func clearFilterText() {
+        var firstCharacterIndex = NSNotFound
+        
+        let notWhitespace = CharacterSet.whitespacesAndNewlines.inverted
+        
+        (textView.text as NSString).enumerateSubstrings(in: NSRange(location: 0, length: textView.text.count), options: .byComposedCharacterSequences, using: { substring, substringRange, enclosingRange, stop in
```
johnxnguyen: Again, use `textView.text.utf16.count` when working between `String` and `NSString`.

johnxnguyen: We can enumerate substrings of `String`. It'd be great to make this whole method more swift idiomatic.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
billypchan: `selectedRange` is `NSRange` so we have to stick with `NSString`'s length. I will create a helper method to access `String`'s equivalent length.

johnxnguyen: From the [docs](https://developer.apple.com/documentation/foundation/nsstring):

> An NSString object encodes a Unicode-compliant text string, represented as a sequence of UTF–16 code units. All lengths, character indexes, and ranges are expressed in terms of 16-bit platform-endian values, with index values starting at 0.

and [here](https://developer.apple.com/documentation/swift/string):

> A string’s utf16 property is a collection of UTF-16 code units, the 16-bit encoding form of the string’s Unicode scalar values.

It's fine to use the `utf16` view.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
billypchan: Actually the only type of `object` is `UserType` and `UserType` is not conform to `Equatable`, but it is a `NSObject`. I am thinking about how to compare 2 `UserType` without casting.

johnxnguyen: The `HashBox` is designed for this purpose.

billypchan: Refactored with `HashBox`

Wire-iOS/Sources/Components/TokenField/String+EnumerateCharacters.swift
```diff
@@ -0,0 +1,28 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension String {
+    var length: Int {
```
typfel: I think an extension is potentially dangerous since it's easy miss that this is actually returning `NSString.length`. I'd suggest to either:

1. Remove extension and use count on utf16 view
2. Rename extension to something like nsStringLength

Wire-iOS/Sources/Components/TokenField/TokenField.swift
billypchan: fixed

Wire-iOS/Sources/Components/TokenField/String+EnumerateCharacters.swift
billypchan: removed this extension and use `String.utf16.count` instead.

Wire-iOS/Sources/Components/TokenField/String+EnumerateCharacters.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+extension String {
+    func enumerateCharacters(range: Range<String.Index>,
+        body: @escaping (String?, Range<String.Index>, Range<String.Index>, inout Bool) -> Void) {
+        enumerateSubstrings(in: range, options: .byComposedCharacterSequences, body)
+    }
```
johnxnguyen: I don't think this is necessary at all, you could simply enumerate over the `unicodeScalars` view of `String`.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -298,31 +778,22 @@ extension TokenField: UITextViewDelegate {
...
+            let range = textRange.upperBound..<textView.text.endIndex
+            textView.text.enumerateCharacters(range: range) { substring, _, _, stop in
+
+                guard let substring = substring,
+                      !substring.isEmpty,
+                      (substring as NSString).character(at: 0) == NSTextAttachment.character else { return }
+                
+                textView.selectedRange = NSRange(location: textView.text.utf16.count, length: 0)
+                stop = true
+            }
```
johnxnguyen: We should make more use of the `String` API. From what I understand, the aim of this is to check if the suffix from the given insertion point contains an attachment. You can achieve this with:

```swift
guard let attachmentCharacter = UnicodeScalar(NSTextAttachment.character) else { ... } 
let textRange = Range(range, in: textView.text) 
let suffix = textView.text.suffix(from: textRange.lowerBound)

if suffix.unicodeScalars.contains(attachmentCharacter) {
  ...
}
```

Wire-iOS/Sources/Components/TokenField/String+EnumerateCharacters.swift
billypchan: refactored with searching`text.unicodeScalars.firstIndex`

Wire-iOS/Sources/Components/TokenField/TokenField.swift
billypchan: fixed

typfel:
I'm good but lets wait until @johnxnguyen has confirmed resolved his comments.

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -430,24 +430,14 @@ final class TokenField: UIView {
...
-        text.enumerateCharacters(range: range) { substring, substringRange, _, stop in
-            guard let substring = substring,
-                substring.isEmpty == false,
-                !substring.trimmingCharacters(in: .whitespaces).isEmpty,
-                (substring as NSString).character(at: 0) != NSTextAttachment.character else { return }
-
-            firstCharacterRange = substringRange
-            stop = true
+        guard let attachmentCharacter = UnicodeScalar(NSTextAttachment.character),
+            let firstCharacterIndex = text.unicodeScalars.firstIndex(where: { $0 != attachmentCharacter && !NSCharacterSet.whitespaces.contains($0) }) else {
```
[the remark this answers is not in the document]
billypchan: fixed

Wire-iOS/Sources/Components/TokenField/TokenField.swift
```diff
@@ -778,20 +768,16 @@ extension TokenField: UITextViewDelegate {
...
-                
-                textView.selectedRange = NSRange(location: textView.text.utf16.count, length: 0)
-                stop = true
-            }
+            let attachmentCharacter = UnicodeScalar(NSTextAttachment.character),
+            let textRange = Range(range, in: textView.text) {            
+                let range = textRange.upperBound..<textView.text.endIndex
+                let subString = textView.text?[range]
+                if subString?.unicodeScalars.contains(attachmentCharacter) == true {
+                    textView.selectedRange = NSRange(location: textView.text.utf16.count, length: 0)
```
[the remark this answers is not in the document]
billypchan: fixed

PR #4195, opened 2020-03-18
Feature: accessibility for button selected state

billypchan:
## What's new in this PR?

Use `ButtonMessageState`'s string value as `Button`'s accessibility value.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Poll/ConversationButtonMessageCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Poll/ConversationButtonMessageCell.swift
@@ -50,7 +50,7 @@ final class ConversationButtonMessageCell: UIView, ConversationMessageCell {
         }
     }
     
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Content/Poll/ConversationButtonMessageCell.swift
```diff
@@ -48,9 +48,11 @@ final class ConversationButtonMessageCell: UIView, ConversationMessageCell {
             button.isEnabled = false
             ///TODO: style for expired state
         }
+        
+        button.accessibilityValue = "\(config.state)"
```
typfel: Isn't the `accessibilityValue` is visible to voice over? I guess we should then localize the value, can be done in separate PR though.

billypchan: Yes, it should be localized. I found quite a lot of `accessibilityValue` uses a fixed string and we should create a ticket for that.

PR #4198, opened 2020-03-19
fix crashes when the app goes to background or sending bug report

billypchan:
### Issues

App crashes when goes to background, or sending a bug report.

### Causes

1. called `UIApplicationDelegate.applicationDidEnterBackground` instead of `Setting.applicationDidEnterBackground`
2. `SelfUser.current` may be nil when sending the debug report.

### Solutions

Fix with calling property method.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings.swift
+++ b/Wire-iOS/Sources/Components/Settings.swift
@@ -136,7 +136,7 @@ final class Settings {
 
         startLogging()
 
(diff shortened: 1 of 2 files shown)


billypchan:
@johnxnguyen Tested send debugging report and works.

johnxnguyen:
@billypchan and putting the app in the BG?

billypchan:
@johnxnguyen Tested background cases.

PR #4201, opened 2020-03-19
Fix: app does not keep message when typing a message and switching apps before sending

billypchan:
### Issues

When the app goes to background and switch to other app and then come back to the app, the draft text is cleared.

### Causes

the draft text is not stored in this case. ( currently we only save the draft text when keyboard dismisses, e.g. switching conversations)

### Solutions

Store the draft text when the app goes to the background.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
@@ -25,13 +25,13 @@ extension ConversationInputBarViewController {
 
         return DraftMessage(text: text, mentions: mentions, quote: quotedMessage as? ZMMessage)
     }
-    
+
     @objc
     func didEnterBackground(_ notification: Notification?) {
         if !inputBar.textView.text.isEmpty {
             conversation.setIsTyping(false)
         }
-        
+
         let draft = draftMessage(from: inputBar.textView)
         delegate?.conversationInputBarViewControllerDidComposeDraft(message: draft)
     }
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -19,7 +19,23 @@ import Foundation
...
+    //MARk: - save draft message
+    func draftMessage(from textView: MarkdownTextView) -> DraftMessage {
+        let (text, mentions) = textView.preparedText
+
+        return DraftMessage(text: text, mentions: mentions, quote: quotedMessage as? ZMMessage)
+    }
     
+    @objc
+    func didEnterBackground(_ notification: Notification?) {
+        guard !inputBar.textView.text.isEmpty else { return }
```
typfel: I think this guard should still be an if statement around the `conversation.setIsTyping(false)`. Like it is now we wouldn't save the draft correctly in the following scenario:

1. Type text
2. Enter background (saves daft)
3. Open app and erase text
4. Enter background (draft is not saved)
5. Kill app
6. Open conversation and draft is back

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -19,7 +19,23 @@ import Foundation
 import MobileCoreServices
 
 extension ConversationInputBarViewController {
+    //MARk: - save draft message
```
typfel:
```suggestion
    // MARK: - Save draft message
```

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
billypchan: Yes, the draft message should be saved no matter it is empty or not.

PR #4202, opened 2020-03-19
feature: localize button state

billypchan:
## What's new in this PR?

Localize button states

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -1747,6 +1747,6 @@
 "account_deleted_session_expired_alert.message" = "The application did not communicate with the server for a long period of time, or your session has been remotely invalidated.";
 
 // Button cell state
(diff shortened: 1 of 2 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1746,3 +1746,7 @@
 "account_deleted_session_expired_alert.title" = "Your session expired";
 "account_deleted_session_expired_alert.message" = "The application did not communicate with the server for a long period of time, or your session has been remotely invalidated.";
 
+// Button cell state
+"buttonMessageCell.state.unselected" = "unselected";
+"buttonMessageCell.state.selected" = "selected";
+"buttonMessageCell.state.confirmed" = "confirmed";
```
johnxnguyen: It would be nice to maintain consistency with the other keys, i.e use snake_case.

billypchan: fixed

PR #4207, opened 2020-03-20
Convert ConversationInputBarViewController+Files and support upload multiple files

billypchan:
## What's new in this PR?

Convert `ConversationInputBarViewController+Files` to Swift. support upload multiple files for document picker.

### TODO: 
- [x] test with iOS 10 device
- [x] unit test
- [x] non UUID file name for the zip file (currently use archive.zip)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+Files.swift
```diff
@@ -17,8 +17,169 @@
...
 import Foundation
+import ZipArchive
+
+extension ConversationInputBarViewController: UINavigationControllerDelegate {}
+
+private let zmLog = ZMSLog(tag: "ConversationInputBarViewController+Files")
 
 extension ConversationInputBarViewController {
+    //TODO: check it is still possible on iOS 10
+    func updateDirectory(itemURL: URL) {
```
marcomaddy1982: Instead of create an extension of the `ConversationInputBarViewController` do you think that we could create an extension of the `FileManager` so we could reuse these methods all around the project more easily?

billypchan: I will move the zipping part outside the view controller for easier testing.

PR #4211, opened 2020-03-23
Convert SplitViewController

billypchan:
## What's new in this PR?

Convert `SplitViewController` to Swift


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.swift
```diff
@@ -67,8 +251,304 @@ extension SplitViewController {
...
+    }
+
+    private func updateLayoutSizeAndLeftViewVisibility() {
+        updateLayoutSize(for: traitCollection)
+        updateLeftViewVisibility()
+    }
+
+    private func updateLeftViewVisibility() {
+        switch layoutSize {
+        case .compact /* fallthrough */, .regularPortrait:
```
johnxnguyen, line 358: The comment is not necessary.

PR #4224, opened 2020-03-27
Fix: [iPad iOS 13] when having SSO code on clipboard, login page gets stuck in loop

billypchan:
### Issues

When a SSO code in copied to clipboard, it repeatedly appear even cancel is pressed.

### Causes

1. The `pollingTimer` keep creating alert even it is already shown
2. After cancel is pressed, it is still trying to show the alert when going to another screen.

### Solutions

1. Check if there is a exist SSO alert before presenting a new one
2. Stop polling after cancel is pressed, and set the flag `cancelSSOPressed` to true to prevent the alert shows again.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
+++ b/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
@@ -172,12 +172,12 @@ extension CompanyLoginController {
             prefilledInput: prefilledInput,
             ssoOnly: ssoOnly,
             error: error,
-            completion: { [weak self] input, succeed in
+            completion: { [weak self] input, success in
                 input.apply(inputHandler)
                 self?.ssoAlert = nil
                 
                 // stop polling loop when cancel is pressed
-                if !succeed {
+                if !success {
                     self?.cancelSSOPressed = true
                 }
             }
--- a/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
+++ b/Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
@@ -19,7 +19,7 @@
 import Foundation
 import WireCommonComponents
 
-@objc protocol CompanyLoginControllerDelegate: class {
+protocol CompanyLoginControllerDelegate: class {
 
     /// The `CompanyLoginController` will never present any alerts on its own and will
     /// always ask its delegate to handle the actual presentation of the alerts.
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -152,15 +163,27 @@ extension CompanyLoginController {
         error: UIAlertController.CompanyLoginError? = nil,
         ssoOnly: Bool = false) {
         
+        // Do not repeatly showing alert if exist
```
David-Henner:
```suggestion
        // Do not repeatly show alert if exist
```

Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
```diff
@@ -152,15 +163,27 @@ extension CompanyLoginController {
...
             ssoOnly: ssoOnly,
             error: error,
-            completion: { input in input.apply(inputHandler) }
+            completion: { [weak self] input, succeed in
+                input.apply(inputHandler)
+                self?.ssoAlert = nil
+                
+                // stop polling loop when cancel is pressed
+                if !succeed {
+                    self?.cancelSSOPressed = true
```
David-Henner:
```suggestion
            completion: { [weak self] input, success in
                input.apply(inputHandler)
                self?.ssoAlert = nil
                
                // stop polling loop when cancel is pressed
                if !success {
                    self?.cancelSSOPressed = true
```

David-Henner: Should we reset `cancelSSOPressed` somewhere?

Wire-iOS/Sources/UserInterface/Company Login/UIAlertController+CompanyLogin.swift
```diff
@@ -86,8 +86,7 @@ extension UIAlertController {
         prefilledInput: String? = nil,
         ssoOnly: Bool = false,
         error: CompanyLoginError? = nil,
-        completion: @escaping (String?) -> Void
-        ) -> UIAlertController {
+        completion: @escaping (String?, Bool) -> Void) -> UIAlertController {
```
David-Henner: could we, if possible, add parameter names for clarity ?

billypchan: added labels.

Wire-iOS/Sources/UserInterface/Company Login/CompanyLoginController.swift
billypchan: `cancelSSOPressed` is reseted when the SSO alert is showing again (when the user pressed the enterprise log in button)

PR #4227, opened 2020-03-30
[Chore] Revert to AVS 5.5.6

johnxnguyen:
### Issues

There were some regressions introduced after integrating AVS 5.6. Some of these regressions can only be fixed with new changes on the AVS side. These changes are included as part of AVS 6, however AVS 6 integration is not complete yet.

So that these regressions don't block the next release, it has been decided to rollback to the previous AVS version.

### Needs Release With

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1194


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -58,7 +58,7 @@ extension ZMEditableUser {
               let userId = selfUser.remoteIdentifier,
               let clientId = selfUser.selfClient()?.remoteIdentifier
         else {
-            fatal("Could not create self user stream which should always exist")
+            fatal("Could create self user stream which should always exist")
```
billypchan: I think the original message is correct?

johnxnguyen: Fixed

PR #4231, opened 2020-03-30
Fix: button spinner animation unstable behaviour

billypchan:
### Issues

1. `SpinnerButton` animation may not start with changing its `isAnimating` property
2. Spinner may be visible when the button has a solid color background

### Causes

1. `ProgressSpinner` has some code related to `UIWindow` and `UIApplication` which causes side effects to animation
2. `SpinnerButton`'s background image is not updated when switching style.

### Solutions

1. Create a simpler `Spinner` class just for `SpinnerButton`
2. Reset`SpinnerButton`'s background image when switching style.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Loading/Spinner.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Loading/Spinner.swift
@@ -17,8 +17,6 @@
 
 import Foundation
 
-import Foundation
-
 final class Spinner: UIView {
 
     var color: UIColor = .white {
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Components/Loading/Spinner.swift
```diff
@@ -0,0 +1,105 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+import Foundation
```
David-Henner: duplicate import

typfel:
What exactly is causing the problems? 

The things happening inside?
```
func applicationDidBecomeActive()
func applicationDidEnterBackground()
```

I'm ok with adding this to get release out, but we should then create a cleanup task for not having two Spinner implementations in the project.

PR #4232, opened 2020-03-30
[Feature] Trigger user data & membership fetch on profile screen

johnxnguyen:
### Issues

The client no longer receives team-wide update events regarding user membership. In order to discover if a user has left the team, the client must explicitly refresh certain metadata. 

### Solutions

One such place this needs to be done is when a user's profile is opened. Upon creating of the `ProfileViewController`, we check if the user is a team member, and if so, we trigger a download of their user data and their membership.

### Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/932


Wire-iOS Tests/DeviceManagement/ProfileViewControllerTests.swift
```diff
@@ -177,4 +177,30 @@ final class ProfileViewControllerTests: XCTestCase {
...
+        sut = ProfileViewController(user: mockUser,
+                                    viewer: selfUser,
+                                    context: .profileViewer)
+
+        // Then
+        XCTAssertEqual(mockUser.refreshDataCount, 1)
+        XCTAssertEqual(mockUser.refreshMembershipCount, 1)
+    }
+
+    func testItDoesNOtRequestsDataRefeshForNonTeamMembers() {
```
typfel:
```suggestion
    func testItDoesNotRequestsDataRefeshForNonTeamMembers() {
```

PR #4233, opened 2020-03-30
Feature: display poll message in conversation list cell status

billypchan:
## What's new in this PR?

Create `PollMatcher` to display poll message in the cell's subtitle.

### TODO
- [x] check why status subtitle is override by empty message. (Self poll creation message is overridden)

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -229,7 +229,7 @@
 "conversation.voiceover.value.active" = "active";
 "conversation.voiceover.value.disabled" = "disabled";
 
-"conversation.status.poll.default" = "New Poll";
+"conversation.status.secutity_alert.default" = "New security alert";
 
 "conversation.status.typing" = "Typing a message…";
 "conversation.status.typing.group" = "%@: typing a message…";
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -341,6 +341,45 @@ final class CallingMatcher: ConversationStatusMatcher {
     var combinesWith: [ConversationStatusMatcher] = []
 }
 
+final class PollMatcher: ConversationStatusMatcher {
```
typfel: We are not adding supports for polls though, This should be a `SecurityAlertMatcher` or something alongs those lines.

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -229,6 +229,8 @@
 "conversation.voiceover.value.active" = "active";
 "conversation.voiceover.value.disabled" = "disabled";
 
+"conversation.status.poll.default" = "New Poll";
```
typfel: This would then be "Security alert"?

Wire-iOS/Sources/UserInterface/ConversationList/ZMConversation+Status.swift
```diff
@@ -693,6 +732,7 @@ final class UnsernameMatcher: ConversationStatusMatcher {
 /*
  Matchers priorities (highest first):
  
+ (Poll)
```
typfel:
```suggestion
 (SecurityAlert)
```

PR #4234, opened 2020-03-31
[Feature] Handle webView response

KaterinaWire:
## What's new in this PR?

- webView was created
- handling the response from the success or filed redirected page

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -81,7 +81,8 @@ class DigitalSignatureVerificationViewController: UIViewController {
 
 extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
     func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
-        guard let url = navigationAction.request.url,
+        guard 
+            let url = navigationAction.request.url,
             let response = parse(url) else {
                 decisionHandler(.allow)
                 return
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -98,7 +98,7 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
         }
     }
     
-    private func parse(_ url: URL) -> VoidResult? {
+    func parse(_ url: URL) -> VoidResult? {
         if let _ = url.absoluteString.range(of: success) {
             return .success
         } else if let _ = url.absoluteString.range(of: failed) {
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -106,11 +106,8 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
             guard let error = errorCode?.value else {
                 return nil
             }
-            if error.contains("authenticationFailed") {
-                return .failure(WebViewError.authenticationFailed)
-            } else {
-                return .failure(WebViewError.internalServerError)
-            }
+            return error.contains("authenticationFailed") ?
+                .failure(WebViewError.authenticationFailed) : .failure(WebViewError.internalServerError)
         } else {
             return nil
         }
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -20,13 +20,13 @@ import UIKit
 import Foundation
 import WebKit
 
-enum WebViewError: Error {
-    case authenticationFailed
-    case internalServerError
-}
-
 class DigitalSignatureVerificationViewController: UIViewController {
 
+    private enum VerificationError: Error {
+        case authenticationFailed
+        case internalServerError
+    }
+    
     var completion: ((_ result: VoidResult?) -> Void)?
     
     private var webView = WKWebView(frame: .zero)
@@ -108,7 +108,7 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
                 return nil
             }
             return error.contains("authenticationFailed") ?
-                .failure(WebViewError.authenticationFailed) : .failure(WebViewError.internalServerError)
+                .failure(VerificationError.authenticationFailed) : .failure(VerificationError.internalServerError)
         } else {
             return nil
         }
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -112,10 +112,11 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
         if let _ = postCode?.value?.range(of: success) {
             return .success
         } else if let _ = postCode?.value?.range(of: failed) {
-            guard let error = postCode?.value else {
+            let errorCode = urlComponents?.queryItems?.first(where: { $0.name == "errorCode" })
+            guard let error = errorCode?.value else {
                 return nil
             }
-            return error.contains("authentication")
+            return error.contains("authenticationFailed")
                 ? .failure(VerificationError.authenticationFailed)
                 : .failure(VerificationError.internalServerError)
         } else {
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
@@ -22,19 +22,25 @@ import WebKit
 
 class DigitalSignatureVerificationViewController: UIViewController {
 
+    typealias Result = ((_ result: VoidResult?) -> Void)
+    
+    // MARK: - Error states
     private enum VerificationError: Error {
         case authenticationFailed
         case internalServerError
     }
     
-    var completion: ((_ result: VoidResult?) -> Void)?
-    
+    // MARK: - Private Property
     private var webView = WKWebView(frame: .zero)
     private var url: URL?
     
     private let success: String = "success"
     private let failed: String = "failed"
     
+    // MARK: - Public Property
+    var completion: Result?
+    
+    // MARK: - Init
     init(url: URL) {
         self.url = url
         super.init(nibName: nil, bundle: nil)
@@ -51,12 +57,13 @@ class DigitalSignatureVerificationViewController: UIViewController {
         loadURL()
     }
     
+    // MARK: - Private Method
     private func setupWebView() {
         webView.translatesAutoresizingMaskIntoConstraints = false
         webView.navigationDelegate = self
         updateButtonMode()
         
-        self.view.addSubview(webView)
+        view.addSubview(webView)
         webView.fitInSuperview()
     }
     
@@ -75,10 +82,11 @@ class DigitalSignatureVerificationViewController: UIViewController {
     }
     
     @objc private func onClose() {
-        self.dismiss(animated: true, completion: nil)
+        dismiss(animated: true, completion: nil)
     }
 }
 
+// MARK: - WKNavigationDelegate
 extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
     func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
         guard 
@@ -107,8 +115,9 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
             guard let error = postCode?.value else {
                 return nil
             }
-            return error.contains("authentication") ?
-                .failure(VerificationError.authenticationFailed) : .failure(VerificationError.internalServerError)
+            return error.contains("authentication")
+                ? .failure(VerificationError.authenticationFailed)
+                : .failure(VerificationError.internalServerError)
         } else {
             return nil
         }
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+import UIKit
+import Foundation
+import WebKit
+
+enum WebViewError: Error {
+    case authenticationFailed
+    case internalServerError
+}
+
+class WebViewViewController: UIViewController {
```
typfel: The name `WebViewViewController` suggest that this a WebView view controller which can be reused in different scenarios, but it's actually only suitable for the digital signature case.

My suggestion:
- Rename it to something along the lines of `DigitalSignatureVerificationViewController` or what you find suitable.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -355,7 +355,8 @@ class SettingsCellDescriptorFactory {
     
     func helpSection() -> SettingsCellDescriptorType {
         
-        let supportButton = SettingsExternalScreenCellDescriptor(title: "self.help_center.support_website".localized, isDestructive: false, presentationStyle: .modal, presentationAction: { 
+        let supportButton = SettingsExternalScreenCellDescriptor(title: "self.help_center.support_website".localized, isDestructive: false, presentationStyle: .modal, presentationAction: {
```
typfel: Irrelevant change?

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    var completion: ((_ result: VoidResult?) -> Void)?
+    
+    private var webView: WKWebView!
```
johnxnguyen: You could initialize the webview immediately with the zero frame, so you don't need to have an implicitly unwrapped optional.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+        buttonItem.accessibilityLabel = "general.done".localized
+        buttonItem.tintColor = UIColor.black
+        navigationItem.leftBarButtonItem = buttonItem
+    }
+    
+    private func createConstraints() {
+        webView.topAnchor.constraint(equalTo: self.view.topAnchor, constant: 0).isActive = true
+        webView.bottomAnchor.constraint(equalTo: self.view.bottomAnchor, constant: 0).isActive = true
+        webView.rightAnchor.constraint(equalTo: self.view.rightAnchor, constant: 0).isActive = true
+        webView.leftAnchor.constraint(equalTo: self.view.leftAnchor, constant: 0).isActive = true
```
johnxnguyen: It better to use `leading` and `trailing` anchors (no assumption about layout direction) instead of `left` and `right` (assumption of an left to right layout direction).

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+    private func updateButtonMode() {
+        let buttonItem = UIBarButtonItem(title: "general.done".localized, style: .done, target: self, action: #selector(WebViewViewController.onClose))
+        buttonItem.accessibilityIdentifier = "DoneButton"
+        buttonItem.accessibilityLabel = "general.done".localized
+        buttonItem.tintColor = UIColor.black
+        navigationItem.leftBarButtonItem = buttonItem
+    }
+    
+    private func createConstraints() {
+        webView.topAnchor.constraint(equalTo: self.view.topAnchor, constant: 0).isActive = true
```
johnxnguyen: I believe we have a utility method on `UIView` to `fitInSuperView` which achieves these four lines in one.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+extension WebViewViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        guard let url = navigationAction.request.url else {
+            decisionHandler(.allow)
+            return
+        }
+        if let _ = url.absoluteString.range(of: "success") {
```
typfel: Move strings to constants.

typfel: You could move the URL parsing logic into a separate method, it could then be more easily tested in a unit test.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
billypchan: We can use `NSLayoutConstraint.activate([...]) to saves assign isActive one-by-one.
`constant` = 0 can be deleted since it is default.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+    
+    private func loadURL() {
+        guard let url = url else { return }
+        let request = URLRequest(url: url)
+        webView.load(request)
+    }
+    
+    @objc private func onClose() {
+        completion?(nil)
+        self.dismiss(animated: true, completion: nil)
```
billypchan: May be `completion` closure can be put inside `dismiss`'s complete parameter?

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+}
+
+extension WebViewViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        guard let url = navigationAction.request.url else {
+            decisionHandler(.allow)
```
billypchan: `completion` should be called in all routes?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
KaterinaWire: There are no changes in this part. These 2 lines are the same.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Done

KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Yes, we have.
Done

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: I use fitInSuperView() method and removed these lines

KaterinaWire: I use fitInSuperView() method and removed these lines

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Yes, but I changed this method. I moved a URL parsing logic into a separate method.

Wire-iOS/Sources/UserInterface/Components/Controllers/WebViewViewController.swift
KaterinaWire: Why?

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import UIKit
+import Foundation
+import WebKit
+
+enum WebViewError: Error {
```
typfel: DigitalSignatureVerificationError?

typfel: Or move the declation inside `DigitalSignatureVerificationViewController` and just call it `VerificationError`

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        case .success:
+            completion?(.success)
+            decisionHandler(.cancel)
+        case .failure(let error):
+            completion?(.failure(error))
+            decisionHandler(.cancel)
+        }
+    }
+    
+    private func parse(_ url: URL) -> VoidResult? {
```
typfel: Let's add unit tests for the for the URLs we would like to parse.

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+        if let _ = url.absoluteString.range(of: success) {
```
typfel: How about making the success and failure parsing a bit more strict by looking the `postCode` query item to determine if had a success or failure.

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -0,0 +1,115 @@
...
+    
+    @objc private func onClose() {
+        self.dismiss(animated: true, completion: nil)
+    }
+}
+
+extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        guard let url = navigationAction.request.url,
+            let response = parse(url) else {
```
typfel:
```suggestion
        guard 
            let url = navigationAction.request.url,
            let response = parse(url) else {
```

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
KaterinaWire: Added

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
KaterinaWire: Done

typfel: Hmm but now you don't determine the error by looking at `errorCode` anymore. Isn't that still useful, especially if we later want to distinguish between different errors.

KaterinaWire: But we have in `postCode` the error `postCode=sas-error-authentication-failed`

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+            return .success
+        } else if let _ = postCode?.value?.range(of: failed) {
+            guard let error = postCode?.value else {
+                return nil
+            }
+            return error.contains("authentication") ?
+                .failure(VerificationError.authenticationFailed) : .failure(VerificationError.internalServerError)
+        } else {
+            return nil
+        }
```
typfel: Couldn't we simplify it to something like this?

```suggestion
        let urlComponents = URLComponents(string: url.absoluteString)
        let postCode = urlComponents?.queryItems?.first(where: { $0.name == "postCode" })
        
        switch postCode?.value {
        case "sas-success":
            return .success
        case "sas-error-authentication-failed"
            return .failure(VerificationError.authenticationFailed)
        default:
            return .failure(VerificationError.internalServerError)
        }
```

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
typfel: Ok, I guess we can keep it like this until we want distinguish the rason why authentication failed, which I guess is what we can find out by looking at `errorCode`

Apart from that why not compare against the entire string "sas-error-authentication-failed". I made a suggestion.

Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
KaterinaWire: Did like this

PR #4239, opened 2020-03-31
Convert ConversationInputBarViewController helpers

billypchan:
Convert keyboard shortcut and UI helper methods to Swift.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
@@ -19,7 +19,7 @@ import Foundation
 import MobileCoreServices
 
 extension ConversationInputBarViewController {
-    
+
     func clearInputBar() {
         inputBar.textView.text = ""
         inputBar.markdownView.resetIcons()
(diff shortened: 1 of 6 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -166,6 +191,59 @@ extension ConversationInputBarViewController {
...
+    // MARK: - notification center
+    @objc //TODO: no objc
+    func setupNotificationCenter() {
+        NotificationCenter.default.addObserver(forName: UIResponder.keyboardDidHideNotification, object: nil, queue: .main) { [weak self] _ in
+            let isRecording = self?.audioRecordKeyboardViewController?.isRecording
+
+            if false == self?.inRotation &&
+               (false == isRecording || nil == isRecording) {
+                self?.mode = .textInput
+            }
```
johnxnguyen, line 234: I think it would make it clearer to read this if these were not optional. Just set both values to `false` if they're optional.

billypchan: fixed

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -226,11 +226,11 @@ extension ConversationInputBarViewController {
...
-            let isRecording = self?.audioRecordKeyboardViewController?.isRecording
-
-            if false == self?.inRotation &&
-               (false == isRecording || nil == isRecording) {
-                self?.mode = .textInput
+            guard let weakSelf = self else { return }
+            
+            if false == weakSelf.inRotation &&
+               false == (weakSelf.audioRecordKeyboardViewController?.isRecording ?? false) {
+                weakSelf.mode = .textInput
```
johnxnguyen, line 233: I was thinking more along the lines of

```swift
let inRotation = self?.inRotation ?? false
let isRecording = self?.audioRecordKeyboardViewController?.isRecording ?? false

if !inRotation && !isRecording {
  self?.mode = .textInput
}
```

PR #4245, opened 2020-04-02
Retire NSItemProvider helper

billypchan:
Retire objc `NSItemProvider` helpers.


Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -134,11 +134,12 @@ final class UnsentImageSendable: UnsentSendableBase, UnsentSendable {
...
         //
         
-        self.attachment.loadItem(forTypeIdentifier: kUTTypeImage as String, options: options, urlCompletionHandler: { [weak self] (url, error) in
+        attachment.loadItem(forTypeIdentifier: kUTTypeImage as String, options: options) { [weak self] (url, error) in
             error?.log(message: "Unable to load image from attachment")
             
             //Tries to load the content from local URL...
-            if let url = url, let imageSource = CGImageSourceCreateWithURL(url as CFURL, nil) {
+            if let url = url,
+                let imageSource = CGImageSourceCreateWithURL(url as! CFURL, nil) {
```
johnxnguyen, line 142: Perhaps we should safely cast url in the line above.

billypchan: I tried casting there, but it gives error:
`Conditional downcast to CoreFoundation type 'CFURL' will always succeed`
The force casting is a must.

billypchan: fixed

PR #4247, opened 2020-04-02
Clean up objc related setting in share extension target

billypchan:
## What's new in this PR?

remove objc related setting in share extension target.

### TODO
- [x] define these in .xcconfig after UI project has no objc code.


johnxnguyen:
How long until the UI project is Objc free?

billypchan:
@johnxnguyen there should be one more PR to clean up .h and .m files.

johnxnguyen:
So maybe we should clean up the .h and .m files first, then we can add these settings to .xcconfig

billypchan:
@johnxnguyen I undo the project file and will change it in a new PR.

PR #4251, opened 2020-04-06
Replace objc loading spinner with Swift version

billypchan:
### Tested cases
- [x] `LegalHoldAlertFactory.makeLegalHoldActivationAlert`
- [x] `ConversationInputBarViewController`: picker a video from document icon button (spinner on `UIImagePicker` is removed)
- [x] Setting: `receiveNewsAndOffers`
- [x] `SettingsSignOutCellDescriptor.logout()`
- [x] remove phone number (`ClearBackgroundNavigationController`)
- [x] `ClientListViewController`


Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -25,15 +25,15 @@ protocol SpinnerCapable: class {
...
         dismissSpinner = presentSpinner(title: title)
     }
 
-    var isSpinnerVisible: Bool {
+    var showLoadingView: Bool {
         set {
             if newValue {
                 // do not show double spinners
-                guard !isSpinnerVisible else { return }
+                guard !showLoadingView else { return }
```
johnxnguyen, line 36: Suggestion: `isLoadingViewVisible`

billypchan: I would prefer keeping original objc method name `showLoadingView ` to prevent unnecessary renaming. We can change it after this PR.

johnxnguyen: But why not do it in this PR? I would argue that it is necessary renaming and easy to achieve the with renaming tool in Xcode. Or am I missing something here?

johnxnguyen: It's also part of our style guide.

billypchan: renamed.

Wire-iOS/Sources/UserInterface/Components/Loading/LoadingSpinner.swift
```diff
@@ -99,3 +99,17 @@ fileprivate final class LoadingSpinnerView: UIView {
...
+        get {
+            return (self as? SpinnerCapableViewController)?.isLoadingViewVisible
+        }
+
+        set {
+            if let newValue = newValue {
+                (self as? SpinnerCapableViewController)?.isLoadingViewVisible = newValue
+            }
+        }
+    }
```
johnxnguyen, line 114: Is this being used? I couldn't see any reference to it. Also, this should be renamed too.

billypchan: fixed, this is used in Setting screen's pushed view controllers.

PR #4264, opened 2020-04-14
Fix: sending a 4K video is allowed and it is not viewable on android

billypchan:
### Issues

When sending a video from camera keyboard, the file can exceed file size limit and not visible on Android devices. (Sending form image picker has no such issue)

### Causes

If pick a 4K video from the camera keyboard, the video is not resized. (Pick from image picker would have default quality, which is 720p)

### Solutions

If the video has a width wider than 1920px, Export the video with `AVAssetExportPreset1920x1080`.
Set default video preset when select a video from image picker.
Set `fileLengthLimit` to self user's max file size limit.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
@@ -313,13 +313,15 @@ class CameraKeyboardViewController: UIViewController, SpinnerCapable {
             })
 
             AVURLAsset.convertVideoToUploadFormat(at: url, fileLengthLimit: Int64(fileLengthLimit)) { resultURL, asset, error in
+                DispatchQueue.main.async(execute: {
+                    self.isLoadingViewVisible = false
+                })
+                
                 guard error == nil,
                     let resultURL = resultURL,
                     let asset = asset else { return }
                 
                 DispatchQueue.main.async(execute: {
-                    self.isLoadingViewVisible = false
-
                     self.delegate?.cameraKeyboardViewController(self, didSelectVideo: resultURL, duration: CMTimeGetSeconds(asset.duration))
                 })
             }
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
@@ -299,7 +299,8 @@ class CameraKeyboardViewController: UIViewController, SpinnerCapable {
 
     fileprivate func forwardSelectedVideoAsset(_ asset: PHAsset) {
         isLoadingViewVisible = true
-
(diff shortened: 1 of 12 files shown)


WireCommonComponents/AVAsset+VideoConvert.swift
```diff
@@ -114,6 +128,10 @@ extension AVURLAsset {
         
         guard let exportSession = AVAssetExportSession(asset: self, presetName: quality) else { return }
         
+        if let fileLengthLimit = fileLengthLimit {
+            exportSession.fileLengthLimit = fileLengthLimit
```
johnxnguyen, line 132: So what does this do? Does it constrain the file size? Or does this refer to video length?

billypchan: From apple document: "The file length that the output of the session should not exceed."
Tested with device and found that setting this value would affect video compression rate.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
```diff
@@ -302,45 +296,45 @@ class CameraKeyboardViewController: UIViewController, SpinnerCapable {
...
+                self.isLoadingViewVisible = true
+            })
+
+            AVURLAsset.convertVideoToUploadFormat(at: url, fileLengthLimit: Int64(ZMUserSession.shared()!.maxUploadFileSize)) { resultURL, asset, error in
+                if error == nil,
+                    let resultURL = resultURL {
+                    DispatchQueue.main.async(execute: {
+                        self.isLoadingViewVisible = false
+
+                        self.delegate?.cameraKeyboardViewController(self, didSelectVideo: resultURL, duration: CMTimeGetSeconds(asset!.duration))
```
johnxnguyen, line 320: It'd be better to safely unwrap the asset with the url.

billypchan: fixed.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/CameraKeyboard/CameraKeyboardViewController.swift
```diff
@@ -302,45 +296,47 @@ class CameraKeyboardViewController: UIViewController, SpinnerCapable {
...
-                return
-            }
+            DispatchQueue.main.async(execute: {
+                self.isLoadingViewVisible = true
+            })
+
+            AVURLAsset.convertVideoToUploadFormat(at: url, fileLengthLimit: Int64(fileLengthLimit)) { resultURL, asset, error in
+                guard error == nil,
+                    let resultURL = resultURL,
+                    let asset = asset else { return }
```
johnxnguyen, line 318: By returning early, does it mean that the loading view is not hidden?

billypchan: put the hide spinner code before the guard.

PR #4280, opened 2020-04-20
[Fix] poll button spinner

David-Henner:
### Issues

GIVEN There is a poll message
WHEN The app becomes active (after being in background)
THEN Spinner doesn't show on selected buttons
AND User needs to navigate back and forth to the conversation to refresh buttons statuses

### Causes

The main conversation controller (`ConversationContentViewController`) reacts to `applicationDidBecomeActive` notification by calling `dataSource.resetSectionControllers()`.

Resetting the section controllers causes the `ConversationMessageSectionController` to be deallocated. Which in turn prevents observation of the `ButtonState` of our buttons. Thus, when a button is selected and its state changes, we don't notice and can't update the UI.

### Solutions

According to a conversation with @typfel:
Originally, the logic in `applicationDidBecomeActive` is to reload everything after coming back from the background, since otherwise we might display out of date data.

At first glance, calling `tableView.reloadData()` should reinstantiate the section controllers.
However, it doesn't work. Instead, calling ` dataSource.loadMessages()` in `applicationDidBecomeActive` does.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/ConversationContentViewController.swift
@@ -138,6 +138,7 @@ final class ConversationContentViewController: UIViewController, PopoverPresente
     private func applicationDidBecomeActive(_ notification: Notification) {
         dataSource.resetSectionControllers()
         tableView.reloadData()
+        dataSource.loadMessages()
     }
     
     override func viewDidAppear(_ animated: Bool) {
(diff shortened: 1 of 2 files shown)


typfel:
I'm wondering what happens now if you: 
1. Scroll back in to the history of a conversation
2. Put app into background
3. Put app into foreground

Calling `loadMessages()` makes me think we will loose our scroll position and jump back to the bottom of the conversation.

Would it be an option to call `calculateSections()` instead?

David-Henner:
After testing your scenario, using `loadMessages()` doesn't make us loose the scroll position.
However, I also tested with `calculateSections()` and it works as well.
Using `calculateSections()` saves us performing a fetch request.

typfel:
Maybe we should call `calculateSections()` from inside `resetSectionControllers()` since only calling `resetSectionControllers()` doesn't do anything useful.

typfel:
> After testing your scenario, using `loadMessages()` doesn't make us loose the scroll position.
> However, I also tested with `calculateSections()` and it works as well.
> Using `calculateSections()` saves us performing a fetch request.

You would need to scroll far back enough so that we increase the message window.

PR #4289, opened 2020-04-22
Fix: partner user: Do not include team member in search result

billypchan:
## What's new in this PR?

If self user is a partner, do not include team members in search query.


typfel:
We do not want exclude all team members for partners only the ones that they are not in an active conversation with. This logic is already implemented here https://github.com/wireapp/wire-ios-sync-engine/blob/e4514eeced8676d51c2ad5283957d936cc245bc4/Source/UserSession/Search/SearchRequest.swift#L49

PR #4299, opened 2020-04-27
Fix: clean up objc protocols

billypchan:
## What's new in this PR?

Clean up unnecessary protocol signatures and `NSObjectProtocol` extension.


michaeldong:
this link suggest use AnyObject not class.
 https://forums.swift.org/t/class-only-protocols-class-vs-anyobject/11507
Do we obey this suggestion?

PR #4303, opened 2020-04-27
[Fix] Trigger team refresh immediately in account screen

johnxnguyen:
### Issues

If the team metadata has changed remotely and the user opens the account screen, they do not see the refreshed data until the dismiss the screen and reopen it.

### Causes

When the account screen is presented, the team is marked as needing to be refreshed. However the refresh won't be triggered until the managed object context is saved, which is probably happening when the screen is dismissed. Therefore opening the screen a second time will show the new metadata.

### Solutions

Trigger the refresh immediately.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
@@ -61,11 +61,14 @@ final class SelfProfileViewController: UIViewController {
      * - parameter userRightInterfaceType: The type of object to determine the user permissions.
      */
 
-    init(selfUser: SettingsSelfUser, userRightInterfaceType: UserRightInterface.Type = UserRight.self) {
+    init(selfUser: SettingsSelfUser,
+         userRightInterfaceType: UserRightInterface.Type = UserRight.self,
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -80,6 +80,9 @@ final class SelfProfileViewController: UIViewController {
 
         if selfUser.isTeamMember {
             selfUser.refreshTeamData()
+
+            // Trigger the refresh immediately.
+            SessionManager.shared?.activeUserSession?.managedObjectContext.saveOrRollback()
```
typfel, line 85: The UI should no access the `managedObjectContext` it's "private". The correct to trigger a save in the UI is to wrap `selfUser.refreshTeamData()` inside a `UserSession.shared?.enqueueChanges { }`

johnxnguyen: Ahhh.... good to know. Thanks for the tip.

typfel:
Tests failed, maybe because `enqueueChanges` are asynchronous. There's a synchronous variant called `performChanges`.

johnxnguyen:
@typfel I think I need to mock the user session.

PR #4304, opened 2020-04-27
Fix: remove unnecessary unboxing in ProfileViewControllerViewModel

billypchan:
## What's new in this PR?

Remove unnecessary unboxing of `bareUser` in `ProfileViewControllerViewModel`.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
@@ -289,10 +289,8 @@ final class ProfileDetailsContentController: NSObject,
             $0.name == (self.isAdminState ? ZMConversation.defaultAdminRoleName : ZMConversation.defaultMemberRoleName)
         }
         
-        guard
-            let role = newParticipantRole,
-            let session = ZMUserSession.shared(),
-            let user = (user as? ZMUser) ?? (user as? ZMSearchUser)?.user
+        guard let role = newParticipantRole,
+              let session = ZMUserSession.shared()
             else { return }
         
         conversation?.updateRole(of: user, to: role, session: session) { (result) in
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/UserProfile/ProfileViewControllerViewModel.swift
```diff
@@ -68,7 +65,7 @@ final class ProfileViewControllerViewModel: NSObject {
     }
     
     var fullUser: ZMUser? {
-        return (bareUser as? ZMUser) ?? (bareUser as? ZMSearchUser)?.user
+        return bareUser as? ZMUser
```
typfel, line 68: This won't work since casting a `ZMSearchUser` to `ZMUser` will fail.

Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -289,10 +289,8 @@ final class ProfileDetailsContentController: NSObject,
             $0.name == (self.isAdminState ? ZMConversation.defaultAdminRoleName : ZMConversation.defaultMemberRoleName)
         }
         
-        guard
-            let role = newParticipantRole,
-            let session = ZMUserSession.shared(),
-            let user = (user as? ZMUser) ?? (user as? ZMSearchUser)?.user
```
typfel, line 295: I would not remove this casting for now since it will break things if the profile is initialised with a `ZMSearchUser`

billypchan: @typfel I will check if it is possible to replace `ZMUser` with `UserType`

typfel: It's fine to keep like it is for now since this code is only executed in the context of conversation where we are guaranteed to always have a real ZMUser.

PR #4308, opened 2020-04-28
[Feature] Call Indicator

marcomaddy1982:
## What's new in this PR?

This PR is still Work In progress.

When an user is in a situation of "broken leg" scenario the client should show a banner that informs the user about the call state.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
@@ -62,12 +62,15 @@ class CallQualityIndicatorViewController: UIViewController {
     }()
 
     private let moreInfoButton: Button = {
-        let button = Button(style: .empty)
-        button.setTitle("call.quality.indicator.more_info.button.text".localized,
+        let button = Button(style: .empty,
+                            titleLabelFont: .smallMediumFont)
+        button.setTitle("call.quality.indicator.more_info.button.text".localized.uppercased(),
                         for: .normal)
         button.textTransform = .none
+        button.backgroundColor = UIColor(white: 1, alpha: 0.24)
+        button.setBorderColor(UIColor(rgb: (230, 6, 6)), for: .normal)
         button.setTitleColor(.white, for: .normal)
-        button.setBorderColor(UIColor(white: 1, alpha: 0.2), for: .normal)
+        button.layer.cornerRadius = 13
         return button
     }()
 
@@ -92,8 +95,8 @@ class CallQualityIndicatorViewController: UIViewController {
     }
     
     private func setUpViews() {
-        view.backgroundColor = UIColor(rgb: (196, 75, 70))
-        view.layer.cornerRadius = 5
+        view.backgroundColor = UIColor(rgb: (230, 6, 6))
+        view.layer.cornerRadius = 4
         view.clipsToBounds = true
 
         for subview in [messageLabel, dismissButton, moreInfoButton] {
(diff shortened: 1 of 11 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
```diff
@@ -0,0 +1,140 @@
...
+import UIKit
+
+protocol CallQualityIndicatorViewControllerDelegate: class {
+    func callQualityIndicatorViewController(_ callQualityIndicatorViewController: CallQualityIndicatorViewController, didTapDismissButton: UIButton)
+    func callQualityIndicatorViewController(_ callQualityIndicatorViewController: CallQualityIndicatorViewController, didTapMoreInfoButton: UIButton)
+}
+
+class CallQualityIndicatorViewController: UIViewController {
+
+    // MARK: - Public Properties
```
billypchan, line 29: should be `internal`?

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
```diff
@@ -0,0 +1,140 @@
...
+    
+    // MARK: - Private Properties
+    private let padding = UIEdgeInsets(top: 16, left: 16, bottom: 8, right: 8)
+    private let spacing: CGFloat = 16
+
+    // MARK: - Components
+
+    private let messageLabel: UILabel = {
+        let label = UILabel()
+        label.text = "Your calling relay is not reachable. This may affect your call experience.".uppercased()
```
billypchan, line 48: The string should be localised.

billypchan:
We should create snapshot testing for the new indicator.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
marcomaddy1982: No they couldn't be.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
marcomaddy1982: Done also for the "More Info" button text

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
```diff
@@ -0,0 +1,159 @@
...
+            moreInfoButton.trailingAnchor.constraint(equalTo: view.trailingAnchor,
+                                                     constant: -padding.trailing),
+            moreInfoButton.bottomAnchor.constraint(equalTo: view.bottomAnchor,
+                                                   constant: -padding.bottom)
+        ])
+    }
+    
+    private func setUpMessageLabelConstraints() {
+        messageLabel.setContentHuggingPriority(.defaultLow, for: .horizontal)
+        messageLabel.setContentHuggingPriority(.defaultLow, for: .vertical)
```
johnxnguyen, line 133: I'm not sure if these are necessary. Some constraints break but recover, so I was trying to avoid them breaking in the first place but I'm not sure if it worked.

marcomaddy1982:
> Does the banner need to adapt to dark mode? I just hardcoded colors while setting up the view and was thinking to make them dynamic later.

I honestly don't know cause it is not specified. I ask in the chat.

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallQualityIndicatorViewController.swift
marcomaddy1982: ok, thanks for the info

marcomaddy1982: @johnxnguyen FYI: everything remains as it is also for dark mode

PR #4311, opened 2020-04-29
fix: SelfProfileViewController observes team name update

billypchan:
## What's new in this PR?

Observes team name update in `SelfProfileViewController` and update the team name label text.


Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -119,6 +121,10 @@ final class SelfProfileViewController: UIViewController {
         navigationItem.rightBarButtonItem = navigationController?.closeItem()
         configureAccountTitle()
         createConstraints()
+        
+        if let team = ZMUser.selfUser()?.team {
```
typfel, line 125: Why add the observe here and not in the `ProfileHeaderViewController` where we want to call the `updateTeamLabel()` method?

johnxnguyen:
do we also need to observe the icon?

PR #4316, opened 2020-05-04
Fix: app crash at fetchController.fetchRequest.fetchOffset > 0

billypchan:
### Issues

App crash at the line `if fetchController.fetchRequest.fetchOffset > 0`

### Causes

`fetchController` has optional type and it may be nil

### Solutions

Use optional unwrap instead.


typfel:
Looks like it was caused by this change: https://github.com/wireapp/wire-ios/pull/4280

PR #4339, opened 2020-05-11
Fix: Xcode 11 - UI is not response to touch event after a call is minimised 

billypchan:
### Issues

After a call screen is minimised, the conversation screen is not responsible to touch events.

### Causes

`CallWindow` is not hidden, and touch evert does not pass though that window.

### Solutions

Hide the window after dismissal


typfel:
Do we unhide the window again when you receive the second call?

billypchan:
@typfel Yes, I tested with simulator. It is handled in another class.

PR #4340, opened 2020-05-11
[Feature] Add informative screen when moving to custom backend

David-Henner:
## What's new in this PR?

New screen informing the user they will be redirected to a custom backend. The screen appears for 5 seconds and then disappears to show the sso login page.


David-Henner:
TODO: Fix colours for dark mode

WireCommonComponents/CircularProgressView.swift
```diff
@@ -45,7 +45,7 @@ public final class CircularProgressView: UIView {
     fileprivate func setupShapeLayer() {
         self.createPath()
         self.shapeLayer.lineWidth = CGFloat(lineWidth)
-        self.shapeLayer.lineCap = .square
+        self.shapeLayer.lineCap = lineCap
```
billypchan, line 48: We should set it with a new properties to prevent affecting existing caller of this class.

Wire-iOS/Sources/UserInterface/Company Login/PreBackendSwitchViewController.swift
```diff
@@ -0,0 +1,250 @@
...
+    
+    let headerView: UIView = {
+        let view = UIView()
+        view.backgroundColor = .black
+        return view
+    }()
+    
+    // Replace by the loading view once ready
+    let progressContainerView: UIView = {
+        let view = UIView(frame: CGRect(x: 0, y: 0, width: 80, height: 80))
```
billypchan, line 54: it would be better to use constraints instead of setting the frame.

David-Henner:
TODO: Accessibility identifiers

Wire-iOS/Sources/UserInterface/Company Login/PreBackendSwitchViewController.swift
```diff
@@ -0,0 +1,250 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+import WireCommonComponents
+import Cartography
```
billypchan, line 22:
```suggestion

```

WireCommonComponents/CircularProgressView.swift
David-Henner: `lineCap` is a new property of `CircularProgressView` set to `.square` by default

typfel:
> TODO: Fix colours for dark mode

Isn't this in the login/registration flow which doesn't have a dark mode?

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1399,6 +1399,9 @@
 "login.sso.error.alert.domain_not_registered.message" = "This email cannot be used for enterprise login. Please enter the SSO code to proceed.";
 "login.sso.error.alert.domain_associated_with_wrong_server.message" = "This email is linked to a different server, but the app can only be connected to one server at a time. Please log out of all Wire accounts on this device and try to login again.";
 "login.sso.error.alert.unknown.message" = "Please try again later.";
+"login.sso.backend_switch.title" = "Leaving Wire";
+"login.sso.backend_switch.subtitle" = "You are being redirected to";
+"login.sso.backend_switch.information" = "Provide credentials only if you are sure it is your Organization log in.";
```
johnxnguyen, line 1404: I'm not sure I understand this copy. Provide your credentials only if you're certain you are trying to log in to your organization?

David-Henner:
> > TODO: Fix colours for dark mode
> 
> Isn't this in the login/registration flow which doesn't have a dark mode?

The title is misleading, the change needed is to set the text color explicitly to black so that it doesn't end up being white text on white background if the user has enabled dark appearance

Wire-iOS/Resources/Base.lproj/Localizable.strings
David-Henner: Yes, the user is being redirected to a custom backend where he will need to enter his credentials to log in, thus needs to make sure he's been redirected to his organization

johnxnguyen: Ok, is `Organization` supposed to be capitalized? Just seems a bit odd to me.

David-Henner: From the design specs yes, but it might be wrong

PR #4341, opened 2020-05-11
[Fix] Colors in share extension when running in dark mode

typfel:
### Issues

Colors are not  readable when running the share extension in dark mode.

### Causes

The share extension is using hard coded UIColors (.white, .black), which doesn't look correct when running in dark mode.

### Solutions

Start using semantic UIColors in the share extension which adapt to the current appearance (dark/light).
### Notes

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13388


typfel:
Tests failing are unrelated, they were disabled in this PR https://github.com/wireapp/wire-ios/pull/4339

PR #4344, opened 2020-05-12
Fix: Xcode 11: Sketch is unusable due to gesture conflict with swipe to dismiss

billypchan:
### Issues

After upgraded to iOS13 SDK, Sketch does not work

### Causes

touch moved gesture is conflict with new modal presentations swipe to dismiss gesture.

### Solutions

Disable the dismiss gesture when touch began on Canvas, and resume it after touch ended.


typfel:
Could we achieve the same thing by setting `isModalInPresentation` to `True`? https://developer.apple.com/documentation/uikit/uiviewcontroller/3229894-ismodalinpresentation

billypchan:
@typfel I will keep this PR open and fix it in another PR. Will open a discussion with design team to see if we want the new swipe to dismiss feature.

billypchan:
@typfel I find this PR would be a simpler solution. There are several points in the project which creates `CanvasViewController` and wrap it in a navigation controller and then present it. Some cases would presents in a cascade way e.g. when sketch a photo. I would prefer this solution since it can fix the issue within `CanvasViewController`

typfel:
> @typfel I find this PR would be a simpler solution. There are several points in the project which creates `CanvasViewController` and wrap it in a navigation controller and then present it. Some cases would presents in a cascade way e.g. when sketch a photo. I would prefer this solution since it can fix the issue within `CanvasViewController`

I see, but I don't think this approach is the right way to go. Take look at this stack overflow answer https://stackoverflow.com/a/57635168

I think option 3 is what we should do and limiting it to only run our own gesture recognizers in the canvas view.

typfel:
Tried it out and it works: https://github.com/wireapp/wire-ios-canvas/pull/24

billypchan:
@typfel Tested and it works! Thanks

PR #4347, opened 2020-05-13
[Fix]: Remove devices step during login includes temporary clients

marcomaddy1982:
### Issues

The temporary device(s) were shown in the remove device scree after the login.

### Causes

Wrong filter applied

### Solutions

Correct filter applied


typfel:
Failing tests seems unrelated, restarted tests.

PR #4358, opened 2020-05-18
Fix: Call screen is not presented when call quality screen is dismissing

billypchan:
### Issues

When a call quality screen is presented and a new incoming call is ringing, the call screen does not present, tap on the top bar does not open the call screen.

### Causes

The dismissal of the call quality screen and the presenting of the call screen starts at the same time and the later one fails to complete.

### Solutions

Delay the presenting of the call screen in this case.

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -19,7 +19,6 @@
 import Foundation
 import UIKit
 import SafariServices
-import WireDataModel
 import WireSyncEngine
 import avs
 import WireCommonComponents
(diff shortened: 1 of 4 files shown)


billypchan:
@johnxnguyen I also thought about that. The problem is the dismissal and presenting are trigged from 2 parts of the project( `CallQualityController` and `CallController`) and triggered by the same observer event. The presenting and the dismissal are not always happens at the same time.

johnxnguyen:
Is the calling survey actually doing anything anymore?

billypchan:
Yes, calling survey is still in use.

johnxnguyen:
> @johnxnguyen I also thought about that. The problem is the dismissal and presenting are trigged from 2 parts of the project( CallQualityController and CallController) and triggered by the same observer event. The presenting and the dismissal are not always happens at the same time.

Isn't the `CallQualitiyController` asking its delegate (`CallQualityController`) to dismiss the survey?

Perhaps we could make only the `CallQualityController` do the dismissal within the `presentCall` method, so we could use the completion handler?

billypchan:
@johnxnguyen rewrite as your suggestion.

PR #4359, opened 2020-05-18
Move Update Licenses File and style icon script to setup.sh

billypchan:
## What's new in this PR?

To save running script "Update Licenses File" every time when running the project, it is moved to `setup.sh`.


johnxnguyen:
Aren't run phase scripts only run when one of the input or output files have changed?

billypchan:
@johnxnguyen They run every time when the project builds, just link "copy resource"/Linking phrases.

setup.sh
```diff
@@ -57,4 +57,23 @@ echo "ℹ️  Doing additional postprocessing..."
...
+echo ""
+
+echo "ℹ️  Update Licenses File..."
+export SCRIPT_INPUT_FILE_0=Cartfile.resolved
+export SCRIPT_INPUT_FILE_1=Carthage/Checkouts
+export SCRIPT_INPUT_FILE_2=EmbeddedDependencies.plist
+
+export SCRIPT_OUTPUT_FILE_0=Wire-iOS/Resources/Licenses.generated.plist
+
+swift run --package-path ./Scripts/update_licenses/
```
johnxnguyen, line 76: I don't think it makes sense to continue using environment vars, since it was a bridge between the Xcode and the script. The input files shouldn't change so I don't see why we couldn't just hardcode the paths in the scripts themselves, it would keep setup.sh nice and clean.

billypchan: will write it. thanks!

Scripts/updateLicenses/main.swift
```diff
@@ -116,28 +107,20 @@ extension String {
...
-        fail("The first input file in Xcode must be the 'Cartfile.resolved' file.")
-    }
-
-    guard let checkoutsPath = getEnvironmentValue(key: "SCRIPT_INPUT_FILE_1") else {
-        fail("The second input file in Xcode must be the 'Cartfile/Checkouts' folder.")
-    }
-
-    guard let embeddedDependenciesPath = getEnvironmentValue(key: "SCRIPT_INPUT_FILE_2") else {
-        fail("The third input file in Xcode must be the 'Cartfile/Checkouts' folder.")
+    guard CommandLine.arguments.count >= 3 else {
```
johnxnguyen, line 110: Does `CommandLine.arguments` include the process name as the first argument? If so, shouldn't we check there are `>= 4` arguments here?

billypchan: replaced with exact argument count check

Scripts/updateLicenses/main.swift
```diff
@@ -222,6 +197,11 @@ func generateFromCartfileResolved(_ content: String, checkoutsDir: URL) -> [Depe
 
 // MARK: - Execution
 
+guard CommandLine.arguments.count == 5 else {
+    fail("Must have 3 input parameters and 1 output parameter. The output file in Xcode must be the 'Wire-iOS/Resources/Licenses.generated.plist' file.")
```
johnxnguyen, line 201: Why do we use the command line arguments when the arguments must be specific values?

billypchan: Seems like we can change the input files some times ago but it is not necessary now.

johnxnguyen: Can we also hardcode the arguments for the license script?

PR #4361, opened 2020-05-19
Fix: Call banner not tappable

billypchan:
### Issues

When call screen is swiped to dismiss, the green return to call bar is not tappable.

### Causes

`CallController.activeCallViewController` is not cleared in this case.

### Solutions

Clean `CallController.activeCallViewController` when dismissed.


billypchan:
Fix the window hiding logic: hide the window when animation is finished instead of pan gesture begins.

Wire-iOS/Sources/UserInterface/Helpers/InteractiveModalTransition.swift
```diff
@@ -131,9 +139,7 @@ final private class ModalInteractionController: UIPercentDrivenInteractiveTransi
             interactionInProgress = true
 
             weak var window = presentationViewController.view.window
```
johnxnguyen, line 141: This property is unused now.

Wire-iOS/Sources/UserInterface/Helpers/InteractiveModalTransition.swift
```diff
@@ -115,6 +115,14 @@ final private class ModalInteractionController: UIPercentDrivenInteractiveTransi
     private var shouldCompleteTransition = false
     private weak var presentationViewController: ModalPresentationViewController!
 
+    override func finish() {
+        super.finish()
+
+        if shouldCompleteTransition {
+            (presentationViewController.view.window as? CallWindow)?.isHidden = true
+        }
```
johnxnguyen, line 123: By placing this here, the window will disappear as soon as the gesture ends, rather that when the dismiss animation completes. So when the calling screen is dragged down past the threshold, as soon as the finger lifts the screen the window disappears suddenly.

billypchan: Thanks for suggestion and I move this part to animation completion block.

johnxnguyen: Did you manually test these changes? I think it's the only way to verify that it works.

billypchan: Yes, tested the several way to pan the screen and seems fine.

Wire-iOS/Sources/UserInterface/Helpers/InteractiveModalTransition.swift
```diff
@@ -98,13 +98,18 @@ final private class ModalDismissalTransition: NSObject, UIViewControllerAnimated
...
             options: [.curveLinear, .allowUserInteraction],
-            animations: animations,
-            completion: transitionContext.complete
-        )
+            animations: animations) { success in
+            transitionContext.complete(success)
+                
+            if success {
+                callWindow?.isHidden = true
+            }
```
johnxnguyen, line 111: `success` here means that the animations finished before the completion handler was called. Perhaps it's also possible that `success` is `true` even if the animation is reversed back to the initial state because the transition was cancelled?

billypchan: Add a check for bounce back case.

PR #4371, opened 2020-05-22
saving Carthage downloading if cache exists

billypchan:
## What's new in this PR?

To saving downloading Carthage downloading on CI server everytime, apply `--cache-builds` to use local cache if exists.

### Note 
if `Carthage/Build` folder is deleted, the cache is gone and the machine has to download again.


billypchan:
Seems this PR only helps local build but not CI build since CI always cleans up Carthage folder every time.

PR #4374, opened 2020-05-25
Feature: observe system apperance change

billypchan:
## What's new in this PR?

Observe `traitCollection.userInterfaceStyle` updates in `AppRootViewController` and update apperance. Add a new dark theme option `auto` to let user to choose follows iOS system style. (`auto` is default value)

Change the Setting screen's dark theme option from switch to option group to allow choosing from 3 values.

### TODO

- [x] Change setting to 3 values: on/off/auto
- [x] test on iPad
- [x] test on iOS 11 device
- [ ] some screens (e.g. conversation detial screen) are not response to system apperance changes


johnxnguyen:
Where does the `auto` copy come from?

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1096,6 +1096,11 @@
 "self.settings.popular_demand.send_button.footer" = "Disable to send via the return key.";
 "self.settings.popular_demand.dark_mode.footer" = "Switch between dark and light theme.";
 
+// dark theme
+"dark_theme.option.dark" = "dark";
+"dark_theme.option.light" = "light";
+"dark_theme.option.auto" = "auto";
```
johnxnguyen, line 1102: Where does this copy come from?

billypchan:
@johnxnguyen It is a temporary copy, I think "Use system setting" would be clearer

Wire-iOS/Settings+ColorScheme.swift
```diff
@@ -57,6 +54,19 @@ extension Settings {
             return .dark
         case "light":
             return .light
+        case "auto":
```
johnxnguyen, line 57: It would be nice to implement the TODO above and move this into an init for `SettingsColorScheme`.

Wire-iOS/Resources/Base.lproj/Localizable.strings
billypchan: How about changing it to "Use system setting" ?

Wire-iOS/Settings+ColorScheme.swift
```diff
@@ -57,6 +54,19 @@ extension Settings {
...
+                switch UIApplication.userInterfaceStyle {
+                case .light:
+                    return .light
+                case .dark:
+                    return .dark
+                default:
+                    return .light
+                }
+            } else {
+                fatal("auto only supported in iOS 13+")
```
johnxnguyen, line 68: iOS 13 or 12?

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsPropertySelectValueCellDescriptor.swift
```diff
@@ -45,24 +44,22 @@ class SettingsPropertySelectValueCellDescriptor: SettingsPropertyCellDescriptorT
...
     }
-    
+
     func select(_ value: SettingsPropertyValue?) {
         do {
-            try self.settingsProperty.set(newValue: self.value)
-        }
-        catch (let e) {
+            try settingsProperty.set(newValue: self.value)
+        } catch (let e) {
```
johnxnguyen, line 59: `let e` here is redundant

Wire-iOS/Settings+ColorScheme.swift
billypchan: fixed

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsPropertySelectValueCellDescriptor.swift
billypchan: fixed

Wire-iOS/Settings+ColorScheme.swift
johnxnguyen: I think it would make more sense to have a new case `SettingsColorScheme` case called `system`. The logic of which variant `light` or `dark` can be moved to `colorSchemeVariant`.

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1096,6 +1096,11 @@
 "self.settings.popular_demand.send_button.footer" = "Disable to send via the return key.";
 "self.settings.popular_demand.dark_mode.footer" = "Switch between dark and light theme.";
 
+// dark theme
+"dark_theme.option.dark" = "dark";
+"dark_theme.option.light" = "light";
+"dark_theme.option.auto" = "Use system setting";
```
johnxnguyen, line 1102: I would suggest `system`. In any case, can you get in touch with design to get some feedback about this? I think we should avoid to introduce new copy without informing design because it can be confusing for other people when they suddenly see new copy.

billypchan: Updated text copy form designers

Wire-iOS/Settings+ColorScheme.swift
billypchan: Merged the enum `DarkThemeOption` to `SettingsColorScheme`

billypchan:
@johnxnguyen changed the copy

Wire-iOS/Settings+ColorScheme.swift
```diff
@@ -16,59 +16,93 @@
...
 import WireSystem
+import UIKit
 
-enum SettingsColorScheme {
-    case light
-    case dark
+enum SettingsColorScheme: Int {
+    case light = 0
+    case dark = 1
+    @available(iOS, introduced: 12.0, message: "system only supported in iOS 12+")
```
johnxnguyen, line 25: Do we really need the `@available` annotation for this enum case?

billypchan: I would prefer keeping this to remind the code readers about this upgrade of `SettingsColorScheme` and testing the backward compatibility of old versions.

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -994,7 +994,7 @@
 
 "self.settings.account_personal_information_group.title" = "Personal Information";
 
-"self.settings.account_picture_group.theme" = "Dark Theme";
+"self.settings.account_picture_group.theme" = "Themes";
```
johnxnguyen, line 997: Actually it looks like `Theme` would be better, to stay consistent with the other option names. I'm not sure if design realized this. I think it's safe to change it. Sorry for the confusion!

Wire-iOS/Settings+ColorScheme.swift
```diff
@@ -16,59 +16,93 @@
...
+            return .system
+        }
+
+        return .light
     }
 
-    var colorSchemeVariant: ColorSchemeVariant {
-        guard let string: String = self[.colorScheme] else {
-            return .light
+    static var allOptions: [SettingsColorScheme] {
```
johnxnguyen, line 75: Can we make `SettingsColorScheme` conform to `CaseIterable`?

Wire-iOS/Settings+ColorScheme.swift
johnxnguyen: But `system` for old versions would simply mean `light`. And also the settings option wouldn't show the `system` option on older versions.

It's just an opportunity to clean up the `#available` conditions, but it's up to you.

Wire-iOS/Settings+ColorScheme.swift
billypchan: Fixed. About https://github.com/wireapp/wire-ios/pull/4374#discussion_r431662738, if we do not mark `.system` only available after iOS12, we may forget to exclude the new enum value in compile time.

johnxnguyen: Ah right I see what you mean now, because we use `.allCases` to populate the option list.

PR #4380, opened 2020-05-29
[Fix] Views being created multiple times

typfel:
### Issues

The message toolbox sometimes contains multiple copies of the same view

### Causes

Views are created when in `configureForMessage()`, which can be called multiple times.

### Solutions

Create views once in the init.

### Note

This might fix the overlapping resend button bug.


billypchan:
`testThatItRendersNewConversationCellWithParticipantsAndNameAllTeamUsersWithGuests` is unstable and we should disable it again for now.

PR #4387, opened 2020-06-03
Fix: Limit file size in share extension

billypchan:
### Issues

When a file is shared via share extension, the file size is not limited, and sending a very big file may crash the app.

### Solutions

Add file size checking and display an error alert in this case

### TODO
- [x] clean up the helper methods (move some method to common components)
- [x] reuse the error message with the app


Wire-iOS Share Extension/Base.lproj/Localizable.strings
```diff
@@ -31,8 +32,14 @@
...
 "share_extension.not_signed_in.title" = "You need to sign into Wire before you can share anything";
 "share_extension.not_signed_in.close_button" = "Close";
+
 "share_extension.timeout.title" = "The connection has timed out.";
 "share_extension.timeout.message" = "Check your Internet connection and try again. The attachments were saved in the conversation.";
+
+"share_extension.error.title" = "Sharing Error";
+"share_extension.error.file_size_too_big.message" = "File size for the selected account is limited to %d MB.";
+"share_extension.error.unsupported_attachment.message" = "Unsupported attachment.";
+
```
johnxnguyen, line 42: Did this go through design?

Wire-iOS Share Extension/SendController.swift
```diff
@@ -93,26 +94,34 @@ final class SendController {
         }
     }
     
+    typealias SendableCompletion = ([Sendable], UnsentSendableError?) -> Void
```
johnxnguyen, line 97: Is it expected that we receive either some sendables or an error? Perhaps it's better to use the `Result` type instead of a tuple.

Wire-iOS Share Extension/SendController.swift
```diff
@@ -196,14 +205,16 @@ final class SendController {
...
 
         unsentSendables.filter {
-            $0.error != .unsupportedAttachment
+            $0.error == nil
         }.forEach {
             sendingGroup.enter()
             $0.send(completion: appendToMessages)
         }
+        
+        let error = unsentSendables.first(where: {$0.error != nil})?.error
```
johnxnguyen, line 214: You could clean this up with:

```swift
let error = unsentSendables.compactMap(\.error).first
```

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -30,6 +30,18 @@ enum UnsentSendableError: Error {
...
     // `UnsentSendables` that report this error should not be sent.
     case unsupportedAttachment
+
+    // The attachment is over file size limitation
+    case fileSizeTooBig
+
+    var localizedString: String {
+        switch self {
+        case .fileSizeTooBig:
+            return String(format: "share_extension.error.file_size_too_big.message".localized, AccountManager.fileSizeLimit / 1048576)
```
johnxnguyen, line 40: What's this magic number `1048576`?

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -357,3 +375,16 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
...
+
+extension AccountManager {
+    // MARK: - Host App State
+    static var sharedAccountManager: AccountManager? {
+        guard let applicationGroupIdentifier = Bundle.main.applicationGroupIdentifier else { return nil }
+        let sharedContainerURL = FileManager.sharedContainerDirectory(for: applicationGroupIdentifier)
+        return AccountManager(sharedDirectory: sharedContainerURL)
+    }
+
+    static var fileSizeLimit: UInt64 {
```
johnxnguyen, line 387: What is the unit of measurement here?

WireCommonComponents/URL+FileSize.swift
```diff
@@ -0,0 +1,36 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension URL {
+    func fileSize() throws -> UInt64? {
```
johnxnguyen, line 21: Are the thrown errors of this method ever considered? If not, you could do just return `nil` if any errors are caught, it would clean up the calling sites bit.

Wire-iOS Share Extension/UnsentSendable.swift
billypchan: 1024^2

Wire-iOS Share Extension/UnsentSendable.swift
billypchan: all units to measure file size is byte.

johnxnguyen: It'd be nice to include that in the name.

Wire-iOS Share Extension/UnsentSendable.swift
johnxnguyen: Can you give it a name to make it clearer what it's purpose is?

billypchan: refactor as a constant.

Wire-iOS Share Extension/Base.lproj/Localizable.strings
billypchan: I found we have same strings in the main app and I will reuse "You can send files up to %@"

Wire-iOS Share Extension/SendController.swift
billypchan: There is a case the would return empty `Sendable` with nil error (when the user cancel the sharing). I would prefer having two parameters in the closure's parameter.

Wire-iOS Share Extension/UnsentSendable.swift
```diff
@@ -30,6 +30,18 @@ enum UnsentSendableError: Error {
     // which does not contain a File, e.g. an URL to a webpage, which instead will also be included in the text content.
     // `UnsentSendables` that report this error should not be sent.
     case unsupportedAttachment
+
+    // The attachment is over file size limitation
+    case fileSizeTooBig
+
+    var localizedString: String {
```
typfel, line 37: You could make this error conform to `LocalizedError`

Wire-iOS Share Extension/SendController.swift
johnxnguyen: An error case for `cancelled` seems reasonable. This approach is (in my opinion) more readable. Which benefit is there to keep two closure arguments?

Wire-iOS.xcodeproj/xcshareddata/xcschemes/Wire-iOS.xcscheme
```diff
@@ -79,6 +79,9 @@
                <Test
                   Identifier = "StartedConversationCellTests/testThatItRendersNewConversationCellWithParticipantsAndNameAllTeamUsersFromSmallTeamWithManyGuests()">
                </Test>
+               <Test
+                  Identifier = "StartedConversationCellTests/testThatItRendersNewConversationCellWithParticipantsAndNameAllTeamUsersWithGuests()">
```
johnxnguyen, line 83: Is this relevant?

billypchan: no, but I also disable the test in another PR to avoid unstable result.

Wire-iOS Share Extension/UnsentSendable.swift
billypchan: done

Wire-iOS Share Extension/SendController.swift
billypchan: refactored with new enum `SendableResult`

johnxnguyen: Why not use Swift's existing `Result` type?

billypchan: Fixed with using Result. Thanks

PR #4392, opened 2020-06-04
Reduce the build time on CI

billypchan:
## What's new in this PR?

The PR reduces the CI step `QA: build for simulator` with the following adjustments:
-  do not clean the project when building again for QA testing
-  add build number when `build for test` (i.e. build product is same as the `QA: build for simulator` step)
- add `skip_detect_devices` for both steps
- remove unnecessary `devices` & `code_coverage` options.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -51,10 +51,7 @@ platform :ios do
             xcversion(version: xcode_version)
         end
 
-        build_number = options[:build_number]
-        if build_number.nil? 
-            build_number = "0"
-        end
+        build = Build.new(options: options)
 
         run_tests(
             scheme: "Wire-iOS",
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -51,8 +51,14 @@ platform :ios do
             xcversion(version: xcode_version)
         end
 
+        build_number = options[:build_number]
```
typfel, line 54: Why don't we retrieve the build number like we do in the other build steps:

```
build = Build.new(options: options)
 xcargs: "BUILD_NUMBER=#{build.build_number}"
```

billypchan: To create the `Build` object, more unnecessary info have to be provided e.g. build_type, commit comment. I would prefer just provide the build number to solve this PR.

typfel: I think it's good if we do all the steps the same way. The `Build` object does some input validation and throws error if a build number is missing for example.

billypchan: fixed.

PR #2204, opened 2018-05-29
Forward all touches in participants list except for taps on show all cell

daehn:
## What's new in this PR?

* Forward all touches in participants list except for taps on show all cell.
* Clean up video placeholder status UILabel creation.

PR #4411, opened 2020-06-15
[Fix] Call participant list not scrollable

johnxnguyen:
### Issues

When viewing the full call participant list (after tapping the "show all" cell), it is not possible to scroll the list.

### Causes

`CallParticipantView` contains code that limits all touch events to the "show all" cell. This was introduced in https://github.com/wireapp/wire-ios/pull/2204/ but it's not clear why. In any case, because the full call participant list never contains the "show all" cell, the view was forwarding all of its touch events and therefore the internal scroll view gesture recognizers were not receiving these events.

### Solutions

Remove the code limiting touch events. 

After several manual tests, there doesn't seem to be any negative side effects. The interactive dismissal of the call UI works (vertical pan) and touching any of the user cells still results in the call info overlay being dismissed, revealing the video grid below.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/CallParticipantsView.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallParticipantsView/CallParticipantsView.swift
@@ -68,12 +68,6 @@ class CallParticipantsView: UICollectionView, Themeable {
         }
     }
     
-    override func point(inside point: CGPoint, with event: UIEvent?) -> Bool {
-        return subviews.any {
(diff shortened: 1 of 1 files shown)


typfel:
My guess is that this code was introduced to workaround an issue with the overlay controls that disappear after a few second but come back if you tap the screen.

Maybe that would be an additional thing to verify that the hide/showing of overlay controls still work was as expected.

johnxnguyen:
> Maybe that would be an additional thing to verify that the hide/showing of the overlay controls still work was as expected.

This is what I mean by "the call info overlay being dismissed". With these changes, you can still dismiss the overlay controls immediately even when tapping one of the user cells.

watcher699 (commented):
Good

PR #4415, opened 2020-06-17
fix: Input field does not preserve text

billypchan:
### Issues

Input field does not preserve text after the app goes to background.

### Causes

After updated to iOS 13 SDK, the `traitCollectionDidChange` is called twice when the app is going to the background (e.g. from `.light` to `.dark` and then from `.dark` to `.light`.) and triggers `.SettingsColorSchemeChanged` twice, which recreates the input field and loading the old preserved text before the app saved the current input bar text.

### Solutions

Do not triggers `.SettingsColorSchemeChanged` when the app going to the background.

### Discussion
Instead of recreating the whole view controls stack, we should use the new adaptive color in the future.


typfel:
Changing system appearance still works after this change?

PR #4418, opened 2020-06-18
[Feature] New video grid

David-Henner:
## What's new in this PR?

New video grid relying on a `CollectionView` instead of a `StackView`.
This allows for displaying more than 4 participants in the grid. 
Using a `CollectionView` introduces flexibility and reduces complexity.

### Note

Tests need to be updated


Wire-iOS/Sources/Helpers/Int+Even.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension Int {
+    var isEven: Bool {
+        return self % 2 == 0
```
billypchan, line 23: We can save an operator with `.isMultiple(of: 2)`

Wire-iOS/Sources/Helpers/Int+Even.swift
```diff
@@ -0,0 +1,29 @@
...
+    }
+    
+    var evened: Int {
+        return Int(round(Double(self) / 2) * 2)
```
billypchan, line 27: To avoid complications
`return isEven ? self : self - 1`

johnxnguyen: If `round()` rounds up, then you'd want to add one, not subtract.

billypchan: I would suggest this method should be named like `roundedEvenly`. But I am not sure the logic wants to be rounded or floored.

johnxnguyen: Or `nextEven`

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -16,99 +16,156 @@
...
-        didSet {
-            axis = layoutDirection
-            lowerHorizontalStackerView.axis = layoutDirection.flipped
-            upperHorizontalStackerView.axis = layoutDirection.flipped
+// MARK: - Segment calculation
+extension GridView {
+    private func calculate(segments segmentType: SegmentType, for indexPath: IndexPath) -> Int {
+        let values: [ParticipantAmount: [SplitType: Int]] = [
+            .moreThanTwo: [
+                .proportionalSplit: videoStreamViews.count.evened / 2,
```
billypchan, line 61: I guess the logic here makes one more empty cell for odd number of participants? It ture we should use evenlyCeiled instead of evened.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -16,99 +16,156 @@
...
+            switch (layoutDirection, segmentType) {
+            case (.vertical, .row), (.horizontal, .column):
+                self = .proportionalSplit
+            case (.horizontal, .row), (.vertical, .column):
+                self = .middleSplit
+            default:
+                return nil
+            }
+        }
+    }
```
johnxnguyen, line 108: So the `SplitType` is used to describe how we should divide the axis? Does `middleSplit` constrain us to always have a layout where there is either two columns or two rows?

David-Henner: Yes, `SplitType` describes how the axis is divided. 
`middleSplit` means that the axis is divided by two.
`proportionalSplit` means that the axis is divided by the number of participants in that column/row.
There are some exceptions for when the number of participants is below 3.

johnxnguyen: Does this implementation support a `proportionalSplit` on both axes? For example, if we're on a iPad, whose screen proportions more closely resembles that of a desktop computer, would it be possible to have a 3 x 4 grid?

David-Henner: It would need some changes in order to achieve that. Because it is assumed that if one axis is `proportionalSplit` the other should be `middleSplit`

johnxnguyen: Ok, I'll wait until the PR is marked as ready to be reviewed to get a better understand of how it works. In any case, for the initial implementation we don't need to worry about how it looks on iPad 👍

David-Henner: I'm also happy to explain 1:1 in a call

David-Henner: For a 3x4 grid on iPad, we could imagine using a different terminology, like `verticalAxis` and `horizontalAxis` (that would be independent of device orientation). I suppose the enum could then calculate the number or rows and columns based on parameters like device type and orientation

David-Henner: For a 3x4 grid on iPad, we could imagine using a different terminology, like `verticalAxis` and `horizontalAxis` (that would be independent of device orientation). I suppose the enum could then calculate the number or rows and columns based on parameters like device type and orientation

johnxnguyen: Or rather we can consider the size class, from which we calculate the grid `m x n`. Given the `m x n` grid and the available content size, we could then calculate the cell sizes.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -16,99 +16,156 @@
...
-        upperHorizontalStackerView.distribution = .fillEqually
+        init?(_ layoutDirection: UICollectionView.ScrollDirection, _ segmentType: SegmentType) {
+            switch (layoutDirection, segmentType) {
+            case (.vertical, .row), (.horizontal, .column):
+                self = .proportionalSplit
+            case (.horizontal, .row), (.vertical, .column):
+                self = .middleSplit
+            default:
+                return nil
+            }
```
johnxnguyen, line 106: I think all possible cases are covered (2 scroll directions * 2 segment types = 4 cases), so you should be able to make this a non fallible initializer.

Wire-iOS Tests/GridViewTests.swift
```diff
@@ -17,101 +17,114 @@
 //
 
 import Foundation
+import XCTest
 @testable import Wire
 
 class GridViewTests: ZMSnapshotTestCase {
```
billypchan, line 23: I would recommend using new Swift snapshot testing.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -16,99 +16,156 @@
...
+        ]
+        
+        let participantAmount = ParticipantAmount(videoStreamViews.count)
+        guard
+            let splitType = SplitType(layoutDirection, segmentType),
+            let value = values[participantAmount]?[splitType] else {
+                return 1
         }
+        
+        return value
```
johnxnguyen, line 77: I think this could be more readable if you use a switch case instead, something like:

```swift
let participantAmount = ParticipantAmount(videoStreamViews.count)
let splitType = SplitType(layoutDirection, segmentType)

switch (participantAmount, splitType) {
case (.moreThanTwo, .proportionalSplit):
  return ...
case (.moreThanTwo, .middleSplit):
  return ...
case (.twoAndLess, .proportionalSplit):
  return ...
case (.twoAndLess, .middleSplit):
  return ...
}
```

This removes the need for the `guard`, avoids recalculating the entire dictionary repeatedly and makes it a bit easier to mentally follow the code path.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridView.swift
```diff
@@ -16,99 +16,156 @@
...
 
-fileprivate extension NSLayoutConstraint.Axis{
-    var flipped: NSLayoutConstraint.Axis {
-        switch self {
-        case .horizontal: return .vertical
-        case .vertical: return .horizontal
-        @unknown default:
-            fatalError()
+class GridView: NSObject {
+    let collectionView: UICollectionView
```
johnxnguyen, line 22: I would expect a class with the `View` suffix to be a subclass of `UIView`. I think it would make more sense to either make `GridView` subclass `UICollectionView` or to be a `UIView` and then make `collectionView` non public and a subview.

PR #4419, opened 2020-06-18
Feature: display iOS 13 link context menu when long press on a link

billypchan:
## What's new in this PR?

Display the iOS link context when long press on a link.
Markdown link is excluded for long tap context menu.

### snapshots: 

detected link:

context menu when long press on the link:


billypchan:
@johnxnguyen tested with markdown link and it also works.

johnxnguyen:
> @johnxnguyen tested with markdown link and it also works.

What do you mean by it works? Does it show the preview? Because the expected behavior is to show an popup asking the user if they want to visit the url.

billypchan:
@johnxnguyen the behavior of tapping on a markdown link is:
normal tap: show the external link alert
long tap: show the context menu with preview (The user can tap the domain name and choose open it or not)

johnxnguyen:
If I remember correctly, the behavior with force touch peeking was that it should also show the confirmation alert. In my opinion, the long tap should behave just like the force touch gesture.

watcher699 (commented):
Good

watcher699 (commented):
Good

PR #4420, opened 2020-06-18
Use SF symbol icons in the context menu

billypchan:
## What's new in this PR?

Use SF symbol icons in the context menu.


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/MessageAction.swift
```diff
@@ -120,6 +121,62 @@ enum MessageAction: CaseIterable {
...
+    @available(iOS 13.0, *)
+    func systemIcon() -> UIImage? {
+        let icon: UIImage?
+        if let imageSystemName = imageSystemName() {
+            icon = UIImage(systemName: imageSystemName)
+        } else {
+            icon = nil
+        }
+
+        return icon
```
johnxnguyen, line 134: You should be able to simplify this:

```swift
return imageSystemName.map(UIImage.init(systemName:))
```

billypchan: fixed

Issue #4398, opened 2020-06-09
Unable to remove second device on device list


zj299792458:
## Checklist (please check before submitting)

* [☑️ ] I checked the existing issues, and the issue was not reported before.

## The problem

Unable to remove two devices in a row on device list, I have to exit the view and re-enter again to delete the second one, please check reproducing steps below.

## Environment

* Wire version that exhibits the issue: AppStore/3.59
* Last Wire version that did not exhibit the issue (if applicable): YES
* iOS version used to run Wire: iOS 13.5.1
* iOS device type: iPhone X
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): Wi-Fi
* Exact date, time and timezone (only for the crash issues):  8 Jun 2020, 11:28 AM

## Details

1. Enter Settings -> Devices, Assume there are multiple devices (not browser) in the device list
2. Swipe left on one of them and click delete
3. Enter password and confirm, wait till success
4. Stay on this view, swipe left on another device and click delete 
5. It failed directly with prompt saying "wrong password", without requesting password again.

Expectation: succeed like the first time 

## Screenshots

See attachment
![IMG_6278](https://user-images.githubusercontent.com/6792825/84103235-0fa29500-aa45-11ea-90a0-0d4f332d8ed0.jpg)

billypchan:
Hi @zj299792458, thanks for your report. We will check this bug. Thanks!

PR #4422, opened 2020-06-19
Fix: Unable to remove the second device on device list 

billypchan:
### Issues

When moving a second device on the client list, the "Wrong password" alert shows even a correct password is entered.

https://github.com/wireapp/wire-ios/issues/4398

### Causes

1. `passwordIsNecessaryForDelete` is not cleared after successful deletion
2. Multiple instants of `ClientRemovalObserver` exist and trying to present multiple alerts, which may cover the correct alert.

### Solutions

1. reset `passwordIsNecessaryForDelete` for all cases.
2. create `ClientRemovalObserverDelegate`, check the `clientRemovalObserver` before UI actions.


typfel:
I think `passwordIsNecessaryForDelete` at some point was introduced since we remember the password between deleted and thus allowed you delete multiple devices without entering the password. I think we shouldn't allow that but I suspect we can remove `passwordIsNecessaryForDelete` now.

billypchan:
@typfel `passwordIsNecessaryForDelete` is introduced in  https://github.com/wireapp/wire-ios/pull/2829 for SSO login failure case. I think it is still necessary to keep.

typfel:
> @typfel `passwordIsNecessaryForDelete` is introduced in #2829 for SSO login failure case. I think it is still necessary to keep.

I think it's tied to some logic where we attempt to delete a device with existing credentials but if that fails we ask for the password. But ok maybe let's not refactor all that logic right now.

PR #4429, opened 2020-06-23
[Feature] Call participants media icons

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13485

#### Introduces new icons to the participants list of a call.

Screenshot 2020-06-23 at 13 33 23

**Video / Screenshare:** avs provides different video states allowing to differentiate between video and screenshare.  Added UI components and presentation logic.
_TODO: add screenshare icon ressource_

**Microphone muted / unmuted:** avs **does not** yet provide microphone states. Thus, the microphone icons have been added to the UI but the presentation logic is not supported in this PR. 

**Guest / Partner:** unrelated to avs. Added UI components and presentation logic.

### Note

_TODO: Record new snapshots_


Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
```diff
@@ -0,0 +1,42 @@
...
+
+protocol IconImageStyle {
+    var icon: StyleKitIcon? { get }
+}
+
+class IconImageView: UIImageView {
+    private(set) var size: StyleKitIcon.Size = .tiny
+    private(set) var color: UIColor = UIColor.from(scheme: .iconGuest)
+    
+    func set(style: IconImageStyle) {
```
billypchan, line 30: How about write this in a property like
`
var style: IconImageStyle {
 get{
...
 }
set{
 ...
}
}
`

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/VideoIconStyle.swift
```diff
@@ -0,0 +1,56 @@
...
+    case video
+    case screenshare
+    case hidden
+    
+    var icon: StyleKitIcon? {
+        switch self {
+        case .hidden:
+            return .none
+        case .screenshare:
+            return .alien
```
David-Henner, line 33: will be replaced by adequate icon

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
David-Henner: I'd need to add a `_style` property to retain its value for the `get {...}`
But we don't need to get `style`, only to set the icon

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/UserTypeIconStyle.swift
```diff
@@ -0,0 +1,55 @@
...
+            self = .external
+        } else if let conversation = conversation {
+            self = !user.isGuest(in: conversation) || user.isSelfUser ? .member : .guest
+        } else {
+            self = !ZMUser.selfUser().isTeamMember
+                || user.isTeamMember
+                || user.isServiceUser
+                || hideIconView
+                ? .member
+                : .guest
```
johnxnguyen, line 52: The context of `hideIconView` seems to have been lost in this refactoring, it's not clear what significance it has in the method. Is it supposed to mean "hide guest icon"?

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/MicrophoneIconStyle.swift
```diff
@@ -0,0 +1,60 @@
...
+class MicrophoneIconImageView: IconImageView {
+    override func set(style: IconImageStyle) {
+        super.set(style: style)
+        guard
+            let style = style as? MicrophoneIconStyle,
+            case style = MicrophoneIconStyle.active else {
+                return
+        }
+        tintColor = .green
+    }
```
johnxnguyen, line 59: `MicrophoneIconImageView` indicates it's only for microphone icons, but it does can accept any icon style (and therefore other icons). I wonder if we can find a way to improve the type enforcement.

johnxnguyen: What about including an optional `tintColor` property in `IconImageStyle`. You then wouldn't need to subclass at all.

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/UserTypeIconStyle.swift
David-Henner: not sure what the original meaning was, but that's what it does, so I renamed it to `hideGuestIcon`

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/MicrophoneIconStyle.swift
David-Henner: done 6b62152

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/UserTypeIconStyle.swift
johnxnguyen: My concern now is why `hideGuestIcon` is only used in the `else` block and not the `else if` block.

Wire-iOS/Sources/UserInterface/StartUI/Common/SectionControllers/Suggestions/DirectorySectionController.swift
```diff
@@ -54,9 +54,9 @@ class DirectorySectionController: SearchSectionController {
         let user = suggestions[indexPath.row]
         let cell = collectionView.dequeueReusableCell(withReuseIdentifier: UserCell.zm_reuseIdentifier, for: indexPath) as! UserCell
         
-        cell.configure(with: user, hideIconView: true)
+        cell.configure(with: user, hideGuestIcon: true)
         cell.showSeparator = (suggestions.count - 1) != indexPath.row
-        cell.guestIconView.isHidden = true
+        cell.userTypeIconView.isHidden = true
```
johnxnguyen, line 59: Is this the only place the `hideGuestIcon` is `true`? If yes, then we may not need this argument at all since the `userTypeIconView` is hidden anyway.

David-Henner: done 8d2190d

johnxnguyen:
@David-Henner unfortunately still some failing tests.

typfel:
Have all the issues been resolved? @billypchan @johnxnguyen ?

PR #4434, opened 2020-06-24
do setup sctipts steps in parallel

billypchan:
## What's new in this PR?

Improve the speed of running `setup.sh` by running unrelated steps in parallel. Tested the running time can be reduced by 40%.


billypchan:
@johnxnguyen yes, since such external script / command do not depend on each other, running them in parallel has the same result.

PR #4439, opened 2020-06-25
Fix: Image viewer is darken for every tap

billypchan:
### Issues

The image viewer is darken for every tap. Menu is not shown when long press on the image.

### Causes

Multiple dark layer is added to the view. 
The `UIMenu` should appear after the image is darkened, but it is not shown. ( seems like an iOS 13 issue)

### Solutions

~~Do not darken the image for iOS13. I will open another PR to solve the menu not shown issue.~~
Remove the darken layer after it is faded out. Update the `UIMenu` deprecated code.


Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.swift
```diff
@@ -494,6 +493,58 @@ final class FullscreenImageViewController: UIViewController {
...
+    func setSelectedByMenu(_ selected: Bool, animated: Bool) {
+        
+        zmLog.debug("Setting selected: \(selected) animated: \(animated)")
+        if selected {
+            // do not add more then one layer
+            if let highlightLayer = highlightLayer {
+                guard imageView?.layer.sublayers?.contains(highlightLayer) == false else {
+                    return
+                }
+            }
```
johnxnguyen, line 507: I suggest to extract this to a private method.

Wire-iOS/Sources/UserInterface/Components/Controllers/FullscreenImageViewController.swift
```diff
@@ -510,22 +561,41 @@ final class FullscreenImageViewController: UIViewController {
...
         let menuController = UIMenuController.shared
         menuController.menuItems = ConversationMessageActionController.allMessageActions
-
-        if let imageView = imageView {
-            menuController.setTargetRect(imageView.bounds, in: imageView)
+        
+        if #available(iOS 13.0, *) {
+            view.window?.makeKey()
+            view.window?.becomeFirstResponder()
+            becomeFirstResponder()
```
johnxnguyen, line 568: Could you elaborate what this does and why we need it?

billypchan: clean up and added the comment to helper method.

Wire-iOS/Sources/UserInterface/Helpers/UIView+Menu.swift
```diff
@@ -0,0 +1,40 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+
+extension UIView {
+        
+    /// The reason why we are touching the window here is to workaround a bug where,
+    /// We now force the window to be the key window and to be the first responder to ensure that we can
```
johnxnguyen, line 25: So what is the bug?

billypchan:
@johnxnguyen the solution is copying a hack form `CollectionCell`. Related discussion: https://stackoverflow.com/questions/59176844/uimenucontroller-is-not-visible-in-ios-13-2/62578001#62578001

Wire-iOS/Sources/UserInterface/Helpers/UIView+Menu.swift
johnxnguyen: Please add the link to the SO post to add some context.

PR #4445, opened 2020-06-30
[Feature] Digital Signature Verification Error Improvements

marcomaddy1982:


Wire-iOS/Sources/UserInterface/Components/Controllers/DigitalSignatureVerificationViewController.swift
```diff
@@ -111,6 +111,7 @@ extension DigitalSignatureVerificationViewController: WKNavigationDelegate {
     }
     
     func parseVerificationURL(_ url: URL) -> VoidResult? {
+        print("VerificationURL: \(url.absoluteString)")
```
KaterinaWire, line 114: Do we need to keep this `print`?

PR #4448, opened 2020-07-02
[Feature] Sort call participants by name

johnxnguyen:
## What's new in this PR?

Participants in the call participant are now sorted by name alphabetically. Previously they were sorted by the join time.


Wire-iOS Tests/CallParticipantsViewTests.swift
```diff
@@ -34,13 +34,11 @@ class CallParticipantsViewTests: ZMSnapshotTestCase {
...
-        for index in 0..<participantCount {
-            rows.append(.callParticipant(user: MockUser.mockUsers()[index], sendsVideo: sendsVideo))
-        }
-        
-        return rows
+        return (0..<participantCount)
+            .lazy
+            .map { MockUser.mockUsers()[$0] }
+            .sorted { $0.name < $1.name }
+            .map { CallParticipantsCellConfiguration.callParticipant(user: $0, sendsVideo: sendsVideo) }
```
johnxnguyen, line 41: By sorting the test participants, we assert the participants are sorted when equating the call info configurations.

PR #4451, opened 2020-07-06
Fix: personal account user should not be able to create a invitation link

billypchan:
### Issues

The "invite people" button is shown when a personal account user created a new conversation. Click the button and he/she can see the guest option screen which is not permitted.

### Causes

`GuestsAllowedCellDescription` is created without checking the current user is a team member or not.

### Solutions

Do not create `GuestsAllowedCellDescription` if the user is not a team member.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
@@ -354,8 +354,8 @@ final class ConversationSystemMessageCellDescription {
             
             /// only display invite user cell for team members
             if SelfUser.current.isTeamMember,
-               selfCanAddUsers(conversation: conversation),
-               isOpenGroup(conversation: conversation) {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -368,6 +368,14 @@ final class ConversationSystemMessageCellDescription {
 
         return []
     }
+    
+    private static func isOpenGroup(conversation: ZMConversation) -> Bool {
+        return conversation.conversationType == .group && conversation.allowGuests
+    }
+
+    private static func selfCanAddUsers(conversation: ZMConversation) -> Bool {
+        return ZMUser.selfUser()?.canAddUser(to: conversation) ?? false
```
typfel, line 377: Can we use `SelfUser.current` here?

billypchan: fixed

PR #4454, opened 2020-07-08
Feature: app lock option in setting screen

billypchan:
## What's new in this PR?

Create app lock option in setting screen

### TODO
- [ ] app lock option should be persisted in keychain?
- [ ] text copy
- [x] read feature flag and display the option for forceCustomAppLock value.


typfel:
Why add a new settings entry for the applock when there’s already an existing one?

billypchan:
@typfel I am thinking "lock with system passcode" and "lock with custom code" should be separated or merged into one option?

typfel:
Let’s clarify that in the feature chat. I was under the impression that we always force a custom passphrase now.

WireCommonComponents/CustomAppLock.swift
```diff
@@ -0,0 +1,59 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireDataModel
+
+public struct CustomAppLockRules: Decodable { ///TODO: merge with AppLockRules?
+    public let forceCustomAppLock: Bool
```
typfel, line 23: We should parse the `session_manager.json` in `SessionManagerConfiguration` and we already have a `forceAppLock` entry which should be enough? Since we are modifying the behaviour of the existing app lock feature and not new configuration.

typfel: Btw I saw now that this code was already like this before.

PR #4455, opened 2020-07-08
[Feature] video tiles media icons

David-Henner:
## What's new in this PR?

- [Add microphone icon to video tiles](https://wearezeta.atlassian.net/browse/ZIOS-13553)
- [Add name label to video tile](https://wearezeta.atlassian.net/browse/ZIOS-13521)
- Small refactoring of the Video Grid implementation


Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -21,60 +21,75 @@ import WireSyncEngine
...
         videoStreams = voiceChannel.videoStreamArrangment.grid
-        isMuted = mediaManager.isMicrophoneMuted && !isOverlayVisible
         networkQuality = voiceChannel.networkQuality
     }
 }
 
 extension VoiceChannel {
     
     private var selfStream: VideoStream? {
+        guard
```
johnxnguyen, line 36: Looks like this isn't used any more, you may be able to delete it.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -23,9 +23,20 @@ import WireDataModel
 import WireSyncEngine
 import avs
 
+struct StreamIdentifier: Equatable {
+    let userId: UUID
+    let clientId: String
+}
+
```
johnxnguyen, line 30: In Sync Engine there is an `AVSClient` struct exactly the same as this and for the same purpose.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridView.swift
```diff
@@ -23,9 +23,20 @@ import WireDataModel
...
 struct Stream: Equatable {
     let userId: UUID
     let clientId: String
```
johnxnguyen, line 33: I suggest to use the bundle the both ids together with `AVSClient` to make their meaning clearer. This would also make the computed property for `streamId` redundant.

David-Henner: done here b65072d

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
David-Henner: it's used below in `videoStreamArrangment`

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoParticipantDetailView.swift
```diff
@@ -0,0 +1,83 @@
...
+        
+        [blurView, microphoneIconView, nameLabel].forEach {
+            $0.translatesAutoresizingMaskIntoConstraints = false
+            $0.clipsToBounds = true
+            addSubview($0)
+        }
+    }
+    
+    func createConstraints() {
+        blurView.fitInSuperview()
```
billypchan, line 72: I would recommend create 4 constraints and activate them in one batch below.

PR #4457, opened 2020-07-08
[Fix]: Correct Error Message For Blocked Domain During Registration

marcomaddy1982:
## What's new in this PR?

When the server was responding that the email during the registration had a blocked domain a generic error message was shown, instead of showing an ad hoc one.

### Dependencies

sync-engine: https://github.com/wireapp/wire-ios-sync-engine/pull/1261


Wire-iOS/Sources/UserInterface/Helpers/UIViewController+ErrorAlert.swift
```diff
@@ -54,6 +54,8 @@ extension UIViewController {
                 message = ""
             case .needsCredentials:
                 message = "error.user.needs_credentials".localized
+            case .domainBlocked:
+                message = code.errorDescription ?? "error.user.unkown_error".localized
```
typfel, line 58: So this is the only error where the localization string comes from the error?

marcomaddy1982: probably I should create a localisable.

typfel: But also we don't to have two version of the localization string.

typfel: There were many errors localized in SE just wondering how they are handled?

Could it be that they are covered by:

```
  } else {
            message = error.localizedDescription
  }
```

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1540,6 +1540,7 @@
 // ZMUserSessionErrorCode mapping. Used during registration and profile edit
 "error.user.unkown_error" = "Something went wrong, please try again";
 "error.user.needs_credentials" = "Please verify your details and try again.";
+"error.user.domain_blocked" = "This email requires a team invitation. Contact your organization's IT administrators to receive an invite.";
```
typfel, line 1543: Now we have this string defined twice. 

I'm fairly sure you don't need add a case to switch statement since the error implements `LocalizedError`.

marcomaddy1982: Actually now it is aligned with all the other cases. I the SE there is the class 'ZMUserSessionErrorCode+Localized' but actually it seems that there is never used.

typfel: I think they are used implicitly through https://github.com/wireapp/wire-ios/blob/0a1dda87f209ead591c27981b39add64b79a2ca0/Wire-iOS/Sources/UserInterface/Helpers/UIViewController%2BErrorAlert.swift#L93

You'll see that the errors which are localised in SE are not included in the switch statement.

marcomaddy1982: > I think they are used implicitly through
> 
> https://github.com/wireapp/wire-ios/blob/0a1dda87f209ead591c27981b39add64b79a2ca0/Wire-iOS/Sources/UserInterface/Helpers/UIViewController%2BErrorAlert.swift#L93

yes yes this I noticed too.

> 
> You'll see that the errors which are localised in SE are not included in the switch statement.

So what should we do?

marcomaddy1982: I would stay aligned to the other cases and merge now s it is and the in a second place try to refactor to have just one localisation and not two in the UI and SE.

typfel: Couldn't you just bump SE no other changes are necessary in the UI project.

marcomaddy1982: We can if we do 'message = code.errorDescription ?? "error.user.unkown_error".localized' but then we are not aligned with the other case. As I wrote previously I would leave as it is now and then refactor all the messages in a second moment.

typfel: Yep, I see that now.

PR #4460, opened 2020-07-09
[Fix] user cell accessibility identifiers

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13610

### Issues

Accessibility identifiers for `guest` and `external user` have been replaced by a single identifier `"img.userType"`, making the icons indistinguishable.

### Causes

`guest` and `external user` icons were two separate instances of `UIImageView` but have been replaced by a single instance of `IconImageView` causing it to have a single accessibility id 

### Solutions

add an `accessibilityIdentifier` property to `IconImageStyle` protocol and associate a different id per icon style

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
@@ -55,6 +55,8 @@ class IconImageView: UIImageView {
         // save size and color if needed
         set(size: size, color: color)
         
+        accessibilityIdentifier = style?.accesibilityIdentifier
+
         guard
             let style = style ?? self.style,
             let icon = style.icon
@@ -66,7 +68,6 @@ class IconImageView: UIImageView {
         isHidden = false
         let color = style.tintColor ?? self.color
         self.setIcon(icon, size: self.size, color: color)
-        self.setUpIconImageView(accessibilityIdentifier: style.accesibilityIdentifier)
         self.style = style
     }
     
--- a/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
@@ -22,7 +22,7 @@ import WireCommonComponents
 protocol IconImageStyle {
     var icon: StyleKitIcon? { get }
     var tintColor: UIColor? { get }
-    var accesibilityIdentifier: String { get }
+    var accessibilityIdentifier: String { get }
     var accessibilityPrefix: String { get }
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
```diff
@@ -22,9 +22,20 @@ import WireCommonComponents
 protocol IconImageStyle {
     var icon: StyleKitIcon? { get }
     var tintColor: UIColor? { get }
+    var accesibilityIdentifier: String? { get }
+    var accessibilityPrefix: String { get }
+    var rawValue: String { get }
```
David-Henner, line 27: chose `rawValue` property because we assume that `IconImageStyle` will be implemented by an `enum` with a `rawValue` of type `String`

We could rename the property to something like `accessibilitySuffix` and have the enums implement it to return the rawValue.

That would make the code clearer and more readable at the cost of some redundancy in the implementation of the protocol.

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
```diff
@@ -54,6 +65,8 @@ class IconImageView: UIImageView {
         // save size and color if needed
         set(size: size, color: color)
         
+        accessibilityIdentifier = style?.accesibilityIdentifier
```
billypchan, line 68: I would suggest overriding. if style is modified, access id is also update:
```
    override var accessibilityIdentifier: String? {
        get {
            return style?.accesibilityIdentifier
        }

        set {
            // no-op
        }
    }
```

johnxnguyen: This is misleading though because it is overriding a read-write property to behave only like a read-only property.

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
johnxnguyen: I think that clarity and readability should be the priority and warrants some redundancy. The benefits of good descriptive names far outweigh a little bit of extra code.

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/IconImageView.swift
billypchan: We have several similar fixes before about `accessibilityIdentifier` is not updated after the context changes. Another choice would be updating `accesibilityIdentifier` in `style`'s `didSet`, but it would be easy to be forgotten in the future.

PR #4461, opened 2020-07-09
[WIP] Feature: store passphrase in the key chain

billypchan:
## What's new in this PR?

Create a temporary request password alert and store the passcode in keychain, when the use enable app lock switch.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
@@ -54,6 +54,8 @@ enum SettingsPropertyError: Error {
 protocol SettingsPropertyFactoryDelegate: class {
     func asyncMethodDidStart(_ settingsPropertyFactory: SettingsPropertyFactory)
     func asyncMethodDidComplete(_ settingsPropertyFactory: SettingsPropertyFactory)
+    
+    func requestNewPassCode()
 }
 
 final class SettingsPropertyFactory {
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -325,12 +327,20 @@ final class SettingsPropertyFactory {
...
+                    return SettingsPropertyValue(CustomAppLock.isActive)
             },
                 setAction: { _, value in
                     switch value {
                     case .number(value: let customAppLock):
-                        CustomAppLock.isActive = customAppLock.boolValue ///TODO:
+                        let isCustomAppLockActive = customAppLock.boolValue
+                        CustomAppLock.isActive = isCustomAppLockActive
+                        
+                        if isCustomAppLockActive {
```
typfel, line 338: AFAIK this should not be configureable and we will always set a pass code.

Wire-iOS/Sources/UserInterface/SelfProfile/SelfProfileViewController.swift
```diff
@@ -185,6 +187,20 @@ final class SelfProfileViewController: UIViewController {
 // MARK: - SettingsPropertyFactoryDelegate
 
 extension SelfProfileViewController: SettingsPropertyFactoryDelegate {
+    func requestNewPassCode() {
+        // ask for new custom pass code
+        let requestPasswordController = RequestPasswordController(context: .newPasscode) { passcode in
+            guard let passcode = passcode else { return }
+            // save to keychain
+            let data = passcode.data(using: .utf8)!
+            ZMKeychain.setData(data, forAccount: SettingsPropertyName.customAppLock.rawValue)
```
typfel, line 196: I'm thinking this is abusing the `ZMKeychain` since app lock is not an "account". I think we should add a new way for setting/deleting passcode from the keychain but I would suggest doing that in a  separate PR. 

I'd suggest adding a TODO for now.

billypchan: I found there is only one place in the UI project using `ZMKeychain.setData()` and the `account` value is a static string (`SettingsPropertyName.lockApp.rawValue`). I guess the 'Account' here means key as there are no other public methods to save the data to `ZMKeychain`.

PR #4255, opened 2020-04-07
Remove objective c bridging header

billypchan:
## What's new in this PR?

Remove objective c bridging header.
Since frameworks is no longer imported via header, adding necessary `import` in each swift files.

PR #4469, opened 2020-07-15
Fix: app crash when entered a number in country selection screen

billypchan:
### Issues

In the country selection screen, app crashes after entered a number in the search field.

### Causes

`@objc` signature of `Country` class is removed in https://github.com/wireapp/wire-ios/pull/4255

### Solutions

Add `@objc` to member `e164`.


Wire-iOS/Sources/Authentication/Helpers/CountryCode/CountryCodeTableViewController.swift
```diff
@@ -140,11 +140,11 @@ extension CountryCodeTableViewController: UISearchBarDelegate {
 
 // MARK: - UISearchResultsUpdating
 
+///TODO: test
```
David-Henner, line 143:
```suggestion
```

PR #4472, opened 2020-07-16
[Fix] Don't display floating video tile in group calls

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13483

- Display floating video tile only in one to one calls (not in group conv/calls)
- Fixed issues with floating video (see below)

### Issues

1. Floating video never shows over the grid

2. Floating video freezes after first grid update

### Causes

1. Floating video relied on `selfPreviewView` property, which was never set

2. Floating video view was removed from cache on `updateState()`

### Solutions

1. Make `selfPreviewView` a computed property returning cached self preview

2. Remove views from cache based on all streams from video configuration (higher source of truth)

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
@@ -80,7 +80,7 @@ extension VoiceChannel {
             return (nil, streamsExcludingSelf)
         }
 
-        if 1 == streamsExcludingSelf.count {
+        if conversation?.conversationType == ZMConversationType.oneOnOne {
             return (selfStream, streamsExcludingSelf)
         } else {
             return (nil, [selfStream] + streamsExcludingSelf)
--- a/Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
@@ -80,7 +80,7 @@ extension VoiceChannel {
             return (nil, streamsExcludingSelf)
         }
 
-        if conversation?.conversationType == ZMConversationType.oneOnOne {
+        if conversation?.conversationType == ZMConversationType.oneOnOne && streamsExcludingSelf.count == 1 {
             return (selfStream, streamsExcludingSelf)
         } else {
             return (nil, [selfStream] + streamsExcludingSelf)
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
@@ -27,12 +27,20 @@ protocol VideoGridConfiguration {
 
 }
 
-// Workaround to make the protocol equatable, it might be possible to conform VideoGridConfiguration
-// to Equatable with Swift 4.1 and conditional conformances. Right now we would have to make
-// the `VideoGridViewController` generic to work around the `Self` requirement of
-// `Equatable` which we want to avoid.
+
 extension VideoGridConfiguration {
 
+    var allStreams: [VideoStream] {
+        guard let stream = floatingVideoStream else {
+            return videoStreams
+        }
+        return videoStreams + [stream]
+    }
+    
+    // Workaround to make the protocol equatable, it might be possible to conform VideoGridConfiguration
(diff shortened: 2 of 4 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -80,7 +80,7 @@ extension VoiceChannel {
             return (nil, streamsExcludingSelf)
         }
 
-        if 1 == streamsExcludingSelf.count {
+        if conversation?.conversationType == ZMConversationType.oneOnOne {
```
johnxnguyen, line 83: Does this mean that if only the self stream exists and it's a one to one conversation, then the self stream is a little thumbnail instead of fullscreen?

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
```diff
@@ -27,12 +27,20 @@ protocol VideoGridConfiguration {
...
-// `Equatable` which we want to avoid.
+
 extension VideoGridConfiguration {
 
+    var allStreams: [VideoStream] {
+        guard let stream = floatingVideoStream else {
+            return videoStreams
+        }
+        return videoStreams + [stream]
+    }
```
johnxnguyen, line 38: I would suggest to make this `var allStreamIds: Set<AVSClient> { ... }`, since it becomes a bit clearer how this property is used.

David-Henner:
> My apologies for causing this bug, hope it didn't take too long to fix!

@johnxnguyen no worries, I should've noticed in the PR :)

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
johnxnguyen: I mean to say rather, that it would help to avoid confusion in the future. Having more than one `[VideoStream]` property is a bit confusing.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
David-Henner: it means that if ~~only~~ the self stream exists and it's a one to one conversation, then the self stream is a thumbnail instead of fullscreen

David-Henner: I think the naming can be confusing here, because `preview` actually means that it will be the `floatingVideoStream`

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
David-Henner: Would using a `Set` be to avoid duplicates?

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
johnxnguyen: The old logic states:

1. If the is no self stream, then display all other streams in a grid.
2. If there is a self stream and one other stream, then display the self stream in the preview and the other person in the grid (fullscreen).
3. If there is a self stream and zero or more than one other streams, then display all streams in the grid.

So according to point 3, if there is only the self stream, it will be displayed in the grid fullscreen.

Contrasting this to the new logic:

1. Same as before
2. If there is a self stream and it's a one to one conversation, display the thumbnail and potentially the other video stream.
3. If there is a self stream and it's a group, display all streams in the grid.

The difference is in 2. The thumbnail is always shown if it's a one to one and there is a self stream, regardless of if there is another stream.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridConfiguration.swift
johnxnguyen: Yes to avoid duplicates. It's not strictly necessary but given that the ids should be unique, I'd say the `Set` is more expressive.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
David-Henner: You're right, good catch

PR #4473, opened 2020-07-16
Add developer setting to send many messages at once, for debugging purposes

marcoconti83:
## What's new in this PR?

I am investigating possible causes for decryption errors when exchanging messages. I added this setting to automate sending messages back and forth between two devices to see if I can reproduce the problem. I am committing this because I believe this to be useful debugging tool that might come in handy at some other time.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
@@ -252,6 +252,14 @@ class SettingsCellDescriptorFactory {
             }
         }
         developerCellDescriptors.append(appendManyMessages)
+        
+        let spamWithMessages = SettingsButtonCellDescriptor(title: "Spam the top conv", isDestructive: true) { _ in
+            
+            self.requestNumber() { count in
+                self.spamWithMessages(amount: count)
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -549,6 +557,31 @@ class SettingsCellDescriptorFactory {
         }
     }
     
+    /// Sends a message that will fail to decode on every other device, on the first conversation of the list
```
johnxnguyen, line 560: Why does it fail to decode on every other device?

marcoconti83: Ops, copy-paste error. This comment is not relevant here

marcoconti83, line 560:
```suggestion
    /// Sends a number of messages to the top conversation in the list, in an asynchronous fashion
```

marcoconti83: Fixed!

PR #4475, opened 2020-07-20
[Feature] Conference calling beta toggle

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13556

- Add Setting property `enableConferenceCallingBeta`
- Add conference calling section to Advanced settings group
- Move Advanced group setup to separate extension & perform small refactor
- Record snapshot


johnxnguyen:
Did you manually test that the switch works?

David-Henner:
> Did you manually test that the switch works?

The switch updates the `useConferenceCalling` property on the call center.
I checked that the right conversation type is used when starting a call (`group` or `conference`)

However, the participant limit stays the same as in the original configuration.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Options.swift
```diff
@@ -19,194 +19,247 @@
...
+        
+        return SettingsSectionDescriptor(
+            cellDescriptors: [VBRToggle],
+            header: .none,
+            footer: "self.settings.vbr.description".localized,
+            visibilityAction: .none
+        )
+    }
+    
+    private var conferenceCallingSection: SettingsSectionDescriptor {
```
David-Henner, line 146: highlighting the new conference calling section

johnxnguyen:
> However, the participant limit stays the same as in the original configuration.

We'll also need to change the participant limit. Does it make sense to supply a new call center config when flipping the switch?

David-Henner:
> > However, the participant limit stays the same as in the original configuration.
> 
> We'll also need to change the participant limit. Does it make sense to supply a new call center config when flipping the switch?

We can change the interface on the `SessionManager` to take a new configuration instance instead of only the `useConferenceCalling` property. The question then is where should the config come from ?

johnxnguyen:
I'm thinking we can get rid of the whole call center config now because so far it has was needed to specify different behavior amongst different builds (mostly drawing a distinction between public and internal). But the new toggle makes this redundant, since the toggle will be released publicly. 

Also, while there is a hard limit of 4 video participants in legacy group calls, there will only be a soft limit in conference calls, meaning that we don't limit the group size but rather the amount of video tiles displayed. This means that we won't need to pass this limit all the way to SE but can just enforce it on the UI side. 

Perhaps we can just use the toggle value when on the UI side to enforce the legacy limit, and pass it to as we do now (and with the CBR setting). Eventually when we remove the toggle, we can hardcode the soft limit (or make it configurable in another way).

David-Henner:
> I'm thinking we can get rid of the whole call center config now because so far it has was needed to specify different behavior amongst different builds (mostly drawing a distinction between public and internal). But the new toggle makes this redundant, since the toggle will be released publicly.
>

Let's then make a separate task for getting rid of the configuration.
Until the configuration is removed, SE will force the call to be audio only if the limit is reached. So we'll need the limit in the configuration to support conference calling, as it is now.

> Also, while there is a hard limit of 4 video participants in legacy group calls, there will only be a soft limit in conference calls, meaning that we don't limit the group size but rather the amount of video tiles displayed. This means that we won't need to pass this limit all the way to SE but can just enforce it on the UI side.
>
> Perhaps we can just use the toggle value when on the UI side to enforce the legacy limit, and pass it to as we do now (and with the CBR setting). Eventually when we remove the toggle, we can hardcode the soft limit (or make it configurable in another way).

In any case, the soft limit on the UI side will be hardcoded unless we make it configurable.

PR #4477, opened 2020-07-20
[Feature] Unlock database with app lock screen

typfel:
## What's new in this PR?

When the database is locked we need to unlock it using biometrics or the passcode and we need to disable any fallbacks which requests that the user to enter his/her password since we can't use it to unlock the database.

* Introduce `AuthenticationScenario` which encapsulates the rules which apply when are unlocking the app.
* Add `isDatabaseLocked `parameter to the `.authenticated` `AppState` to signal that the database is locked.

### DIFF
--- a/WireCommonComponents/AppLock.swift
+++ b/WireCommonComponents/AppLock.swift
@@ -68,11 +68,7 @@ public class AppLock {
             case .screenLock(requireBiometrics: let requireBiometrics, grantAccessIfNoPasscodeIsSet: _):
                 return requireBiometrics ? .deviceOwnerAuthenticationWithBiometrics : .deviceOwnerAuthentication
             case .databaseLock:
-                #if targetEnvironment(simulator)
-                    return .deviceOwnerAuthentication
-                #else
-                    return .deviceOwnerAuthenticationWithBiometrics
-                #endif
+                return .deviceOwnerAuthentication
                 
             }
         }
--- a/WireCommonComponents/AppLock.swift
+++ b/WireCommonComponents/AppLock.swift
@@ -58,41 +58,77 @@ public class AppLock {
         /// Biometrics failed and account password is needed instead of device PIN
         case needAccountPassword
     }
+    
+    public enum AuthenticationScenario {
+        case screenLock(requireBiometrics: Bool, grantAccessIfNoPasscodeIsSet: Bool)
+        case databaseLock
+        
+        var policy: LAPolicy {
+            switch self {
+            case .screenLock(requireBiometrics: let requireBiometrics, grantAccessIfNoPasscodeIsSet: _):
+                return requireBiometrics ? .deviceOwnerAuthenticationWithBiometrics : .deviceOwnerAuthentication
+            case .databaseLock:
+                #if targetEnvironment(simulator)
+                    return .deviceOwnerAuthentication
+                #else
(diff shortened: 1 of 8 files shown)


WireCommonComponents/AppLock.swift
```diff
@@ -58,41 +58,73 @@ public class AppLock {
...
+            switch self {
+            case .screenLock(requireBiometrics: let requireBiometrics, grantAccessIfNoPasscodeIsSet: _):
+                return requireBiometrics ? .deviceOwnerAuthenticationWithBiometrics : .deviceOwnerAuthentication
+            case .databaseLock:
+                return .deviceOwnerAuthentication
+                
+            }
+        }
+        
+        var supportsUserFallback: Bool {
```
David-Henner, line 76: nice

WireCommonComponents/AppLock.swift
```diff
@@ -58,41 +58,73 @@ public class AppLock {
...
         AppLock.weakLAContext = context
         
-        let policy: LAPolicy = useBiometricsOrAccountPassword ? LAPolicy.deviceOwnerAuthenticationWithBiometrics : LAPolicy.deviceOwnerAuthentication
-        let canEvaluatePolicy = context.canEvaluatePolicy(policy, error: &error)
-        
-        if useBiometricsOrAccountPassword && (BiometricsState.biometricsChanged(in: context) || !canEvaluatePolicy) {
-            callback(.needAccountPassword)
+        let canEvaluatePolicy = context.canEvaluatePolicy(scenario.policy, error: &error)
+                
+        if scenario.supportsUserFallback && (BiometricsState.biometricsChanged(in: context)) || !canEvaluatePolicy {
```
David-Henner, line 110:
```suggestion
        if scenario.supportsUserFallback && (BiometricsState.biometricsChanged(in: context) || !canEvaluatePolicy) {
```

WireCommonComponents/AppLock.swift
```diff
@@ -58,41 +58,73 @@ public class AppLock {
...
-                if useBiometricsOrAccountPassword, let laError = error as? LAError, laError.code == .userFallback {
+                if scenario.supportsUserFallback, let laError = error as? LAError, laError.code == .userFallback {
                     authResult = .needAccountPassword
                 }
                 
-                callback(authResult)
+                callback(authResult, context)
             })
         } else {
             // If there's no passcode set automatically grant access unless app lock is a requirement to run the app
```
David-Henner, line 126: This comment doesn't reflect well the cases covered in the `else` statement and it'd be clearer to rename `grantAccessIfNoPasscodeIsSet` to something like `grantAccessIfPolicyCannotBeEvaluated`

Because we might use one of those two policies:
- `deviceOwnerAuthentication`:  can be evaluated if biometrics **or** passcode are enabled
- `deviceOwnerAuthenticationWithBiometrics`: can be evaluated **only** if biometrics are enabled

This means that `canEvaluatedPolicy` can be `false` not only if passcode isn't enabled, but also if biometrics aren't enabled.

David-Henner: Or 
```
var resultForUnevaluablePolicy: AuthenticationResult
```

typfel: Yep, I agree that more accurate and less confusing.

PR #4478, opened 2020-07-21
Feature: custom app lock unlock screen

billypchan:
## What's new in this PR?

Unlock screen for custom app lock feature.

### TODO
- [ ] update icon after updated icon file
- [x] update red color
- [x] clean up style constant to new file
- [ ] finish viewModel after passcode storage is merged


Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/UnlockViewController.swift
```diff
@@ -0,0 +1,241 @@
...
+
+        /// unlock with passcode
+        /// - Returns: true if succeed
+        func unlock(passcode: String) -> Bool {
+            //TODO: logic
+            return false
+        }
+    }
+
+    private let viewModel: UnlockViewModel = {
```
typfel, line 35: This AppLock feature is currently implemented using the VIPER pattern so I think we should continue to use that pattern for the `UnlockViewController` as well. See `AppLockInteractor`.

billypchan: I will keep it simple with MVC first, then split it to VIPER later.

billypchan: Added VIPER skeleton

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/UnlockViewController.swift
```diff
@@ -0,0 +1,247 @@
...
+        button.isEnabled = false
+
+        button.addTarget(self, action: #selector(onUnlockButtonPressed(sender:)), for: .touchUpInside)
+
+        return button
+    }()
+
+    private let revealIcon: StyleKitIcon = .cross
+    lazy var accessoryTextField: AccessoryTextField = {
+        let textField = AccessoryTextField(kind: .passcode,
```
typfel, line 58: Are we validating the input when entering the passcode? It's usually best **not** to validate the input when entering a password because the rules can change from the time when the user created the password and when he needs to enter it.

tl:dr: we should only validate the password when it's created.

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/UnlockViewController.swift
```diff
@@ -0,0 +1,247 @@
...
+import WireCommonComponents
+
+protocol UnlockUserInterface: class {
+}
+
+extension UnlockViewController: UnlockUserInterface {
+    
+}
+
+// This VC should be wrapped in KeyboardAvoidingViewController as the "unlock" button would be covered on 4 inch iPhone
```
typfel, line 29:
```suggestion
/// UnlockViewController
/// 
/// This VC should be wrapped in KeyboardAvoidingViewController as the "unlock" button would be covered on 4 inch iPhone
```

Let's make it a documentation comment so it's easy to Discover this hint.

typfel:
NOTE: I've only reviewed the view part since it's currently not integrated with the existing app lock implementation.

billypchan:
@typfel The logics are in other draft PRs. I will update them after the dummy UI in those PRs are replaced by this and other PRs.

typfel:
> @typfel The logics are in other draft PRs. I will update them after the dummy UI in those PRs are replaced by this and other PRs.

So will you merge this PR first and then update the other PRs and then can we can review them as well?

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/UnlockViewController.swift
billypchan: Here I reuse the logic for login with a password. It returns validation error `.tooShort(.passcode)` when the passcode is empty(And the unlock button is disabled).

typfel: 👍 So we only validate that the user at least something.

PR #4483, opened 2020-07-23
[Fix] Video Grid UI issues

David-Henner:
### Issues

- Wrong microphone icon color in video tiles
- Wrong media icons ordering in participants list
- Non-respect of bottom safe area in video tiles user info overlay

### Causes

- Icon color should be `white`
- Media icons were order right-to-left instead of left-to-right
- Use of `bottomAnchor` instead of `safeBottomAnchor` in constraints

### Solutions

- Set microphone icon color to `white`
- Order icons in left-to-right fashion
- Use `safeBottomAnchor` in constraints


Wire-iOS/Sources/UserInterface/Components/Views/UserCell.swift
```diff
@@ -135,7 +135,7 @@ class UserCell: SeparatorCollectionViewCell, SectionListCellType {
         avatarSpacer.addSubview(avatar)
         avatarSpacer.translatesAutoresizingMaskIntoConstraints = false
         
-        iconStackView = UIStackView(arrangedSubviews: [verifiedIconView, userTypeIconView, microphoneIconView, videoIconView, connectButton, checkmarkIconView, accessoryIconView])
+        iconStackView = UIStackView(arrangedSubviews: [videoIconView, microphoneIconView, userTypeIconView, verifiedIconView, connectButton, checkmarkIconView, accessoryIconView])
```
johnxnguyen, line 138: Just a side note: it would be nice to separate the calling related icons to a subclass of `UserCell`. Is it just these icons that would need to be extracted or is there more logic?

David-Henner: only these icons would need to be extracted. I think only microphone and video icons are used only in calling UI and the rest is used in other use cases

PR #4486, opened 2020-07-27
[Feature] limit number of video participants

David-Henner:
## What's new in this PR?

- Limit number of video participants based on the configuration of conference calling


Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -92,6 +94,20 @@ extension VoiceChannel {
         }
     }
 
+    var limitedSortedActiveVideoStreams: [VideoStream] {
+        var videoStreams = sortedActiveVideoStreams
+        
+        // Place self stream first to avoid cutting it off
+        if let selfStreamIndex = videoStreams.firstIndex(where: { $0.stream.streamId == selfStreamId }) {
```
johnxnguyen, line 101: If you change the sorting of `sortedParticipants` to:

```swift
private var sortedParticipants: [CallParticipant] {
  return participants.sorted { 
    $0.user.isSelfUser || $0.user.name?.lowercased() < $1.user.name?.lowercased() 
  }
}

Then you wouldn't need to pull out the self stream here.
```

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoCallType.swift
```diff
@@ -0,0 +1,45 @@
...
+    
+    var videoStreamsLimit: Int {
+        switch self {
+        case .conferenceCalling:
+            return 12
+        case .legacyGroupCalling:
+            return 4
+        }
+    }
+}
```
johnxnguyen, line 45: Where else do we check for the participant limit?

David-Henner: Only the one time in `VideoConfiguration.swift`. I was wondering if we also need to prevent participants from enabling their video if the limit is reached

johnxnguyen: Yeah, it's in `CallInfoConfiguration.swift`. Search for `canUpgradeToVideo`.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
David-Henner: shouldn't we be comparing the user id + client id to make sure the self stream is kept and not a stream from the same user on another client ?

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -67,11 +69,11 @@ extension VoiceChannel {
     
     fileprivate var videoStreamArrangment: (preview: VideoStream?, grid: [VideoStream]) {
         guard isEstablished else { return (nil, selfStream.map { [$0] } ?? [] ) }
+
+        let videoStreams = limitedSortedActiveVideoStreams
```
johnxnguyen, line 73: Given that you adjust `sortedParticipants`, I think we can remove `limitedSortedActiveVideoStreams` altogether and replace it with:

```swift
let videoStreams = sortedActiveVideoStreams.prefix(VideoConfiguration.MaxVideoStreams)
```

where `MaxVideoStreams` is private static computed property.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -245,14 +245,19 @@ extension CallParticipantState {
 fileprivate extension VoiceChannel {
     
     var canUpgradeToVideo: Bool {
-        guard let conversation = conversation, conversation.conversationType != .oneOnOne else { return true }
-        guard conversation.localParticipants.count <= ZMConversation.maxVideoCallParticipants else { return false }
-
-        if ZMConversation.callCenterConfiguration.useConferenceCalling {
+        guard !ZMConversation.useConferenceCalling else {
+            return true
+        }
```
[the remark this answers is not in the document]
johnxnguyen: In this [PR](https://github.com/wireapp/wire-ios-sync-engine/pull/1274/files#diff-e7c09a6ae824aa9178e0a55e75591fbeR120) I've added a `isConferenceCall` property to `VoiceChannel`. You can use this to replace `ZMConversation.useConferenceCalling`.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
johnxnguyen: Yes good point, but I see you already did this :)

PR #4493, opened 2020-07-29
[Fix] Conference calling blockers

johnxnguyen:
### Issues

1. Calling is completely broken: for legacy calls, the calls never connect, for conference calls, there is no audio or video
2. If a client has conference calling enabled and is in a legacy group call with more than 4 participants, she is is able to enabled video (when she should not be able to).
### Causes

1. With AVS 6.2.11 targeted message was enabled. In the request strategy, we were incorrectly setting the `ignore_missing` parameter, so the backend was reporting missing clients (when it shouldn't) and therefore all call messages were not being sent.
2. The check to disable the video button incorrectly checks the video participant limit based on whether the conference calling public beta toggle is enabled or not. Because the toggle is enabled, the max limit is the 500, but it should be 4.

### Solutions

1. Fix the query parameter.
2. Since we're in a legacy call, simply apply the legacy group limit.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/234


Wire-iOS/Sources/Helpers/syncengine/Conversation+Participants.swift
```diff
@@ -29,7 +29,11 @@ extension ZMConversation {
     }
     
     static var maxVideoCallParticipants: Int {
-        return useConferenceCalling ? maxParticipants : 4
+        return useConferenceCalling ? maxParticipants : legacyGroupVideoParticipantLimit
+    }
+
+    static var legacyGroupVideoParticipantLimit: Int {
+        return 4
     }
```
billypchan, line 37:
```suggestion
    static let legacyGroupVideoParticipantLimit: Int = 4
```

LUGO022 (commented):
Thank you

PR #4499, opened 2020-07-30
[Feature] Add launch argument for conference calling

David-Henner:
## What's new in this PR?

- Launch argument that overrides the conference calling toggle

### DIFF
--- a/Wire-iOS/Sources/Components/Settings.swift
+++ b/Wire-iOS/Sources/Components/Settings.swift
@@ -73,8 +73,8 @@ final class Settings {
     // MARK: - subscript
     subscript<T>(index: SettingKey) -> T? {
         get {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/Components/Settings.swift
```diff
@@ -73,6 +73,10 @@ final class Settings {
     // MARK: - subscript
     subscript<T>(index: SettingKey) -> T? {
         get {
+            if case .conferenceCalling = index, let override = AutomationHelper.sharedHelper.useConferenceCalling as? T {
```
billypchan, line 76: `override` is a Swift keyword and I would suggest rename to `overrideSetting`

David-Henner: fixed

PR #4500, opened 2020-07-30
Feature: store applock passcode in keychain

billypchan:
## What's new in this PR?

Store custom passcode in keychain with `EncryptionKeys` helper methods

## dependency
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1008
- [x] https://github.com/wireapp/wire-ios-utilities/pull/94


Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeSetupInteractor.swift
```diff
@@ -35,8 +38,17 @@ final class PasscodeSetupInteractor {
...
+enum KeychainAccount: String {
+    case customAppLock = "customAppLock"
+}
+
 // MARK: - Interface
 extension PasscodeSetupInteractor: PasscodeSetupInteractorInput {
+    func storePasscode(passcode: String) { ///TODO: return
+        let data = passcode.data(using: .utf8)!
+        //TODO: error handle with setData return false
+        ZMKeychain.setData(data, forAccount: KeychainAccount.customAppLock.rawValue)
```
johnxnguyen, line 50: Do we call `ZMKeychain.setData(:forAccount:)` elsewhere in the codebase? If yes, can we unify how we specify the account? Perhaps adding new cases to `KeychainAccount`? We should avoid to do the same thing if different ways.

billypchan: the another case in the project is saving app lock setting:
```
            let data = (newValue ? "YES" : "NO").data(using: .utf8)!
            ZMKeychain.setData(data, forAccount: SettingsPropertyName.lockApp.rawValue)
```

The account param name is confusing and I agree we should rename it.

johnxnguyen: I guess we have to ensure the rawValue is the same though, so we don't break existing keychain data.

typfel: We are abusing the `ZMKeyChain` class here and yes we were already abusing it before when storing `SettingsPropertyName.lockApp.rawValue`.

The original  purpose of `ZMKeyChain` was to store the login cookie an account.

We should re-factor the key chain code but maybe not this this PR. This is how I solved it for encryption at rest https://github.com/wireapp/wire-ios-data-model/blob/56ed7e2fa4fb98d4d9c981b002cc27845f9821f3/Source/Utilis/EncryptionKeys.swift#L28

billypchan: update this PR with https://github.com/wireapp/wire-ios-data-model/pull/1008

billypchan: Moved keychain helper to https://github.com/wireapp/wire-ios-utilities/pull/94

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeSetupInteractor.swift
```diff
@@ -37,6 +40,11 @@ final class PasscodeSetupInteractor {
 
 // MARK: - Interface
 extension PasscodeSetupInteractor: PasscodeSetupInteractorInput {
+    func storePasscode(passcode: String) {
+        guard let data = passcode.data(using: .utf8) else { return }
+
+        try? Keychain.updateItem(PasscodeKeychainItem.passcode, value: data)
```
johnxnguyen, line 46: We should probably catch and handle failures.

billypchan: added store passcode failure handling (do not switch on the option.)

PR #4501, opened 2020-07-31
Fix/replace icon

billypchan:
## What's new in this PR?

Replace placeholder icons with new style kit icons.


Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeError.swift
```diff
@@ -43,19 +45,20 @@ enum PasscodeError: CaseIterable {
...
+        
+        let attributedString: NSAttributedString = NSAttributedString(string: message) && font
+        
+        return NSAttributedString(attachment:textAttachment) + attributedString
 
-        return attributedString
+    }
+    
+    var descriptionWithInvalidIcon: NSAttributedString {
+        return description(icon: .circleCross, color: UIColor(rgb: (150, 153, 156)), font: .smallRegularFont)
```
johnxnguyen, line 58: Please name this color.

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeError.swift
```diff
@@ -43,19 +45,20 @@ enum PasscodeError: CaseIterable {
...
     }
 
-    //TODO paint code icon
     var descriptionWithPassedIcon: NSAttributedString {
-
-        let attributedString: NSAttributedString = NSAttributedString(string: "✅" + message)
-
-        return attributedString
+        return description(icon: .circleTick, color: UIColor(rgb: (2, 193, 1)), font: .smallSemiboldFont)
```
johnxnguyen, line 62: Please name this color.

Wire-iOS/Sources/UserInterface/Components/CGFloat+Const.swift
```diff
@@ -26,6 +26,13 @@ extension Float {
     }
 }
 
+extension StyleKitIcon.Size {
+    enum CreatePasscode {
+        static let iconSize: StyleKitIcon.Size = 11
```
typfel, line 31: Why are these custom sizes needed? Are we compensating for something?

typfel: I took a look at the Paintcode file and the new eyeSlash icon looks squished vertically. Could we maybe fix this my transforming it to match the existing eye icon?

billypchan: This is for the icon size only use in the text attachment. I changed to `.custom(11)` for readability

billypchan: I will add one more new eye icon for this screen

PR #4502, opened 2020-07-31
[Feature] Inform user when call is using variable bit rate

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13547

- Inform user of which bit rate the call is using (variable / constant) if CBR was enabled in the settings

### TODO

- [ ] Update Tests


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -157,7 +158,8 @@ struct CallInfoConfiguration: CallInfoViewControllerInput  {
         preferedVideoPlaceholderState: CallVideoPlaceholderState,
         permissions: CallPermissionsConfiguration,
         cameraType: CaptureDevice,
-        mediaManager: AVSMediaManagerInterface = AVSMediaManager.sharedInstance()
+        mediaManager: AVSMediaManagerInterface = AVSMediaManager.sharedInstance(),
+        settings: Settings = Settings.shared
```
johnxnguyen, line 162: Why not inject the value for `userEnabledCBR` here? It would save you having to mock the `Settings`.

PR #4508, opened 2020-08-05
[Fix] video button disabled message

David-Henner:
## What's new in this PR?

- Changed alert displayed when user taps video button in a call & video permissions are not granted. Now displays the correct message and actions instead of the incorrect alert ("Too many people for Video")

- Small refactor of the camera permission alert

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -27,6 +27,7 @@
 "general.edit" = "Edit";
 "general.done" = "Done";
 "general.confirm" = "OK";
+"general.later" = "Later";
 "general.skip" = "Not Now";
 "general.accept" = "Accept";
 "general.decline" = "No, thanks";
@@ -363,7 +364,7 @@
 "image.edit_image" = "Edit image";
 
 // Camera access
-"camera_access.denied" = "Wire needs access to the Camera.";
+"camera_access.denied" = "Wire needs access to the camera";
 "video_call.camera_access.denied" = "Wire doesn’t have access to the camera";
 "camera_access.denied.instruction" = "";
 "camera_access.denied.open_settings" = "Enable it in Wire Settings";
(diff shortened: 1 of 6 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -27,6 +27,7 @@
 "general.edit" = "Edit";
 "general.done" = "Done";
 "general.confirm" = "OK";
+"general.later" = "Later";
 "general.skip" = "Not Now";
```
David-Henner, line 31: added the `"general.later"` string to fit the design requirements, but `"general.skip"` could also be used

Wire-iOS/Sources/Helpers/UIAlertController+Permissions.swift
```diff
@@ -0,0 +1,59 @@
...
+extension UIAlertController {
+    class func cameraPermissionAlert(with completion: AlertActionHandler? = nil) -> UIAlertController {
+        let alert = UIAlertController(
+            title: "voice.alert.camera_warning.title".localized,
+            message: "NSCameraUsageDescription".infoPlistLocalized,
+            preferredStyle: .alert
+        )
+        
+        alert.addAction(.actionLater(with: completion))
+        alert.addAction(.actionSettings(with: completion))
```
johnxnguyen, line 32: What's the intended use of `completion` if it's passed to both actions?

David-Henner: good question! I assumed it was to execute an action after the alert controller was dismissed, but it turns out it's actually not needed at all

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -674,9 +675,6 @@
 
 "call.video.paused" = "Video paused";
 
-"call.video.too_many.alert.title" = "Too many people for Video";
-"call.video.too_many.alert.message" = "Video calls only work in groups of 4 or less.";
-
```
johnxnguyen, line 679: We can only remove this copy when we drop support for legacy group calls. Are you planning to park this PR until we can merge?

Wire-iOS/Sources/Helpers/UIAlertController+Permissions.swift
David-Henner: To be executed once the alert is dismissed regardless of the action tapped

Wire-iOS/Resources/Base.lproj/Localizable.strings
David-Henner: Even if I don't remove the copy, it wouldn't be used anywhere, because the new logic for the alert is to only show the camera permissions alert when permissions aren't granted. 

I should either update the logic to support legacy group calls, or wait until we drop support

johnxnguyen: Is it a lot of work to add the original alert (with the correct triggering logic)? That way we could ship out the fix and then remove the legacy alert at a later time.

David-Henner: added the original alert 060e565
If camera permissions are disabled and legacy alert has to be displayed, camera permissions alert will be displayed.

PR #4515, opened 2020-08-06
[Refactor] permissions alerts

David-Henner:
## What's new in this PR?

- Refactor the permissions alert for a cleaner implementation
- Make alerts dismiss button consistent using `Later` instead of `OK`

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIAlertController+Permissions.swift
+++ b/Wire-iOS/Sources/Helpers/UIAlertController+Permissions.swift
@@ -20,7 +20,7 @@ import Foundation
 import UIKit
 import WireCommonComponents
 
-extension UIAlertController {    
+extension UIAlertController {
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/Helpers/UIAlertController+Permissions.swift
```diff
@@ -22,9 +22,32 @@ import WireCommonComponents
...
+            completion: completion
+        )
+    }
+
+    class var microphonePermissionAlert: UIAlertController {
+        return permissionAlert(
+            title: "voice.alert.microphone_warning.title".localized,
+            message: "NSMicrophoneUsageDescription".infoPlistLocalized
+        )
+    }
```
johnxnguyen, line 37: On a side note, I'm starting to really like this style of splitting long method calls over multiple lines and have started to do it myself :)

David-Henner: easy way to make things more readable :)

PR #4516, opened 2020-08-06
[Chore] add accessibility identifier to GridCell

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13627

- Add accessibility identifier to `GridCell`


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/GridCell.swift
```diff
@@ -22,6 +22,8 @@ import UIKit
 class GridCell: UICollectionViewCell {
     static let reuseIdentifier = String(describing: GridCell.self)
 
+    override var accessibilityIdentifier: String? = GridCell.reuseIdentifier
```
billypchan, line 25:
```suggestion
    override var accessibilityIdentifier: String? {
    get {
     return GridCell.reuseIdentifier
     }
     set{
     //no-op
     }
```

PR #4517, opened 2020-08-06
[Fix] conference calling launch argument override

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13665

### Issues

When we're using `-UseConferenceCalling 1` launch argument, and switch off the conference calling toggle, UI will behave as if conference calling is enabled but calls will be using legacy group calling

### Causes

The conference calling override was only overriding the `get` of the `Setting` property for conference calling, while the `set` would still change the conference calling configuration in the SE 

### Solutions

Redefine the launch argument behaviour to only override the `Setting` property once after launch.

### DIFF
--- a/Wire-iOS/Sources/Helpers/AutomationHelper+ConferenceCalling.swift
+++ b/Wire-iOS/Sources/Helpers/AutomationHelper+ConferenceCalling.swift
@@ -1,0 +1,38 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/Helpers/AutomationHelper+ConferenceCalling.swift
```diff
@@ -0,0 +1,38 @@
...
+import Foundation
+import WireCommonComponents
+
+extension AutomationHelper {
+    /// Whether Conference Calling should be used
+    var useConferenceCalling: Bool? {
+        let key = "UseConferenceCalling"
+        guard UserDefaults.standard.object(forKey: key) != nil else {
+            return nil
+        }
```
johnxnguyen, line 28: This guard is redundant, you can safely remove it.

David-Henner: If I remove it, and use only `UserDefaults.standard.bool(forKey: key)`, it will return `false` if there is no object for the key. However, I need it to return `nil` in that case.

johnxnguyen: Ah right, sorry I misread `object` and `bool`.

PR #4520, opened 2020-08-10
[Feature]: Feature Flag File posting actions

marcomaddy1982:
### Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13658


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+CreateViews.swift
```diff
@@ -47,6 +47,10 @@ extension ConversationInputBarViewController {
     }
 
     private func setupInputBar() {
+        gifButton.isHidden = !SecurityFlags.gifAction.isEnabled
+        videoButton.isHidden = !SecurityFlags.videoPickerAction.isEnabled
```
johnxnguyen, line 51: I'm pretty sure the video button just leads to the camera, which is allowed I believe. Do you know what the video picker is referring to?

marcomaddy1982: I honestly have no idea!!!

marcomaddy1982:
> Don't forget to add snapshot tests.

On my machine everything was working fine.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+CreateViews.swift
marcomaddy1982: It seems that they should be the action in the image. I asked to the product team

Wire 2020-08-11 at 09-17-35

marcomaddy1982: the fix is already done.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+CreateViews.swift
```diff
@@ -35,18 +35,29 @@ extension ConversationInputBarViewController {
...
+               locationButton,
+               videoButton]
+            : [photoButton,
+               mentionButton,
+               sketchButton,
+               audioButton,
+               pingButton,
+               uploadFileButton,
+               locationButton,
+               videoButton]
```
johnxnguyen, line 55: I think it would be cleaner to do something like this: 

```swift 
return [
  photoButton,
  mentionButton,
  ...,
  SecurityFlags.gifAction.isEnabled ? gifButton : nil,
  audioButton,
  ...    
].compactMap { $0 }
```

marcomaddy1982:
> I think it would make more sense to just remove the file button altogether, maybe even have a single security flag called `externalFilePicker` which encompasses the camera roll and files app. The focus here is that we add the possibility to prevent all external files from entering the app, not just some. What do you think?

I think that it is a good idea use just one flag and remove the file button altogether.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+FilesPopover.swift
```diff
@@ -89,7 +89,7 @@ extension ConversationInputBarViewController {
                                     mediaTypes: [kUTTypeMovie as String], allowsEditing: true,
                                     pointToView: self.videoButton.imageView)
         }
-
+
```
billypchan, line 92: This file can be reseted since no code is changed.

Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
```diff
@@ -22,6 +22,8 @@ public enum SecurityFlags {
     case clipboard
     case saveMessage
     case profileCameraRoll
+    case gifAction
+    case externalFilePicker
     
     public var bundleKey: String {
```
billypchan, line 28: This class is only used in main app. All `public` marker can be removed.

PR #4526, opened 2020-08-13
[Feature]: Feature Flag CBR Enabled

marcomaddy1982:
### Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13684

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -195,7 +195,7 @@ final class AppRootViewController: UIViewController, SpinnerCapable {
             self.sessionManager?.switchingDelegate = self
             self.sessionManager?.urlActionDelegate = self
             sessionManager.updateCallNotificationStyleFromSettings()
-            sessionManager.useConstantBitRateAudio = SecurityFlags.CBR.isEnabled
+            sessionManager.useConstantBitRateAudio = SecurityFlags.forceConstantBitRateCalls.isEnabled
                 ? true
                 : Settings.shared[.callingConstantBitRate] ?? false
             sessionManager.useConferenceCalling = Settings.shared[.conferenceCalling] ?? false
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -195,7 +195,9 @@ final class AppRootViewController: UIViewController, SpinnerCapable {
             self.sessionManager?.switchingDelegate = self
             self.sessionManager?.urlActionDelegate = self
             sessionManager.updateCallNotificationStyleFromSettings()
-            sessionManager.useConstantBitRateAudio = Settings.shared[.callingConstantBitRate] ?? false
+            sessionManager.useConstantBitRateAudio = SecurityFlags.CBR.isEnabled
+                ? true
+                : Settings.shared[.callingConstantBitRate] ?? false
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
```diff
@@ -24,6 +24,7 @@ enum SecurityFlags {
     case profileCameraRoll
     case gifAction
     case externalFilePicker
+    case CBR
```
johnxnguyen, line 27: Name suggestion: `forceConstantBitRateCalls`.

Wire-iOS/Sources/Components/Settings.swift
```diff
@@ -90,7 +90,9 @@ class Settings {
             case .disableCallKit:
                 SessionManager.shared?.updateCallNotificationStyleFromSettings()
             case .callingConstantBitRate:
-                SessionManager.shared?.useConstantBitRateAudio = newValue as? Bool ?? false
+                SessionManager.shared?.useConstantBitRateAudio = SecurityFlags.CBR.isEnabled
+                    ? true
+                    : newValue as? Bool ?? false
```
billypchan, line 95: It would be better make this into a method and use it also in `AppRootViewController`.

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -195,7 +195,9 @@ final class AppRootViewController: UIViewController, SpinnerCapable {
             self.sessionManager?.switchingDelegate = self
             self.sessionManager?.urlActionDelegate = self
             sessionManager.updateCallNotificationStyleFromSettings()
-            sessionManager.useConstantBitRateAudio = Settings.shared[.callingConstantBitRate] ?? false
+            sessionManager.useConstantBitRateAudio = SecurityFlags.CBR.isEnabled
+                ? true
+                : Settings.shared[.callingConstantBitRate] ?? false
```
johnxnguyen, line 200: Isn't this redundant since the Setting is overridden?

marcomaddy1982: yes it is but I put there just to be sure, anyway I can remove it.

David-Henner: The Setting `set` is overridden, not the `get`

Wire-iOS/Sources/Components/Settings.swift
marcomaddy1982: Cause I will not change anything anymore I would leave here as it is. But thanks for pointing it out.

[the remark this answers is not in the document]
marcomaddy1982: I prefer the ternary operator in this case but no problem to use your suggestion if you like more :-D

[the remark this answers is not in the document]
David-Henner: I think it depends on what behaviour is best in this case.
If you use the ternary, it will trigger a chain of action because the `useConstantBitRateAudio` property has a `didSet` block. I don't know if it's necessary if the value isn't expected to change at any other time.
Whereas using the `where` ensures that `useConstantBitRateAudio` isn't changed at all.

Wire-iOS/Sources/AppRootViewController.swift
marcomaddy1982: Actually We need that cause we are never passing through the settings.

PR #4530, opened 2020-08-13
[Feature]: Feature Flag Enable Link Preview

marcomaddy1982:
### Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13682

### DIFF
--- a/Wire-iOS/Sources/Components/Settings.swift
+++ b/Wire-iOS/Sources/Components/Settings.swift
@@ -171,10 +171,12 @@ class Settings {
 
     static var disableLinkPreviews: Bool {
         get {
-            return ExtensionSettings.shared.disableLinkPreviews
+            return SecurityFlags.avoidLinkPreview.isEnabled
+                ? true
+                : ExtensionSettings.shared.disableLinkPreviews
         }
         set {
-            ExtensionSettings.shared.disableLinkPreviews = newValue
+            ExtensionSettings.shared.disableLinkPreviews = SecurityFlags.avoidLinkPreview.isEnabled ? true : newValue
         }
     }
 
--- a/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
+++ b/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
@@ -24,6 +24,7 @@ enum SecurityFlags {
     case profileCameraRoll
     case gifAction
     case externalFilePicker
+    case avoidLinkPreview
     
     var bundleKey: String {
         switch self {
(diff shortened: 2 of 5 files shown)


Wire-iOS/Sources/Components/Settings.swift
```diff
@@ -171,10 +171,12 @@ class Settings {
...
-            return ExtensionSettings.shared.disableLinkPreviews
+            return SecurityFlags.avoidLinkPreview.isEnabled
+                ? true
+                : ExtensionSettings.shared.disableLinkPreviews
         }
         set {
-            ExtensionSettings.shared.disableLinkPreviews = newValue
+            ExtensionSettings.shared.disableLinkPreviews = SecurityFlags.avoidLinkPreview.isEnabled ? true : newValue
         }
     }
```
johnxnguyen, line 181: I think this is dead code. Perhaps we should remove it.

Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
```diff
@@ -24,6 +24,7 @@ enum SecurityFlags {
     case profileCameraRoll
     case gifAction
     case externalFilePicker
+    case avoidLinkPreview
```
johnxnguyen, line 27: Suggestion: `generateLinkPreviews`

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Options.swift
```diff
@@ -40,7 +40,7 @@ extension SettingsCellDescriptorFactory {
             popularDemandSendButtonSection,
             popularDemandDarkThemeSection,
             appLockSection,
-            linkPreviewSection
+            SecurityFlags.avoidLinkPreview.isEnabled ? linkPreviewSection : nil
```
johnxnguyen, line 43: This logic mismatches the name, I think you want to return nil when the flag is is enabled. However, I think it's better to invert the meaning of the name in this case (see name suggestion).

Wire-iOS/Sources/Components/Settings.swift
```diff
@@ -171,7 +171,9 @@ class Settings {
 
     static var disableLinkPreviews: Bool {
         get {
-            return ExtensionSettings.shared.disableLinkPreviews
+            return !SecurityFlags.generateLinkPreviews.isEnabled
+                ? true
+                : ExtensionSettings.shared.disableLinkPreviews
```
johnxnguyen, line 176: I think we should remove this static var altogether, if no one is using it. But that means you still need to override the subscript for `Settings`.

johnxnguyen: My bad, it is being used :)

PR #4541, opened 2020-08-17
[Fix] Badge items on conversation message shows "Cut" and "Paste"

KaterinaWire:
### Issues
Badge items on conversation message shows `"Cut"` and `"Paste"` actions.
**Actual result**
the badge items show:` Cut Copy Paste Copy Reply > Edit Delete Share Like`

**Expected result**
the badge items show: `Copy Reply Edit Delete Share Like`
### Solutions
To hide `"Cut", "Paste" and "Copy"` actions only if the SecurityFlags.clipboard is disabled.

### Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13695

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
@@ -63,15 +63,15 @@ final class MarkdownTextView: NextResponderTextView {
     override func canPerformAction(_ action: Selector,
                                    withSender sender: Any?) -> Bool {
         switch action {
-        case #selector(UIResponderStandardEditActions.paste(_:)) where !SecurityFlags.clipboard.isEnabled,
-             #selector(UIResponderStandardEditActions.cut(_:)) where !SecurityFlags.clipboard.isEnabled,
-             #selector(UIResponderStandardEditActions.copy(_:)) where !SecurityFlags.clipboard.isEnabled:
-             return SecurityFlags.clipboard.isEnabled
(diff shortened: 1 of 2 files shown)


Wire-iOS/Sources/UserInterface/Components/Views/MarkdownTextView.swift
```diff
@@ -63,9 +63,9 @@ final class MarkdownTextView: NextResponderTextView {
     override func canPerformAction(_ action: Selector,
                                    withSender sender: Any?) -> Bool {
         switch action {
-        case #selector(UIResponderStandardEditActions.paste(_:)),
-             #selector(UIResponderStandardEditActions.cut(_:)),
-             #selector(UIResponderStandardEditActions.copy(_:)):
+        case #selector(UIResponderStandardEditActions.paste(_:)) where !SecurityFlags.clipboard.isEnabled,
+             #selector(UIResponderStandardEditActions.cut(_:)) where !SecurityFlags.clipboard.isEnabled,
+             #selector(UIResponderStandardEditActions.copy(_:)) where !SecurityFlags.clipboard.isEnabled:
```
johnxnguyen, line 68: To reduce duplication and increase readability, you could do something like:

```swift
switch action {
case #selector(UIResponderStandardEditActions.paste(_:)), ...:
    guard SecurityFlags.clipboard.isEnabled else { return false }
    fallthrough
default:
    return super.canPerformACtion(action, withSender: sender)
}
```

KaterinaWire: Done

PR #4545, opened 2020-08-18
Fix: show passcode setup screen when login or unlock

billypchan:
## What's new in this PR?

Clean up unused `allowsUnwind`.
Fix the back button is shown on passcode screen issue by changing the action to `.transition(.passcodeSetup, mode: .reset)`
Unlock screen - support tap keyboard return button to unlock

Tested cases:

When force app lock is true:
- registration on new device - show set up passcode screen
- login second acct on the same device - set up passcode screen not shown
- logout second acct - unlock screen is shown
- logout first acct - unlock screen is not shown
- login second acct on the again same device- set up passcode screen not shown (since passcode is still stored )
- (When the user upgrade from a version not support custom passcode) start the app after upgrade - show set up passcode screen, when the code is created, show the unlock screen

When force app lock is false:
- login with an acct - set up passcode screen not shown
- go to setting -> enable app lock - show set up passcode screen
- register a new acct - set up passcode screen not shown

### DIFF
--- a/Wire-iOS/Sources/Helpers/AppLock/AppLock+PasscodeExistence.swift
+++ b/Wire-iOS/Sources/Helpers/AppLock/AppLock+PasscodeExistence.swift
@@ -21,7 +21,7 @@ import WireCommonComponents
 import WireUtilities
 
 extension AppLock {
-    static var isCustomPassCodeNotSet: Bool {
+    static var isCustomPasscodeNotSet: Bool {
         return AppLock.rules.useCustomCodeInsteadOfAccountPassword &&
                Keychain.fetchPasscode() == nil
     }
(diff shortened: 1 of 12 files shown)


Wire-iOS/Sources/Helpers/AppLock/AppLock+PasscodeExistence.swift
```diff
@@ -0,0 +1,28 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireCommonComponents
+import WireUtilities
+
+extension AppLock {
+    static var isCustomPassCodeNotSet: Bool {
```
johnxnguyen, line 24:
```suggestion
    static var isCustomPasscodeNotSet: Bool {
```

johnxnguyen (approved):

PR #4547, opened 2020-08-19
[Fix]: NSValueTransformer Warning

marcomaddy1982:
### Issues

At runtime we have the following warning:

CoreData: warning: Property 'payload' on Entity 'StoredUpdateEvent' is using nil or an insecure NSValueTransformer. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead.

This is because using NSValueTransformer for core data model attribute is insecure and deprecated.

More info: https://www.kairadiagne.com/2020/01/13/nssecurecoding-and-transformable-properties-in-core-data.html

### Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1017
sync-engine: https://github.com/wireapp/wire-ios-sync-engine/pull/1288


Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -120,6 +120,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         setupTracking()
         NotificationCenter.default.addObserver(self, selector: #selector(userSessionDidBecomeAvailable(_:)), name: Notification.Name.ZMUserSessionDidBecomeAvailable, object: nil)
         
+        if #available(iOS 12.0, *) {
+            NSDictionaryValueTransformer.register()
+        }
```
billypchan, line 125: We should put this in sync engine?

johnxnguyen:
We have to wait for encryption at rest to be re-merged to develop.

Wire-iOS/Sources/AppDelegate.swift
marcomaddy1982: but where in the sync-engine?

marcomaddy1982: After removing `NSDictionaryValueTransformer` form the sync engine this will be removed and th PR closed but let's wait until the encryption will be re-merged to develop.

PR #4554, opened 2020-08-21
[Refactor] Add push notification strategy

KaterinaWire:
## What's new in this PR?

- Use NotificationEngine to fetch events and convert them to ZMLocalNotification objects
- Hide user notification if no user notifications needs to be displayed
- Show fallback message if multiple user notifications needs to be displayed


Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,114 @@
...
+                    titleText = notification.title ?? ""
+                }
+            default:
+                bodyText = "\(localNotifications.count) Notifications"
+            }
+            bestAttemptContent.body = bodyText
+            bestAttemptContent.title = titleText
+
+
+            //            if localNotifications.count > 1 {
```
typfel, line 97: Handled by switch statement above and can be deleted?

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,114 @@
...
+            let localNotifications = processEvents(events, liveEvents: true, prefetchResult: nil, moc: moc).compactMap { $0 }
+            var bodyText = ""
+            var titleText = ""
+            switch localNotifications.count {
+            case 0:
+                let emptyContent = UNNotificationContent()
+                contentHandler!(emptyContent)
+            case 1:
+                if let notification = localNotifications.first {
+                    print(notification)
```
typfel, line 86: Add a `// TODO katerina debugging` so that we don't forgot to remove this later.

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,114 @@
...
         self.contentHandler = contentHandler
         bestAttemptContent = (request.content.mutableCopy() as? UNMutableNotificationContent)
         
-        if let bestAttemptContent = bestAttemptContent {
-            // Modify the notification content here...
-            bestAttemptContent.title = "\(bestAttemptContent.title) [modified]"
-            
-            contentHandler(bestAttemptContent)
-        }
+//        if let bestAttemptContent = bestAttemptContent {
```
typfel, line 38: Add a `TODO katerina` describing why this is commented out

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,114 @@
...
-class NotificationService: UNNotificationServiceExtension {
+public class NotificationService: UNNotificationServiceExtension {
 
     var contentHandler: ((UNNotificationContent) -> Void)?
     var bestAttemptContent: UNMutableNotificationContent?
+    var notificationSession: NotificationSession?
 
-    override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+    public override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        notificationSession = try! self.createNotificationSession()
```
typfel, line 33: I don't think we should assume that the `NotificationService` is restarted everytime we get a notification request and thus recreate the notification session.

Either add a `TODO katerina` or handle  the fact the session might already exist.

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Removed

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Removed

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Removed

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Added `TODO` comment and added to the `Technical debts` Jira ticket

Wire Notification Service Extension/Info.plist
```diff
@@ -15,17 +17,26 @@
...
 	<key>NSExtension</key>
 	<dict>
 		<key>NSExtensionPointIdentifier</key>
 		<string>com.apple.usernotifications.service</string>
 		<key>NSExtensionPrincipalClass</key>
 		<string>$(PRODUCT_MODULE_NAME).NotificationService</string>
 	</dict>
+	<key>WireGroupId</key>
+	<string>${WIRE_BUNDLE_ID}</string>
+	<key>NSAppTransportSecurity</key>
```
typfel, line 36: We should not allow arbitrary loads

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,65 @@
...
     var bestAttemptContent: UNMutableNotificationContent?
+    var notificationSession: NotificationSession?
 
-    override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+    public override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
+        if notificationSession == nil {
+            notificationSession = try? self.createNotificationSession()
+        }
+        
         self.contentHandler = contentHandler
```
typfel, line 36: There's no need to interact with the `notificationSession` here? Like updating the `PushNotificationStatus`?

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,65 @@
...
+}
 
+extension NotificationService: LocalNotificationsDelegate {
+    public func shouldCreateNotificationWith(_ alert: (title: String, body: String), showNotification: Bool) {
+        if showNotification {
+            if let bestAttemptContent = bestAttemptContent {
+                bestAttemptContent.title = alert.title
+                bestAttemptContent.body = alert.body
+                
+                contentHandler!(bestAttemptContent)
```
typfel, line 74:
```suggestion
                contentHandler?(bestAttemptContent)
```

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,65 @@
...
+            }
+        } else {
+            let emptyContent = UNNotificationContent()
+            contentHandler!(emptyContent)
```
typfel, line 78:
```suggestion
            contentHandler?(emptyContent)
```

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,65 @@
...
-        if let bestAttemptContent = bestAttemptContent {
-            // Modify the notification content here...
-            bestAttemptContent.title = "\(bestAttemptContent.title) [modified]"
-            
-            contentHandler(bestAttemptContent)
-        }
     }
     
-    override func serviceExtensionTimeWillExpire() {
+    public override func serviceExtensionTimeWillExpire() {
```
typfel, line 40: If we don't manage to download the notification we still not show the original notification since it won't make sense to the user. We could either show some fallback message  *or* we hide the notification.

This is something we should discuss with product/design. I'd say add a `TODO` about this and hide the notification for now.

typfel (changes requested):

Wire Notification Service Extension/Info.plist
KaterinaWire: Removed

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Done

typfel: We are still not hiding the notification by setting an empty body and title.

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,64 @@
...
+        if let bestAttemptContent = bestAttemptContent {
+            switch messageCount {
+            case 0:
+                let emptyContent = UNNotificationContent()
+                contentHandler?(emptyContent)
+            case 1:
+                bestAttemptContent.title = alert.title
+                bestAttemptContent.body = alert.body
+            default:
+                bestAttemptContent.body = "\(messageCount) " + "self.settings.notifications.push_notification.title".localized
```
typfel, line 76: We should include the messageCount as a localization string argument instead of concatenating  them together, because we can't assume that all languages have the number in the beginning of the sentence.

Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -131,6 +136,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         zmLog.info("application:didFinishLaunchingWithOptions END \(String(describing: launchOptions))")
         zmLog.info("Application was launched with arguments: \(ProcessInfo.processInfo.arguments)")
         
+        let notificationTypes: UIUserNotificationType = [UIUserNotificationType.alert,UIUserNotificationType.badge, UIUserNotificationType.sound]
```
typfel, line 139: Add a `TODO` reminding us that this code is only for debugging and should be moved to SE.

KaterinaWire: Added

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: I added a new localization string. Do I have to do something more to have all translations in the future?

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Ah yes...
I moved the logic from the part where we receive empty events, so there won't be any notifications.

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -17,30 +17,65 @@
...
         // Called just before the extension will be terminated by the system.
         // Use this as an opportunity to deliver your "best attempt" at modified content, otherwise the original push payload will be used.
         if let contentHandler = contentHandler, let bestAttemptContent =  bestAttemptContent {
             contentHandler(bestAttemptContent)
         }
     }
+    
+    private func createNotificationSession() throws -> NotificationSession? {
+        guard let applicationGroupIdentifier = Bundle.main.applicationGroupIdentifier,
+            let accountIdentifier = accountManager?.selectedAccount?.userIdentifier
```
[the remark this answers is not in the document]
KaterinaWire: I added TODO comment

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1602,7 +1602,7 @@
 // Those are used by the backend to localize APNS
 "push.notification.new_user" = "%@ joined Wire";
 "push.notification.new_message" = "New message";
-
+"push.notifications.push_notification.title" = "%d Notifications";
```
typfel, line 1605: I would name this:

push.notifications.bundled_message.title

typfel: Also does this work? Wouldn't you need add new `.strings` file to the notification service extension? Similar to how we have a `.strings` file in the share extension.

typfel: And then we would also need to update the crowding jobs on Jenkins

KaterinaWire: If we anyway need add new .strings file and update the crowding jobs on Jenkins, does it make sense to do it in Notification engine, where it supposed to be?

typfel: No, it's still better to keep it in the UI project then we just do need to modify an line in existing Jenkins job instead of creating an entirely new job.

KaterinaWire: I added new `.strings` file to the notification service extension.

KaterinaWire: Should we update the crowding jobs on Jenkins after a merge in develop?

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: I don't think I understand what you mean.

KaterinaWire: I implemented it in this PR:
https://github.com/wireapp/wire-ios-notification-engine/pull/6

typfel: 👍

Wire Notification Service Extension/NotificationService.swift
```diff
@@ -16,31 +16,79 @@
...
         
-        if let bestAttemptContent = bestAttemptContent {
-            // Modify the notification content here...
-            bestAttemptContent.title = "\(bestAttemptContent.title) [modified]"
-            
-            contentHandler(bestAttemptContent)
+        if let session = notificationSessions[accountIdentifier] {
+            currentNotificationSession = session
+        } else {
+            let notificationSession = try? self.createNotificationSession(request.content as? UNMutableNotificationContent)
```
typfel, line 41: Why not pass the `accountIdentifier` as the argument instead?

typfel (changes requested):

Wire-iOS/Resources/Base.lproj/Localizable.strings
typfel: Yes we should

Wire Notification Service Extension/NotificationService.swift
KaterinaWire: Because we also need to get the event ID from the payload.

PR #4558, opened 2020-08-24
Fix/zios 13735 when pressing cancel on wipe date screen it still wipes data 

billypchan:
### Issues

- When login/register, passcode setup screen is not keyboard avoiding
- When wiping database confirm dialog is shown, pressing cancel also wipes the database

### Causes

- the passcode screen is not wrapped in keyboard avoiding view controller like when presented from app lock screen
- missing confirm word check

### Solutions

- create `KeyboardAvoidingAuthenticationCoordinatedViewController` to wrap the passcode screen in keyboard avoiding view controlle
- check the confirm keyboard before wiping.


Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/WipeDatabaseViewController.swift
```diff
@@ -98,6 +98,10 @@ final class WipeDatabaseViewController: UIViewController {
     func presentConfirmAlert() {
         let confirmController = RequestPasswordController(context: .wiping, callback: { [weak self] confirmText in
             
+            if confirmText != "wipe_database.alert.confirm_input".localized {
+                return
+            }
+
```
johnxnguyen, line 104: A `guard` would be more appropriate and readable here.

johnxnguyen (approved):

PR #4560, opened 2020-08-25
[Feature] Security flag to force encryption at rest

johnxnguyen:
## What's new in this PR?

Add a new security flag which allows forcing encryption at rest to be enabled. If the flag is enabled, then encryption at rest is enabled when the user session is created. Additionally, the option in the account settings is hidden to prevent the feature being disabled.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-build-configuration/pull/43

### DIFF
--- a/Wire-iOS/Sources/AppRootViewController.swift
+++ b/Wire-iOS/Sources/AppRootViewController.swift
@@ -556,6 +556,10 @@ extension AppRootViewController: SessionManagerCreatedSessionObserver, SessionMa
                 soundEventListeners[accountId] = SoundEventListener(userSession: userSession)
             }
         }
+
+        if SecurityFlags.forceEncryptionAtRest.isEnabled && !userSession.encryptMessagesAtRest {
+            userSession.encryptMessagesAtRest = true
+        }
     }
 
     func sessionManagerCreated(unauthenticatedSession: UnauthenticatedSession) {
--- a/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
+++ b/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
@@ -60,7 +60,7 @@ enum SecurityFlags {
         case .backup:
             return "BackupEnabled"
         case .forceEncryptionAtRest:
-            return "ForceEncryptionAtRest"
+            return "ForceEncryptionAtRestEnabled"
         }
     }
     
--- a/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
+++ b/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
@@ -31,7 +31,7 @@ enum SecurityFlags {
     case cameraRoll
     case backup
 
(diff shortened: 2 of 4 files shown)


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -556,6 +556,10 @@ extension AppRootViewController: SessionManagerCreatedSessionObserver, SessionMa
                 soundEventListeners[accountId] = SoundEventListener(userSession: userSession)
             }
         }
+
+        if SecurityFlags.forceEncryptionAtRest.isEnabled && !userSession.encryptMessagesAtRest {
+            userSession.encryptMessagesAtRest = true
+        }
```
johnxnguyen, line 562: @typfel I'm wondering if this is the best place to do this: when we will migrate the existing database to and from EAR, we might need to display some progress indication UI. I'm wondering if it's better to do this in the `SessionManager` where we have more control. What do you think?

billypchan: create a method like `userSession.setEncryptMessagesAtRestIfNeeded()` would be better.

Wire-iOS/Wire-Info.plist
```diff
@@ -206,5 +206,7 @@
 	<string>${CAMERA_ROLL_ENABLED}</string>
 	<key>BackupEnabled</key>
 	<string>${BACKUP_ENABLED}</string>
+	<key>ForceEncryptionAtRest</key>
+	<string>$(FORCE_ENCRYPTION_AT_REST)</string>
```
marcomaddy1982, line 210: Maybe FORCE_ENCRYPTION_AT_REST_ENABLED

Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
```diff
@@ -55,6 +59,8 @@ enum SecurityFlags {
             return "CameraRollEnabled"
         case .backup:
             return "BackupEnabled"
+        case .forceEncryptionAtRest:
+            return "ForceEncryptionAtRest"
```
marcomaddy1982, line 63: Maybe "ForceEncryptionAtRestEnabled"

marcomaddy1982 (approved):

Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
johnxnguyen: Done

Wire-iOS/Wire-Info.plist
johnxnguyen: Done

Wire-iOS/Sources/AppRootViewController.swift
typfel: Yes, I agree it's probably better to handle this in `SessionManager`, but let's refactor it when we add the migration.

typfel (approved):

PR #4563, opened 2020-08-25
Fix: disable auto fill passcode

billypchan:
## What's new in this PR?
By default, when the user forgets his/her passcode code, the user can fill in the text field store passcode with touch id/face id and then screen lock passcode. 
Prevent the user auto fill passcode from keyboard accessory. A hack is applied by setting `textContentType = .oneTimeCode`.

### DIFF
--- a/Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
+++ b/Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
@@ -249,9 +249,17 @@ final class AccessoryTextField: UITextField, TextContainer, Themeable {
             autocapitalizationType = .none
             returnKeyType = isNew ? .default : .continue
             if #available(iOS 12, *) {
-                textContentType = isNew ? .newPassword : .password
+                if isNew {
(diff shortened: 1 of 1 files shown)


johnxnguyen (commented):
What is the context of this? Is it a requirement that the user can't use autofill?

What is the difference between `.oneTimeCode` and `.init(rawValue: "")`?

billypchan:
@johnxnguyen `.oneTimeCode` is only supported for iOS 12+. This PR fixed this issue that the user can fill the passcode field from iOS keyboard accessory.

johnxnguyen:
@billypchan I mean what is the behavior of `.init(rawValue: "")`? Why would we choose `.oneTimeCode` when we can use `.init(rawValue: "")`?

My concern is that we are misusing `.oneTimeCode`.

billypchan:
`.oneTimeCode` would be the correct choice since it is not stored in iOS system keychain.

related discussion:
https://stackoverflow.com/questions/45452170/ios-11-disable-password-autofill-accessory-view-option

johnxnguyen (approved):

PR #4569, opened 2020-09-01
Moving debug actions to their own file

marcoconti83:
## What's new in this PR?

As part of an investigation about decryption errors with the security team, I want to add more debugging capabilities to the app. In particular, we want to add a debug console to the app that will allow us to enter specific commands with parameters (such as the user ID or conversation ID that should be affected by that command).

This PR does some ground work, by adding more options to the troubleshooting public menu. As I was adding these menu items, I also noticed a lot of code related to (internal) debug options.

I thought it would be cleaner to have debug logic in its own file, instead of inside `SettingsCellDescriptorFactory`. I moved that code to its own file. In the process I removed some duplicated code (e.g. code for alerts) by refactoring to common functions.

I created this PR before I continue with the next steps that are needed for my investigation, so that changes are easier to track. I will open more PRs in the future.

### Main changes
- Moved debug menu actions to their own file `DebugActions` (I'm open to suggestions on naming or to where to put this file)
- Added placeholder debug menu item to enter debug commands
- Added debug menu item to print user ID

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
@@ -1,0 +1,292 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
```diff
@@ -0,0 +1,292 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireSyncEngine
+import AppCenterCrashes
+
+
+class DebugActions {
```
johnxnguyen, line 24: I suggest using an enum if you want to namespace some static functions because then you can't instantiate an instance of `DebugActions`.

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
marcoconti83: Thanks, will do

PR #4571, opened 2020-09-04
Feature: install Countly SDK

billypchan:
## What's new in this PR?

Install Countly SDK and restore `Analystic` class methods.

### TODO
- [x] fork Countly/countly-sdk-ios
- [x] fix testThatItSetsOptOutOnAppCenter


Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -0,0 +1,77 @@
...
+    override init() {
+        if let countlyAppKey = Bundle.countlyAppKey,
+            let countlyHost = Bundle.countlyHost {
+            let config: CountlyConfig = CountlyConfig()
+            config.appKey = countlyAppKey
+            config.host = countlyHost
+            config.deviceID = CLYTemporaryDeviceID //TODO: wait for ID generation task done
+            config.manualSessionHandling = true
+            Countly.sharedInstance().start(with: config)
+        }
```
johnxnguyen, line 53: Perhaps we should make the initializer fallible so that we prevent creating an instance that isn't configured.

Wire-iOS/Sources/Analytics/Analytics.swift
```diff
@@ -19,29 +19,47 @@
...
-    }
+    private static var sharedAnalytics: Analytics?
+    
+    static var shared: Analytics = {
+        return sharedAnalytics!
+    }()
 
-    class func loadShared(withOptedOut optedOut: Bool) {
-        //no-op
+    static func loadShared(with optedOut: Bool) {
```
johnxnguyen, line 38: This argument label obfuscates the significance of the `Bool`.

```swift
// What does this mean?
Analytics.loadShared(with: true)

// This is conveys meaning
Analytics.loadShared(withOptedOut: true)

// Or even perhaps better
Analytics.loadShared(optedOut: Bool)
```

johnxnguyen: Actually, I think the term "optedOut" could be replaced with something clearer too. Perhaps "isActivated", to signify that Analytics is turned on and broadcasting events.

Wire-iOS/Sources/Analytics/Analytics.swift
```diff
@@ -19,29 +19,47 @@
...
+    private func setupObserver() {
+         NotificationCenter.default.addObserver(self, selector: #selector(userSessionDidBecomeAvailable(_:)), name: Notification.Name.ZMUserSessionDidBecomeAvailable, object: nil)
     }
     
+    @objc
+    private func userSessionDidBecomeAvailable(_ note: Notification?) {
+//        callingTracker = AnalyticsCallingTracker(analytics: self)
+        //TODO:
+//        decryptionFailedObserver = AnalyticsDecryptionFailedObserver(analytics: self)
+        setTeam(ZMUser.selfUser().team)
```
johnxnguyen, line 60: What happens if two different team accounts are logged in? Which team do we set?

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -0,0 +1,77 @@
...
+import WireSystem
+import Countly
+
+private let zmLog = ZMSLog(tag: "Analytics")
+
+final class AnalyticsCountlyProvider: NSObject, AnalyticsProvider {
+    
+    private var sessionBegun: Bool = false
+    
+    var isOptedOut: Bool {
```
johnxnguyen, line 29: Here the name `isActivated` will make more sense than `isOptedOut`. See my previous comment.

Wire-iOS/Sources/Analytics/AnalyticsProviderFactory.swift
```diff
@@ -36,16 +36,16 @@ final class AnalyticsProviderFactory: NSObject {
...
             return AnalyticsConsoleProvider()
-        }
-        else if AutomationHelper.sharedHelper.useAnalytics {
+        } else if AutomationHelper.sharedHelper.useAnalytics {
             // Create & return valid provider, when available.
-            return nil
-        }
-        else {
+            let provider = AnalyticsCountlyProvider()
+            provider.isOptedOut = false
```
johnxnguyen, line 46: Shouldn't we be passing in this value as a parameter?

Wire-iOS/Sources/Analytics/Analytics.swift
billypchan: This is copied from the obj-c code, I think it is no longer necessary and can be replaced with a `init`

Wire-iOS/Sources/Analytics/Analytics.swift
billypchan: This line should be called every time when `Notification.Name.ZMUserSessionDidBecomeAvailable` is fired.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: It is restored from the objc code and I will check is it make sence to rename. It is related to user opted out in privarcy setting.

Wire-iOS/Sources/Analytics/Analytics.swift
johnxnguyen: Right, which happens every time the user switches account.

johnxnguyen: So we only broadcast events for the foreground account?

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
johnxnguyen: Yes but I would argue that the term "opt in / out"  is specific to the settings domain (these are the terms that the user will see), but not as precise in the context of the analytics provider, where it really means "activate / deactivate" or something like that.

Wire-iOS/Sources/Analytics/Analytics.swift
billypchan: I will review the restored objc code in the coming PR to see how does it works.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: I would prefer to keep the name for now since there are some event tags have the "opted out" keyword.

Wire-iOS/Sources/Analytics/AnalyticsProviderFactory.swift
```diff
@@ -42,9 +42,7 @@ final class AnalyticsProviderFactory: NSObject {
             return AnalyticsConsoleProvider()
         } else if AutomationHelper.sharedHelper.useAnalytics {
             // Create & return valid provider, when available.
-            let provider = AnalyticsCountlyProvider()
-            provider.isOptedOut = false
-            return provider
+            return AnalyticsCountlyProvider(optedOut: false)
```
johnxnguyen, line 45: I mean, the value of `optedOut` should be passed in as an argument to the enclosing factory method.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
johnxnguyen: I'm not sure I follow, why does the tag affect how we name the property?

billypchan: In `Analytics+OptOut.swift`, we have a line `tagEvent("settings.opted_out_tracking")` when this property set. I will check is it still necessary.

billypchan: `isOptedOut` is a property of `AnalyticsProvider` protocol. I would suggest not to rename it until finished all the redo of analytics objc code.

Wire-iOS/Sources/Analytics/AnalyticsProviderFactory.swift
billypchan: actually we don't need this paramenter, I removed it.

johnxnguyen: I'm confused how the opt out works. The `TrackingManager` is the central place to opt in / out of metric collection, including `Analytics`. But this opt out value doesn't get passed to the analytics provider.

Furthermore, when we create `Analytics`, the `provider` will only be created if we opt out is `false`. So, it suggests that a provider exists only when the user doesn't opt out. But the provider has an `optOut` property which will start / end the analytics session when the value is changed. This suggests that there can be a provider even when the user has opted out, but it doesn't do anything in that case.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
johnxnguyen: Do as you think is best 👍

Wire-iOS/Sources/Analytics/AnalyticsProviderFactory.swift
billypchan: tested with the setting screen again and made some changes.
the opt out is controlled by`TrackingManager.disableCrashAndAnalyticsSharing` and passed to the analytics provider.
When the user opted out,  it will call the provider's `flush` method, and then provider will set to nil.

billypchan:
@johnxnguyen tested with the setting screen's switch and it should works as expected now.

johnxnguyen (approved):

watcher699 (commented):
Y

PR #4599, opened 2020-09-22
Set user when switching accounts

billypchan:
## What's new in this PR?

Update Countly user properties when the user switches between different accounts.
- [x] fix the issue "when switch to a non team user, the user info is not updated."


Wire-iOS/Sources/Analytics/Analytics.swift
```diff
@@ -49,12 +49,17 @@ final class Analytics: NSObject {
 //        callingTracker = AnalyticsCallingTracker(analytics: self)
         //TODO:
 //        decryptionFailedObserver = AnalyticsDecryptionFailedObserver(analytics: self)
-        setTeam(ZMUser.selfUser().team)
+        selfUser = ZMUser.selfUser()
```
typfel, line 52: Can't we use `SelfUser.current`? Same question applies for other places where you get the self user.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -73,10 +78,20 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
         zmLog.info("AnalyticsCountlyProvider \(self) deallocated")
     }
 
-    //TODO: call this after switched account
-    func updateUserProperties() {
-        guard let selfUser = SelfUser.provider?.selfUser as? ZMUser,
-              let team = selfUser.team else {
+    private var shouldTracksEvent: Bool {
+        guard let selfUser = selfUser as? ZMUser,
+            let _ = selfUser.team else {
```
typfel, line 83: What are we checking here, if the self user belongs to a team? If so you could use `isTeamMember` instead and avoid the cast to `ZMUser`.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -73,10 +78,20 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
...
+        }
+
+        return true
+    }
+    
+    /// update user properties after self user changes
+    private func updateUserProperties() {
+        guard shouldTracksEvent,
+            let selfUser = selfUser as? ZMUser,
+            let team = selfUser.team else {
```
typfel, line 94: Is the only thing we query from the `Team` objects the number of members? Then we could consider exposing that on the `UserType` or expose a `TeamType` on the `UserType`.

typfel (changes requested):
So we can only track events for the active account? I guess that's fine we only generate events when the application is in the foreground.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: fixed

Wire-iOS/Sources/Analytics/Analytics.swift
billypchan: fixed

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: fixed.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -73,10 +78,28 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
...
+                
+            //clean up
+            ["team_team_id",
+             "team_user_type",
+             "team_team_size",
+             "user_contacts"].forEach() {
+                Countly.user().unSet($0)
+            }
+                Countly.user().set("user_id", value: (self.selfUser as? ZMUser )?.userId.uuid.zmSHA256Digest().zmHexEncodedString() ?? "")
```
typfel, line 98: What is this `user_id` property we are setting here? Seems strange that we are sending the `user_id` of the self user when are going through the trouble of generating a random analytics identifier.

billypchan: the hashed user id is required in the spec, I will discuss again to see if the user id need to be unset also for non team users.

billypchan: after discussion, this case is not likely to happen and since we do not track for non-team users, I will clean the user id.

typfel: Do we send the `analyticsIdentifier` in a different way?

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -19,63 +18,161 @@
...
-    
+
     func flush(completion: Completion?) {
         isOptedOut = true
         completion?()
     }
 }
+
+extension UserType {
+    var userIdHash: String? {
```
typfel, line 173: This was outdated and needs to be replaced by the `analyticsIdentifier`

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: add a TODO for now and usr_id will be filled in next PR.

typfel (approved):

PR #4600, opened 2020-09-23
[Feature] show popup for outdated clients

David-Henner:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/ZIOS-13745

- Warn the user a call cannot be placed when their avs version is outdated (blacklisted)


Wire-iOS/Sources/UserInterface/Calling/CallController/CallController.swift
```diff
@@ -52,39 +52,45 @@ final class CallController: NSObject {
...
         guard let userSession = ZMUserSession.shared() else { return }
-
-        if let priorityCallConversation = userSession.priorityCallConversation {
-            topOverlayCall = priorityCallConversation
-
-            if priorityCallConversation == minimizedCall {
-                minimizeCall(in: priorityCallConversation)
-            } else {
-                let animated: Bool
-                if SessionManager.shared?.callNotificationStyle == .callKit {
```
billypchan, line 71: no need to check `SessionManager.shared?.callNotificationStyle` now?

David-Henner: we're still checking in `shouldAnimate(call: ZMConversation)` at line `237` :)

billypchan (approved):

PR #4602, opened 2020-09-23
Feature: restore analytics calling tracker

billypchan:
## What's new in this PR?

Restore AnalyticsCallingTracker and add new screen share tracking.


Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -34,14 +32,39 @@ struct CallInfo {
     let video: Bool
 }
 
+extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
+    func callParticipantsDidChange(conversation: ZMConversation, participants: [CallParticipant]) {
```
typfel, line 36: This will get called for many reason not only when screen sharing is turned on or off, for example when a participant mutes himself/herself.

- You would need guard against only updating `screenSharingStartTime` the first time you discover `videoState == .screenSharing` is on.
- What about if someone turn screen sharing on or off many times during a call?
- What about if the call ends while screen sharing is on? Then you wouldn't have any `screenSharingEndTime` set?
- What if multiple people screen shared during a call? Should we maybe send multiple events during a call?

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -102,15 +133,34 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
                 callInfo.establishedDate = Date()
                 analytics.tag(callEvent: .established, in: conversation, callInfo: callInfo)
             }
+            
+            guard let userSession = ZMUserSession.shared() else {
+                Log.calling.error("UserSession not available when .established call")
```
typfel, line 138: I would never except this to happen. If you still want to log something I think we should do it under a separate Analytics tag.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -102,15 +133,34 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
...
+            }
+
+            callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
+
         case .terminating(reason: let reason):
             if let callInfo = callInfos[conversationId] {
                 analytics.tag(callEvent: .ended(reason: reason.analyticsValue), in: conversation, callInfo: callInfo)
+
+                if let start = screenSharingStartTime {
+                    analytics.tag(callEvent: .screenSharing(duration: (screenSharingEndTime ?? Date()).timeIntervalSince(start)),
```
typfel, line 149: As mentioned above maybe this is not right place to send this event?

Wire-iOS Tests/Analytics/AnalyticsCountlyProviderTests.swift
```diff
@@ -20,8 +20,30 @@ import Foundation
...
     func testThatLogRoundedConvertNumberIntoBuckets() {
         XCTAssertEqual([0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 50, 100].map({$0.logRound()}), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 9, 46, 91])
     }
+    
+    func testThatCountlyAttributesFromConverationIsGenerated() {
+        let mockConversation = ZMConversation.createOtherUserConversation(moc: coreDataFixture.uiMOC, otherUser: otherUser)
+
+        let convertedDictionary = mockConversation.attributesForConversation.countlyStringValueDictionary
+        
+        XCTAssertEqual(convertedDictionary, ["conversation_guests_wireless": "0", "is_allow_guests": "False", "conversation_type": "one_to_one", "conversation_guests_pro": "0", "user_type": "user", "with_service": "False", "conversation_size": "2", "conversation_services": "0", "is_global_ephemeral": "False", "conversation_guests": "0"])
```
typfel, line 47: Formatting?

typfel (changes requested):

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
billypchan: the current logic design just records the last screen share timing from any other remote user. I would discuss this with the PM.

typfel: Still would need to account for the fact the `callParticipantsDidChange` is also called when the user mutes himself.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
billypchan: fixed

typfel: How is this fixed? I see no change.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
billypchan: rewrite the screen duration recording and it tags event when screen sharing is ended.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -34,32 +32,72 @@ struct CallInfo {
...
+            if let conversationId = conversation.remoteIdentifier,
+               let callInfo = callInfos[conversationId] {
+                analytics.tag(callEvent: .screenSharing(duration: -screenSharingInfo.startTime.timeIntervalSinceNow),
+                              in: conversation,
+                              callInfo: callInfo)
+
+                screenSharingInfos.remove(screenSharingInfo)
+            }
+
+        }
```
typfel, line 62: This approach is much better! It has some bugs though:

- Multiple `ScreenSharingInfo` will get inserted for the same user when `callParticipantsDidChange()` called repeatedly.
- Having a `screenSharingInfos` set makes me think we want to support tracking multiple users screen sharing at the same time but that doesn't work due to the `first(where:)` usage. Either make it work for multiple users or replace the set with a single value.
- Code is hard to read, maybe break it up separate statements instead of one big nested if statement.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
typfel: I would also log this an analytics error.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -80,16 +79,15 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
...
-        userProperties["user_id"] = selfUser.userId.uuid.zmSHA256Digest().zmHexEncodedString()
-
         let teamSize = team.members.count.logRound()
-        userProperties["team_team_size"] = teamSize
-        userProperties["user_contacts"] = teamSize
 
-        let convertedAttributes = convertToCountlyDictionary(dictioary: userProperties)
+        let userProperties: [String: Any] = ["team_team_id": selfUser.hasTeam,
+                                             "team_user_type": selfUser.teamRole,
+                                             "user_id": selfUser.userId.uuid.zmSHA256Digest().zmHexEncodedString(),
```
typfel, line 86: This is still wrong but I guess will be fixed in https://github.com/wireapp/wire-ios/pull/4599/files ?

typfel (changes requested):

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
billypchan: fixed

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: fixed after merging

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
billypchan: - the `ScreenSharingInfo` object is inserted only when the participant has a video state == `.screenSharing` and it will be removed from `screenSharingInfos` set when screen sharing ends. I would add a test to show that.
- when multiple screen share starts, it will insert several `ScreenSharingInfo` objects with different `clientId` and tracking them is possible. (I will add a test)
- I will refactor the code.

typfel (approved):

PR #4605, opened 2020-09-24
[Feature]: New Account Creation Screen 

marcomaddy1982:
## What's new in this PR?

UI design for Account Creation Screen for the Custom Backend Scenario


billypchan:
Do we have a snapshot for this case?

marcomaddy1982:
> Do we have a snapshot for this case?

No, no snapshot tests where there.

billypchan (approved):

PR #4618, opened 2020-09-29
[Feature]: Analytics add UserID

marcomaddy1982:
## What's new in this PR?

in order to identify an user on Analytics we added an unique id in core data for the User object and set as value in the "user_id" of the Analytics userProperties.

### Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13750


johnxnguyen (changes requested):
Don't forget to bump the data model component.

billypchan (approved):

billypchan:
@marcoconti83 You may also need to bump Sync engine since UI project links to data modal via Sync engine.

billypchan (changes requested):
bump sync engine and check the tests would pass or not.

marcomaddy1982:
> Don't forget to bump the data model component.

Done

marcomaddy1982:
> bump sync engine and check the tests would pass or not.

I guess this is not necessary for now,  cause the data-model dependency is pointing to feature/Analytics?

PR #4631, opened 2020-10-06
[Feature]: AppStateController Refactoring

marcomaddy1982:
## What's new in this PR?

The context of this PR is the `App Launching Routing Refactoring`. The idea is separate it in smaller chunks that are easier to review. I thought that a reasonable starting point could be refactor the `AppState` and the way we calculate it in the `AppStateController`. 
Previously in the `AppStateController` you could have found a lot of variables used to calculate the current `AppState`. These variables were set in the `SessionManagerDelagate` methods and used in the `calculateAppState()` method to find the current appState. This approach is really verbose and it have been produced a lot of not necessary code. 
The idea behind this PR is set directly the appState in the `SessionManagerDelagate` call backs and get rid of all the variables used in the  `calculateAppState()` and of the method itself. 
Moreover to decouple the `AppStateController` from the `SessionManagerDelagate` callbacks (and make it lighter) I decided to create the  `AppStateSessionManagerObserver`. The scope of this class is to inform the `AppStateController` of the new `AppState`, in this manner the only dirty of the `AppStateController` will be transit (when necessary) from a state to another.

It seems to works fine but It would be nice if some of you at some point could do other manual test.

P.S: Any other idea is super well accepted

### DIFF
--- a/Wire-iOS/Sources/AppStateController.swift
+++ b/Wire-iOS/Sources/AppStateController.swift
@@ -18,7 +18,6 @@
 
 import Foundation
 import WireSyncEngine
-//import WireSystem
 
 private let zmLog = ZMSLog(tag: "AppState")
 
--- a/Wire-iOS/Sources/AppStateController.swift
+++ b/Wire-iOS/Sources/AppStateController.swift
@@ -39,26 +39,10 @@ final class AppStateController : NSObject {
     
     // MARK - Private Set Property
     private(set) var sessinManagerObeserver = AppStateSessionManagerObserver()
+    private(set) var previousAppState: AppState = .headless
     private(set) var appState: AppState = .headless {
         willSet {
-            lastAppState = appState
-        }
-    }
-    private(set) var lastAppState: AppState = .headless
-    
-    // MARK - Test Property (??????)
-    private var isRunningTests = ProcessInfo.processInfo.isRunningTests {
-        didSet {
-            appState = !isRunningTests
-                ? .headless
-                : .unauthenticated(error: nil)
-        }
-    }
-    var isRunningSelfUnitTest = false {
-        didSet {
-            appState = isRunningSelfUnitTest
-                ? .headless
-                : .unauthenticated(error: nil)
+            previousAppState = appState
         }
     }
     
@@ -78,11 +62,10 @@ extension AppStateController: AppStateSessionManagerObserverDelegate {
     }
     
     // MARK - Private Helpers
-    
     private func transition(to appState: AppState,
                             completion: (() -> Void)? = nil) {
         self.appState = appState
-        if lastAppState != appState {
+        if previousAppState != appState {
             zmLog.debug("transitioning to app state: \(appState)")
             delegate?.appStateController(transitionedTo: appState) {
                 completion?()
@@ -100,7 +83,7 @@ extension AppStateController: AppStateSessionManagerObserverDelegate {
     }
 }
 
-// TO DO: Ask why AuthenticationCoordinatorDelegate implements AuthenticationStatusProvider. For the scope of knowing the appState it is not necessary. We could get rid off all the AuthenticationStatusProvider properties
+// TO DO: Ask why AuthenticationCoordinatorDelegate implements AuthenticationStatusProvider. For the scope of knowing the appState it is not necessary. We could get rid off all the AuthenticationStatusProvider properties.
 
 // MARK - AuthenticationCoordinatorDelegate
 
--- a/Wire-iOS/Sources/AppStateController.swift
+++ b/Wire-iOS/Sources/AppStateController.swift
@@ -18,112 +18,72 @@
 
 import Foundation
 import WireSyncEngine
-import WireSystem
+//import WireSystem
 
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/AppStateController.swift
```diff
@@ -18,112 +18,54 @@
...
-    fileprivate var loadingAccount : Account?
-    fileprivate var authenticationError : NSError?
-    fileprivate var isRunningTests = ProcessInfo.processInfo.isRunningTests
-    var isRunningSelfUnitTest = false
-    var databaseEncryptionObserverToken: Any? = nil
-
-    /// The state of authentication.
-    fileprivate(set) var authenticationState: AuthenticationState = .undetermined
+    // MARK - Private Set Property
+    private(set) var sessinManagerObeserver = AppStateSessionManagerObserver()
```
johnxnguyen, line 40:
```suggestion
    private(set) var sessionManagerObserver = AppStateSessionManagerObserver()
```

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -138,100 +80,13 @@ final class AppStateController : NSObject {
...
-
 }
 
-extension AppStateController : SessionManagerDelegate {
-    
-    func sessionManagerWillLogout(error: Error?, userSessionCanBeTornDown: (() -> Void)?) {
-        authenticationError = error as NSError?
-        authenticationState = .loggedOut
-        databaseEncryptionObserverToken = nil
+// TO DO: Ask why AuthenticationCoordinatorDelegate implements AuthenticationStatusProvider. For the scope of knowing the appState it is not necessary. We could get rid off all the AuthenticationStatusProvider properties.
```
marcomaddy1982, line 85: @typfel @billypchan @johnxnguyen: Do you know why `AuthenticationCoordinatorDelegate` implements `AuthenticationStatusProvider`

Wire-iOS/Sources/AppStateSessionManagerObserver.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+
+import Foundation
+import WireSyncEngine
+
+protocol AppStateSessionManagerObserverDelegate: class {
+    func sessionManagerObserver(_: AppStateSessionManagerObserver,
+                                willTransitionTo _: AppState,
+                                completion: (() -> Void)?)
+}
```
johnxnguyen, line 26: I think we should find a better name for this method, because it's not the session manager observer that is transitioning to the new app state, but rather it wants to communicate to the delegate that it calculated a new app state.

Wire-iOS/Sources/AppStateSessionManagerObserver.swift
```diff
@@ -0,0 +1,133 @@
...
+
+class AppStateSessionManagerObserver: SessionManagerDelegate {
```
johnxnguyen, line 28: Also I think this name can be improved. `AppStateSessionManagerObserver` doesn't reveal the intent of the class. I suggest something more literal such as `AppStateCalculator`.

Wire-iOS/Sources/AppStateController.swift
johnxnguyen: My guess: various information (as declared by `AuthenticationStatusProvider`) is scattered in the code base, most of it in sync engine. The `AppStateController` probably seemed like a good place to collate all that information into one place since it exists early enough and relates to the app state.

Wire-iOS/Sources/AppStateSessionManagerObserver.swift
marcomaddy1982: Right, What do you thing 

```swift
func appStateSessionManagerObserver(_: AppStateSessionManagerObserver,
                                                     didFind _: AppState,
                                                     completion: (() -> Void)?)
```
or 

```swift
func appStateSessionManagerObserver(_: AppStateSessionManagerObserver,
                                                     didCalculate _: AppState,
                                                     completion: (() -> Void)?)
```

Wire-iOS/Sources/AppStateSessionManagerObserver.swift
marcomaddy1982: I really like it

marcomaddy1982: So the the delegate method could be 

```swift
func appStateCalculator(_: AppStateCalculator,
                                didCalculate _: AppState,
                                completion: (() -> Void)?)
```

Wire-iOS/Sources/AppStateController.swift
marcomaddy1982: Honestly it seems to me a bit out of scope, cause we use none of them to calculate the the `AppState`. The idea is move the `AuthenticationCoordinatorDelegate` call backs in the `AppStateCalculator` without `AuthenticationStatusProvider` protocol properties and find a better place to collate all that information. 
What do you thinks about?

johnxnguyen: Yeah sounds good.

PR #4632, opened 2020-10-06
[Refactoring]: AppStateController Refactoring

marcomaddy1982:
## What's new in this PR?

The context of this PR is the `App Launching Routing Refactoring`. The idea is separate it in smaller chunks that are easier to review. I thought that a reasonable starting point could be refactor the `AppState` and the way we calculate it in the `AppStateController`.
Previously in the `AppStateController` you could have found a lot of variables used to calculate the current `AppState`. These variables were set in the `SessionManagerDelagate` call back methods and used in the `calculateAppState()` method to find the current `appState`. This approach is really verbose and it has produced a lot of not necessary code.
The idea behind this PR is set directly the `appState` in the `SessionManagerDelagate` call backs and get rid of all the variables used in the `calculateAppState()` and of the method itself.
Moreover after clean up the `AppStateController`, I renamed it in `AppStateCalculator`. The scope of the `AppStateCalculator` class is to inform the `AppStateController` of the new `AppState`.

It seems to works fine but It would be nice if some of you at some point could do other manual test.

P.S: Any other idea is super well accepted

### Dependencies 
sync-engine: https://github.com/wireapp/wire-ios-sync-engine/pull/1321


johnxnguyen (commented):
So now all the app state calculation is done in separate class, what is the role of the `AppStateController`?

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -18,112 +18,54 @@
...
-    fileprivate var loadingAccount : Account?
-    fileprivate var authenticationError : NSError?
-    fileprivate var isRunningTests = ProcessInfo.processInfo.isRunningTests
-    var isRunningSelfUnitTest = false
-    var databaseEncryptionObserverToken: Any? = nil
-
-    /// The state of authentication.
-    fileprivate(set) var authenticationState: AuthenticationState = .undetermined
+    // MARK - Private Set Property
+    private(set) var appStateCalculator = AppStateCalculator()
```
johnxnguyen, line 40: Can be `private let`.

marcomaddy1982: I could not use `private let` cause it is passed through dependency injection in the `SessionManager.create`. Moreover in swift it is not possible (for obvious reasons) define a property as `private(set) let`.

johnxnguyen: Ah sorry, I mean it can be `let` since it's never mutated.

marcomaddy1982: Ok we can make it public and and as a `let`. The idea was encapsulate it and protect from external set, that's why I used `private(set) let` but using `let` it is super fine as well :-D

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -18,112 +18,54 @@
...
-    }
-    
-    deinit {
-        NotificationCenter.default.removeObserver(self)
+        appStateCalculator.delegate = self
     }
+}
 
-    func calculateAppState() -> AppState {
-        guard !isRunningTests || isRunningSelfUnitTest else { return .headless }
```
johnxnguyen, line 78: Do we not need these flags anymore?

marcomaddy1982: It seems no, cause when the tests are running at the start the appStae is always `headless`, moreover all the test are passing as before.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -138,100 +80,13 @@ final class AppStateController : NSObject {
...
-
 }
 
-extension AppStateController : SessionManagerDelegate {
-    
-    func sessionManagerWillLogout(error: Error?, userSessionCanBeTornDown: (() -> Void)?) {
-        authenticationError = error as NSError?
-        authenticationState = .loggedOut
-        databaseEncryptionObserverToken = nil
+// TO DO: Ask why AuthenticationCoordinatorDelegate implements AuthenticationStatusProvider. For the scope of knowing the appState it is not necessary. We could get rid off all the AuthenticationStatusProvider properties.
```
typfel, line 85: True, it's not necessary for calculating the AppState. The `AuthenticationStatusProvider` is was creating as way to abstract the registration / login code from accessing the SessionManager directly for easier testing. 

`AppStateController` doesn't necessarily need to implement `AuthenticationStatusProvider` but then something else need to take over that responsibility.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -259,13 +114,9 @@ extension AppStateController : AuthenticationCoordinatorDelegate {
     }
 
     func userAuthenticationDidComplete(addedAccount: Bool) {
```
typfel, line 116: It kind of breaks the abstraction that this not inside the `AppStateCalculator`.

typfel (changes requested):
Good changes, I agree that the logic the is easier to follow now without all the boolean flags and don't quite see benefit of extracting the AppState calculation into a separate class. Why not keep `SessionManagerDelegate` implementation inside the `AppStateController`?

Wire-iOS/Sources/AppStateController.swift
marcomaddy1982: The other PR  (https://github.com/wireapp/wire-ios/pull/4635) will cover this part.

Wire-iOS/Sources/AppStateController.swift
marcomaddy1982: The other PR (#4635) will cover this part.

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        loadingAccount = nil
+        let appState: AppState = .unauthenticated(error: authenticationError)
+        delegate?.appStateCalculator(self,
+                                     didCalculate: appState,
+                                     completion: nil)
+    }
+        
+    func sessionManagerDidBlacklistCurrentVersion() {
```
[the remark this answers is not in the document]
marcomaddy1982: Ok, I will take a note

> Good changes, I agree that the logic the is easier to follow now without all the boolean flags and don't quite see benefit of extracting the AppState calculation into a separate class. Why not keep `SessionManagerDelegate` implementation inside the `AppStateController`?

Actually after this other PR (#4635) I think that there is no need to extracting the `AppState` calculation into a separate class. Anyway I would rename the `AppStateController` in `AppStateCalculator`. To do this I should merge the (#4635) into this.

marcomaddy1982:
> Good changes, I agree that the logic the is easier to follow now without all the boolean flags and don't quite see benefit of extracting the AppState calculation into a separate class. Why not keep `SessionManagerDelegate` implementation inside the `AppStateController`?

Actually after this other PR (#4635) I think that there is no need to extracting the `AppState` calculation into a separate class. Anyway I would rename the `AppStateController` in `AppStateCalculator`. To do this I should merge the (#4635) into this.

Wire-iOS/Sources/AppStateCalculator.swift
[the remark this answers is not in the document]
marcomaddy1982: We should take a note to this. But for now I guess we should keep the stuffs in the UI as they are. What do you think about?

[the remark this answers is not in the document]
typfel: Yes, we can keep things are as they are in the UI but only if we make the change I suggested in sync engine.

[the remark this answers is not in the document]
marcomaddy1982: Ok should we change the sync engine before merging this PR?

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -138,100 +80,13 @@ final class AppStateController : NSObject {
...
-            self?.isDatabaseLocked = isDatabaseLocked
-            self?.updateAppState()
-        })
-    }
-    
-}
-
-extension AppStateController {
-    
-    @objc func applicationDidBecomeActive() {
```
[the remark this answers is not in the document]
marcomaddy1982: Done

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    // MARK - Private Property
+    private var loadingAccount : Account?
+    private var databaseEncryptionObserverToken: Any? = nil
+    
+    // MARK - Private Implemetation
+    @objc
+    private func applicationDidBecomeActive() {
+        delegate?.appStateCalculator(self,
+                                     didCalculate: .headless,
```
typfel, line 53: This not correct, we should move away from .headless when the app becomes active.

typfel: I guess is hard to do without keeping the current appState in a variable or something.

Wire-iOS/Sources/AppStateCalculator.swift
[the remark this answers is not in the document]
typfel: Why not do it now so that you don't forget about it, I think it will be a fairly small change.

Wire-iOS/Sources/AppStateCalculator.swift
marcomaddy1982: so if I understood correctly we should keep the lastAppState and in the `applicationDidBecomeActive()` propagate it. Am I right?

Wire-iOS/Sources/AppStateCalculator.swift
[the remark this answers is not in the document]
marcomaddy1982: Of course.

Wire-iOS/Sources/AppStateCalculator.swift
typfel: Yes

typfel: But you also need make sure that we we don't propagate any app state changes before the application has become active.

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -0,0 +1,177 @@
...
+//
+
+import Foundation
+import WireSyncEngine
+
+
+private let zmLog = ZMSLog(tag: "AppState")
+
+extension AppStateCalculator {
+    static let appStateDidTransit = Notification.Name(rawValue: "AppStateDidTransit")
```
typfel, line 26: Was this intentionally renamed from `appStateDidTransition`?

Wire-iOS/Sources/AppStateCalculator.swift
marcomaddy1982: I'm a bit lost now. So which is your suggestion?

Wire-iOS/Sources/AppStateCalculator.swift
marcomaddy1982: Yes, I guess now reading again it sound pretty bad. Any suggestion?

typfel: Change is back to `appStateDidTransition`? That what would be inline with how the methods are named.

Wire-iOS/Sources/AppStateCalculator.swift
marcomaddy1982: Done

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -0,0 +1,183 @@
...
+        }
+        propagate(appState: appState, completion: completion)
+    }
+    
+    private func propagate(appState: AppState, completion: (() -> Void)? = nil) {
+        ZMSLog(tag: "AppState").debug("transitioning to app state: \(appState)")
+        delegate?.appStateCalculator(self, didCalculate: appState, completion: {
+            completion?()
+        })
+    }
```
johnxnguyen, line 77: A couple things: 

1. It would be cleaner to remove the replace the stored property `forceTransition` with an argument to `transition`. It's one less mutable state to maintain. 
2. It's possible for `propagate` to be called twice, if the `appState` is different and the transition is force. This might have unintended consequences.
3. `forceTransition` seems to really mean "force transition propagation", because we update the appState regardless. I think we can express intent more clearly by propagating the appState every time we update it's value, and only update if it changes or is forced. This also allows you to remove some code duplication.

```swift
private func transition(to appState: AppState, force: Bool = false, completion: ...) {
  guard self.appState != appState || force else {
    completion?()
    return
  }
  
  self.appState = appState
  delegate?.appStateCalculator(self, didCalculate: appState, completion: ...)
}

```

marcomaddy1982: I really like it. Thanks. Much better in this manner.

marcomaddy1982: Btw in your suggestion it is not taking in consideration the `previousAppState.

marcomaddy1982: Done

typfel (approved):

PR #4634, opened 2020-10-06
[Fix] Accept 2 file extensions

KaterinaWire:
### Issues
When you sent a backup file through the email, the Gmail app changes the file extension.(xxx.**ios_wbu** to xxx.**ios-wbu**).

### Causes
There is the Gmail app issue.
https://support.google.com/mail/thread/15716995?hl=en

### Solutions
To be able to accept both file extensions.

### Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13821


marcomaddy1982 (approved):

Wire-iOS/Sources/Authentication/Backup/BackupRestoreController.swift
```diff
@@ -30,6 +30,7 @@ protocol BackupRestoreControllerDelegate: class {
 
 final class BackupRestoreController: NSObject {
     static let WireBackupUTI = "com.wire.backup-ios"
+    static let WireBackupAdditionalUTI = "com.wire.backup-ios-additional"
```
billypchan, line 33: We can create a static array like 
```suggestion
    static let WireBackupUTIs = ["com.wire.backup-ios", "com.wire.backup-ios-additional"]
```

johnxnguyen (changes requested):
I think we can improve the naming here to avoid confusion in the future. Are we able to modify the existing UTI (not the extension, just the naming)? Something like `com.wire.backup-ios-underscore` and `com.wire.backup-ios-hyphen`. I think it would also need a comment to explain why we have these two types.

KaterinaWire:
> I think we can improve the naming here to avoid confusion in the future. Are we able to modify the existing UTI (not the extension, just the naming)? Something like `com.wire.backup-ios-underscore` and `com.wire.backup-ios-hyphen`. I think it would also need a comment to explain why we have these two types.

@johnxnguyen Yes, it's possible to modify the  existing UTI name. I've made it and added a comment.

johnxnguyen (approved):

Cartfile.resolved
```diff
@@ -22,7 +22,7 @@ github "wireapp/wire-ios-link-preview" "27.1.2"
 github "wireapp/wire-ios-protos" "24.0.0"
 github "wireapp/wire-ios-request-strategy" "206.0.0"
 github "wireapp/wire-ios-share-engine" "191.0.0"
-github "wireapp/wire-ios-sync-engine" "339.1.1"
+github "wireapp/wire-ios-sync-engine" "fix/accept-a-second-file-extension-for-backup"
```
billypchan, line 25: This needs to be restored before merging.

billypchan (changes requested):

Cartfile.resolved
KaterinaWire: Done

billypchan (approved):

PR #4635, opened 2020-10-07
[Feature]: AuthenticationCoordinatorDelegate and AuthenticationStatusProvider Refactoring

marcomaddy1982:
## What's new in this PR?

IMHO it seems to me that the `AuthenticationCoordinatorDelegate` shouldn't implement the `AuthenticationStatusProvider` cause we use none of the property contained in the `AuthenticationStatusProvider` to calculate the the `AppState`. 
Moreover cause now it is the `AppStateCalculator` that find the `AppState` the idea is move there the `AuthenticationCoordinatorDelegate` call back.

Anyway after moving out the `AuthenticationStatusProvider` protocol from the `AuthenticationCoordinatorDelegate`, we need another way to instantiate at the app launch the information contained in the `AuthenticationStatusProvider`. The idea is adding a default implementation of the protocol and create a class that implement it (`AuthenticationStatusProviderObject`)

P.s I know that the `AuthenticationStatusProviderObject` is not a really good naming, so please help is you have better ideas.


Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -56,9 +56,7 @@ class AuthenticationCoordinator: NSObject, AuthenticationEventResponderChainDele
     let eventResponderChain: AuthenticationEventResponderChain
 
     /// Shortcut for accessing the authentication status provider (returns the delegate).
-    var statusProvider: AuthenticationStatusProvider? {
-        return delegate
-    }
+    var statusProvider: AuthenticationStatusProvider?
```
johnxnguyen, line 59: Probably doesn't need to be optional. Also please update the comment (you could even delete it).

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
```diff
@@ -23,7 +23,7 @@ import WireSyncEngine
  * Provides context about the current authentication stack.
  */
 
-protocol AuthenticationStatusProvider: class {
+public protocol AuthenticationStatusProvider: class {
```
johnxnguyen, line 26: It's public now because it's used in a public type?

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
```diff
@@ -70,3 +70,33 @@ protocol AuthenticationStatusProvider: class {
...
+    }
+
+    var authenticatedUserNeedsEmailCredentials: Bool {
+        guard let emailAddress = selfUser?.emailAddress else { return false }
+        return emailAddress.isEmpty
+    }
+
+    var sharedUserSession: ZMUserSession? {
+        return ZMUserSession.shared()
+    }
```
johnxnguyen, line 87: You could take this opportunity to use this property in the other properties.

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
```diff
@@ -70,3 +70,33 @@ protocol AuthenticationStatusProvider: class {
...
+    var selfUser: UserType? {
+        return ZMUserSession.shared()?.selfUser
+    }
+
+    var numberOfAccounts: Int {
+        return SessionManager.numberOfAccounts
+    }
+}
+
+class AuthenticationStatusProviderObject: AuthenticationStatusProvider { }
```
johnxnguyen, line 102: I suppose we could just convert `AuthenticationStatusProvider` into a class, then creating a mock subclass. I don't see any real benefit to keeping it a protocol anymore.

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
marcomaddy1982: thanks for pointing it out. I added but forgot to remove it. Thanks a lot. I remove the `public` access modifier immediately.

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
marcomaddy1982: Yep you are right. Thanks

Wire-iOS/Sources/Authentication/Delegates/AuthenticationStatusProvider.swift
marcomaddy1982: Good point. Let me try :-)

marcomaddy1982: Done, please have a look once you have 2 mins if it is that what you meant.

Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -55,18 +55,16 @@ class AuthenticationCoordinator: NSObject, AuthenticationEventResponderChainDele
...
     // MARK: - State
 
     /// The displayed view controller.
     var currentViewController: AuthenticationStepViewController?
 
     /// The object controlling the state of authentication.
     let stateController: AuthenticationStateController
+    
+    /// Shortcut for accessing the authentication status provider.
+    var statusProvider: AuthenticationStatusProvider?
```
johnxnguyen, line 67: Still this doesn't make much sense to keep optional since we always provide a default value. But it does make sense to keep it mutable since we may need to set a mock.

johnxnguyen (commented):
Was the status provider ever mocked in tests? We may need to update the tests to fit the new API.

Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
marcomaddy1982: Done.

marcomaddy1982:
> Was the status provider ever mocked in tests? We may need to update the tests to fit the new API.

I can't find any place where it is user the `statusProvider` in the tests, am I missing something?

Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -108,10 +122,13 @@ class AuthenticationCoordinator: NSObject, AuthenticationEventResponderChainDele
...
 
     /// Creates a new authentication coordinator with the required supporting objects.
-    init(presenter: UINavigationController & SpinnerCapable, sessionManager: ObservableSessionManager, featureProvider: AuthenticationFeatureProvider) {
+    init(presenter: UINavigationController & SpinnerCapable,
+         sessionManager: ObservableSessionManager,
+         featureProvider: AuthenticationFeatureProvider) {
         self.presenter = presenter
         self.sessionManager = sessionManager
         self.stateController = AuthenticationStateController()
+        self.statusProvider = AuthenticationStatusProvider()
```
typfel, line 131: Let's inject it through the initialiser like we do for the `AuthenticationFeatureProvider`? That way we can make the `statusProvider` property a `let` and still replace it in the tests.

typfel (changes requested):

Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
marcomaddy1982: Done

typfel (approved):

PR #4645, opened 2020-10-09
[Refactoring]: AppRootViewController Transition Animation Refactoring

marcomaddy1982:
# Context

This PR will try to refactor some parts of the `App Launch Routing`:

## 1 - App Root ViewController Initialisation
Currently the Application root view controller of the window is created and set programmatically in the `AppRootViewController`. I found this way really verbose and "archaic". A much easier and more diffused way is set the rootViewController of the application in the `Main.storyborad`. As you can notice the initial view controller of the `Main.storyboard` is type of `RootViewController`. 

## 2 - Application Transition Animation After AppState Changes
Currently the application animation transition are performed in the `AppRootViewController`, which is pretty messy and difficult to maintain. The idea is trying to decouple it till we can get rid of it. In this PR I extracted from it:
- The part of code that perform the transition animation when the `AppState` moving it in the  `RootViewController`
- The navigation logic moving it in the `AppRootRouter`.

## 3 - Moreover
The `Navigator` class it is added to the project. We already discussed it in the past iOS Collectives. Currently the `Navigator` class is not used but it will. So I thought it makes sense to add it and you can try to review it as well. (Btw you could find an example of how it is used here: https://github.com/marcomaddy1982/NavigationStack)

P.S: You will find a lot of commented code cause this PR is a little chunk of the App Lunch Refactoring.  You will also noticed that instead of the tabbar controller after the login you will find a red screen. Don't worry :-D. Step by Step we will arrive to show it again. Please focus your review mostly on the "AppRootRouter" for the navigation and the "RootViewController" for the animation transitions.


Wire-iOS/Sources/Navigator.swift
```diff
@@ -0,0 +1,149 @@
...
+    public func pop(_ animated: Bool) {
+        let vc = navigationController.popViewController(animated: animated)
+        vc.flatMap { runCompletion(for: $0) }
+    }
+
+    public func dismiss(_ viewController: UIViewController?, animated: Bool) {
+        guard let viewController = viewController else { return }
+        runCompletion(for: viewController)
+        viewController.dismiss(animated: animated, completion: nil)
+    }
```
johnxnguyen, line 83: What's the use case for passing in `nil`?

Wire-iOS/Sources/Navigator.swift
```diff
@@ -0,0 +1,149 @@
...
+
+    public func present(_ viewController: UIViewController, animated: Bool, onComplete: (() -> Void)?) {
+        navigationController.present(viewController, animated: animated, completion: onComplete)
+    }
+
+    public func root(_ viewController: UIViewController, animated: Bool) {
+        closures.forEach { $0.value() }
+        closures = [:]
+        navigationController.viewControllers = [viewController]
+    }
```
johnxnguyen, line 93: I think a better name for this is `setRoot(_ viewController: UIViewController, animated: Bool)`.

Wire-iOS/Sources/Navigator.swift
```diff
@@ -0,0 +1,149 @@
...
+        shouldHideBackButton(for: viewController)
+    }
+
+    public override func setViewControllers(_ viewControllers: [UIViewController], animated: Bool) {
+        super.setViewControllers(viewControllers, animated: animated)
+        viewControllers.forEach(shouldHideBackButton(for:))
+    }
+
+    private func shouldHideBackButton(for viewController: UIViewController) {
+        viewController.navigationItem.backBarButtonItem = UIBarButtonItem(title: "",
```
johnxnguyen, line 144: A better name for this is `hideBackButton(for viewController: UIViewController)`.

Wire-iOS/Sources/Navigator.swift
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        super.pushViewController(viewController, animated: animated)
+        shouldHideBackButton(for: viewController)
+    }
+
+    public override func setViewControllers(_ viewControllers: [UIViewController], animated: Bool) {
+        super.setViewControllers(viewControllers, animated: animated)
+        viewControllers.forEach(shouldHideBackButton(for:))
```
johnxnguyen, line 140: Do you need to hide the back button here and when `viewControlers` is set?

Wire-iOS/Sources/NotificationObserver.swift
```diff
@@ -0,0 +1,92 @@
...
+            observer.applicationDidBecomeActive()
+        })
+
+        addObserverToken(NotificationCenter.default.addObserver(forName: UIApplication.willEnterForegroundNotification,
+                                                                object: nil,
+                                                                queue: nil) { [weak self] _ in
+            guard let observer = self else { return }
+            observer.applicationWillEnterForeground()
+        })
+    }
```
johnxnguyen, line 55: There are some opportunities to remove code duplication here.

Wire-iOS/Sources/Navigator.swift
marcomaddy1982: This is the idea, but at the same time this is just an helper class, it could be that we will remove the class at the end of the refactoring.

johnxnguyen: I mean, it looks like we're hiding the back button repeatedly, in this this method and also when the property `viewControllers` is set.

marcomaddy1982: I got it now. Yes I can remove this one. Thanks

Wire-iOS/Sources/Navigator.swift
marcomaddy1982: Done

Wire-iOS/Sources/Navigator.swift
marcomaddy1982: What do you mean?

marcomaddy1982: Probably is the name `runCompletion` that confuse you?

marcomaddy1982: probably you mean that it is better `runCompletion` in the `completion` method. I guess you are right.

Wire-iOS/Sources/RootViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+    func removeChild(_ viewController: UIViewController?) {
+        viewController?.willMove(toParent: nil)
+        viewController?.view.removeFromSuperview()
+        viewController?.removeFromParent()
+    }
+
+    func childViewController<T: UIViewController>(with type: T.Type) -> T? {
+        return children.compactMap { $0 as? T }.first
+            ?? children.compactMap { $0.childViewController(with: type).flatMap { $0 } }.first
+    }
```
johnxnguyen, line 149: This will visit every single view controller in the hierarchy. Maybe it's worth to stop searching once we find a child of the desired type.

johnxnguyen (commented):
In general looks good to me. I'm wondering though how I should consider the commented out code? Is this intended to be addressed in separate PRs?

Wire-iOS/Sources/Navigator.swift
```diff
@@ -0,0 +1,145 @@
...
+public final class NoBackTitleNavigationController: UINavigationController {
+    public override var viewControllers: [UIViewController] {
+        didSet {
+            viewControllers.forEach(hideBackButton(for:))
+        }
+    }
+
+    public override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        super.pushViewController(viewController, animated: animated)
+        hideBackButton(for: viewController)
```
johnxnguyen, line 136: Same here... do we need to hide the back button explicitly or will the `didSet` on `viewControllers` handle this case?

Wire-iOS/Sources/Navigator.swift
johnxnguyen: I mean when would I ever call `dismiss(nil, animated: ...)`?

johnxnguyen: Yes, maybe it also makes sense to call `runCompletion` in the when `viewController.dismiss` completes.

marcomaddy1982:
> In general looks good to me. I'm wondering though how I should consider the commented out code? Is this intended to be addressed in separate PRs?

Yes of course, there are too stuffs to consider and it is impossible to have a unique PR. If you notice also the snapshot tests are failing. We should add piece by pieces, also cause sometime I have to understand what is the duty of a certain part of code and mostly  it is not immediate.

Wire-iOS/Sources/Navigator.swift
marcomaddy1982: In this case I guess we have to hide it explicitly, cause I think that the set `viewControllers` is not called when we push a new view controller. Or am I wrong?

johnxnguyen: I don't know, probably not hard to find out though.

marcomaddy1982: I was checking that right now!!! Actually the `didSet` of the `override viewControllers` is NOT called here 

```swift
public override func setViewControllers(_ viewControllers: [UIViewController], animated: Bool { }
``` 

and here

```swift
public override func pushViewController(_ viewController: UIViewController, animated: Bool) { }
``` 

So I guess if should be added again in both the methods.

marcomaddy1982: I found out debugging this project  https://github.com/marcomaddy1982/NavigationStac

johnxnguyen:
> Yes of course, there are too stuffs to consider and it is impossible to have a unique PR. If you notice also the snapshot tests are failing. We should add piece by pieces, also cause sometime I have to understand what is the duty of a certain part of code and mostly it is not immediate.

@marcomaddy1982 yep small and concise PRs are the way to go, but it's also important to try structure PRs to contain a single logical change. It's much easier to review because the reviewer doesn't need to filter what is considered important and what is a temporary stash of changes. When not everything in the PR is review with the same standard then some things slip in to production that shouldn't have. It's not always an easy thing to do but it's worth the effort I think.

For example, I think you could have opened a PR for each new class defined, then one or more PRs to integrate them with the existing code.

Wire-iOS/Sources/Navigator.swift
johnxnguyen: Ah ok good to know! 👍

marcomaddy1982:
> > Yes of course, there are too stuffs to consider and it is impossible to have a unique PR. If you notice also the snapshot tests are failing. We should add piece by pieces, also cause sometime I have to understand what is the duty of a certain part of code and mostly it is not immediate.
> 
> @marcomaddy1982 yep small and concise PRs are the way to go, but it's also important to try structure PRs to contain a single logical change. It's much easier to review because the reviewer doesn't need to filter what is considered important and what is a temporary stash of changes. When not everything in the PR is review with the same standard then some things slip in to production that shouldn't have. It's not always an easy thing to do but it's worth the effort I think.
> 
> For example, I think you could have opened a PR for each new class defined, then one or more PRs to integrate them with the existing code.

@johnxnguyen I understand your point. Probably adding the `Navigator` part was wrong in this PR cause it is still not used. I was confident that there were no "hurt" introducing it cause we discuss it another couple of times during the collective.
I also understand point regarding the temporary stash of changes, unfortunately it is not easy don't have temporary commented code or temporary behaviour of the app during this refactoring if we wanna preceded with small refactoring chunks (at least for now).

marcomaddy1982:
> > > Yes of course, there are too stuffs to consider and it is impossible to have a unique PR. If you notice also the snapshot tests are failing. We should add piece by pieces, also cause sometime I have to understand what is the duty of a certain part of code and mostly it is not immediate.
> > 
> > 
> > @marcomaddy1982 yep small and concise PRs are the way to go, but it's also important to try structure PRs to contain a single logical change. It's much easier to review because the reviewer doesn't need to filter what is considered important and what is a temporary stash of changes. When not everything in the PR is review with the same standard then some things slip in to production that shouldn't have. It's not always an easy thing to do but it's worth the effort I think.
> > For example, I think you could have opened a PR for each new class defined, then one or more PRs to integrate them with the existing code.
> 
> @johnxnguyen I understand your point. Probably adding the `Navigator` part was wrong in this PR cause it is still not used. I was confident that there were no "hurt" introducing it cause we discuss it another couple of times during the collective.
> I also understand point regarding the temporary stash of changes, unfortunately it is not easy don't have temporary commented code or temporary behaviour of the app during this refactoring if we wanna preceded with small refactoring chunks (at least for now).

@johnxnguyen btw I'm going to remove from this PR some little stuff to make it lighter and open separated PRs

Wire-iOS/Sources/RootViewController.swift
marcomaddy1982: I tried to clean up this using the forEach but I found them really verbose. Honestly cause we are not looping big data array so I would stay with something more readable. Are you agree or did you have in mind a different way to write this method?

johnxnguyen: Here's a simple lazy BFS search algorithm. Is this similar to what you had?

```swift
/// Lazy breadth first search.
func firstChild<T: UIViewController>(ofType type: T.Type) -> T? {
    // Check all the children first.
    for child in children {
      if let result = child as? T {
        return result
      }
    }

    // Then check next layer down.
    for child in children {
      if let result = child.firstChild(ofType: type) {
        return result
      }
    }

    return nil
  }
```

One other thing came to mind: this method in general will return the first child that can be casted to `T`, not the first child that is exactly of type `T`. This means that we can get a child as a superclass. Not sure if that is a problem?

marcomaddy1982: Actually the method was exactly the same, I just used children.forEach. 

 > This means that we can get a child as a superclass. Not sure if that is a problem? 
Good point, we can leave in this manner I guess.

marcomaddy1982: Done

johnxnguyen (approved):

PR #4648, opened 2020-10-12
[Refactoring]: Decouple Responsibility From AppRootRouter

marcomaddy1982:
## What's new in this PR?

In this PR I tried to decouple some of the responsibility of the `AppRootRouter` creating the `SessionManagerLifeCycleObserver` and the `ForegroundNotificationDecisioner` classes. 
Indeed before the call backs of the  `SessionManagerCreatedSessionObserver` and `SessionManagerDestroyedSessionObserver` were implemented in the `AppRootViewController`. Same situation for the  `ForegroundNotificationOberver` method. With the introduction of the `AppRootRouter`, which will take care just of the navigation, I thought that was more logic separate that call backs in separated classes.


Wire-iOS/Sources/SessionManagerLifeCycleObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    func sessionManagerCreated(unauthenticatedSession: UnauthenticatedSession) { }
+
+    // MARK: - SessionManagerDestroyedSessionObserver
+    func sessionManagerDestroyedUserSession(for accountId: UUID) {
+        resetSoundEventLister(for: accountId)
+    }
+    
+    // MARK: - Private Implementation
+    private func setSoundLister(for userSession: ZMUserSession) {
```
johnxnguyen, line 40: Typo: `setSoundListener` or better yet `setSoundEventListener`.

Wire-iOS/Sources/SessionManagerLifeCycleObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+            SecurityFlags.forceEncryptionAtRest.isEnabled,
+            userSession.encryptMessagesAtRest == false
+        else {
+            return
+        }
+        
+        userSession.encryptMessagesAtRest = true
+    }
+    
+    private func resetSoundEventLister(for accountID: UUID) {
```
johnxnguyen, line 59: Typo also here.

Wire-iOS/Sources/ForegroundNotificationDecisioner.swift
```diff
@@ -0,0 +1,61 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import WireSyncEngine
+
+//TO DO: Ask for the logic, not clear when a notification shuld be presented
+class ForegroundNotificationDecisioner: ForegroundNotificationResponder {
```
johnxnguyen, line 22: I suggest a name such as `ForegroundNotificationFilter`.

Also, what do you want to know about this logic?

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -27,6 +27,10 @@ public class AppRootRouter: NSObject {
     private let navigator: NavigatorProtocol
     private var appStateCalculator = AppStateCalculator()
     private var authenticationCoordinator: AuthenticationCoordinator?
+    private var sessionManagerCreatedSessionObserverToken: Any?
+    private var sessionManagerDestroyedSessionObserverToken: Any?
```
johnxnguyen, line 31: I don't think there's a need to retain the tokens separately, you could put them in a single `observerTokens` array.

Wire-iOS/Sources/ForegroundNotificationDecisioner.swift
marcomaddy1982: I actually don't understand this part, cause the it is not logic. 

```swift
if clientVC.isConversationListVisible {
            return false
        }
        
guard clientVC.isConversationViewVisible else {
           return true
}
```

Moreover I would like to know if there a ticket or something that describe when a notification should be shown, so in case I could try to write a bit cleaner that code.

johnxnguyen: > cause the it is not logic.

Not sure what you mean here. But we can step through the conditions:

```swift
if clientVC.isConversationListVisible {
     return false
}
```

If the conversation list is visible, then we don't want to display a user notification banner because that information will be visible from the list.

```swift
guard clientVC.isConversationViewVisible else {
    return true
}
```

The rest of the conditions depend on the conversation view being visible, so we assert it here. If the conversation view isn't visible, then it means we must be somewhere else in the app, in which case we should show the banner.

Afterwards we check that the visible conversation view is for a different conversation than the one the new message notification originates from. There's no point showing the new message banner for a conversation if that conversation is open and visible to the user.

This whole method is poorly expressed (I think I wrote it a few years ago) because of the cascading conditions. Perhaps it would be nicer to delegate the decision to view controllers much in the same way the preferred status bar style is delegate to child view controllers.

As far as I know there's no documentation for this. The general principle though is that the in app notification should only be shown when it's not redundant because some existing and visible UI is conveying the information.

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -27,6 +27,9 @@ public class AppRootRouter: NSObject {
     private let navigator: NavigatorProtocol
     private var appStateCalculator = AppStateCalculator()
     private var authenticationCoordinator: AuthenticationCoordinator?
+    private var observerTokens: [Any?] = []
```
johnxnguyen, line 30: Can be `[Any]`.

Wire-iOS/Sources/ForegroundNotificationDecisioner.swift
marcomaddy1982: ok, now it is a bit clearer. Thanks.

marcomaddy1982: > Perhaps it would be nicer to delegate the decision to view controllers much in the same way the preferred status bar style is delegate to child view controllers.

About this comment, it is a good idea. the scope of for now is trying to refactoring the App Launch. Let's focus on that for now and later we could clean this kind of stuffs. 
Are you agreed?

johnxnguyen: Yeah of course, was just an idea for the future 😄

johnxnguyen (commented):
Looks good, but you need to fix the tests.

marcomaddy1982:
> Looks good, but you need to fix the tests.

This is not so simple, the part of the tests is still not refactored. I mean, I guess it is the part regarding the calls and video calls and there is nothing in the AppRouter regarding calls and video calls.

johnxnguyen:
> This is not so simple, the part of the tests is still not refactored. I mean, I guess it is the part regarding the calls and video calls and there is nothing in the AppRouter regarding calls and video calls.

There are a lot of failed snapshot tests. I guess this is because those tests can't render their view properly yet.

johnxnguyen (approved):

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -74,22 +78,34 @@ public class AppRootRouter: NSObject {
...
-                    ? true
-                    : Settings.shared[.callingConstantBitRate] ?? false
-                sessionManager.useConferenceCalling = true
+
+                self.setCallingSettting(for: sessionManager)
                 sessionManager.start(launchOptions: launchOptions)
         }
     }
+    
+    private func createLifeCycleObeserverTokens(for sessionManager: SessionManager) {
```
typfel, line 94: Doesn't it make more sense for this method to live inside the `SessionManagerLifeCycleObserver`?

typfel, line 94: typo `createLifeCycleObeserverTokens` > `createLifeCycleObserverTokens `

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -74,22 +78,34 @@ public class AppRootRouter: NSObject {
...
+        let createdSessionObserverToken = sessionManager.addSessionManagerCreatedSessionObserver(sessionManagerLifeCycleObserver)
+        observerTokens.append(createdSessionObserverToken)
+                    
+        let destroyedSessionObserverToken = sessionManager.addSessionManagerDestroyedSessionObserver(sessionManagerLifeCycleObserver)
+        observerTokens.append(destroyedSessionObserverToken)
+    }
+    
+    private func setCallingSettting(for sessionManager: SessionManager) {
```
typfel, line 102: typo: `setCallingSettting` > `setCallingSetting`

also I think it make more sense to use plural here since we are setting multiple options: `setCallingSetttings(for:)`

Wire-iOS/Sources/AppRootRouter.swift
marcomaddy1982: Yep, you are right. I will address the change in the next PR. This is already merged :-)

Wire-iOS/Sources/AppRootRouter.swift
marcomaddy1982: Yep, you are right. I will address the change in the next PR. This is already merged :-)

Wire-iOS/Sources/AppRootRouter.swift
marcomaddy1982: same here, I will apply the changes to the new PR

marcomaddy1982: same here, I will apply the changes to the new PR

David-Henner (commented):
Thanks :)

marcomaddy1982:
> Thanks :)

You are welcome!!! :-)

PR #4649, opened 2020-10-13
[Fix] Add new app state

KaterinaWire:
### Issues
When the applock is enable and client accepts a call through CallKit, the client has no visual elements about the current call.

### Solutions
As part of solving the current issue, add a new app state `locked`.

### Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13792


Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -295,11 +295,16 @@ final class AppRootViewController: UIViewController, SpinnerCapable {
                 Analytics.shared.selfUser = SelfUser.current
 
                 viewController = clientViewController
+                if let pendingCallKitCall = SessionManager.shared?.callKitManager?.pendingCallKitCall {
+                    pendingCallKitCall()
+                }
```
johnxnguyen, line 300: Do we need to switch to the account for that pending call? Or is this logic already present in session manager?

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -130,6 +131,7 @@ extension AppLockPresenter {
     }
     
     private func requestAccountPassword(with message: String) {
+        NotificationCenter.default.post(name: .appLocked, object: self, userInfo: nil)
```
johnxnguyen, line 134: So this is triggering the change of app state to `.appLocked`. I would expect it to be the other way around, that when we transition to `.appLocked` then we display the app lock ui.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -229,6 +247,18 @@ extension AppStateController {
         updateAppState()
     }
     
+    @objc func applicationDidBecomeLocked() {
```
billypchan, line 250:
```suggestion
    @objc 
    private func applicationDidBecomeLocked() {
```

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: This was changed as we discussed during the call.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -263,6 +278,20 @@ extension AppStateController : AuthenticationCoordinatorDelegate {
...
+    }
+    
+    private var isLockTimeoutReached: Bool {
+        let lastAuthDate = AppLock.lastUnlockedDate
+        
+        // The app was authenticated at least N seconds ago
+        let timeSinceAuth = -lastAuthDate.timeIntervalSinceNow
+        let isWithinTimeoutWindow = (0..<Double(AppLock.rules.appLockTimeout)).contains(timeSinceAuth)
+        return !isWithinTimeoutWindow
+    }
```
johnxnguyen, line 294: I suggest to move this to a senate object and extract a protocol from it. This will improve your tests.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -65,7 +66,15 @@ final class AppStateController : NSObject {
...
-        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
+        NotificationCenter.default.addObserver(self,
+            selector: #selector(applicationDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil)
+        
+        NotificationCenter.default.addObserver(self,
+        selector: #selector(applicationDidBecomeUnlocked),
+        name: .appUnlocked,
+        object: .none)
```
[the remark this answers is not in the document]
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -61,12 +60,6 @@ extension AppLockInteractor: AppLockInteractorInput {
...
     
-    var isAuthenticationNeeded: Bool {
-        let screenLockIsActive = appLock.isActive && isLockTimeoutReached && isAppStateAuthenticated
-        
-        return screenLockIsActive || isDatabaseLocked
-    }
-    
     var isDimmingScreenWhenInactive: Bool {
         return AppLock.isActive || userSession?.encryptMessagesAtRest == true
     }
```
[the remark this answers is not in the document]
KaterinaWire: No.
Removed

Wire-iOS/Sources/AppState.swift
```diff
@@ -22,6 +22,7 @@ import WireDataModel
 enum AppState : Equatable {
     
     case headless
+    case appLocked
```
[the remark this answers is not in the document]
KaterinaWire: Done

Wire-iOS/Sources/AppRootViewController.swift
KaterinaWire: Ok, I will move it to the next PR

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: I'm not 100% sure what you meant, but I moved this logic into a separate protocol. You can check it in my last commit.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -28,7 +29,30 @@ protocol AppStateControllerDelegate : class {
...
+        let lastAuthDate = AppLock.lastUnlockedDate
+        
+        // The app was authenticated at least N seconds ago
+        let timeSinceAuth = -lastAuthDate.timeIntervalSinceNow
+        let isWithinTimeoutWindow = (0..<Double(AppLock.rules.appLockTimeout)).contains(timeSinceAuth)
+        return !isWithinTimeoutWindow
+    }
+}
+
+final class AppStateController : NSObject, ScreenLockHelperProtocol {
```
johnxnguyen, line 55: I meant to move this code into some concrete type, abstract it with a protocol and refer to the protocol in `AppStateController`. So it would look something like:

```swift
protocol AppLockTimerProtocol {
  var shouldLockApp: Bool { get }
}

class AppStateController {

  // This should be injectable, so when testing we can provide a different implementation.
  let appLockTimer: AppLockTimerProtocol
  
  func calculateAppState() {
    ...
    switch authenticationState {
    case .loggedIn where isDatabaseLocked || appLockTimer.shouldLockApp:
     appState = .locked
      ...
    }
  }
}
```

You could then provide a test implementation that can return true of false as you wish (no time calculation at all) and inject that when testing.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -206,58 +195,23 @@ extension AppLockPresenter {
...
     @objc func applicationWillResignActive() {
-        if appLockInteractorInput.isDimmingScreenWhenInactive {
-            userInterface?.setContents(dimmed: true)
-        }
+        showReauth(visible: false)
     }
     
     @objc func applicationDidEnterBackground() {
         if self.authenticationState == .authenticated {
             AppLock.lastUnlockedDate = Date()
```
typfel, line 206: Don't we need to move this somewhere else since the `AppLockPresenter` will not be instantiated when the app is unlocked and the app enters the background.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -23,6 +23,7 @@ import WireSyncEngine
 
 extension Notification.Name {
     static let appUnlocked = Notification.Name("AppUnlocked")
+    static let appLocked = Notification.Name("AppLocked")
```
typfel, line 26: NOTE: Can't comment on correct line so I'm commenting here.

isn't the `AuthenticationState.authenticated` state unnecessary now since we won't show the AppLock at all when authenticated?

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.swift
```diff
@@ -19,30 +19,9 @@
 import UIKit
 
 final class NotificationWindowRootViewController: UIViewController {
```
typfel, line 21: This class does nothing now?

typfel (changes requested):

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.swift
KaterinaWire: Seems like yes, but removing this class requires removing instances from other parts of the code. I would prefer to do it in another PR.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: Yes.
I moved this observer to the `AppStateController`

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: Yes, we don't need it now

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -61,16 +59,6 @@ extension AppLockInteractor: AppLockInteractorInput {
         return AppLock.isCustomPasscodeNotSet
     }
     
-    var isAuthenticationNeeded: Bool {
```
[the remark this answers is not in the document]
KaterinaWire: Yes, we don't need it anymore.

Wire-iOS Tests/AppLock/AppLockViewControllerSnapshotTests.swift
```diff
@@ -35,7 +35,6 @@ final class AppLockViewControllerSnapshotTests: XCTestCase {
     }
     
     func testDimmedState() {
-        sut.setContents(dimmed: true)
         verify(matching: sut)
     }
```
johnxnguyen, line 39: Is this test needed anymore?

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -28,6 +29,10 @@ protocol AppStateControllerDelegate : class {
     
 }
 
+protocol AppLockProtocol : class {
+    var shouldLockScreen: Bool { get }
+}
+
```
johnxnguyen, line 35: Better to move this to a separate file so as not to clutter up this one.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -262,7 +298,24 @@ extension AppStateController : AuthenticationCoordinatorDelegate {
...
+    }
+    
+    private var isLockTimeoutReached: Bool {
+        let lastAuthDate = AppLock.lastUnlockedDate
+        
+        // The app was authenticated at least N seconds ago
+        let timeSinceAuth = -lastAuthDate.timeIntervalSinceNow
+        let isWithinTimeoutWindow = (0..<Double(AppLock.rules.appLockTimeout)).contains(timeSinceAuth)
+        return !isWithinTimeoutWindow
+    }
```
johnxnguyen, line 318: It's better to create another type to implement this protocol, since the point is to able to inject a mock implementation to make testing easier.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -59,15 +64,30 @@ final class AppStateController : NSObject {
...
+                                               name: .appUnlocked,
+                                               object: .none)
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(applicationDidEnterBackground),
+                                               name: UIApplication.didEnterBackgroundNotification,
+                                               object: .none)
         
         appState = calculateAppState()
+        appLock = self
```
johnxnguyen, line 90: There's not much sense for `self.appLock` to hold a reference to itself. This value should be injectable (for tests), but it we can of course initialize a default value.

Wire-iOS/Sources/AppStateController.swift
johnxnguyen: The name could be improved by describing the behavior of the abstraction. If this is an abstraction over a timer object that tells us whether we should lock the screen, then `AppLockTimerProtocol` would be more suitable.

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -24,36 +24,19 @@ import WireCommonComponents
 private let zmLog = ZMSLog(tag: "UI")
 
 final class AppLockViewController: UIViewController {
-    private var lockView: AppLockView!
+    private var lockView: AppLockView?
```
johnxnguyen, line 27: Why did this become optional?

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -49,7 +54,7 @@ final class AppStateController : NSObject {
     private(set) var lastAppState : AppState = .headless
     weak var delegate : AppStateControllerDelegate? = nil
     
-    fileprivate var isDatabaseLocked = false
+    internal var isDatabaseLocked = false
```
typfel, line 57: Why did this become internal? Are you modifying it in a test?

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -112,26 +98,14 @@ extension AppLockInteractor: AppLockInteractorInput {
...
-        appState = newState
-    }
 }
 
 // MARK: - Helpers
 extension AppLockInteractor {
     
     private var authenticationScenario: AppLock.AuthenticationScenario {
-        if isDatabaseLocked {
-            return .databaseLock
```
typfel, line 130: Why was the .databaseLock scenario removed? When the database is locked we can only unlock with biometric or passcode **not** the account password.

Wire-iOS/Sources/AppState.swift
```diff
@@ -22,6 +22,7 @@ import WireDataModel
 enum AppState : Equatable {
     
     case headless
+    case locked
     case authenticated(completedRegistration: Bool, databaseIsLocked: Bool)
```
typfel, line 26: I think we should remove the `databaseIsLocked` associated value or move it to the `.locked` case.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -228,6 +251,19 @@ extension AppStateController {
         hasEnteredForeground = true
         updateAppState()
     }
+
+    @objc func applicationDidBecomeUnlocked() {
+        updateAppState()
+    }
+    
+    @objc func applicationDidEnterBackground() {
```
typfel, line 259: I think when AppLock is active we always dim the screen when app will enter the background regardless if the lock timer as expired. That way you can't read any message from screenshot in the iOS task manager.

typfel (changes requested):

typfel:
Have you tested this on a real device? With and without EAR enabled?

Wire-iOS Tests/AppLock/AppLockViewControllerSnapshotTests.swift
KaterinaWire: No. I deleted it.

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: Changed the name and moved it to a separate file.

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: Removed this extension. I created a helper class `AppLockTimer` that implements `AppLockTimerProtocol`.

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: In init() of the AppStateController class I added a  default value:
appLockTimer = AppLockTimer()
(`AppLockTimer` is a class that  implements `AppLockTimerProtocol`)

Wire-iOS Tests/AppStateControllerTests.swift
```diff
@@ -116,4 +117,87 @@ final class AppStateControllerTests: XCTestCase {
...
+    }
+    
+}
+
+extension AppStateControllerTests {
+    func set(appLockActive: Bool, timeoutReached: Bool, authenticatedAppState: Bool, databaseIsLocked: Bool) {
+        AppLock.isActive = appLockActive
+        AppLock.rules = AppLockRules(useBiometricsOrAccountPassword: false, useCustomCodeInsteadOfAccountPassword: false, forceAppLock: false, appLockTimeout: 900)
+        let timeInterval = timeoutReached ? -Double(AppLock.rules.appLockTimeout)-100 : -10
+        AppLock.lastUnlockedDate = Date(timeIntervalSinceNow: timeInterval)
```
johnxnguyen, line 193: Now that you have a protocol abstraction of the app lock timer, you can define a mock timer and use that during testing. This means you don't' need any of this logic in your tests anymore.

For example:

```swift
struct MockAppLockTimer: AppLockTimerProtocol {
    var shouldLock: Bool

    var shouldLockApp: Bool {
        return shouldLock
    }
}

File: Wire-iOS/Sources/AppStateController.swift [hunk 23]
```diff
@@ -59,15 +60,30 @@ final class AppStateController : NSObject {
     fileprivate var isRunningTests = ProcessInfo.processInfo.isRunningTests
     var isRunningSelfUnitTest = false
     var databaseEncryptionObserverToken: Any? = nil
+    var appLockTimer: AppLockTimerProtocol?
```
johnxnguyen, line 63: This doesn't need to be optional since you always provide a value.

Wire-iOS Tests/AppStateControllerTests.swift
KaterinaWire: Done.
Also I added tests for `AppLockTimer` class.

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: I've made `appLockTimer` as non optional and set a default value in the definition otherwise it requires to initialize it in `super.init `.

(`var appLockTimer: AppLockTimerProtocol = AppLockTimer()`)

KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
KaterinaWire: Previously we created an instance of `AppLockPresenter` an instance of `AppLockView`. After refactoring the `AppLockPresenter`, we decided to hide or show the `lockView` in the `AppLockPresenter` initializer.
I rolled back `lockView` changes. Now it's again non-optional. But there are also changes in `AppLockViewController`:
```
1. lockView = AppLockView()
2. self.appLockPresenter = AppLockPresenter(userInterface: self)
```

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: No, I am not modifying it. These were unnecessary changes.

Wire-iOS/Sources/AppState.swift
KaterinaWire: Removed

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: @johnxnguyen @typfel To roll back this logic I added an observer to `NotificationWindowRootViewController` for `UIApplication.willResignActiveNotification` event. In this moment I create `AppLockViewController` screen.
Also in `AppRootViewController` for `ApplicationDidBecomeActive` event I hide dimmed screen by `AppDelegate.shared.notificationsWindow?.isHidden = true`

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
KaterinaWire: @johnxnguyen @typfel It has been removed together with an `appState` property from the `AppLockInteractor`.
Since we do not currently have an `appState` property, I added the `isDatabaseLocked`  parameter in the `AppLockInteractor` initializer. This parameter will be passed from the `AppStateViewController` so that ʻAppLockInteractor` knows whether the database is locked or not.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -40,14 +40,13 @@ protocol AppLockUserInterface: class {
...
     func setSpinner(animating: Bool)
-    func setContents(dimmed: Bool)
     func setReauth(visible: Bool)
 }
 
 enum AuthenticationState {
     case needed
+    case noNeeded
     case cancelled
-    case authenticated
```
johnxnguyen, line 50: I'm a bit confused what this represents and why it's no longer needed.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -40,14 +40,13 @@ protocol AppLockUserInterface: class {
     func presentCreatePasscodeScreen(callback: ResultHandler?)
     
     func setSpinner(animating: Bool)
-    func setContents(dimmed: Bool)
     func setReauth(visible: Bool)
 }
 
 enum AuthenticationState {
     case needed
+    case noNeeded
```
johnxnguyen, line 48: What is the purpose of this new case? To force the app lock view controller to never display the reauthorization ui?

johnxnguyen (changes requested):
In my opinion, reusing the `AppLockViewController` to block the screen for task switching is introducing unnecessary complexity. A better approach would be to just display a simple view in the notification window. 

The notification window seems to have no other responsibility so you could repurpose it for the sole task of hiding the app when entering the background.

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -300,6 +300,8 @@ final class AppRootViewController: UIViewController, SpinnerCapable {
             viewController = LaunchImageViewController()
         case .loading(account: let toAccount, from: let fromAccount):
             viewController = SkeletonViewController(from: fromAccount, to: toAccount)
+        case .locked:
+            viewController = AppLockViewController(isDatabaseLocked: appStateController.isDatabaseLocked)
```
johnxnguyen, line 304: I suggest to pass `isDatabaseLocked` as an associated value on the `locked` case so you avoid reaching back to the `appStateController`.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: Yes

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: This state is unnecessary now. We used it previously in the `applicationDidEnterBackground` event. In this event we changed the `lastUnlockedDate` if user is authenticated. It has now been moved to `AppStateController`.

Wire-iOS/Sources/AppRootViewController.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
KaterinaWire: According to your suggestion, I added a separate view to block the screen for task switching, and this state is no longer needed.

KaterinaWire:
> In my opinion, reusing the `AppLockViewController` to block the screen for task switching is introducing unnecessary complexity. A better approach would be to just display a simple view in the notification window.
> 
> The notification window seems to have no other responsibility so you could repurpose it for the sole task of hiding the app when entering the background.

Added a `DimmedViewController` for this purpose.

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -228,6 +247,19 @@ extension AppStateController {
...
+    }
+    
+    @objc func applicationDidEnterBackground() {
+        switch authenticationState {
+        case .loggedIn where appState != .locked(databaseIsLocked: isDatabaseLocked):
+            AppLock.lastUnlockedDate = Date()
+        default:
+            break
+        }
+    }
```
johnxnguyen, line 262: Setting the `lastUnlockedDate` should be the responsibility of the `AppLockTimer`. The problem with having `AppLock.lastUnlockedDate` publicly accessible and mutable is that it can be modified from anywhere and indeed it is from more than one place (here and in the `AppLockPresenter`. This can lead to programming errors that are difficult to catch before they are released.

I suggest to encapsulate all app lock date logic inside the `AppLockTimer`. This means to remove `AppLock.lastUnlockDate` and implementing the `AppLockTimer` in a way similar to:

```swift
class AppLockTimer {

  private var isLocked = true
  private var startDate = Date.distantPast

  var shouldLockApp: Bool {
    return isAppLockFeatureEnabled && isLocked
  }

  func appDidBecomeUnlocked() {
    isLocked = false
  }

  func appDidEnterForeground() {
    let elpasedTime = startDate.timeIntervalSinceNow
    isLocked = elpasedTime > 10 // <- some threshold
  }

  func appDidEnterBackground() {
    guard !isLocked else { return }
    startDate = Date()
  }

}
```

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -31,31 +31,16 @@ final class AppLockViewController: UIViewController {
...
-
-            if dimContents {
-                AppDelegate.shared.notificationsWindow?.makeKey()
-            } else {
-                AppDelegate.shared.notificationsWindow?.isHidden = !dimContents
-                AppDelegate.shared.window?.makeKey()
-            }
-        }
-    }
+    private var isDatabaseLocked: Bool = false
```
johnxnguyen, line 34: Is there a need to store this? If you simply need it to instantiated the `AppLockPresenter`, then you should create the presenter in `init`.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.swift
```diff
@@ -16,33 +16,50 @@
...
 
+import Cartography
 import UIKit
+import WireCommonComponents
 
 final class NotificationWindowRootViewController: UIViewController {
-    private(set) var appLockViewController: AppLockViewController?
-
+    
+    private(set) var dimmedViewController: DimmedViewController = DimmedViewController()
```
johnxnguyen, line 25: Does the notification window do anything beside holding the `dimmedViewController`? If not, then I suggest to repurpose the notification window and it's root view controller to be the dimmed view controller / window. As a side note, I think a more intent-revealing name is `ScreenObfuscation`.

For instance:

```swift
final class ScreenObfuscationWindow: UIWindow {

    super.init(frame: frame)
    rootViewController = RootViewController(frame: frame)
  }

  func applicationWillResignActive() {
    isHidden = true
  }

  func applicationDidEnterForeground() {
    isHidden = !AppLock.isActive
  }

}

private extension ScreenObfuscationWindow {

  class RootViewController: UIViewController {

    override func loadView() {
      view = // some opaque view
    }

  }

}
```

This way you only need to hide the window and not care about adding / removing a separate view controller.

johnxnguyen (commented):
Are there any tests for the `AppLockTimer`?

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
johnxnguyen: The method that ends up hiding / showing the lock view is `requireAuthenticationIfNeeded()` and is called at the end of the initializer for the `AppLockPresenter`. You could remove this invocation from the initializer and call it from the `viewDidLoad` method:

```swift
override func viewDidLoad() {
  super.viewDidLoad()
  lockView = AppLockView()
  presenter.requireAuthenticationIfNeeded()
}
```

Or, you could also just initialize the lock view when declaring it: `private let lockView = AppLockView()`

KaterinaWire:
> AppLockTimer

Yes, there's a `AppLockTimerTests` class

KaterinaWire:
> Are there any tests for the `AppLockTimer`?

Yes, there's a `AppLockTimerTests` class

Wire-iOS/Sources/AppRootViewController.swift
```diff
@@ -546,6 +548,7 @@ extension AppRootViewController {
     @objc fileprivate func applicationDidBecomeActive() {
         updateOverlayWindowFrame()
         teamMetadataRefresher.triggerRefreshIfNeeded()
+        AppDelegate.shared.notificationsWindow?.isHidden = true
```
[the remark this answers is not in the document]
KaterinaWire: Replaced

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: Modified

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
KaterinaWire: Yes. 
I moved into `init`

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindowRootViewController.swift
KaterinaWire: Changed the NotificationWindow

Wire-iOS/Sources/Helpers/AppLock/AppLockTimerProtocol.swift
```diff
@@ -0,0 +1,57 @@
...
+//
+
+import Foundation
+import WireCommonComponents
+
+protocol AppLockTimerProtocol {
+    var shouldLockScreen: Bool { get }
+    func appDidBecomeUnlocked()
+    func appDidEnterForeground()
+    func appDidEnterBackground()
```
johnxnguyen, line 26: I was thinking more that these methods were not part of the protocol but rather callbacks to notifications (defined only in `AppLockTimer`).

Wire-iOS/Sources/AppStateController.swift
```diff
@@ -226,9 +245,24 @@ extension AppStateController {
...
+
+    @objc func applicationDidBecomeUnlocked() {
+        appLockTimer.appDidBecomeUnlocked()
         updateAppState()
     }
     
+    @objc func applicationDidEnterBackground() {
+        switch authenticationState {
+        case .loggedIn:
+            appLockTimer.appDidEnterBackground()
```
johnxnguyen, line 260: If the `AppLockTimer` does its own observation "application did enter background", then this is unnecessary, including the check of the authentication state.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindow.swift
```diff
@@ -17,22 +17,60 @@
 //
 
 import Foundation
+import WireSyncEngine
 import UIKit
+import WireCommonComponents
+import Cartography
 
-final class NotificationWindow: UIWindow {
+final class ScreenObfuscationWindow: UIWindow {
```
johnxnguyen, line 25: Don't forget to rename the file.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindow.swift
```diff
@@ -17,22 +17,60 @@
...
     
     override init(frame: CGRect) {
         super.init(frame: frame)
-        rootViewController = NotificationWindowRootViewController()
+        rootViewController = RootViewController()
         backgroundColor = .clear
         accessibilityIdentifier = "ZClientNotificationWindow"
         accessibilityViewIsModal = true
         windowLevel = UIWindowLevelNotification // status bar level - 1
         isOpaque = false
```
johnxnguyen, line 34: You may want to update these properties:

I would say that background color is irrelevant now, the accessibility identifier should change (check with QA though because they're probably using it), and the window is now opaque.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindow.swift
```diff
@@ -17,22 +17,60 @@
 //
 
 import Foundation
+import WireSyncEngine
 import UIKit
+import WireCommonComponents
+import Cartography
```
johnxnguyen, line 23: We're trying to get rid of Cartography, but it looks like it's not used in this file anyway. I think you can safely remove this import.

KaterinaWire: Yes, you are right, we don't need it now.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindow.swift
KaterinaWire: Done

Wire-iOS/Sources/AppStateController.swift
KaterinaWire: Added an observer to the `AppLockTimer` and removed from the `AppStateController`.

Wire-iOS/Sources/UserInterface/Overlay/NotificationWindow.swift
KaterinaWire: I will ask about `accessibilityIdentifier`.
The rest of the irrelevant  setting I removed.

KaterinaWire: Victor confirmed that they don't use this identifier for their tests.

Wire-iOS/Sources/Helpers/AppLock/AppLockTimerProtocol.swift
KaterinaWire: As we discussed during the call, I can't use observers `appDidBecomeUnlocked` and `appDidEnterForeground` in `AppLockTimer` because there are the same observers in `AppStateController`.
Since `AppLockTimer` is a helper class that conform `AppLockTimerProtocol` and in `AppStateController` we use an instance by this protocol. That's why I need to have these methods as a part of the `AppLockTimerProtocol`

Wire-iOS Tests/AppLock/AppLockTimerTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        super.setUp()
+        sut = AppLockTimer()
+    }
+    
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
+    
+    func testThatShouldLockScreenReturnsTrueIfNeeded() {
```
johnxnguyen, line 36:
```suggestion
    func testThatScreenShouldBeLockedIfAppLockIsActiveAndTimeoutIsReached() {
```

Wire-iOS Tests/AppLock/AppLockTimerTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        //when / then
+        XCTAssertFalse(sut.shouldLockScreen)
+    }
+}
+
+extension AppLockTimerTests {
+    func set(appLockActive: Bool, timeoutReached: Bool) {
+        AppLock.isActive = appLockActive
+        AppLock.rules = AppLockRules(useBiometricsOrAccountPassword: false, useCustomCodeInsteadOfAccountPassword: false, forceAppLock: false, appLockTimeout: 900)
+        timeoutReached ? nil : sut.appDidBecomeUnlocked()
```
johnxnguyen, line 65: It would be more readable to use an `if` statement.

Wire-iOS Tests/AppStateControllerTests.swift
```diff
@@ -18,16 +18,36 @@
 
 import XCTest
 @testable import Wire
+@testable import WireCommonComponents
 
+private final class MockAppLockTimer: AppLockTimerProtocol {
+    var shouldLock: Bool = true
+    var lastUnlockedDate = Date.distantPast
```
johnxnguyen, line 25: Is this last unlock date property necessary?

Wire-iOS/Sources/UserInterface/Overlay/ScreenObfuscationWindow.swift
```diff
@@ -0,0 +1,70 @@
...
+                                               selector: #selector(applicationWillResignActive),
+                                               name: UIApplication.willResignActiveNotification,
+                                               object: .none)
+        NotificationCenter.default.addObserver(self, selector: #selector(applicationDidBecomeActive),
+                                               name: UIApplication.didBecomeActiveNotification,
+                                               object: nil)
+    }
+    
+    @objc func applicationWillResignActive() {
+        isHidden = !(AppLock.isActive || (ZMUserSession.shared()?.isDatabaseLocked ?? false))
```
johnxnguyen, line 48: This seems like duplicated code (see comment in the other PR about CallKit).

Wire-iOS Tests/AppLock/AppLockPresenterTests.swift
```diff
@@ -210,73 +187,49 @@ final class AppLockPresenterTests: XCTestCase {
...
-        assert(contentsDimmed: false, reauthVisibile: false)
+        assert(reauthVisibile: false)
     }
     
     func testThatPasswordVerifiedWithNotValidatedResultDimsContentsIfAuthNeeded() {
-        //given
-        appLockInteractor._isAuthenticationNeeded = true
-        
-        //when
+         //when
```
johnxnguyen, line 202: There seem to be quite a few tests here that referred to dimmed contents. You may want to update the test names to be more accurate.

Wire-iOS/Sources/UserInterface/Overlay/ScreenObfuscationWindow.swift
typfel, line 48: We need check if EAR is enabled not if the database is locked since that's never the case when main UI is shown.

```suggestion
        isHidden = !(AppLock.isActive || (ZMUserSession.shared()?.encryptMessagesAtRest ?? false))
```

typfel (changes requested):

typfel (approved):

Wire-iOS/Sources/UserInterface/Overlay/ScreenObfuscationWindow.swift
KaterinaWire: Yes, I created
```
var isLockScreenEnabled: Bool {
         return AppLock.isActive || (ZMUserSession.shared()?.encryptMessagesAtRest ?? false)
     }
```
in the `SessionManager` extension in my another PR which is (currently) based on this PR because they are related.
But it makes sense to have it here.
What @typfel @johnxnguyen do you think about having `isLockScreenEnabled` in the `SessionManager` extension?
(I will remove it from another PR)

Wire-iOS Tests/AppStateControllerTests.swift
KaterinaWire: Don't need it anymore

Wire-iOS Tests/AppLock/AppLockTimerTests.swift
KaterinaWire: Fixed

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
KaterinaWire: I initialized `lockView` when declaring it.

Wire-iOS Tests/AppLock/AppLockPresenterTests.swift
KaterinaWire: Renamed

johnxnguyen (approved):
Looks good, still need to fix a failing test though.

billypchan (approved):

billypchan:
@KaterinaNikolaenko I read through the code and seems no issue, but you need to update this PR.

PR #4651, opened 2020-10-13
Fix: restore launch flag --use-app-center

billypchan:
### Issues

launch flag `--use-app-center` not work

### Causes

The flag is removed after App center intergration

### Solutions

Remove old Hockey app related code and redo fro AppCenter.

App center is enabled or disabled with these checkings:
priority 1:
check launching option `--use-app-center`'s value (0/1)

priority 2:
check UserDefaults KVP `UseAppCenter`

priority 3:
check Bundled info plist has `UseAppCenter` (form .xcconfig key `USE_APP_CENTER_KEY`)


WireCommonComponents/AutomationHelper.swift
```diff
@@ -89,14 +96,24 @@ public final class AutomationHelper: NSObject {
...
+        
+        let value = arguments.flagValueIfPresent(AutomationKey.useAppCenter.rawValue)
+        switch value {
+        case "0":
+            useAppCenterLaunchOption = false
+        case "1":
+            useAppCenterLaunchOption = true
+        default:
+            break
+        }
```
johnxnguyen, line 114: Why do we set the `useAppCenterLaunchOption` differently from the rest?

billypchan: There is an existing `useAppCenter` property which gets a stored value from `UserDefaults`. The new `useAppCenterLaunchOption` property has a higher priority than that.

johnxnguyen (commented):
@billypchan can you describe what the intended behavior should be because it's a bit confusing from the code.

billypchan:
@johnxnguyen updated the description to explain the user app center flag in launch option, user default and bundled plist

johnxnguyen (commented):
So we only care about negative values? The behavior I see is this:

```swift
let useAppCenter = launchArgumentValue && userDefaultsValue && bundleValue
```

Which means that the launch argument is `--use-app-center` is `true`, but the user defaults value is `false`, then app center would be disabled. Is this correct?

billypchan:
@johnxnguyen I will clean up the code and put all logic in one place.

Wire-iOS/Sources/AppDelegate+AppCenter.swift
```diff
@@ -32,9 +32,10 @@ extension AppDelegate {
 
     func setupAppCenter(completion: @escaping () -> Void) {
 
-        let shouldUseAppCenter = AutomationHelper.sharedHelper.useAppCenter || Bundle.useAppCenter
+        // launch option has higher priority than bundle option
+        let shouldUseAppCenter = AutomationHelper.sharedHelper.useAppCenter && Bundle.useAppCenter
```
billypchan, line 36: @johnxnguyen `AutomationHelper.sharedHelper.useAppCenter` here has `launchArgumentValue && userDefaultsValue` check inside

johnxnguyen:
> @johnxnguyen I will clean up the code and put all logic in one place.

I was asking a question about what the flags mean. It's not clear at all what effect a single flag is supposed to have. Is each flag really only about whether we should disabled app center?

billypchan:
@johnxnguyen the flags are coming from different requirements:

`launchArgumentValue` - only for QA, when running the app with automation
`userDefaultsValue` - for manually control app center with user default, for testing
`bundleValue` - build time force disable depends on config.

For the question "Which means that the launch argument is --use-app-center is true, but the user defaults value is false, then app center would be disabled. Is this correct?"

The use case of `--use-app-center` for QA is set it to false to force disable app center. Set it to ture is equal to not to set it and will let other 2 booleans make the decision.

johnxnguyen (approved):

PR #4656, opened 2020-10-15
[Feature]: AppCenterOperation

marcomaddy1982:
## What's new in this PR?

Same topic of this PR https://github.com/wireapp/wire-ios/pull/4655 but applied to the AppCenter launch configuration. 

### NOTA BENE ###
Let's review in detail ti.


billypchan (approved):

Wire-iOS/Sources/LaunchSequenceOperation.swift
```diff
@@ -98,3 +101,88 @@ final class FileBackupExcluderOperation: LaunchSequenceOperation {
...
+        
+        MSAppCenter.setLogLevel(.verbose)
+        
+        // This method must only be used after Services have been started.
+        MSAppCenter.setTrackingEnabled(true)
+        
+    }
+}
+
+extension AppCenterOperation: MSDistributeDelegate {
```
typfel, line 135: It's bit counter intuitive that the `AppCenterOperation` implements in the `MSDistributeDelegate`. If something is called `operation ` I expect it to perform an operation and then be done.

Maybe the delegates could implemented somewhere else and then be injected into the `AppCenterOperation`?

typfel (changes requested):

Wire-iOS/Sources/LaunchSequenceOperation.swift
marcomaddy1982: I would find a bit verbose something like that honestly. :-D

PR #4660, opened 2020-10-19
[Refactoring]: DeepLinks

marcomaddy1982:
## What's new in this PR?

In this PR can find:

# Application Launching Refactoring Issue Fix:
When the app was in background and then for some reasons were going in foreground the `AppStateCalculator` was forcing to load the `appState`. This is wrong cause if we have no variation in the `appState` this is not necessary and we don't have to force it. In fact in the scenario we were opening a deep link and the user was authenticated we were forcing the app to the same state it already was (`.authenticated`) and all the transition were performed and the ZMClientController were returning to its initial state and show the conversation list. Removing the `force` parameter the transition are not execute when not necessary.

# DeepLink Sync-Engine communication Refactoring:  
Regarding this topic see the PR https://github.com/wireapp/wire-ios-sync-engine/pull/1327

## Dependenties
sync-engine: https://github.com/wireapp/wire-ios-sync-engine/pull/1327


johnxnguyen (commented):
Can you add a test for this?

Cartfile
```diff
@@ -1,7 +1,7 @@
 github "wireapp/wire-ios-ziphy" ~> 14.0
 github "wireapp/Cartography" "4.0.0-xcode_11_4_1"
-github "wireapp/wire-ios-sync-engine" ~> 341.0
-github "wireapp/wire-ios-share-engine" ~> 192.0
+github "wireapp/wire-ios-sync-engine" ~> 342.0
```
billypchan, line 3: we can bump this after feature branch of sync engine is merged.

marcomaddy1982: yes of course, cause I merge from develop, I could actually use there the feature branch

marcomaddy1982:
> Can you add a test for this?

Regarding the Tests you can notice that the behaviour of the DeepLinks and the behaviour of the `PresentationDelegate` is pretty tested in the sync-engine. See PR wireapp/wire-ios-sync-engine#1327

johnxnguyen:
> Regarding the Tests you can notice that the behaviour of the DeepLinks and the behaviour of the PresentationDelegate is pretty tested in the sync-engine.

@marcomaddy1982 the test I was referring to was for the fix rather than the refactoring which followed.

marcomaddy1982:
> > Regarding the Tests you can notice that the behaviour of the DeepLinks and the behaviour of the PresentationDelegate is pretty tested in the sync-engine.
> 
> @marcomaddy1982 the test I was referring to was for the fix rather than the refactoring which followed.

Yep It's true I got confused a bit with the comments. I'm thinking how I could test it.

marcomaddy1982:
> > Regarding the Tests you can notice that the behaviour of the DeepLinks and the behaviour of the PresentationDelegate is pretty tested in the sync-engine.
> 
> @marcomaddy1982 the test I was referring to was for the fix rather than the refactoring which followed.

It is not so immediate write a test for this scenario. So basically we should simulate these steps:
1) the app launch
2) the user opens a conversation
3) the app goes in background
4) the app become active

and test that the topViewController in the hierarchy is not changes (so basically it is the conversation opened).

johnxnguyen:
> It is not so immediate write a test for this scenario.

@marcomaddy1982 what would it require to be able to write a test?

marcomaddy1982:
> > It is not so immediate write a test for this scenario.
> 
> @marcomaddy1982 what would it require to be able to write a test?

So basically we should test that the when the app come out from the background state to the foreground state the topViewController on the screen is not changed if the appState didn't as well. I tried to write a test and try to find a similar scenario around the tests written but without success.

johnxnguyen:
> So basically we should test that the when the app come out from the background state to the foreground state the topViewController on the screen is not changed if the appState didn't as well. I tried to write a test and try to find a similar scenario around the tests written but without success.

@marcomaddy1982 correct me if I'm wrong, but it seems like the core of what we want to test is that invoking the `AppStateCalculator` method `transition(to appState: AppState)` will return early (an not invoke the delegate callback) if the given app state is the same as the current app state? I think a simple unit test like this is sufficient enough.

marcomaddy1982:
> > So basically we should test that the when the app come out from the background state to the foreground state the topViewController on the screen is not changed if the appState didn't as well. I tried to write a test and try to find a similar scenario around the tests written but without success.
> 
> @marcomaddy1982 correct me if I'm wrong, but it seems like the core of what we want to test is that invoking the `AppStateCalculator` method `transition(to appState: AppState)` will return early (an not invoke the delegate callback) if the given app state is the same as the current app state? I think a simple unit test like this is sufficient enough.

@johnxnguyen we could also test this of course.

johnxnguyen:
@marcomaddy1982 a small correction. We would want to test that invoking `applicationDidBecomeActive()` on the `AppStateCalculator` doesn't trigger the delegate callback for the app state transition.

marcomaddy1982:
> @marcomaddy1982 a small correction. We would want to test that invoking `applicationDidBecomeActive()` on the `AppStateCalculator` doesn't trigger the delegate callback for the app state transition.

I don't know if is is so easy at it seems but I can give a try of course.

marcomaddy1982:
@johnxnguyen could we merge so I can start writing the tests related to the ticket we created during the planning?

johnxnguyen:
> @johnxnguyen could we merge so I can start writing the tests related to the ticket we created during the planning?

What about the test we discussed for this PR?

marcomaddy1982:
> > @johnxnguyen could we merge so I can start writing the tests related to the ticket we created during the planning?
> 
> What about the test we discussed for this PR?

It will be included there.

marcomaddy1982:
> > > @johnxnguyen could we merge so I can start writing the tests related to the ticket we created during the planning?
> > 
> > 
> > What about the test we discussed for this PR?
> 
> It will be included there.

But if you prefer I can separate them.

marcomaddy1982:
> > @johnxnguyen could we merge so I can start writing the tests related to the ticket we created during the planning?
> 
> What about the test we discussed for this PR?

Done.

Wire-iOS Tests/AppStateCalculatorTests.swift
```diff
@@ -106,4 +106,51 @@ final class AppStateCalculatorTests: XCTestCase {
...
+        sut.sessionManagerDidFailToLogin(account: nil, error: error)
+        
+        let appRootRouter = AppRootRouterMock()
+        sut.delegate = appRootRouter
+        
+        sut.applicationDidEnterBackground()
+        sut.applicationDidBecomeActive()
+        
+        // WHEN
+        XCTAssertFalse(appRootRouter.isAppStateCalculatorCalled)
```
johnxnguyen, line 125: Should be "Then".

Wire-iOS Tests/AppStateCalculatorTests.swift
```diff
@@ -106,4 +106,51 @@ final class AppStateCalculatorTests: XCTestCase {
...
+        let appRootRouter = AppRootRouterMock()
+        sut.delegate = appRootRouter
+        
+        sut.applicationDidEnterBackground()
+        // AppState changes when the app is in background
+        sut.sessionManagerDidBlacklistCurrentVersion()
+        sut.applicationDidBecomeActive()
+        
+        // WHEN
+        XCTAssertTrue(appRootRouter.isAppStateCalculatorCalled)
```
johnxnguyen, line 145: Should be "Then".

Wire-iOS Tests/AppStateCalculatorTests.swift
```diff
@@ -106,4 +106,51 @@ final class AppStateCalculatorTests: XCTestCase {
...
+        sut.delegate = appRootRouter
+        
+        sut.applicationDidEnterBackground()
+        sut.applicationDidBecomeActive()
+        
+        // WHEN
+        XCTAssertFalse(appRootRouter.isAppStateCalculatorCalled)
+    }
+    
+    func testApplicationTransitIfAppStateDontChangeWhenAppBecomeActive() {
```
johnxnguyen, line 128: Do you mean `testApplicationTransitIfAppStateChangesWhenAppBecomesActive`?

johnxnguyen (approved):

PR #4662, opened 2020-10-22
Fix: upload a video (via share extension) but it is stuck in this screen

billypchan:
### Issues

When a user sharing a ~70MB .mov, it takes several minutes and the converted video is ~140MB.

### Causes

The converssion from .mov to .mp4 may increase the file size due to quality config

### Solutions

Calculate the estimated file size, if the size exceeded the account limit, reduce the export quality

### DIFF
--- a/Wire-iOS Share Extension/UnsentSendable.swift
+++ b/Wire-iOS Share Extension/UnsentSendable.swift
@@ -312,7 +312,8 @@ class UnsentFileSendable: UnsentSendableBase, UnsentSendable {
                                       fileURL: URL,
                                       completion: @escaping SendingCompletion) {
         if UTTypeConformsTo(UTI as CFString, kUTTypeMovie) {
-            AVURLAsset.convertVideoToUploadFormat(at: fileURL) { (url, _, error) in
+            AVURLAsset.convertVideoToUploadFormat(at: fileURL,
+                                                  fileLengthLimit: Int64(AccountManager.fileSizeLimitInBytes)) { (url, _, error) in
                 completion(url, error)
             }
         } else {
--- a/Wire-iOS.xcodeproj/project.pbxproj
+++ b/Wire-iOS.xcodeproj/project.pbxproj
@@ -5658,6 +5658,55 @@
 			path = Cells;
 			sourceTree = "<group>";
 		};
+		A96E072525416F87001F7658 /* video */ = {
+			isa = PBXGroup;
+			children = (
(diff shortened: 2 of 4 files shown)


johnxnguyen (commented):
It seems strange that converting a 71mb mov file to a high quality mp4 almost doubles the file size, yet converting it to medium quality reduces the file size by a factor of 10. Is this consistent behavior?

billypchan:
@johnxnguyen checked that in the above sample case the video is reduced to low quality and the size is reduced to 320x568.

johnxnguyen (commented):
Prioritizing smaller file sizes over quality seems like a product decision. I'm not sure if it's a good idea to degrade the quality past a certain point, but rather indicate to the user that the file is simply too large.

@typfel what are your thoughts?

billypchan:
@johnxnguyen due to the limited amount of memory allowed for share extension, without downsizing the video the user has to wait too long(or even crash) for UI response. We already have a downsize logic for 4K video to HD size and further downsizing should be also applied.

typfel:
So we didn't event set the `fileLengthLimit` when sending a video from the share extension. Wasn't it enough to start using the `fileLengthLimit`?

billypchan:
> `fileLengthLimit`

For the API comment it said: `Indicates the file length that the output of the session should not exceed.  Depending on the content of the source asset, it is possible for the output to slightly exceed the file length limit.  The length of the output file should be tested if you require that a strict limit be observed before making use of the output.  See also maxDuration and timeRange.`

Tested with devices and I found that setting `fileLengthLimit` when exporting the video does not affect the file size. We still need to determine what quality preset for the target file size.

johnxnguyen:
> Tested with devices and I found that setting `fileLengthLimit` when exporting the video does not affect the file size. 

@billypchan is it a bug? It might be worth to try to get it to work.

billypchan:
@johnxnguyen I tested that on the device but it does not work.

typfel:
@billypchan @johnxnguyen I tested it on the simulator with the file size limit set to 5Mb, I tried to send a 25 Mb video and it got converted to a 5,6 Mb video.

billypchan:
@typfel You may check new `testThatLongVideoIsConvertedToLowQualityFormat`, the compression rate is variable and may affected by the content of the video.

typfel:
I tested it again by setting the `fileLengthLimit` to 10Mb and sent your test file on a real device. The video that was sent was 9.4 Mb.

PR #4664, opened 2020-10-23
[Refactoring]: AppRootRouter Clean Up

marcomaddy1982:
## What's new in this PR?

the scope of this PR is clean up a bit the `AppRootRouter` trying to encapsulate as possible its properties and moving not necessary code from it.

### DIFF
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -52,6 +52,12 @@ public class AppRootRouter: NSObject {
     private(set) var sessionManager: SessionManager? {
         didSet {
             guard let sessionManager = sessionManager else { return }
+            switchingAccountRouter.sessionManager = sessionManager
+            urlActionRouter.sessionManager = sessionManager
+            sessionManagerLifeCycleObserver.sessionManager = sessionManager
+            foregroundNotificationFilter.sessionManager = sessionManager
+            quickActionsManager.sessionManager = sessionManager
+            
             sessionManager.foregroundNotificationResponder = foregroundNotificationFilter
             sessionManager.switchingDelegate = switchingAccountRouter
             sessionManager.presentationDelegate = urlActionRouter
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -161,11 +161,11 @@ public class AppRootRouter: NSObject {
     }
     
     private func setCallingSettings() {
-        SessionManager.shared?.updateCallNotificationStyleFromSettings()
-        SessionManager.shared?.useConstantBitRateAudio = SecurityFlags.forceConstantBitRateCalls.isEnabled
+        sessionManager?.updateCallNotificationStyleFromSettings()
+        sessionManager?.useConstantBitRateAudio = SecurityFlags.forceConstantBitRateCalls.isEnabled
             ? true
             : Settings.shared[.callingConstantBitRate] ?? false
-        SessionManager.shared?.useConferenceCalling = true
+        sessionManager?.useConferenceCalling = true
     }
 }
 
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -32,18 +32,17 @@ public class AppRootRouter: NSObject {
     let callWindow = CallWindow(frame: UIScreen.main.bounds)
     let overlayWindow = NotificationWindow(frame: UIScreen.main.bounds)
     
-    // TO DO: this shoud be private
-    private(set) var switchingAccountRouter: SwitchingAccountRouter?
-    
     // MARK: - Private Property
     private let navigator: NavigatorProtocol
     private var appStateCalculator = AppStateCalculator()
     private var deepLinkURL: URL?
     
     private var authenticationCoordinator: AuthenticationCoordinator?
+    private var switchingAccountRouter: SwitchingAccountRouter
     private var urlActionRouter: URLActionRouter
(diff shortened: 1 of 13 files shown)


Wire-iOS/Sources/QuickActionsManager.swift
```diff
@@ -53,7 +50,7 @@ public final class QuickActionsManager: NSObject {
     @objc func performAction(for shortcutItem: UIApplicationShortcutItem, completionHandler: ((Bool)->())?) {
         switch shortcutItem.type {
         case UIApplicationShortcutItem.markAllAsReadType:
-            sessionManager.markAllConversationsAsRead {
+            SessionManager.shared?.markAllConversationsAsRead {
```
johnxnguyen, line 53: Won't relying on the singleton make this harder to test?

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -149,12 +160,12 @@ public class AppRootRouter: NSObject {
...
-    private func setCallingSettings(for sessionManager: SessionManager) {
-        sessionManager.updateCallNotificationStyleFromSettings()
-        sessionManager.useConstantBitRateAudio = SecurityFlags.forceConstantBitRateCalls.isEnabled
+    private func setCallingSettings() {
+        SessionManager.shared?.updateCallNotificationStyleFromSettings()
+        SessionManager.shared?.useConstantBitRateAudio = SecurityFlags.forceConstantBitRateCalls.isEnabled
             ? true
             : Settings.shared[.callingConstantBitRate] ?? false
-        sessionManager.useConferenceCalling = true
+        SessionManager.shared?.useConferenceCalling = true
```
johnxnguyen, line 168: `AppRootRouter` has the reference to the session manager already, why not use it?

Wire-iOS/Sources/QuickActionsManager.swift
marcomaddy1982: Actually my fix is wrong is not changing anything cause `SessionManager.shared` is  `AppDelegate.shared.appRootRouter?.sessionManager` :-D. I will just do `sessionManager?.markAllConversationsAsRead`.

Wire-iOS/Sources/AppRootRouter.swift
marcomaddy1982: exactly, I got confused and as I wrote in the previous comment `SessionManager.shared` is `AppDelegate.shared.appRootRouter?.sessionManager`.

Wire-iOS/Sources/QuickActionsManager.swift
marcomaddy1982: `SessionManager.shared` === `AppDelegate.shared.appRootRouter?.sessionManager` explains also the reason why in the other object I'm not injecting the `sessionManager` after it is created.

johnxnguyen: I'm not sure I understand

marcomaddy1982: Basically I meant that `SessionManager.shared` and `AppDelegate.shared.appRootRouter?.sessionManager` are exactly the same instance cause 

```swift
static var shared : SessionManager? {
        return AppDelegate.shared.appRootRouter?.sessionManager
    }
```

johnxnguyen: My initial comment was about testability. By using a singleton (which relies on another singleton), we're making it pretty difficult to test. Wouldn't it be better to inject the session manager? Also note there is a `SessionManagerType` that abstracts it to a protocol.

marcomaddy1982: You comment makes sense, probably it would be better revert part and inject the `SessionManager` for a better testability.

marcomaddy1982: Done

Wire-iOS/Sources/ForegroundNotificationFilter.swift
```diff
@@ -18,8 +18,21 @@
 
 import WireSyncEngine
 
+final class ForegroundNotificationFilter {
+    
+    // MARK: - Public Property
+    var sessionManager: SessionManager?
```
johnxnguyen, line 24: Does this (and the other references to `SessionManager` need to be optional?

marcomaddy1982: Cause my idea is to instantiate the class in the init of the AppRootRouter and then set the sessionManager property when we have it. 
This is optional in case we need this init method for better testability.

johnxnguyen: At which point do we set the session manager on the `AppRootRouter`? Are we not able provide the session manager when initializing the router?

marcomaddy1982: In the `AppRootRouter` the `sessionManger` in the call back when we create it, line 163 of this PR.  Then in the `didSet` of the `sessionManger` I set the `sessionManager` also for all the other references in cascade. Of course we could set the `sessionManager` directly when the other instances are initialised in the the init of the `AppRootRouter` cause it passed as reference type.

marcomaddy1982: But honestly I would prefer to set it directly when the `sessionManager` is created.

johnxnguyen: It looks like you can make all `sessionManager` properties non optional if you create the session manager when you initialize the app router. It's up to you though.

johnxnguyen (approved):

PR #4666, opened 2020-10-27
tag app open event after self user is set

billypchan:
## What's new in this PR?

Tag for "app.open" event once after `selfUser` property of `AnalyticsProvider` is set (tagging requires user id, and it is assigned after `selfUser` is assigned when the user session is created).


typfel:
How about making this solution a bit more generic to not only solve the case for when the app is opened? We could for example store tags in an array if the `selfUser` is not set and then forward the tags later when the selfUser is set.

billypchan:
@typfel I discussed this feature with PM and confirmed that the event has to be tagged every time the app becomes active. I will open another PR.

PR #4668, opened 2020-10-27
tag app.open when app became active

billypchan:
## What's new in this PR?

For every time the app became active, tag `app.open` event.
If selfuser is not set, i.e. before the user session is created, store the event and send it after selfuser is set.


Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -119,8 +133,16 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
...
-                  attributes: [String: Any]) {
+    func tagEvent(_ event: String, attributes: [String: Any]) {
+        //store the event before self user is assigned, send it later when self user is ready.
+        if selfUser == nil {
+            if storedEvents == nil {
+                storedEvents = []
+            }
+
+            storedEvents?.append(StoredEvent(event: event, attributes: attributes))
+        }
```
johnxnguyen, line 144: I suggest to use a guard statement to make the intent a bit clearer:

```swift
guard selfUser != nil else {
  storedEvents.append(...)
  return
}
```

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -32,6 +32,14 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
     /// flag for recording session is begun
     private var sessionBegun: Bool = false
 
+    private struct StoredEvent {
+        let event: String
+        let attributes: [String: Any]
+    }
+
+    /// store the events before selfUser is assigned. Send them and clear after selfUser is set
+    private var storedEvents: [StoredEvent]?
```
johnxnguyen, line 41: Is there a particular reason to make this optional?

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: fixed.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: removed the optional and simplied the code, thx!

johnxnguyen (changes requested):
It would be great to have a test for this new behavior.

Wire-iOS Tests/Analytics/AnalyticsCountlyProviderTests.swift
```diff
@@ -54,4 +54,27 @@ final class AnalyticsCountlyProviderTests: XCTestCase, CoreDataFixtureTestHelper
...
+
+        //THEN
+        XCTAssertEqual(analyticsCountlyProvider.storedEventsCount, 1)
+
+        //WHEN
+        sut.selfUser = coreDataFixture.selfUser
+
+        //THEN
+        XCTAssertEqual(analyticsCountlyProvider.storedEventsCount, 0)
+    }
```
billypchan, line 79: @johnxnguyen test added for stored events

johnxnguyen: If you abstract out `Countly.sharedInstance()` from `AnalyticsCountlyProvider`, you would be able to test that events "recorded" only once the self user is set. Also, that would open up the possibility of more focused testing.

For instance, suppose you had a protocol:

```swift
protocol CountlyInstance {
  func recordEvent(_ key: String, segmentation: [String : String]?)
}
```

Which can be conformed to by `Countly`, then you could inject a test instance and verify invocations to this method rather than exposing the concept of stored events and counting them.

billypchan: adding `MockCountly` class for tests

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -46,19 +65,32 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
...
+            }
+
+            storedEvents.removeAll()
         }
     }
+    
+    var countlyInstanceType: CountlyInstance.Type
 
-    init() {
+    init?(countlyInstanceType: CountlyInstance.Type = Countly.self) {
```
johnxnguyen, line 85: I think it makes more sense to pass in the instance itself, not the type.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
```diff
@@ -80,7 +112,7 @@ final class AnalyticsCountlyProvider: AnalyticsProvider {
         config.appKey = countlyAppKey
         config.host = countlyURL.absoluteString
         config.manualSessionHandling = true
-
+        
         config.deviceID = analyticsIdentifier
         Countly.sharedInstance().start(with: config)
```
johnxnguyen, line 117: Ideally we remove all concrete instances of `Countly`, and just have the protocol type.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: passing the type is better since there are some other places calling another static methods, e.g. `Countly.user().set()`

johnxnguyen: can we abstract those too?

billypchan: I will create another PR to test the switching account requirement. It would be better to write tests when abstracting the methods.

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: I will open another PR for testing switching account and non-team member behavior

Wire-iOS/Sources/Analytics/AnalyticsCountlyProvider.swift
billypchan: Added test and mocked `start()` method. I will add more mocks if necessary.
https://github.com/wireapp/wire-ios/pull/4680

johnxnguyen (approved):

PR #4670, opened 2020-10-27
[Fix] Hide Callkit when AppLock is enabled

KaterinaWire:
### Issues
If a CallKit call is answered while the device is locked and AppLock is enabled, then Wire would not be opened and presented to the user. 

### Solutions
Do not use CallKit when AppLock is enabled.

### Dependencies
https://wearezeta.atlassian.net/browse/ZIOS-13792

### TODO
- Tests
- Agree with the design team

### DIFF
--- a/Wire-iOS/Sources/Helpers/syncengine/SessionManager+Shared.swift
+++ b/Wire-iOS/Sources/Helpers/syncengine/SessionManager+Shared.swift
@@ -20,10 +20,15 @@ import Foundation
 import WireSyncEngine
 import AVFoundation
 import WireSyncEngine
+import WireCommonComponents
 
 extension SessionManager {
     static var shared : SessionManager? {
         return AppDelegate.shared.sessionManager
+    }
+    
+    var isLockScreenEnabled: Bool {
+        return AppLock.isActive || (ZMUserSession.shared()?.isDatabaseLocked ?? false)
     }
     
     func updateCallNotificationStyleFromSettings() {
@@ -31,7 +36,7 @@ extension SessionManager {
         let hasAudioPermissions = AVCaptureDevice.authorizationStatus(for: AVMediaType.audio) == AVAuthorizationStatus.authorized
         let isCallKitSupported = !UIDevice.isSimulator
         
-        if isCallKitEnabled && isCallKitSupported && hasAudioPermissions {
+        if isCallKitEnabled && isCallKitSupported && hasAudioPermissions && !isLockScreenEnabled {
             self.callNotificationStyle = .callKit
         }
         else {
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
@@ -41,6 +41,7 @@ protocol AppLockUserInterface: class {
     
     func setSpinner(animating: Bool)
     func setReauth(visible: Bool)
+    func setIncomingCallHeader(visible: Bool, from callerDisplayName: String)
 }
 
 enum AuthenticationState {
@@ -97,8 +98,15 @@ final class AppLockPresenter {
     func requireAuthenticationIfNeeded() {
         switch authenticationState {
         case .needed:
-            showReauth(visible: false)
-            appLockInteractorInput.evaluateAuthentication(description: AuthenticationMessageKey.deviceAuthentication)
+            if let callConversation = ZMUserSession.shared()?.priorityCallConversation {
+                showReauth(visible: true)
+                showIncomingCall(visible: true, for: callConversation)
+            } else {
+                showReauth(visible: false)
+                showIncomingCall(visible: false)
+                appLockInteractorInput.evaluateAuthentication(description: AuthenticationMessageKey.deviceAuthentication)
+            }
+            
         case .cancelled:
             showReauth(visible: true)
         case .pendingPassword:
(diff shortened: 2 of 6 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsPropertyToggleCellDescriptor.swift
```diff
@@ -64,6 +65,12 @@ final class SettingsPropertyToggleCellDescriptor: SettingsPropertyCellDescriptor
             toggleCell.switchView.isOn = boolValue
             toggleCell.switchView.accessibilityLabel = identifier
             toggleCell.switchView.isEnabled = self.settingsProperty.enabled
+            switch settingsProperty.propertyName {
+            case .disableCallKit:
+                toggleCell.switchView.isHidden = SessionManager.shared?.isLockScreenEnabled ?? false
+            default:
+                toggleCell.switchView.isHidden = false
+            }
```
KaterinaWire, line 73: In this implementation, I hide the `disableCallKit` settings section if AppLock is enabled. I will ask the design team about this implementation .It was done because after we disabled AppLock, we don't know what the previous state of the `disableCallKit` section was.

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -98,8 +99,15 @@ final class AppLockPresenter {
...
-            appLockInteractorInput.evaluateAuthentication(description: AuthenticationMessageKey.deviceAuthentication)
+            if let callConversation = ZMUserSession.shared()?.priorityCallConversation {
+                showReauth(visible: true)
+                showIncomingCall(visible: true, for: callConversation)
+            } else {
+                showReauth(visible: false)
+                showIncomingCall(visible: false)
+                appLockInteractorInput.evaluateAuthentication(description: AuthenticationMessageKey.deviceAuthentication)
+            }
+
```
johnxnguyen, line 110: So, if we have an incoming call, then the user must tap a button to start the authentication? The nice thing about having FaceID or TouchID is that authenticate happens as soon as you open the app.

KaterinaWire: But what about the AppLock screen where we can to see that there's an incoming call. Do you think we should show this screen when the authentication process is failed or canceled?

johnxnguyen: I'm not sure, maybe speak with design.

Wire-iOS/Sources/Helpers/syncengine/SessionManager+Shared.swift
```diff
@@ -20,18 +20,23 @@ import Foundation
...
 
 extension SessionManager {
     static var shared : SessionManager? {
         return AppDelegate.shared.sessionManager
     }
     
+    var isLockScreenEnabled: Bool {
+        return AppLock.isActive || (ZMUserSession.shared()?.isDatabaseLocked ?? false)
+    }
+
```
johnxnguyen, line 33: Is this duplicated logic? Surely we already have this defined somewhere?

Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
```diff
@@ -50,6 +70,19 @@ final class AppLockView: UIView {
             self.authenticateButton.isHidden = !showReauth
         }
     }
+    
+    var showIncomingCall: Bool = false {
+        didSet {
+            self.incomingCallLabel.isHidden = !showIncomingCall
+            self.callerDisplayNameLabel.isHidden = !showIncomingCall
```
johnxnguyen, line 77: Probably you can just hide the container instead of the subviews.

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -133,4 +133,13 @@ extension AppLockViewController: AppLockUserInterface {
...
     }
+    
+    func setIncomingCallHeader(visible: Bool) {
+        if let callConversation = ZMUserSession.shared()?.priorityCallConversation, visible {
+            lockView.showIncomingCall = true
+            lockView.callerDisplayName = (callConversation.voiceChannel?.initiator?.name) ?? "conversation.status.someone".localized //TODO: ask the design team
+        } else {
+            lockView.showIncomingCall = false
+        }
+    }
```
johnxnguyen, line 144: The logic regarding the call conversation and the content of the views is the responsibility of the present, not the view. The presenter should as the the view to show incoming call header passing along the caller display name, or to hide it in case it is not needed.

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsPropertyToggleCellDescriptor.swift
johnxnguyen: I think it would be better to disable the toggle rather than hide it altogether. I guess it's up to design though what they want to do.

In any case, I wouldn't put this logic here but move it closer to the settings property configuration for `disableCallKit`.

johnxnguyen (commented):
So we want to disable call kit whenever the app lock screen would become visible, which isn't necessarily when the app lock setting is enabled, right? This could happen if the app lock feature is forced, but also if encryption at rest is enabled and the database is locked.

In this case I think we to define a single source of truth about whether the app lock screen will be shown or not.

Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
```diff
@@ -138,7 +184,22 @@ final class AppLockView: UIView {
             authenticateButton.leadingAnchor.constraint(equalTo: contentContainerView.leadingAnchor, constant: CGFloat.PasscodeUnlock.buttonPadding),
             authenticateButton.topAnchor.constraint(equalTo: authenticateLabel.bottomAnchor, constant: 24),
             authenticateButton.trailingAnchor.constraint(equalTo: contentContainerView.trailingAnchor, constant: -CGFloat.PasscodeUnlock.buttonPadding),
-            authenticateButton.bottomAnchor.constraint(equalTo: contentContainerView.safeBottomAnchor, constant: -CGFloat.PasscodeUnlock.buttonPadding)])
+            authenticateButton.bottomAnchor.constraint(equalTo: contentContainerView.safeBottomAnchor, constant: -CGFloat.PasscodeUnlock.buttonPadding),
```
billypchan, line 187: I would suggest add snapshot tests for this UI changes.

PR #4681, opened 2020-10-30
[Fix] Make email and phone fields non-editable when re-authenticating

typfel:
### Issues

When your session expires you are asked to login with your email / phone number fields pre-filled. It's possible to edit these fields but those edits are ignored when trying to login, which is confusing.

### Causes

Fields are editable but the email / phone number is not taken from the input field when submitting the login request.

### Solutions

Make input fields non-editable when re-authenticating.


johnxnguyen (approved):
Looks like you need to update a snapshot.

typfel:
@johnxnguyen fixed the reason for the failed snapshot instead.

PR #4682, opened 2020-10-30
[Refactoring]: Merged Develop Into the Feature Branch

marcomaddy1982:
#### EVERYBODY THAT MERGED SOMETHING ON DEVELOP DURING THE LAST PERIOD PLEASE HAVE A LOOK ####

The scope of this PR is align the feature branch with the last changes of develop.


johnxnguyen (commented):
My changes were correctly merged. 👍

typfel:
In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.

I can't see that change here. Are we not setting the `tintColor` of the window anymore?

marcomaddy1982:
> In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.
> 
> I can't see that change here. Are we not setting the `tintColor` of the window anymore?

It got probably lost. I will add again.

marcomaddy1982:
> > In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.
> > I can't see that change here. Are we not setting the `tintColor` of the window anymore?
> 
> It got probably lost. Cause there is no `mainWind` anymore there. Let's check on Monday.



> > In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.
> > I can't see that change here. Are we not setting the `tintColor` of the window anymore?
> 
> It got probably lost. Cause there is no `mainWind` anymore there. Let's check on Monday.

@typfel let's do in this manner: I will do this for speed up the merging

```swift
AppDelegate.shared.window?.tintColor = UIColor.Wire.primaryLabel
```

but let's re-think in the future the way we do this kind of stuffs'.

typfel:
> > > In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.
> > > I can't see that change here. Are we not setting the `tintColor` of the window anymore?
> > 
> > 
> > It got probably lost. Cause there is no `mainWind` anymore there. Let's check on Monday.
> 
> > > In my [PR](https://github.com/wireapp/wire-ios/pull/4678) I was changing the `tintColor` of the window when transition to the unauthenticated state.
> > > I can't see that change here. Are we not setting the `tintColor` of the window anymore?
> > 
> > 
> > It got probably lost. Cause there is no `mainWind` anymore there. Let's check on Monday.
> 
> @typfel let's do in this manner: I will do this for speed up the merging
> 
> ```swift
> AppDelegate.shared.window?.tintColor = UIColor.Wire.primaryLabel
> ```
> 
> but let's re-think in the future the way we do this kind of stuffs'.
> Moreover, could you check if there is something still missing form your PRs. Please?

Wasn't part of my PR but looks like you lost setting the `tintColor` when transitioning to the authenticated state. See previous implementation https://github.com/wireapp/wire-ios/blob/fe906fb95c231c1a1fb59d246bacb118c413d8b0/Wire-iOS/Sources/AppRootViewController.swift#L277

typfel:
> but let's re-think in the future the way we do this kind of stuffs'.

Couldn't you access the `window` property on the `rootViewController` instead of going through the AppDelegate?

billypchan (approved):
The merge seems completed. But I will double check when this feature branch merge back to develop

marcomaddy1982:
> > but let's re-think in the future the way we do this kind of stuffs'.
> 
> Couldn't you access the `window` property on the `rootViewController` instead of going through the AppDelegate?

What do you mean? As far as I remember a `UIViewController` doesn't have a property for the `window`. Am I missing something? Or are you referring that the root view controller's superview is the window of the app?

marcomaddy1982:
> The merge seems completed. But I will double check when this feature branch merge back to develop

Ok, thanks. But actually the idea is double check before merging on develop branch to avoid possible derive bugs before a possible candidate release.

typfel:
> > > but let's re-think in the future the way we do this kind of stuffs'.
> > 
> > 
> > Couldn't you access the `window` property on the `rootViewController` instead of going through the AppDelegate?
> 
> What do you mean? As far as I remember a `UIViewController` doesn't have a property for the `window`. Am I missing something? Or are you referring that the root view controller's superview is the window of the app?

You are correct but the `UIViewController` has a view which has a window when it's added to view hierarchy.

marcomaddy1982:
> > The merge seems completed. But I will double check when this feature branch merge back to develop
> 
> Ok, thanks. But actually the idea is double check before merging on develop branch to avoid possible derive bugs before a future candidate release.

What I mean is this: if you don't have much time today for review it deeply it is not a problem. Take your time but IMHO it's better avoid merging on develop something not correct and after then fix it.

marcomaddy1982:
> > > > but let's re-think in the future the way we do this kind of stuffs'.
> > > 
> > > 
> > > Couldn't you access the `window` property on the `rootViewController` instead of going through the AppDelegate?
> > 
> > 
> > What do you mean? As far as I remember a `UIViewController` doesn't have a property for the `window`. Am I missing something? Or are you referring that the root view controller's superview is the window of the app?
> 
> You are correct but the `UIViewController` has a view which has a window when it's added to view hierarchy.

Ok we can do in this manner, but in future we could investigate if we can set the color on the view of the viewController we are going to fade for that particular state.

typfel:
> Ok we can do in this manner, but in future we could investigate if we can set the color on the view of the viewController we are going to fade for that particular state.

Ah yes, that's much cleaner. Why not do it like that right away.

typfel (approved):

marcomaddy1982:
> > Ok we can do in this manner, but in future we could investigate if we can set the color on the view of the viewController we are going to fade for that particular state.
> 
> Ah yes, that's much cleaner. Why not do it like that right away.

Cause I should investigate if will somehow introduce bugs :-). I would not like to risk it, if you are agree of course.

marcomaddy1982:
> > > Ok we can do in this manner, but in future we could investigate if we can set the color on the view of the viewController we are going to fade for that particular state.
> > 
> > 
> > Ah yes, that's much cleaner. Why not do it like that right away.
> 
> Cause I should investigate if will somehow introduce bugs :-). I wouldn't like risking introduce them, if you are agree of course.

typfel:
> > > Ok we can do in this manner, but in future we could investigate if we can set the color on the view of the viewController we are going to fade for that particular state.
> > 
> > 
> > Ah yes, that's much cleaner. Why not do it like that right away.
> 
> Cause I should investigate if will somehow introduce bugs :-). I would not like to risk it, if you are agree of course.

OK, that's safer.

typfel:
> The merge seems completed. But I will double check when this feature branch merge back to develop

@marcomaddy1982 is right. We should be more careful now that what we merge into develop is not causing any regressions. @billypchan if you have any doubts about the merge please check it on this branch.

billypchan:
I would prefer to do not to branch from branches and it would hard to track the changes.

marcomaddy1982:
> I would prefer to do not to branch from branches and it would hard to track the changes.

Ok but how could I help you? I mean, what can I do to make you in the condition to check as better as possible the merge?

billypchan (changes requested):
To save a merge, I would recommend these step:
1. merge `develop` to ` feature/ZIOS-13827_AppLaunchingRoutingRefactoring `
2. check the changes in`https://github.com/wireapp/wire-ios/pull/4688` does not conflict with the feature branch
3. update other branches of the feature branch 

`develop` branch is updated frequently. Creating a branch from `develop` and matching it back will would increase the amount of merging.

marcomaddy1982:
> To save a merge, I would recommend these step:
> 
> 1. merge `develop` to `feature/ZIOS-13827_AppLaunchingRoutingRefactoring`
> 2. check the changes in`https://github.com/wireapp/wire-ios/pull/4688` does not conflict with the feature branch
> 3. update other branches of the feature branch
> 
> `develop` branch is updated frequently. Creating a branch from `develop` and matching it back will would increase the amount of merging.

@billypchan this is true and I will follow your hint but it is also true that this PR is there since Friday morning and this kind of huge merge should be reviewed a bit faster for the reason you mentioned before that the develop branch changes frequently. 
The idea is: when a refactoring is merged on develop the owners of the ongoing feature branch will take care of solving the conflict born from the refactoring cause they know better how the refactoring is going to impact on its own code. 

P.s don't get me wrong this is not a critic I'm trying just to find a way how to speed up this kind of situations :-D

billypchan:
@marcomaddy1982 you can merge this PR first and then update the feature branch with `develop` branch with latest updates later.

PR #4687, opened 2020-11-02
[Refactoring]: Create the AuthenticateRouter

marcomaddy1982:
## What's new in this PR?

The scope of this PR is preparing to start developing the part of the refactoring which goal is removing the `callWindow`. 
Cause the app will show the call screen just if the user is logged in, the idea is move all the navigation about it in the router that will take care of the navigation of  the authentication cases.  So the first step to achieve this part is create the `AuthenticatedRouter`.


typfel:
I thought we were wrapping up the first part routing refactoring and were ready to merge it to develop. In that case we shouldn't merge more changes to the feature branch.

marcomaddy1982:
> I thought we were wrapping up the first part of the routing refactoring and were ready to merge it to develop. In that case we shouldn't merge more changes to the feature branch.

yes you thought right, but I'm still waiting for the double check of @billypchan regarding that merging PR  so I started do work on something else in the meantime. I choose feature/ZIOS-13827_MergeDevelop branch for the merge cause I needed one to open the PR. Probably better if I add the label `don't merge`.

billypchan:
I would recommend draft a PR that merging into `develop` from `feature/ZIOS-13827_MergeDevelop`, to track the changes of this feature and comfirm that the feature branch is not too far away form `develop`.

marcomaddy1982:
> I would recommend draft a PR that merging into `develop` from `feature/ZIOS-13827_MergeDevelop`, to track the changes of this feature and comfirm that the feature branch is not too far away form `develop`.

I guess you wanted to comment the other [PR](https://github.com/wireapp/wire-ios/pull/4682). Am I right? :-)

johnxnguyen (commented):
This appears to have the call window removed. I thought those changes were already approved / merged in a separate PR?

marcomaddy1982:
> This appears to have the call window removed. I thought those changes were already approved / merged in a separate PR?

@johnxnguyen: the call Window PR destination branch was this one. Actually I can not merge stuff on the feature branch cause I'm waiting for the approval to merge on develop. Does the current situation of this branch create issue to you for reviewing it? Could I do something that will help you with the review process?

johnxnguyen:
> Does the current situation of this branch create issue to you for reviewing it?

@marcomaddy1982 it's just a little bit confusing because the PR says that it's to support removing the call window, but in fact the call window is already removed here. I would have expected to see this PR first, then the one about removing the call window. 

> Could I do something that will help you with the review process?

It's harder to review now because the diff shows changes that aren't related to the new router. In general I think we should aim to keep the changes focused to the scope of the PR.

marcomaddy1982:
> > Does the current situation of this branch create issue to you for reviewing it?
> 
> @marcomaddy1982 it's just a little bit confusing because the PR says that it's to support removing the call window, but in fact the call window is already removed here. I would have expected to see this PR first, then the one about removing the call window.
> 
> > Could I do something that will help you with the review process?
> 
> It's harder to review now because the diff shows changes that aren't related to the new router. In general I think we should aim to keep the changes focused to the scope of the PR.

@johnxnguyen you are absolutely right. I opened this PR first and then I for mistake merged the call window one's here when it was approved. It was my bad. Sorry.

johnxnguyen:
>  I opened this PR first and then for mistake I merged the call window one here when it was approved. It was my bad. Sorry.

@marcomaddy1982 all good 😄

johnxnguyen (approved):

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -47,7 +47,13 @@ class AppStateCalculator {
     
     // MARK: - Public Property
     weak var delegate: AppStateCalculatorDelegate?
-    
+    var wasUnautheticated: Bool {
```
David-Henner, line 50: typo `wasUnauthenticated`

marcomaddy1982: Fixed. Thanks

David-Henner (approved):

PR #4691, opened 2020-11-04
[Feature] Support for EAR in share extension

typfel:
## What's new in this PR?

Ask the user to authenticate if the database is locked using encryption at rest. This allows users which have EAR enabled to use the share extension, which was previously broken.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/183

### Notes

The user experience for authenticating on the share extension is not that great but that is probably best tackled when re-designing share extension.


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -469,29 +469,35 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
             return
         }
         
-        // TODO jacob handle database lock scenario
+        let scenario: AppLock.AuthenticationScenario
         
-        let scenario: AppLock.AuthenticationScenario =
-            .screenLock(requireBiometrics: AppLock.rules.useBiometricsOrAccountPassword,
-                        grantAccessIfPolicyCannotBeEvaluated: !AppLock.rules.forceAppLock)
+        if sharingSession?.encryptMessagesAtRest == true {
```
johnxnguyen, line 487: I suggest to store `sharingSession?.encryptMessagesAtRest == true` in a computed property or local variable to avoid duplication.

marcomaddy1982 (approved):

PR #4695, opened 2020-11-06
fix: appcenter is not started

billypchan:
### Issues

App center is not started.

### Causes

After https://github.com/wireapp/wire-ios/pull/4651, we fixed the appcenter is always started case but it stops app center starting on normal case.

### Solutions

Fix the check for `UserDefaults` KVP. When "UseAppCenter" key is not exist, default value is true and app center can start.


WireCommonComponents/AutomationHelper.swift
```diff
@@ -43,7 +43,16 @@ public final class AutomationHelper: NSObject {
...
+        
+        //when UserDefaults's useAppCenter is not set, default is true to allow app center start
+        let useAppCenter: Bool
+        if UserDefaults.standard.object(forKey: "UseAppCenter") != nil {
+            useAppCenter = UserDefaults.standard.bool(forKey: "UseAppCenter")
+        } else {
+            useAppCenter = true
+        }
+
+        return useAppCenterLaunchOption && useAppCenter
```
typfel, line 55: To me this logic would only make sense if `useAppCenterLaunchOption` is optional and only set if `--user-app-center` is present.

Then we could write something like this:

```
if let useAppCenterLaunchOption {
    return useAppCenterLaunchOption
} else {
    return UserDefaults.standard.bool(forKey: "UseAppCenter") ?? true
}
```

typfel (changes requested):

WireCommonComponents/AutomationHelper.swift
billypchan: The use case of `useAppCenterLaunchOption` is set it to false to disable app center.

If the code is refactored as suggestion, it would have difference result:

when `useAppCenterLaunchOption` is false, it will check `UserDefaults.standard.bool(forKey: "UseAppCenter")`

The original code would let `useAppCenterLaunchOption` has higher priority, when `useAppCenterLaunchOption` is false, app center would not start no matter `UserDefaults.standard.bool(forKey: "UseAppCenter")`'s value is true or false.

johnxnguyen: How would it have a different result? Jacob's suggestion is much more readable.

marcomaddy1982:
@billypchan I would suggest to merge dev on this branch in order to have all the changes regarding the App Launch Refactoring and have a look to the `AppCenterOperation` class.  I know your change is made in the `AutomationHelper.swift` but IMHO for testing it is better being alined with develop.

marcomaddy1982:
> @billypchan I would suggest to merge dev on this branch in order to have all the changes regarding the App Launch Refactoring and have a look to the `AppCenterOperation` class. I know your change is made in the `AutomationHelper.swift` but IMHO for testing it is better being alined with develop.

WireCommonComponents/AutomationHelper.swift
billypchan: refactor the code for readibility

billypchan:
@marcomaddy1982 retested after merging

marcomaddy1982 (approved):

WireCommonComponents/AutomationHelper.swift
typfel: > If the code is refactored as suggestion, it would have difference result:

You're correct, I had a typo in my example (I updated it).

If `useAppCenterLaunchOption` is optional and we do:

```
if let useAppCenterLaunchOption = self.useAppCenterLaunchOption {
    return useAppCenterLaunchOption
} else {
    return UserDefaults.standard.bool(forKey: "UseAppCenter") ?? true
}
```

the `--user-app-center` will have priority if present and otherwise we rely on the value from user defaults.

typfel (changes requested):
See my new comment.

typfel (approved):

PR #4698, opened 2020-11-09
[Refactoring]: CallRouter Tests

marcomaddy1982:
## What's new in this PR?

PR for CallController/CallQualityController/ActiveCallRouter Tests


Wire-iOS Tests/Routers/CallRouterTests.swift
```diff
@@ -0,0 +1,105 @@
...
+    override func tearDown() {
+        sut = nil
+        callController = nil
+        super.tearDown()
+    }
+    
+    func testThatVersionAlertIsNotPresented_WhenCallStateIsTerminatedAndReasonIsNotOutdatedClient() {
+        // GIVEN
+        let callState: CallState = .terminating(reason: .canceled)
+        let conversation = ZMConversation()
```
typfel, line 44: I don't know why this works in other tests. I'd suggest to use the `MockConversation` and then convert it to a `ZMConversation` using `convertToRegularConversation`.

marcomaddy1982: Yep honestly I didn't understand as well. :-D. I found another way using `CoreDataFixture` but I will try your suggestion too. Thanks.

marcomaddy1982: Also using `convertToRegularConversation` it doesn't like. No idea honestly.

typfel: What do you mean doesn't like? You get a runtime error?

marcomaddy1982: Same error as before. The one we were discussing in the Chapter::iOS chat.

typfel: Doesn't make any sense.

Are doing this
```
let mockConversation = MockConversation.oneOnOneConversation()
mockConversation.convertToRegularConversation()
```

marcomaddy1982: Yes, exactly that!!!!

marcomaddy1982: Also for me doesn't make any sense. I double checked again after your comment (I could have messed something) but unfortunately nothing changed: same issue.

Wire-iOS/Sources/UserInterface/Calling/CallController/CallController.swift
```diff
@@ -42,6 +41,10 @@ final class CallController: NSObject {
     
     // MARK: - Public Implementation
     func updateActiveCallPresentationState() {
+        priorityCallConversation = ProcessInfo.processInfo.isRunningTests
```
marcomaddy1982, line 44: I don't like much this solution, so if you have something better in mind I will be happy to change this part.

marcomaddy1982: Also this could be a solution probably more readable. 

```swift
private var priorityCallConversation: ZMConversation? {
        set {
            test_priorityCallConversation = newValue
        }
        get {
            return ProcessInfo.processInfo.isRunningTests
                ? test_priorityCallConversation
                : ZMUserSession.shared()?.priorityCallConversation
        }
    }
    private var test_priorityCallConversation: ZMConversation? // This is used just for testing purpose
```

typfel: Another solution would be to introduce a protocol for the calling interface on `ZMUserSession`, which then can be mocked in the tests.

See https://github.com/wireapp/wire-ios-sync-engine/blob/7e61d97a1a563e42c62d92453b7ef4d478b86ce9/Source/UserSession/ZMUserSession%2BEncryptionAtRest.swift#L23 and https://github.com/wireapp/wire-ios/blob/b74e82a302c483956d9a21ad03873b80a23c3b81/Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift#L24

Wire-iOS Tests/Routers/ActiveCallRouterTests.swift
```diff
@@ -0,0 +1,360 @@
...
+//
+
+import Foundation
+import XCTest
+@testable import Wire
+
+final class ActiveCallRouterTests: XCTestCase, CoreDataFixtureTestHelper {
+    
+    var coreDataFixture: CoreDataFixture!
+    var sut: ActiveCallRouterMock!
```
David-Henner, line 26: I find it a bit confusing that the test class is named `ActiveCallRouterTests` when overall, it seems like this class is testing `CallController`'s behaviour

Wire-iOS Tests/Routers/ActiveCallRouterTests.swift
```diff
@@ -0,0 +1,360 @@
...
+        callController.testHelper_setPriorityCallConversation(nil)
+        
+        // WHEN
+        callController_callCenterDidChange(callState: callState, conversation: conversation)
+        
+        // THEN
+        XCTAssertTrue(sut.hideCallTopOverlayIsCalled)
+    }
+    
+    func testThatCallTopOverlayIsHidden_When() {
```
David-Henner, line 120: looks like the name is incomplete

Wire-iOS/Sources/Helpers/syncengine/ZMUserSession+Calling.swift
```diff
@@ -18,28 +18,41 @@
 
 import WireSyncEngine
 
+protocol UserSessionCallConversationSet {
```
marcomaddy1982, line 21: The name is not good at all. Any advice?

Wire-iOS Tests/Routers/ActiveCallRouterTests.swift
marcomaddy1982: Fine for you if I rename it in `CallControllerTests`?

Wire-iOS Tests/Routers/ActiveCallRouterTests.swift
marcomaddy1982: We don't need this test actually cause it want to test the same of the previous one. I will remove it. Thanks.

Wire-iOS Tests/CallControllerTests.swift
```diff
@@ -0,0 +1,217 @@
...
+    var userSession: MockUserSession!
+    
+    override func setUp() {
+        super.setUp()
+        coreDataFixture = CoreDataFixture()
+        router = ActiveCallRouterProtocolMock()
+        conversation = ZMConversation.createOtherUserConversation(moc: coreDataFixture.uiMOC,
+                                                                  otherUser: otherUser)
+        userSession = MockUserSession()
+        userSession.priorityCallConversation = conversation
```
johnxnguyen, line 38: I suggest to set the priority call conversation in the tests explicitly, because it's part of the "given".

Wire-iOS/Sources/Helpers/syncengine/ZMUserSession+Calling.swift
johnxnguyen: `CallConversationProvider`?

David-Henner: How about `CallConversationProviding` ?

David-Henner (approved):

Wire-iOS Tests/CallControllerTests.swift
marcomaddy1982: Done

Wire-iOS/Sources/Helpers/syncengine/ZMUserSession+Calling.swift
marcomaddy1982: Maybe a bit better `CallConversationProvider`. :-)

marcomaddy1982: Done to `CallConversationProvider`.

Wire-iOS/Sources/UserInterface/Calling/CallController/CallController.swift
```diff
@@ -23,7 +23,7 @@ final class CallController: NSObject {
 
     // MARK: - Public Implentation
     weak var router: ActiveCallRouterProtocol?
-    var userSession: UserSessionCallConversationSet?
+    var userSession: CallConversationProvider?
```
johnxnguyen, line 26: A little nitpick: it would be better to rename this to 

```swift
var callConversationProvider: CallConversationProvider?
```

johnxnguyen (approved):

PR #4709, opened 2020-11-13
[Fix] Analytic sessions are not being updated

johnxnguyen:
### Issues

Countly integration is broken since 3.68. The majority of iOS users have zero sessions recorded and therefore are not showing any recorded events. Some users however do show recorded events and sessions. 

### Causes

Manual session handling is enabled, however the sessions were never being explicitly updated nor explicitly ended. As a result, Countly was recognizing users but never received any events or session information from those users. In some cases though events would be uploaded, but (I believe) this would only happen when the self user changes (due to an account switch).

### Solutions

Explicitly start, update and end the session. The lifetime of the session matches the default behavior of Countly (when manual session handling is disabled), which means sessions begin when the app is foregrounded, updated every 60 seconds, and ended when the app is backgrounded.

### Notes

There is substantial refactoring in this PR to make this work and it has been done under time pressure, so there is certainly a lot more that I would refactor to clean up and make it more robust and better tested. For now, however, manual testing has indicated that it is working as expected.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
@@ -224,6 +224,9 @@ class SettingsCellDescriptorFactory {
         let triggerSlowSyncButton = SettingsButtonCellDescriptor(title: "Trigger slow sync", isDestructive: false, selectAction: DebugActions.triggerSlowSync)
         developerCellDescriptors.append(triggerSlowSyncButton)
 
+        let showAnalyticsIdentiferButton = SettingsButtonCellDescriptor(title: "What's my analytics id?", isDestructive: false, selectAction: DebugActions.showAnalyticsIdentifier)
+        developerCellDescriptors.append(showAnalyticsIdentiferButton)
+
         return SettingsGroupCellDescriptor(items: [SettingsSectionDescriptor(cellDescriptors:developerCellDescriptors)], title: title, icon: .robot)
     }
     
--- a/Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
@@ -145,6 +145,25 @@ enum DebugActions {
             ZMUserSession.shared()?.requestSlowSync()
         }
     }
+
+    static func showAnalyticsIdentifier(_ type: SettingsCellDescriptorType) {
+        guard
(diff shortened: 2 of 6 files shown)


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -224,6 +224,9 @@ class SettingsCellDescriptorFactory {
         let triggerSlowSyncButton = SettingsButtonCellDescriptor(title: "Trigger slow sync", isDestructive: false, selectAction: DebugActions.triggerSlowSync)
         developerCellDescriptors.append(triggerSlowSyncButton)
 
+        let showAnalyticsIdentiferButton = SettingsButtonCellDescriptor(title: "What's my analytics id?", isDestructive: false, selectAction: DebugActions.showAnalyticsIdentifier)
```
billypchan, line 227: the title needs to be localized?

Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
```diff
@@ -145,6 +145,25 @@ enum DebugActions {
...
+            let controller = UIApplication.shared.topmostViewController(onlyFullScreen: false),
+            let userSession = ZMUserSession.shared()
+        else {
+            return
+        }
+
+        let selfUser = ZMUser.selfUser(inUserSession: userSession)
+
+        let alert = UIAlertController(
+            title: "Analytics identifier",
```
billypchan, line 160: This is for debug only or also for public release?

typfel (approved):

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
johnxnguyen: No, it's only for debug purposes (in the developer menu).

Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
johnxnguyen: Debug only.

johnxnguyen:
Apologies @typfel I needed to push one more change after your approval.

billypchan (approved):

typfel (approved):

johnxnguyen:
I merged the PR with some failing tests to get a head start on triggering the RC build. The 4 failing tests are unrelated to analytics and may be flaky, I'll check.

PR #4711, opened 2020-11-13
Fix: app crash at zm conversation stable random participants

billypchan:
### Issues

redo of https://github.com/wireapp/wire-ios/pull/4697
App crash after applied above fix

### Causes

`generator.rand()` may return negative result. Cast to Int may cause a crash when accessing a negative index.

### Solutions

Keep the original logic and do not cast to Int.
Tested with real account and checked that the refactored code will produce the same result.

### DIFF
--- a/Wire-iOS Tests/RandomGeneratorFromDataTests.swift
+++ b/Wire-iOS Tests/RandomGeneratorFromDataTests.swift
@@ -20,7 +20,7 @@ import Foundation
 import XCTest
 @testable import Wire
 
-class RandomGeneratorFromDataTests: ZMSnapshotTestCase {
+final class RandomGeneratorFromDataTests: XCTestCase {
     func testThatItGeneratesAPseudorandom() {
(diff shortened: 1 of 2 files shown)


typfel (approved):
Looks OK,

But I'm wondering why we are re-implementing this when there's  a `shuffled()` method in the standard library https://developer.apple.com/documentation/swift/array/2994759-shuffled

The our custom random generator also looks too complicated and dangerous with that pointer magic. Couldn't we use the  `RandomNumberGenerator` protocol https://developer.apple.com/documentation/swift/randomnumbergenerator and use the `UUID` as the seed For an existing random number generator. 

Feel free to merge this first and tackle my proposal in separate PR.

billypchan:
@typfel I had a look at the random generator and I think it is a legacy code before Swift random generator is introduced. I would suggest align this to other platfor to see how they implement this.

David-Henner (approved):

PR #4718, opened 2020-11-18
[Feature] maximise / minimise video stream on double tap

David-Henner:
## What's new in this PR?

Maximise / minimise video stream on double tap.

### current behaviour

**given** we have a video grid containing one or more video tiles
**when** user double taps on a video tile
**then** the video tile's fill mode is toggled (either fills or fits the available space in the tile)

### new behaviour

**given** we have a video grid containing one or more video tiles
**when** user double taps on a video tile that is **not** maximised
**then** the video tile is maximised to **fit** the **video grid**

**given** we have a video grid containing one or more video tiles
**when** user double taps on a video tile that is maximised
**then** the video tile is minimised to allow all video tiles to be shown

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
@@ -116,8 +116,9 @@ final class VideoGridViewController: UIViewController {
     }
 
     // MARK: - Public Interface
-
-    public func switchFillMode(location: CGPoint) {
+    
+    public func handleDoubleTap(gesture: UIGestureRecognizer) {
+        let location = gesture.location(in: gridView)
         toggleMaximized(view: streamView(at: location))
     }
     
@@ -282,15 +283,10 @@ final class VideoGridViewController: UIViewController {
     }
     
     private func streamView(at location: CGPoint) -> BaseVideoPreviewView? {
-        let displayedStreams = dataSource.compactMap { $0.stream }
-        
-        return viewCache.values.lazy
-            .compactMap { $0 as? BaseVideoPreviewView }
-            .first {
-                let isShown = displayedStreams.contains($0.stream)
-                let isAtLocation = self.view.convert($0.frame, from: $0.superview).contains(location)
-                return isShown && isAtLocation
+        guard let indexPath = gridView.indexPathForItem(at: location) else {
+            return nil
         }
+        return streamView(for: dataSource[indexPath.row].stream) as? BaseVideoPreviewView
     }
 
     private func stream(with streamId: AVSClient) -> Stream? {
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
@@ -127,7 +127,7 @@ final class VideoGridViewController: UIViewController {
         let stream = view?.stream
         
         maximizedView = isMaximized(stream: stream) ? nil : view
-        (view as? VideoPreviewView)?.shouldFill = !isMaximized(stream: stream)
+        (view as? VideoPreviewView)?.isMaximized = isMaximized(stream: stream)
         updateVideoGrid(with: videoStreams)
     }
     
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
@@ -40,7 +40,7 @@ final class VideoGridViewController: UIViewController {
     }
 
     private var dataSource: [VideoStream] = []
-    private var maximizedStream: Stream?
+    private var maximizedView: BaseVideoPreviewView?
     private let gridView = GridView()
     private let thumbnailViewController = PinnableThumbnailViewController()
     private let networkConditionView = NetworkConditionIndicatorView()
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
```diff
@@ -110,11 +118,31 @@ final class VideoGridViewController: UIViewController {
...
+    }
+    
+    private func stream(at location: CGPoint) -> Stream? {
+        let displayedStreams = dataSource.compactMap { $0.stream }
+        
+        return viewCache.values.lazy
+            .compactMap { $0 as? BaseVideoPreviewView }
+            .filter { displayedStreams.contains($0.stream) }
+            .first(where: { self.view.convert($0.frame, from: $0.superview).contains(location) })?
+            .stream
```
David-Henner, line 131: not quite sure how performance heavy this operation is.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoPreviewView.swift
```diff
@@ -30,11 +30,15 @@ final class VideoPreviewView: BaseVideoPreviewView {
         }
     }
 
-    var shouldFill: Bool = true {
+    var isMaximized: Bool = false {
```
johnxnguyen, line 33: I would suggest a name like `isFocused` as it describes more the intention rather than the appearance. Additionally, "maximized" suggests that only one view can be maximum, but there's nothing to prevent several views having this property set to true. But it makes sense to have a group of focused views, even if for now we will only expect to have one at a time.

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
```diff
@@ -257,6 +280,18 @@ final class VideoGridViewController: UIViewController {
...
+        let displayedStreams = dataSource.compactMap { $0.stream }
+        
+        return viewCache.values.lazy
+            .compactMap { $0 as? BaseVideoPreviewView }
+            .first {
+                let isShown = displayedStreams.contains($0.stream)
+                let isAtLocation = self.view.convert($0.frame, from: $0.superview).contains(location)
+                return isShown && isAtLocation
+        }
+    }
```
johnxnguyen, line 294: Can't we use the collection view delegate methods to determine which tile was tapped?

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
```diff
@@ -32,14 +32,22 @@ final class VideoGridViewController: UIViewController {
...
-    private var videoStreams: [VideoStream] = []
+    private var videoStreams: [VideoStream] {
+        if let videoStream = configuration.videoStreams.first(where: { isMaximized(stream: $0.stream) }) {
+            return [videoStream]
+        }
+        return configuration.videoStreams
+    }
+
+    private var dataSource: [VideoStream] = []
+    private var maximizedView: BaseVideoPreviewView?
```
johnxnguyen, line 43: Just to explore the idea we briefly discussed: we could generalize the concept of displaying a single maximized view to displaying a specific selection of views with a filter. 

Suppose we have two video grid modes, where one is the normal mode (no filtering) and the other is the focused mode (filtering with some predicate):

```swift
enum GridMode {
  case normal
  case focused
}

private var gridMode: GridMode = ...

// This is all available video streams.
private var allVideoStreams: [VideoStream] = ...

// It's a func to indicate that it's not O(1) complexity.
private func visibleVideoStreams() -> [VideoStream] {
  switch gridMode {
    case .normal:
      return allVideoStreams
    case .focused:
      return allVideoStreams.filtered(\.isFocused)
  }
}
```

You would consider `allVideoStreams` to be your master datasource, but it's not the thing you would pass to the collection view. Instead you would pass the result of `visibleVideoStreams()`.

I think the benefit of this approach is that you minimize state that needs to be maintained and synchronized; you don't need to keep track of a `maximizedView`, nor check if a view satisfies `isMaximized(stream:)` because this state should already be inside the view (`isFocused`).

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
David-Henner: It should be possible, but I think the tap gesture recognisers on the parent view controller (single tap for overlay and double tap for toggling fill mode) are interfering with the collection view's. 
I'll look into it

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
```diff
@@ -108,13 +116,29 @@ final class VideoGridViewController: UIViewController {
...
+        updateVideoGrid(with: videoStreams)
+    }
+    
+    private func isMaximized(stream: Stream?) -> Bool {
+        guard
+            let streamId = stream?.streamId,
+            let maximizedStreamId = maximizedView?.stream.streamId
+        else { return false }
+        
+        return streamId == maximizedStreamId
```
johnxnguyen, line 141: The `maximizedView` property seems to only exist as a way to keep track of which stream id is considered maximized. Why not simply use the `isMaximized` property defined in the view? Indeed, there'd need to be some changes to make this work, but I think if we can reduce the amount of new state the better.

Imagine for instance that instead of defining `isMaximized` as a property of the view, you add it as a property to `VideoStream`. You would then replace the `stream` property on the view with `videoStream: VideoStream`, which would give you access to the `isMaximized` property so you can still adjust the fill mode.

With this in place, you could then do something like:

```swift
class BaseVideoPreviewView {
  var isMaximized: Bool {
    get { videoStream.isMaximized }
    set { videoStream.isMaximized = newValue }
  }
}

func toggleMaximized(view: BaseVideoPreviewView) {
  view.isMaximized.toggle()
  updateVideoGrid(with: videoStreams)
}

var videoStreams: [VideoStream] {
  if let maximizedStream = allVideoStreams.first(where: \.isMaximized) {
    return [maximizedStream]
  } else {
    return allVideoStreams
  }
}
```

David-Henner: Great suggestion! I'll make the changes

David-Henner:
@johnxnguyen please kindly pick a second reviewer :)

johnxnguyen:
@David-Henner oops forgot!

billypchan (approved):

marcomaddy1982 (approved):

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
David-Henner: After evaluation, doing this change is possible but it implies a bigger refactoring than initially estimated. If time and resources allows, I will tackle it in another PR.

johnxnguyen (approved):

PR #4719, opened 2020-11-19
[Fix] Network indicator not shown on conversation list

typfel:
### Issues

Network indicator is not shown on the conversation list.

### Causes

Constraints are breaking for the height of the indicator.

`TopBar` is connected to `contentContainer` and the same time `networkStatusViewController.view` is connected the `contentContainer`and the `topBarView`. This leaves zero space for the network indicator to grow when needed.

### Solutions

Detach the `TopBar` from the `contentContainer` and only leave it connected to the bottom of the network indicator, so that the network indicator can push the `TopBar` down when needed.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusViewController.swift
@@ -111,16 +111,6 @@ final class NetworkStatusViewController : UIViewController {
         }
     }
 
-    func createConstraintsInParentController(bottomView: UIView, controller: UIViewController) {
-        constrain(bottomView, controller.view, view) { bottomView, containerView, networkStatusViewControllerView in
-            networkStatusViewControllerView.leading == containerView.leading
-            networkStatusViewControllerView.trailing == containerView.trailing
-            bottomView.top == networkStatusViewControllerView.bottom
-        }
-        
-        view.topAnchor.constraint(equalTo: controller.safeTopAnchor).isActive = true
-    }
-
     func showOfflineAlert() {
         let offlineAlert = UIAlertController(title: "system_status_bar.no_internet.title".localized,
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusViewController.swift
```diff
@@ -111,16 +111,6 @@ final class NetworkStatusViewController : UIViewController {
         }
     }
 
-    func createConstraintsInParentController(bottomView: UIView, controller: UIViewController) {
```
billypchan, line 114: nice to remove mix using of Cartography and normal constraint creation!

Wire-iOS/Sources/UserInterface/Conversation/ConversationRootViewController.swift
```diff
@@ -86,24 +85,34 @@ final class ConversationRootViewController: UIViewController {
...
-            conversationViewControllerView.edges == contentView.edges
-        }
         
-        contentView.bottomAnchor.constraint(equalTo: self.safeBottomAnchor).isActive = true
+        [contentView,
+         conversationViewController.view,
+         networkStatusViewController.view
+        ].forEach() {
+            $0?.translatesAutoresizingMaskIntoConstraints = false
+        }
```
billypchan, line 94:
```suggestion
        ].disableAutoresizingMaskTranslation()
```

Wire-iOS/Sources/UserInterface/Components/NetworkStatus/NetworkStatusViewController.swift
typfel: Yes, mixing the two is very confusing since Cartography automatically disables `translatesAutoresizingMaskIntoConstraints`.

billypchan (approved):

KaterinaWire (approved):

PR #4721, opened 2020-11-19
clean up SelfUser.provider when deinit

billypchan:
### Issues

When testing with Xcode 12.2, the test crashes when setting `SelfUser.provider` to nil.

### Causes

`EXC_BAD_ACCESS` occurs when running the test.
It seems like `SelfUser.provider` is corrupted after accessing second time.

### Solutions

When `CoreDataFixture` deinit, clean `SelfUser.provider`. (It may had side effect to other tests, which may assume `SelfUser.provider` is set)
Refactor `SettingsCellDescriptorFactory.dataUsagePermissionsGroup()` to prevent getting default parameter argument from  `SelfUser.current`

### Discussion
Should we protect `SelfUser.provider` in a queue, to prevent racing condition?

### DIFF
--- a/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
+++ b/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
@@ -1,6 +1,6 @@
 //
 // Wire
-// Copyright (C) 2019 Wire Swiss GmbH
+// Copyright (C) 2020 Wire Swiss GmbH
 //
 // This program is free software: you can redistribute it and/or modify
 // it under the terms of the GNU General Public License as published by
--- a/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
+++ b/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
@@ -48,12 +48,12 @@ final class SettingsTableViewControllerSnapshotTests: XCTestCase {
 	}
 
     func testForSettingGroup() {
-        let group = settingsCellDescriptorFactory.settingsGroup()
+        let group = settingsCellDescriptorFactory.settingsGroup(isTeamMember: coreDataFixture.selfUser.isTeamMember)
         verify(group: group)
     }
 
     func testForAccountGroup() {
-        let group = settingsCellDescriptorFactory.accountGroup()
+        let group = settingsCellDescriptorFactory.accountGroup(isTeamMember: coreDataFixture.selfUser.isTeamMember)
         verify(group: group)
     }
 
--- a/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
+++ b/Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
@@ -60,7 +60,7 @@ final class SettingsTableViewControllerSnapshotTests: XCTestCase {
     func testForAccountGroupWithDisabledEditing() {
 		MockUserRight.isPermitted = false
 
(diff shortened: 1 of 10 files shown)


Wire-iOS/Sources/Analytics/Analytics.swift
```diff
@@ -48,7 +48,9 @@ final class Analytics: NSObject {
     @objc
     private func userSessionDidBecomeAvailable(_ note: Notification?) {
         callingTracker = AnalyticsCallingTracker(analytics: self)
-        selfUser = SelfUser.current
+        if SelfUser.provider != nil {
+            selfUser = SelfUser.current
+        }
```
johnxnguyen, line 53: This would be more clearly expressed as

```swift
if let selfUser = SelfUser.provider?.selfUser {
  ...
}
```

Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
```diff
@@ -48,19 +48,19 @@ final class SettingsTableViewControllerSnapshotTests: XCTestCase {
 	}
 
     func testForSettingGroup() {
-        let group = settingsCellDescriptorFactory.settingsGroup()
+        let group = settingsCellDescriptorFactory.settingsGroup(isTeamMember: coreDataFixture.selfUser.isTeamMember)
```
johnxnguyen, line 51: Instead of injecting the value all the way down, why not set the SelfUser.provider in the test?

Wire-iOS Tests/SwiftSnapshotTesting/CoreDataFixture.swift
```diff
@@ -166,6 +166,7 @@ final class CoreDataFixture {
     }
 
     deinit {
+        SelfUser.provider = nil
```
johnxnguyen, line 169: The idea of the self user provider was to ensure that no test assumed the existence of a self user, but had to ensure that they provided a value during set up (and consequently, tore it down afterwards). 

It would achieve this by hard crashing any test that resulted in a code path to `SelfUser.current` if it hadn't explicitly set the `provider`.

I guess it doesn't hurt to clear the provider in this deinit, but I wouldn't say it is the responsibility of `CoreDataFixture`.

billypchan: I agree that `CoreDataFixture` should not assign to `SelfUser` which may cause side effect to other tests. But we added ` SelfUser.provider = selfUserProvider` in the init method of `CoreDataFixture` so it should take the responsibility after created it.

Wire-iOS Tests/SettingsTableViewControllerSnapshotTests.swift
billypchan: Some tests get the `isTeamMember` property from injected self user. To prevent the confusion of injected self user and singleton self user, I would prefer injecting the value.

Wire-iOS/Sources/Analytics/Analytics.swift
billypchan: fixed

Wire-iOS Tests/SwiftSnapshotTesting/CoreDataFixture.swift
johnxnguyen: Ah since it sets it in init, then yes it should clear it in deinit. 👍

johnxnguyen (approved):

David-Henner (approved):

PR #4725, opened 2020-11-23
fix: crash when calling waitForGroupsToBeEmpty([MediaAssetCache.defaultImageCache.dispatchGroup])

billypchan:
### Issues

On Xcode12, test crashes when calling `waitForGroupsToBeEmpty([MediaAssetCache.defaultImageCache.dispatchGroup])`

### Causes

unknown, but it seems that multiple callings to that method may causing racing condition.

### Solutions

call that method in a queue

refactor `ConversationReplyCellTests` with swift snapshot framework, to prevent a crash with running FB snapshot testing.


Wire-iOS Tests/ConversationMessageCell/ConversationCellSnapshotTestCase.swift
```diff
@@ -75,58 +75,67 @@ class ConversationCellSnapshotTestCase: XCTestCase, CoreDataFixtureTestHelper {
...
-            
-            if waitForImagesToLoad {
-                XCTAssert(self.waitForGroupsToBeEmpty([MediaAssetCache.defaultImageCache.dispatchGroup]))
-            }
-            
+
             if waitForTextViewToLoad {
                 // We need to run the run loop for UITextView to highlight detected links
                 let delay = Date().addingTimeInterval(1)
                 RunLoop.main.run(until: delay)
```
billypchan, line 92: TODO: this line crash on Xcode 12.2

PR #4726, opened 2020-11-23
[Feature] Combine local and remote app lock configs

KaterinaWire:
### Issues
We should use AppLock configs from two sources ("session_manager" and BE).

### Causes
Because we need to use the new feature flag endpoints and support the old client settings.

### Solutions
- combine two AppLock configs in AppLock class;
- add observer with feature changes in `applicationDidBecomeActive()` event.

Needs releases with:

- https://github.com/wireapp/wire-ios-request-strategy/pull/256


WireCommonComponents/AppLock.swift
```diff
@@ -136,6 +168,26 @@ public class AppLock {
...
 
+extension AppLock {
+    private func configureObservers() {
+        NotificationCenter.default.addObserver(forName: FeatureController.featureConfigDidChange, object: nil, queue: nil) { [weak self] (note) in
+            self?.updateAppLockFeature(note)
+        }
+    }
+    
+    public func updateAppLockFeature(_ notification: Notification) {
+//        var applock: AppLockConfig?
```
marcomaddy1982, line 179: Do we still need this?

WireCommonComponents/AppLock.swift
```diff
@@ -18,16 +18,42 @@
...
+    public var appLockTimeout: UInt?
+}
+
 public class AppLock {
     // Returns true if user enabled the app lock feature.
     
-    public static var rules = AppLockRules.fromBundle()
-
+    static var rulesFromBundle = AppLockRules.fromBundle()
+    static var rulesFromCoreData: Feature.AppLock?
```
johnxnguyen, line 37: I have some concerns about this approach: if the self user changes, the app lock configuration may change because of different teams, but these static properties don't account for that.

Perhaps it would be more suitable to do something like:

```swift
static var rulesFromBundle: AppLockRules.fromBundle()

static var rules: AppLockRules {
  guard 
    let selfUser = SelfUser.provider?.selfUser,
    let appLock = selfUser?.team.feature(for: Feature.AppLock.self)
   else { 
    return rulesFromBundle 
  }
  
  var result = rulesFromBundle
  // override using appLock
  return result
}
```

WireCommonComponents/AppLock.swift
```diff
@@ -18,16 +18,42 @@
...
 
 private let zmLog = ZMSLog(tag: "UI")
 private let UserDefaultsDomainStateKey = "DomainStateKey"
 
+public struct AppLockConfig {
+    public var status: Feature.Status
+    public var forceAppLock: Bool?
+    public var appLockTimeout: UInt?
+}
+
```
johnxnguyen, line 32: What is this being used for?

WireCommonComponents/AppLock.swift
KaterinaWire: SelfUser.provider?.selfUser is a `UserType` and it doesn't have a `team` property.

WireCommonComponents/AppLock.swift
KaterinaWire: Removed

WireCommonComponents/AppLock.swift
```diff
@@ -26,8 +27,25 @@ private let UserDefaultsDomainStateKey = "DomainStateKey"
 public class AppLock {
     // Returns true if user enabled the app lock feature.
     
-    public static var rules = AppLockRules.fromBundle()
+    internal static var rulesFromBundle = AppLockRules.fromBundle()
+    internal static var rulesFromCoreData: FeatureConfigResponse<Feature.AppLock>?
```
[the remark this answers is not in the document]
KaterinaWire: Fixed

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -98,6 +99,7 @@ extension AppStateCalculator: ApplicationStateObserving {
     
     func applicationDidBecomeActive() {
         hasEnteredForeground = true
+        configureObservers()
```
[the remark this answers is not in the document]
KaterinaWire: Removed

WireCommonComponents/AppLock.swift
KaterinaWire: It is not necessary now. I forgot to remove it.

PR #4728, opened 2020-11-24
[Feature] Add notification to fetch feature config

KaterinaWire:
## What's new in this PR?

Trigger app lock config fetch when app becomes active.

###  Dependencies
https://wearezeta.atlassian.net/browse/SQSERVICES-83


johnxnguyen (approved):

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -98,6 +99,7 @@ extension AppStateCalculator: ApplicationStateObserving {
     
     func applicationDidBecomeActive() {
         hasEnteredForeground = true
+        notifyToFetchFeatureConfig()
```
johnxnguyen, line 102: @marcomaddy1982 is this a good place to put this?

marcomaddy1982: So the idea is that the `AppStateCalculator` would just calculate the `appState` and not doing any there business logic. You could try to do that in the `AppRootRouter`. Anyway, the `AppRootRouter` it is not completed and one day the idea is moving out from there all the business logic not related to navigation.

johnxnguyen: Yup makes sense, so we thought another place would be in the `AppLockPresenter`.

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -174,6 +176,13 @@ extension AppStateCalculator: AuthenticationCoordinatorDelegate {
     }
 }
 
+// MARK: - Post notifications
+extension AppStateCalculator {
+    private func notifyToFetchFeatureConfig() {
+        NotificationCenter.default.post(name: FeatureConfigRequestStrategy.needsToFetchFeatureConfigNotificationName, object: nil)
```
johnxnguyen, line 182: As discussed, you may need to post a `NotificationInContext` object because this is the kind of notification being observed in the request strategy.

johnxnguyen (changes requested):

Wire-iOS/Sources/AppStateCalculator.swift
marcomaddy1982: Yes, I guess I saw in the other PR. :-)

Wire-iOS/Sources/AppStateCalculator.swift
KaterinaWire: I will keep the same notification for now

Wire-iOS/Sources/AppStateCalculator.swift
KaterinaWire: @johnxnguyen We talked about adding a feature change observer to the `AppLockPresenter`.
But post notification about fetching feature we still need do it during the `applicationDidBecomeActive` event. 
I put it in `AppRootRouter` , as @marcomaddy1982 suggested.

johnxnguyen (approved):
If we use a downstream object sync for fetching single features, we would need to change this. But for now 👍

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -418,6 +419,7 @@ extension AppRootRouter: ApplicationStateObserving {
     func applicationDidBecomeActive() {
         updateOverlayWindowFrame()
         teamMetadataRefresher.triggerRefreshIfNeeded()
+        updateTeamFeature()
```
typfel, line 422: Is the long term plan to refetch all feature config every time the app becomes active

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -314,6 +314,7 @@ extension AppRootRouter {
         }
         
         self.authenticatedRouter = authenticatedRouter
+        updateTeamFeature()
```
typfel, line 317: You want to fetch the feature configs after logging/registration? We usually do this in our slow sync.

typfel (changes requested):

Wire-iOS/Sources/AppRootRouter.swift
KaterinaWire: It's a temp solution while we do not have update events from the BE.

Wire-iOS/Sources/AppRootRouter.swift
KaterinaWire: It's a temp solution.

typfel (approved):

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -466,6 +472,12 @@ extension AppRootRouter: AudioPermissionsObserving {
     }
 }
 
+// MARK: - Post notifications
+extension AppRootRouter {
```
marcomaddy1982, line 476: So I suppose that also this a temp solution?

KaterinaWire: Yes, I added a comment to remove it in the future

PR #4729, opened 2020-11-24
[Feature] Hide / show app lock

KaterinaWire:
## What's new in this PR?

Show App-lock feature in Wire settings based on BE App-lock status value.

### Dependencies

https://wearezeta.atlassian.net/browse/SQSERVICES-56


WireCommonComponents/AppLock.swift
```diff
@@ -171,6 +171,7 @@ public struct AppLockRules: Decodable {
     public let useCustomCodeInsteadOfAccountPassword: Bool
     public let forceAppLock: Bool
     public let appLockTimeout: UInt
+    public var isEnabled: Bool?
```
johnxnguyen, line 174: I think it would be fine to make `isEnabled` non optional and default to `true`. We should think of these default values as being applicable for non team users too, because in their case they will never get the config from the BE.

KaterinaWire: Updated

johnxnguyen (changes requested):
It would be good to have a test for this.

KaterinaWire:
> It would be good to have a test for this.

@johnxnguyen 
It's no so easy, because the logic of show or hide App Lock feature is based on the App Lock rules, which I can't mock.

typfel (approved):

marcomaddy1982:
@KaterinaWire since @johnxnguyen is off I took the freedom to review your PR and approve it. There is just test failing `testForOptionsGroupScrollToBottom` that could be a flaky tests.

marcomaddy1982 (approved):

johnxnguyen (approved):

PR #4730, opened 2020-11-24
[Fix]: On iPad no Spinner and Network Alert is shown when logout

marcomaddy1982:
### Issues

on iPad when there is no connectivity and the user is logging out, after entering password and tapping OK dialog is closed, but nothing happens, no spinner is shown and no alert is shown.

### Causes

The `topMostViewController` (`KeyboardAvoidingViewController`) is not of type `SpinnerCapable`

### Solutions

`KeyboardAvoidingViewController` is now `SpinnerCapable`

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/KeyboardAvoidingViewController.swift
@@ -19,7 +19,10 @@
 import Foundation
 import UIKit
 
-class KeyboardAvoidingViewController: UIViewController {
+class KeyboardAvoidingViewController: UIViewController, SpinnerCapable {
+
+    //MARK: SpinnerCapable
+    var dismissSpinner: SpinnerCompletion?
     
     let viewController: UIViewController
     var disabledWhenInsidePopover: Bool = false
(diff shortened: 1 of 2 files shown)


typfel (changes requested):
Would it be an option to make `KeyboardAvoidingViewController` implement `SpinnerCapable`? I guess we could avoid the if statement then.

David-Henner (commented):
I agree with Jacob's comment. 
If not possible, then I think it would make sense to extract the `if` statement out of the method for clarity

marcomaddy1982:
> SpinnerCapable



> Would it be an option to make `KeyboardAvoidingViewController` implement `SpinnerCapable`? I guess we could avoid the if statement then.

Actually this was my first idea but then I opted for the solution I've implemented. Cause I found not really logic having a `KeyboardAvoidingViewController` as `SpinnerCapable` :-). Anyway, of course we could try to go for this other solution (if it works cause I didn't try).

marcomaddy1982:
@typfel @David-Henner: `SpinnerCapable` works.

typfel (approved):
<deleted>

David-Henner (approved):

typfel:
@marcomaddy1982 ready to merge this?

PR #4731, opened 2020-11-24
remove UserImageView strong ref to Managed object

billypchan:
### Issues

On Xcode 12, test(ConversationTextMessageTests) crashes when `UserImageView ` dealloc
### Causes

App crash at when NSManagedObject dealloc

### Solutions

weaken the ref of `ZMUser` in `UserImageView`


typfel (changes requested):
I think the `UserImageView` should hold a strong reference to the user object. We want to release all instances of `ZMUser ` we should tear down the views.

These issues in the tests I suspect is just a side effect of another problem.

billypchan:
check that this PR does not fix the issue.

PR #4743, opened 2020-11-27
[Refactor] Update SharingSession initializer

KaterinaWire:
## What's new in this PR?

Parse `session_manager.json` file and pass the config file to the `SharingSession`.

This PR is one in a series for refactoring for AppLock. Full description of refactoring you can find here:
wireapp/wire-ios-data-model

Needs releases with:

- [https://github.com/wireapp/wire-ios-share-engine/pull/187] 
- [https://github.com/wireapp/wire-ios-data-model/pull/1078]


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -147,14 +147,22 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
+            let accountIdentifier = account?.userIdentifier,
+            let url = Bundle.main.url(forResource: "session_manager", withExtension: "json"),
+            let configuration = SharingSessionConfiguration.load(from: url)
             else { return }
-
+        
+        let appLockConfig = AppLockController.Config(useBiometricsOrAccountPassword: configuration.useBiometricsOrAccountPassword,
+                                                     useCustomCodeInsteadOfAccountPassword: configuration.useCustomCodeInsteadOfAccountPassword,
+                                                     forceAppLock: configuration.forceAppLock,
+                                                     timeOut: configuration.appLockTimeout)
```
johnxnguyen, line 158: The `SharingSessionConfiguration` appears to only be a way to load the json and pass it to create the app lock config. Why don't we use the existing `AppLockRules` struct instead?

KaterinaWire: I thought we will remove `AppLock` and `AppLockRules` from the UI.
But yes, I can use `AppLockRules` and it will be used only by Share extension.

johnxnguyen (approved):

Wire-iOS Share Extension/ShareExtensionViewController.swift
KaterinaWire: Replaced it with `AppLockRules`

typfel (approved):

PR #4748, opened 2020-12-02
[Refactor] Use Applock in user session

KaterinaWire:
## What's new in this PR?

Instead of using the static properties of the `AppLock` class, we will use the applock instance of `UsserSession`.

This PR is one in a series for refactoring for AppLock. Full description of refactoring you can find here:
wireapp/wire-ios-data-model/pull/1078

Needs releases with:

- [ wireapp/wire-ios-sync-engine/pull/1351 ]


Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
```diff
@@ -52,34 +56,42 @@ private final class UserSessionMock: AppLockInteractorUserSession {
...
         completion(result)
     }
+    
+    init(moc: NSManagedObjectContext) {
+        let config = AppLockController.Config(useBiometricsOrAccountPassword: false,
+                                                     useCustomCodeInsteadOfAccountPassword: false,
+                                                     forceAppLock: false,
+                                                     timeOut: 900)
+        self.appLockController = AppLockMock(config: config, selfUser: ZMUser.selfUser(in: moc))
+    }
```
johnxnguyen, line 66: You only need the self user so it would be better to just pass in that through the initializer rather than a context.

Wire-iOS/Sources/Authentication/Event Handlers/Initial Sync/AuthenticationInitialSyncEventHandler.swift
```diff
@@ -49,14 +49,7 @@ final class AuthenticationInitialSyncEventHandler: NSObject, AuthenticationEvent
...
             actions.append(.transition(nextStep, mode: .reset))
         } else {
-            // append passcode step if new registered or first time login but passcode is not store.(Upgrade from pervious version)
-            if AppLock.isCustomPasscodeNotSet &&
-               (AppLock.rules.forceAppLock ||
-                AppLock.isActive) {
-                actions.append(.transition(.passcodeSetup, mode: .reset))
-            } else {
-                actions.append(postAction)
-            }
```
johnxnguyen, line 59: Note, we are removing the the custom passcode flow here, but still need to do it elsewhere. It would be good to create a ticket so we don't forget about this.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
```diff
@@ -52,34 +56,42 @@ private final class UserSessionMock: AppLockInteractorUserSession {
...
+final class AppLockInteractorTests: ZMSnapshotTestCase {
     var sut: AppLockInteractor!
     private var appLockInteractorOutputMock: AppLockInteractorOutputMock!
     private var userSessionMock: UserSessionMock!
     
     override func setUp() {
         super.setUp()
         appLockInteractorOutputMock = AppLockInteractorOutputMock()
-        userSessionMock = UserSessionMock()
+        userSessionMock = UserSessionMock(moc: uiMOC)
```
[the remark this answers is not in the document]
typfel: Some options: 
- Change the `AppLockController` to take a `UserType` instead of `ZMUser`
- Expose protocol type of the `AppLockController` on the ZMUserSession.

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+AppLock.swift
```diff
@@ -0,0 +1,45 @@
...
+//
+
+import WireSyncEngine
+
+// MARK: - AppLock helper
+extension SettingsPropertyFactory {
+    
+    private var appLock: AppLockController? {
+        return ZMUserSession.shared()?.appLockController
+    }
```
johnxnguyen, line 26: This is one of those cases where we probably can reasonably expect there to always be a user session available, but the use of singletons doesn't allow us to be certain about this. I wonder if we can initialize an instance of `SettingsPropertyFactory` with the current user session. @typfel thoughts?

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -53,26 +55,56 @@ final class AppLockInteractor {
     }
     
     var appState: AppState?
+
+    var appLock: AppLockController? {
```
johnxnguyen, line 59: As a note, if we had a `locked` app state, we could create the app lock module only when necessary and inject an instance of `AppLockController`, then we wouldn't have to worry about optionals. But that's an improvement to think about later.

Wire-iOS/Sources/Authentication/Event Handlers/Initial Sync/AuthenticationInitialSyncEventHandler.swift
```diff
@@ -49,14 +49,7 @@ final class AuthenticationInitialSyncEventHandler: NSObject, AuthenticationEvent
         if let nextStep = nextRegistrationStep {
             actions.append(.transition(nextStep, mode: .reset))
         } else {
-            // append passcode step if new registered or first time login but passcode is not store.(Upgrade from pervious version)
```
typfel, line 52: You decided not to ask the user to set passcode during login/registration?

typfel (changes requested):

johnxnguyen (approved):

Wire-iOS/Sources/Authentication/Event Handlers/Initial Sync/AuthenticationInitialSyncEventHandler.swift
typfel: John answered this question.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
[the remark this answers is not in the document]
KaterinaWire: The reason I used `ZMUser` instead of `UserType` in `AppLockController` is because we need to get a `team` from the user. The `UserType` protocol does not provide a `team`.

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+AppLock.swift
typfel: Yes, that would make sense.

Seems like we already initialise the `SettingsPropertyFactory` with the current user session, it's abstracted by the `ZMUserSessionInterface`though, which doesn't expose much functionality. It was probably added for mocking purposes in the tests.

I would propose to use the same strategy as in the interactors and inject a composed protocol of all the necessary user session protocols.

WireCommonComponents/AppLock.swift
```diff
@@ -17,124 +17,6 @@
...
-            zmLog.error("Local authentication error: \(String(describing: error?.localizedDescription))")
-        }
-    }
-    
-    public class func persistBiometrics() {
-        BiometricsState.persist()
-    }
-}
 
 public struct AppLockRules: Decodable {
```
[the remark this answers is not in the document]
KaterinaWire: Yes, we use it now for `ShareExtension`, because we need to provide a bundle configuration to initialise the `ApplockController`.

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
typfel: Wasn't this `locked` app state already introduced by another PR from @KaterinaNikolaenko that was never merged. Will you revisit that PR soon and then also make this improvement?

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
KaterinaWire: Fixed

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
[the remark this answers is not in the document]
typfel: Then option 2 would be the way to go.

or you could expose the `team` on the `UserType` and update `TeamType` to expose the properties you need. But be overkill since you don't want to test the internals of `AppLockController` in the UI...

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
KaterinaWire: Yes, we discussed with @johnxnguyen to have that PR become part of this refactoring. But it will take time to fix all merge conflicts.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
```diff
@@ -52,34 +56,42 @@ private final class UserSessionMock: AppLockInteractorUserSession {
     func verify(password: String, completion: @escaping (VerifyPasswordResult?) -> Void) {
         completion(result)
     }
+    
+    init(selfUser: ZMUser) {
+        let config = AppLockController.Config(useBiometricsOrAccountPassword: false,
+                                                     useCustomCodeInsteadOfAccountPassword: false,
```
marcomaddy1982, line 62: It is not align well here.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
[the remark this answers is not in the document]
KaterinaWire: Do I understand you correctly about the second option:
instead of heaving 
`var appLockController: AppLockController` in ZMUserSession

to have something like this:

```
protocol AppLockType {
    var isActive: Bool { get set }
    var lastUnlockedDate: Date { get set }
    var isCustomPasscodeNotSet: Bool { get }
    var config: AppLockController.Config { get }
    
    func persistBiometrics()
}

var appLockController: AppLockType?

```

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+AppLock.swift
KaterinaWire: Yes, we initialise `SettingsPropertyFactory`  with the current user session, but `userSession` is optional too.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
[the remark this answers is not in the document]
typfel: Exactly

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+AppLock.swift
typfel: > Yes, we initialise `SettingsPropertyFactory` with the current user session, but `userSession` is optional too.

We can change it to be not optional. The unwrapping can happen higher up somewhere.

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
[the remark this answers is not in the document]
KaterinaWire: Implemented

Wire-iOS Tests/AppLock/AppLockInteractorTests.swift
KaterinaWire: Fixed

marcomaddy1982:
@KaterinaWire since @johnxnguyen  is off I took the freedom to review your PR and approve it.

marcomaddy1982 (approved):

Wire-iOS/Sources/Authentication/Event Handlers/Initial Sync/AuthenticationInitialSyncEventHandler.swift
KaterinaWire: Done 
https://wearezeta.atlassian.net/browse/SQSERVICES-103

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory+AppLock.swift
KaterinaWire: I would do this refactoring later, because we also initialise `SettingsPropertyFactory` with `nil`parameters in `AVSMediaManager`. These changes will covert some another classes not related to this PR.

I created a ticket not to forget about it:
https://wearezeta.atlassian.net/browse/SQSERVICES-104

PR #4749, opened 2020-12-02
[Feature] add pulsing ability to microphone icon

David-Henner:
## What's new in this PR?

- Created a `PulsingIconImageView` to allow any icon to pulse.
- Updated the microphone icon to be of type `PulsingIconImageView` in participant list & participant details view

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/PulsingIconImageView.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/PulsingIconImageView.swift
@@ -1,0 +1,75 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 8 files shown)


typfel (approved):

Wire-iOS/Sources/UserInterface/Components/Views/IconImageView/PulsingIconImageView.swift
```diff
@@ -0,0 +1,75 @@
...
+
+protocol PulsingIconImageStyle {
+    var shouldPulse: Bool { get }
+}
+
+class PulsingIconImageView: IconImageView {
+        
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        NotificationCenter.default.addObserver(
```
marcomaddy1982, line 30: For this we introduced in a PR the `ApplicationStateObserving` that does all the work to observer, remove the observer (that I don't see here) and dispatch call backs for the application states. You can have a look into the `AppRootRouter` is you are interested to use it.

David-Henner: If my understanding is correct, it's not necessary to remove the observer starting from iOS 9. It should get cleaned up automatically. 

Anyways, thanks for the suggestion! I will look into it

marcomaddy1982: You are right, I forgot about that!!

marcomaddy1982 (approved):

PR #4752, opened 2020-12-03
xcode12 catalina snapshots update

billypchan:
## What's new in this PR?

Update snapshots generated with iOS simulator 14.2


carthageXcode12.sh
```diff
@@ -0,0 +1,22 @@
+     
+
+#!/usr/bin/env bash
+
+# carthage.sh
+# Usage example: ./carthage.sh build --platform iOS
```
johnxnguyen, line 6: So does this mean we have to run this script instead of `carthage bootstrap` each time?

johnxnguyen (commented):
Can you tell me a bit more about the test plan and how it's configured?

carthageXcode12.sh
billypchan: Yes, it is a workaround from Carthage issue discussion and still wait for the update
This PR is focused on images update, I will remove this change.

PR #4759, opened 2020-12-08
clean up tests and do not set up observer when tests

billypchan:
## What's new in this PR?

- Do not create unnecessary observers when testing
- Clean up observer tokens when deinit
- Clean up tests which do not release resources
- Clean up redundant checkers

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
@@ -268,8 +268,10 @@ final class ConversationInputBarViewController: UIViewController,
 
         super.init(nibName: nil, bundle: nil)
 
-        conversationObserverToken = ConversationChangeInfo.add(observer:self, for: conversation)
-        typingObserverToken = conversation.addTypingObserver(self)
+        if !ProcessInfo.processInfo.isRunningTests {
+            conversationObserverToken = ConversationChangeInfo.add(observer:self, for: conversation)
+            typingObserverToken = conversation.addTypingObserver(self)
+        }
 
         setupNotificationCenter()
         setupInputLanguageObserver()
@@ -280,6 +282,14 @@ final class ConversationInputBarViewController: UIViewController,
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    
+    deinit {
+        textfieldObserverToken = nil
+        conversationObserverToken = nil
+        typingObserverToken = nil
+        conversationObserverToken = nil
+        userObserverToken = nil
+    }///TODO: test crash
 
     // MARK: - view life cycle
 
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
```diff
@@ -308,13 +318,15 @@ final class ConversationInputBarViewController: UIViewController,
...
+                conversationObserverToken = ConversationChangeInfo.add(observer:self, for: conversation)
+            }
 
-        if let connectedUser = conversation.connectedUser,
-            let userSession = ZMUserSession.shared() {
-            userObserverToken = UserChangeInfo.add(observer:self, for: connectedUser, in: userSession)
+            if let connectedUser = conversation.connectedUser,
+                let userSession = ZMUserSession.shared() {
+                userObserverToken = UserChangeInfo.add(observer:self, for: connectedUser, in: userSession)
+            }
```
David-Henner, line 329: If the observer registration is not required to be done before the lines below (where we call the `update...()` methods) then I think we could move observer registration to the end of `viewDidLoad`. Then assess if tests are running in a guard statement and return early if they are. That way we can avoid nesting `if` statements

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -45,20 +43,26 @@ final class BackgroundViewController: UIViewController {
         self.userSession = userSession
         super.init(nibName: .none, bundle: .none)
         
-        if let userSession = userSession {
-            self.userObserverToken = UserChangeInfo.add(observer: self, for: user, in: userSession)
+        if !ProcessInfo.processInfo.isRunningTests {
```
David-Henner, line 46: I'd also change this with a guard statement

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController.swift
billypchan: I moved set up observer code to a new method.

David-Henner: even better 👍

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
billypchan: also moved to a seperated method

David-Henner (approved):

Wire-iOS/Sources/UserInterface/ConversationList/BackgroundViewController.swift
```diff
@@ -59,6 +62,10 @@ final class BackgroundViewController: UIViewController {
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    
+    deinit {
```
typfel, line 66: Also redundant

typfel (changes requested):
I think the nilling in the  `deinit` is unnecessary and you anyway not starting any observers in the tests.

billypchan:
@typfel removed token clean up in deinit

typfel (approved):

PR #4765, opened 2020-12-09
Fix: remove unnecessary member variables from background view controller

billypchan:
### Issues

On Xcode 12, tests crash sometimes and the stack trace stops at background view controller

### Causes

These are some unknown reason of crashing  when dealloc ZMUser and ZMSession
Dispatch group is sometimes not released, any it is only used in tests.

### Solutions

Remove unnecessary member variables ZMUser and ZMSession
Use closure instead of dispatch group to monitor user image is loaded in tests

### discussion
It would be easy to do snapshot tests and test update logics if we refactor this VC to VIPER


marcomaddy1982 (approved):

typfel (approved):

billypchan:
@typfel any bugs for revented this PR?

PR #4766, opened 2020-12-09
fix: CallInfoConfigurationTests without ZMUser

billypchan:
### Issues

On XCode 12, test suite may crash at `CallInfoViewControllerAccessoryType` is deallocing

### Causes

Unknown, but we should not use `ZMUser` for tests to prevent any side effect

### Solutions

Use `MockUserType` instead of `ZMUser` for testing.
Inject self user to `canUpgradeToVideo` method to match the mock self user in the tests

### dependency

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1357


Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
```diff
@@ -22,8 +22,21 @@ import WireDataModel
...
+        case (.incoming(let lhsUser), .incoming(let rhsUser)):
+            return (lhsUser as? NSObject) == (rhsUser as? NSObject)
+        case (.outgoing(let lhsUser), .outgoing(let rhsUser)):
+            return (lhsUser as? NSObject) == (rhsUser as? NSObject)
+        case (.none, .none):
+            return true
+        default:
+            return false
+        }
+    }
```
David-Henner, line 39: is this necessary because of the change from `ZMUser` to `UserType` ?

billypchan: I will add a similar change to https://github.com/wireapp/wire-ios-sync-engine/pull/1358/files, to remove the unnecessary casting.

billypchan: @David-Henner the equal checker is needed sine UserType is not necessary to be hashable.

David-Henner (approved):

Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
```diff
@@ -22,8 +22,23 @@ import WireDataModel
...
+                   (lhsUser as? NSObject) == (rhsUser as? NSObject)
+        case (.outgoing(let lhsUser), .outgoing(let rhsUser)):
+            return (lhsUser as? ZMUser) == (rhsUser as? ZMUser) ||
+                (lhsUser as? NSObject) == (rhsUser as? NSObject)
+        case (.none, .none):
+            return true
+        default:
+            return false
+        }
+    }
```
johnxnguyen, line 41: I think there's an issue with the way you're casting and checking equality. Consider

```swift
let user1: ZMSearchUser = ...
let user2: ZMSearchUser = ...

// This evaluates to true, but it shouldn't
.incoming(user1) == .incoming(user2)
```

In this case, casting both `lhsUser` and `rhsUser` to `ZMUser` will fail, so the expression will be `nil == nil`. I think what you want to do is assert that if the left side is a `ZMUser`, then try to cast the right size as `ZMUser`. 

Similarly with `NSObject`, if the user type doesn't inherit from `NSObject` then you may get in the same situation of comparing `nil == nil`.

johnxnguyen (changes requested):

Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
johnxnguyen: I see that `UserType` extends `NSObjectProtocol`, so you can expect the cast to `NSObject` to always succeed.

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
billypchan: To get rid of the compare method, I use generic for the associated values. Since both `ZMUser` and `MockUserType` are subclasses of `NSObject`, we can use `NSObject` as the common generic type.

Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
```diff
@@ -20,10 +20,10 @@ import Foundation
 import UIKit
 import WireDataModel
 
-enum CallDegradationState: Equatable {
+enum CallDegradationState<User: Equatable>: Equatable {
     case none
-    case incoming(degradedUser: ZMUser?)
-    case outgoing(degradedUser: ZMUser?)
+    case incoming(degradedUser: User?)
+    case outgoing(degradedUser: User?)
```
johnxnguyen, line 26: This approach means there's a lot of casting since we lose all information about what the original user class was. Did you consider to use `HashBox` or something like that to give equatable conformance to `UserType`?

billypchan: Refactored this PR with HashBoxed usertype

David-Henner (approved):

David-Henner:
There are two approvals on this PR but I would suggest to wait for @johnxnguyen 's review

Wire-iOS/Sources/UserInterface/Calling/CallDegradationViewController/CallDegradationController.swift
```diff
@@ -52,7 +52,7 @@ final class CallDegradationController: UIViewController {
     fileprivate func updateState() {
         switch state {
         case .outgoing(degradedUser: let degradeduser):
-            visibleAlertController = UIAlertController.degradedCall(degradedUser: degradeduser, confirmationBlock: { [weak self] (continueDegradedCall) in
+            visibleAlertController = UIAlertController.degradedCall(degradedUser: degradeduser as? UserType, confirmationBlock: { [weak self] (continueDegradedCall) in
```
johnxnguyen, line 55:
```suggestion
            visibleAlertController = UIAlertController.degradedCall(degradedUser: degradeduser?.value, confirmationBlock: { [weak self] (continueDegradedCall) in
```

johnxnguyen (approved):

PR #4770, opened 2020-12-14
[Refactor] Update Applock flow

KaterinaWire:
## What's new in this PR?

According to the design flow
(https://zpl.io/bArgMpp or zpl://screen?sid=5fbbd5b02956135e8e9f3849&pid=5f0f1143f783828d19da7721)

we should provide a screen with a custom passcode


Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -494,12 +494,11 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
         if sharingSession.encryptMessagesAtRest {
             scenario = .databaseLock
         } else {
-            scenario = .screenLock(requireBiometrics: sharingSession.appLockController.config.useBiometricsOrAccountPassword,
-                                   grantAccessIfPolicyCannotBeEvaluated: !sharingSession.appLockController.config.forceAppLock)
```
johnxnguyen, line 498: What was `grantAccessIfPolicyCannotBeEvaluated` being used for? Was this to allow the app lock to be bypassed because there was no device passcode set and no requirement to use a custom / account passcode?

KaterinaWire: It was a code in  `evaluateAuthentication(scenario: AuthenticationScenario,...) `method.
```
if canEvaluatePolicy {
            context.evaluatePolicy(scenario.policy, localizedReason: description, reply: { (success, error) -> Void in
                var authResult: AuthenticationResult = success ? .granted : .denied
            
                if scenario.supportsUserFallback, let laError = error as? LAError, laError.code == .userFallback {
                    authResult = .needAccountPassword
                }
                
                callback(authResult, context)
            })
        } else {
            // If the policy can't be evaluated automatically grant access unless app lock
            // is a requirement to run the app. This will for example allow a user to access
            // the app if he/she has disabled his/her passcode.
            callback(scenario.grantAccessIfPolicyCannotBeEvaluated ? .granted : .unavailable, context)
            zmLog.error("Local authentication error: \(String(describing: error?.localizedDescription))")
        }
```

```
var grantAccessIfPolicyCannotBeEvaluated: Bool {
            if case .screenLock(requireBiometrics: _, grantAccessIfPolicyCannotBeEvaluated: true) = self {
                return true
            }
            return false
  }
```

johnxnguyen: So why is it gone now? Because we will always have the custom passcode as a fallback?

KaterinaWire: Yes

johnxnguyen: So, if I disable my passcode while app lock is enabled, then Wire will ask me to create a custom passcode when I return to the app?

KaterinaWire: Yes.
If you disable biometrics and passcode,  then Wire will ask you to create a custom passcode.
Here is a code from the DM:

```
let canEvaluatePolicy = context.canEvaluatePolicy(scenario.policy, error: &error)
        
        if scenario.supportsUserFallback && (BiometricsState.biometricsChanged(in: context) || !canEvaluatePolicy) {
            callback(.needCustomPasscode, context)
            return
        }
```
I will send videos in 1:1 chat.

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory+Options.swift
```diff
@@ -230,9 +230,7 @@ extension SettingsCellDescriptorFactory {
             cellDescriptors: [appLockToggle],
             headerGenerator: { return nil },
             footerGenerator: { return self.appLockSectionSubtitle },
-            visibilityAction: { _ in
-                return LAContext().canEvaluatePolicy(LAPolicy.deviceOwnerAuthentication, error: nil)
-            }
+            visibilityAction: .none
```
David-Henner, line 233: `visibilityAction` default value is `.none` so I think you can remove this line

KaterinaWire: Yes, makes sense.
Done.

David-Henner (approved):

PR #4775, opened 2020-12-16
bump sync engine: change CallParticipant.user type to UserType

billypchan:
## What's new in this PR?

Apply necessary change for `CallParticipant`.`user` type change.

https://github.com/wireapp/wire-ios-sync-engine/pull/1358

Further update will continue in https://github.com/wireapp/wire-ios/pull/4766


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
```diff
@@ -264,9 +264,9 @@ extension VoiceChannel {
     var degradationState: CallDegradationState {
         switch state {
         case .incoming(video: _, shouldRing: _, degraded: true):
-            return .incoming(degradedUser: firstDegradedUser)
+            return .incoming(degradedUser: firstDegradedUser as! ZMUser)
```
johnxnguyen, line 267: I thought the idea was to accept a `UserType`, are you not going to do this anymore?

johnxnguyen:
@billypchan please add a more descriptive title and description, it's not really clear what the intention of this PR is.

David-Henner (commented):
Shouldn't the UI be updated to use `UserType` instead of casting to `ZMUser` ? 
Or at least there should be a task associated to it otherwise this might go unnoticed and stay this way for a while

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallInfoConfiguration.swift
billypchan: I will keep minimal changes in this bump PR and handle the type changing in https://github.com/wireapp/wire-ios/pull/4766

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -37,7 +37,7 @@ struct VideoConfiguration: VideoGridConfiguration {
 
 extension CallParticipant {
     var streamId: AVSClient {
-        return AVSClient(userId: user.remoteIdentifier, clientId: clientId)
+        return AVSClient(userId: (user as! ZMUser).remoteIdentifier, clientId: clientId)
```
johnxnguyen, line 40: By force casting, we won't be able to have a mock user in the call participant. But isn't that the point of these changes, to be able to do such things?

billypchan: `UserType` is missing the `remoteIdentifier` property. We can add it in another PR to prevent the casting.

johnxnguyen (approved):

David-Henner (approved):

PR #4779, opened 2020-12-18
[Fix] Restore setting passcode after login

johnxnguyen:
### Issues

We temporarily removed logic to set up the custom passcode during the login flow.

### Causes

The app lock feature is now associated with the active user session, which we don't have until after the login flow.

### Solutions

Move the logic that request setting a custom passcode when the conversation list appears. 

### Notes

This isn't an ideal solution because it doesn't prevent conflict with other possible UI presentation (such as the call overlay). I'm investigating ways to improve this by making use of the (not yet merged) `locked` app state.


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+PasscodeSetup.swift
```diff
@@ -0,0 +1,40 @@
...
+
+import Foundation
+import WireSyncEngine
+
+extension ZClientViewController {
+
+    func setUpCustomPasscodeIfNeeded(then block: @escaping () -> Void) {
+        guard
+            let appLock = ZMUserSession.shared()?.appLockController,
+            appLock.config.useBiometricsOrCustomPasscode,
```
KaterinaWire, line 27: Why do we need to check `useBiometricsOrCustomPasscode` ?
I think it should be something like this:
`(appLock.config.useBiometricsOrCustomPasscode && AuthenticationType.current == .passcode) || AuthenticationType.current == .unavailable`

johnxnguyen: `useBiometricsOrCustomPasscode` means use biometrics if available, otherwise fallback on the custom passcode. But biometric authentication always needs to have a fallback set up, otherwise you can be locked out of your device if you can't use biometry for some reason (such as a face mask).

So `useBiometricsOrCustomPasscode` really means to set a custom passcode and use biometry if possible.

johnxnguyen: But indeed you're right that we need to check the case where `useBiometricsOrCustomPasscode` is false and there is no device passcode set. So the condition should probably be:

```swift
let requiresCustomPasscode = useBiometricsOrCustomPasscode || AuthenticationType.current == .unavailable
guard appLock.isEnforced && requiresCustomPasscode && customPasscodeIsNotSet else { .. }

```

KaterinaWire: Yes

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -165,9 +165,13 @@ final class ConversationListViewController: UIViewController {
 
         if !viewDidAppearCalled {
             viewDidAppearCalled = true
-            ZClientViewController.shared?.showDataUsagePermissionDialogIfNeeded()
-            ZClientViewController.shared?.showAvailabilityBehaviourChangeAlertIfNeeded()
+
+            ZClientViewController.shared?.setUpCustomPasscodeIfNeeded(then: {
```
billypchan, line 169: I would suggest moving these logic to interactor if we refactor this VC. For now it is acceptable to put these logics here but not easy to test it.

billypchan (approved):

Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
johnxnguyen: I agree, it's a bit of a hack now but it's out of scope for this PR to fix it unfortunately.

PR #4785, opened 2020-12-24
ConnectRequestsViewController: do not setup observers when testing

billypchan:
## What's new in this PR?

Do not set up observers when testing which may cause a crash when destructing `ConnectRequestsViewController`

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.swift
@@ -38,7 +38,8 @@ final class ConnectRequestsViewController: UIViewController, UITableViewDataSour
         tableView.delegate = self
         tableView.dataSource = self
         
-        if let userSession = ZMUserSession.shared() {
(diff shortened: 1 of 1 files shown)


johnxnguyen (commented):
What is the cause of the crash exactly?

billypchan:
@johnxnguyen observers may not be released after the test is torn down, which may cause bad access crash in some cases.

johnxnguyen:
> @johnxnguyen observers may not be released after the test is torn down, which may cause bad access crash in some cases.

@billypchan does this just affect certain types of notifications? I'm wondering if by disabling certain code for tests we're just stepping around a deeper problem and also affecting the tests in a negative way, since no observation code will execute.

marcomaddy1982 (approved):

billypchan:
I found the crash is related to `ConnectRequestsViewControllerSnapshotTests`, which creates `ZMConversation` and `ZMUser`. The unreleased objects should be `ConnectRequestsViewController.connectionRequests` (`[ZMConversation]` type). I willl open another PR to fix that issue.

PR #4787, opened 2020-12-24
change Analytics.share to optional

billypchan:
### Issues

When testing `SettingsTableViewControllerSnapshotTests.testForSettingGroup` solely, app crashes.

### Causes

`Analytics.share` is nil in this case.

### Solutions

Since we would not set up Analytics for testing, change the `share` to optional.


Wire-iOS/Sources/UserInterface/Settings/CellDescriptors/SettingsCellDescriptorFactory.swift
```diff
@@ -214,7 +214,7 @@ class SettingsCellDescriptorFactory {
         let showStatistics = SettingsExternalScreenCellDescriptor(title: "Show database statistics", isDestructive: false, presentationStyle: .navigation, presentationAction: {  DatabaseStatisticsController() })
         developerCellDescriptors.append(showStatistics)
 
-        if !Analytics.shared.isOptedOut &&
+        if false == Analytics.shared?.isOptedOut &&
```
johnxnguyen, line 217: Is this the line that was crashing the tests?

billypchan: Yes

johnxnguyen (commented):
Could we perhaps mock the singleton for the test? I feel that making it optional complicates the code.

PR #4793, opened 2020-12-28
[Feature] Show account indicator in app lock screen

johnxnguyen:
### Issues

When asked to enter the app lock passcode, the user does not know which passcode to enter.

### Causes

Each account may now have its own app lock passcode, but the unlock screen doesn't indicate which account is being unlocked.

### Solutions

Add an account indicator to the top of the unlock screen.


Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
```diff
@@ -92,7 +92,8 @@ final class AppLockViewController: UIViewController {
     private func presentCustomPassCodeUnlockScreenIfNeeded(message: String,
                                                            callback: @escaping RequestPasswordController.Callback) {
         if unlockViewController == nil {
-            let viewController = UnlockViewController()
+            // TODO: [John] Avoid static methods.
+            let viewController = UnlockViewController(selfUser: ZMUser.selfUser(), userSession: ZMUserSession.shared())
```
billypchan, line 96: We can inject self user and session from init?

billypchan (commented):
Remove todo

Wire-iOS/Sources/UserInterface/Overlay/AppLockViewController.swift
johnxnguyen: Unfortunately not, since the `AppLockViewController` is a singleton. I want to address this in a separate PR so that the scope of this one stays small.

johnxnguyen:
> Remove todo

@billypchan I'll address this in a separate PR (see other comment).

billypchan (approved):

marcomaddy1982 (approved):

PR #4794, opened 2020-12-28
[Refactor]: SQCORE-285 remove session manager async initialisation

marcomaddy1982:
## What's new in this PR?
According to the [RFC](https://wearezeta.atlassian.net/wiki/spaces/IOS/pages/159646019/Simplify+API+for+Login+Registration) about the login and registration API one of the improvement we would like to add is removing the async creation of the `SessionManager` in the `AppRootRouter`. With the changes of the current PR the `SessionManager` is now created in the `AppDelegate` and injected in the AppRootRouter without being an option value. 
Moreover,  differently from what written in the RFC, the `start` of the `SessionManager` will be excuses at the start of the `AppRootRouter`.

### Dependencies

[SE](https://github.com/wireapp/wire-ios-sync-engine/pull/1367)


Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -40,6 +41,8 @@ var defaultFontScheme: FontScheme = FontScheme(contentSizeCategory: UIApplicatio
 
 class AppDelegate: UIResponder, UIApplicationDelegate {
     
+    // MARK: - Private Property
+    private static var sharedAppDelegate: AppDelegate!
```
johnxnguyen, line 45: Why is the purpose of this when there already exists `shared`?

Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -213,21 +218,60 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             ZMUserSession.shared()?.application(application, handleEventsForBackgroundURLSession: identifier, completionHandler: completionHandler)
         }
     }
+    
+    func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
+        createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
+    }
```
johnxnguyen, line 224: Does this get called automatically or do you need to observer something?

Wire-iOS/Sources/AppDelegate.swift
marcomaddy1982: I have no idea honestly, I just moved it. it was place are below in the class. One of the action we could do after this PR is check where is used and remove it cause I really don't like it.

Wire-iOS/Sources/AppDelegate.swift
marcomaddy1982: As far as I understood during these days this should be called automatically, anyway I could be wrong. Unfortunately I can't debug on the device cause I have installed iOS14.2 on my device and I'm working with Xcode11.x.

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -76,33 +62,55 @@ public class AppRootRouter: NSObject {
...
     }
     
     // MARK: - Public implementation
     
     public func start(launchOptions: LaunchOptions) {
-        transition(to: .headless, completion: {
-            Analytics.shared.tagEvent("app.open")
-        })
-        createAndStartSessionManager(launchOptions: launchOptions)
+        guard let deepLinkURL = deepLinkURL else { showInitial(launchOptions: launchOptions); return; }
```
billypchan, line 105:
```suggestion
        guard let deepLinkURL = deepLinkURL else { showInitial(launchOptions: launchOptions) 
        return 
        }
```

billypchan (commented):
I would recommend running swift lint to clean up the code.

johnxnguyen (commented):
> As far as I understood during these days this should be called automatically, anyway I could be wrong. 

@marcomaddy1982 in the SE PR I found:

```swift
public func executeWhenFileSystemIsAccessible(_ block: @escaping () -> Void) {		
         if isProtectedDataAvailable || ZMPersistentCookieStorage.hasAccessibleAuthenticationCookieData() {		
             block()		
         } else {		
             var token: Any? = nil		
             token = NotificationCenter.default.addObserver(forName: UIApplication.protectedDataDidBecomeAvailableNotification, object: nil, queue: nil) { _ in		
                 block()		
                 NotificationCenter.default.removeObserver(token!)		
             }		
         }		
     }
```

So I think you need to listen want to do the same thing in the app delegate.

marcomaddy1982:
> > As far as I understood during these days this should be called automatically, anyway I could be wrong.
> 
> @marcomaddy1982 in the SE PR I found:
> 
> ```swift
> public func executeWhenFileSystemIsAccessible(_ block: @escaping () -> Void) {		
>          if isProtectedDataAvailable || ZMPersistentCookieStorage.hasAccessibleAuthenticationCookieData() {		
>              block()		
>          } else {		
>              var token: Any? = nil		
>              token = NotificationCenter.default.addObserver(forName: UIApplication.protectedDataDidBecomeAvailableNotification, object: nil, queue: nil) { _ in		
>                  block()		
>                  NotificationCenter.default.removeObserver(token!)		
>              }		
>          }		
>      }
> ```
> 
> So I think you need to listen want to do the same thing in the app delegate.

@johnxnguyen ok, as far I understood this way is used in all the other classes different from the AppDelegate,  cause in the appDelegate when the device is unlocked it is triggered 

```swift
func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
```

Anyway, I think you are right anyway but for another reason: the `executeWhenFileSystemIsAccessible` just when we were creating the `sessionManager` and so just once at the start of the app when it was created in the `AppRootRouter`. 
So the following manner the `createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)` will be called every time the screen is unlocked and that isn't the behaviour we expected. 

```swift
func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
         createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
     }
```

What do you think?

johnxnguyen:
@marcomaddy1982 true, you'd only want to run it once. But I guess you could always do a guard that the `appRouter` is `nil` before trying to start it. So something like:

```swift
func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
  guard appRootRouter == nil else { return }
  createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
}
```

marcomaddy1982:
> @marcomaddy1982 true, you'd only want to run it once. But I guess you could always do a guard that the `appRouter` is `nil` before trying to start it. So something like:
> 
> ```swift
> func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
>   guard appRootRouter == nil else { return }
>   createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
> }
> ```



> @marcomaddy1982 true, you'd only want to run it once. But I guess you could always do a guard that the `appRouter` is `nil` before trying to start it. So something like:
> 
> ```swift
> func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
>   guard appRootRouter == nil else { return }
>   createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
> }
> ```

@johnxnguyen right, I was just writing the same :-) :-) . But thanks for pointing it out!!!!

Wire-iOS/Sources/AppDelegate.swift
marcomaddy1982: So removing `private static var sharedAppDelegate: AppDelegate!` I should instead the following all around the code

```swift
 (UIApplication.shared.delegate as? AppDelegate)?.property
```

marcomaddy1982: Should I do it in a separate PR?

johnxnguyen: You could use `AppDelegate.shared`, if you make `shared` a stored property:

```swift
class AppDelegate {
  
  // Delete this
  private static var sharedAppDelegate: AppDelegate!

  // This is now a stored property, not a computed property.
  private(set) static var shared: AppDelegate!

  override init() {
    super.init()
    Self.shared = self
  }
}
```

marcomaddy1982: At the end I just did 

```swift
static var shared: AppDelegate {
        return UIApplication.shared.delegate as! AppDelegate
    }
```

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -102,7 +102,10 @@ public class AppRootRouter: NSObject {
     // MARK: - Public implementation
     
     public func start(launchOptions: LaunchOptions) {
-        guard let deepLinkURL = deepLinkURL else { showInitial(launchOptions: launchOptions); return; }
+        guard let deepLinkURL = deepLinkURL else {
+            showInitial(launchOptions: launchOptions);
```
billypchan, line 106:
```suggestion
            showInitial(launchOptions: launchOptions)
```

Wire-iOS/Sources/AppDelegate.swift
```diff
@@ -213,21 +216,61 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             ZMUserSession.shared()?.application(application, handleEventsForBackgroundURLSession: identifier, completionHandler: completionHandler)
         }
     }
+    
+    func applicationProtectedDataDidBecomeAvailable(_ application: UIApplication) {
+        guard appRootRouter == nil else { return }
+        createAppRootRouterAndInitialiazeOperations(launchOptions: launchOptions)
+    }
```
johnxnguyen, line 223: Do you think it's possible that this gets called before the app finishes launching? Because if it could, then it means that we pass in empty launch options.

Wire-iOS/Sources/AppRootRouter.swift
marcomaddy1982: Done!!!

Wire-iOS/Sources/AppDelegate.swift
marcomaddy1982: Good question, I honestly don't know but I'm going to investigate.

johnxnguyen (approved):

billypchan (approved):

PR #4795, opened 2020-12-29
[Refactor] Wipe database

johnxnguyen:
### Context

It is now possible that each account has its own app lock passcode. Currently, if the user forgets their passcode, then the only option for them is to wipe the database and log in again.

### Issues

If the user forgets a passcode and proceeds to wipe the database, then all logged in accounts are deleted, rather than just the account associated with the forgotten passcode.

### Causes

We explicitly delete all accounts. This made sense when there could only exist one passcode for the entire app.

### Solutions

Delete only the affected account using changes in https://github.com/wireapp/wire-ios-sync-engine/pull/1368

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/1368
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1098


Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/WipeDatabase/WipeDatabaseInteractor.swift
```diff
@@ -33,13 +31,14 @@ final class WipeDatabaseInteractor {
...
+            let sessionManager = SessionManager.shared,
+            let account = sessionManager.accountManager.selectedAccount
+        else {
+            return
+        }
 
-    func deletePasscode() {
-        // TODO: [John] Inject the app lock controller.
-        guard let appLock = ZMUserSession.shared()?.appLockController else { return }
-        try? appLock.deletePasscode()
```
johnxnguyen, line 43: Passcode deletion is now taken care of when the account is deleted.

marcomaddy1982 (approved):

billypchan (approved):

PR #4800, opened 2020-12-30
[Chore] Define Viper protocols

johnxnguyen:
## What's new in this PR?

I've opened this PR primarily to return to the discussion started in https://github.com/wireapp/ios-architecture/pull/99, but with the intention of merging the result.

### Context

We want to adopt the Viper presentation pattern in our codebase and have been discussing it for quite some time. There already exists some efforts to use Viper (such as the app lock related classes), but the implementations are incomplete and not consistent.

### Problem

There are a myriad of Viper implementations, each with slight variations in structure (e.g how components are connected), style (e.g naming and terminology), and how they are written in code (e.g use of protocols and type safety). If we wish to avoid implementing the Viper pattern is different ways, then we need to:

1) define a convention, and
2) design a way to enforce that convention.

### Solutions

The convention I am proposing is one that I've developed for a personal project. First I will explain the way in which the convention is enforced, then I will add some notes about the choice of terminology.

The router, interactor, presenter and view components are modeled with "interface" protocols which describe the role of that component in relation to other components. 

For instance, the `InteractorInterface` has a single requirement for a reference to the `presenter`. Similarly, the `PresenterInterface` requires reference a reference to the `interactor`. The types of these references are generic and will need to be specified by the concrete classes that conform to these interface protocols.

The directional relationships between each of these components are also modeled with base protocols. These protocols have no requirements, but are intended to be inherited by module specific relationships and used to specified the references of the interface protocols.

Finally, there is the a special component that represents the module as a whole and is described by `ModuleInterface`. This protocol acts as a description of the concrete types used in the module. It also provides a method to assemble the module by connecting the components together. A type conforming to the module interface could be seen a "wireframe", which is a lightweight object responsible for instantiating a module.

This cluster of protocols and associated types essentially leverages Swift's type system to ensure that a module is strictly typed and correctly connected. For instance, note that the default implementation of the `assemble` method will intentionally crash when it can't satisfy a relationship requirement. Additionally, these protocols also help to enforce naming conventions of relationships.

### Example

To help elucidate the purpose of these protocols, consider the following example module.

```swift
final class FooModule: ModuleInterface {

    typealias Router = FooRouter
    typealias Interactor = FooInteractor
    typealias Presenter = FooPresenter
    typealias View = FooView

    func build() -> View {
        let router = Router()
        let interactor = Interactor()
        let presenter = Presenter()
        let view = View()

        assemble(router: router, interactor: interactor, presenter: presenter, view: view)

        return view
    }

}

// MARK: - Router / Presenter

protocol FooRouterPresenterInterface: RouterPresenterInterface {

    /// Router API exposed to the presenter.

}

// MARK: - Interactor / Presenter

protocol FooPresenterInteractorInterface: PresenterInteractorInterface {

    /// Presenter API exposed to the interactor.

}

protocol FooInteractorPresenterInterface: InteractorPresenterInterface {

    /// Interactor API exposed to the presenter.

}

// MARK: - View / Presenter

protocol FooViewPresenterInterface: ViewPresenterInterface {

    /// View API exposed to the presenter.

}

protocol FooPresenterViewInterface: PresenterViewInterface {

    /// Presenter API exposed to the view.

}
```

Here we define the module and each of the interfaces for the module components. We then need to define the concrete components. For brevity, here's just one for the interactor:

```swift
final class FooInteractor: InteractorInterface {

    weak var presenter: FooPresenterInteractorInterface!

}

// MARK: - API for prsenter

extension FooInteractor: FooInteractorPresenterInterface {}
```

If we comment out the conformance for `FooRouterPresenterInterface`, then we will still build but the app will intentionally crash when trying to assemble the module.

### Template files

Each Viper module consists of 5 files, `FooModule`, `FooRouter`, `FooInteractor`, `FooPresenter`, and `FooView`. Xcode templates can be used to generate all of these files given a module name. This allows easy creation of a new module with minimal effort, all whilst maintaining naming and style consistency.

### Notes on terminology

I propose the term `module` over `wireframe` in order to reduce the amount of terminology. A viper module refers to the collection of components for a specific part of the app. By reusing the term "module", we can think of the module class `FooModule` as the collection of components specific to this module, leading to expressive method calls such as `FooModule().build()`.

I propose to name relationships in the form `<module name><implementer><caller>` for brevity but also for clarity when using autocompletion. For example, consider:

```swift
FooPresenterInteractorInterface
FooPresenterViewInterface
```

We are able to search "FooPresenter" and see autocompletion for each of these protocols in a clean way. This is useful when you know you need to search for a particular protocol, which I imagine will happen a lot. It follows the train of thought "I'm looking for the protocol in the **Foo** module, implemented by the **Presenter** and called by the **Interactor**".


marcomaddy1982:
@johnxnguyen really interesting your proposal!! Thanks for doing that 👍 
I have just one question, if I remember correctly the relation between `Presenter` and `Router` is not bidirectional. So basically the `Presenter` has an instance of the `Router` but not vice versa. 
I expect that from the `Router` any action shouldn't be done on the `Presenter` but in case we need to do something particular we should instead have a `FooViewRouterInterface`, so basically  the `View` should expose some methods to the `Router` and than in case `View`  will interact with the `Presenter`.

Anyway, we are free to have whatever approach we like. So let discuss about this point.

johnxnguyen:
>  have just one comment, if I remember correctly the relation between Presenter and Router is not bidirectional. So basically the Presenter has an instance of the Router but not vice versa.

@marcomaddy1982 ah yeah in fact, I wasn't so sure about that! I think then it's appropriate to delete it and remove the `presenter` requirement on the `RouterInterface`.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
```diff
@@ -0,0 +1,120 @@
...
+
+protocol PresenterInterface: PresenterInteractorInterface & PresenterViewInterface {
+
+    associatedtype RouterPresenter
+    associatedtype InteractorPresenter
+    associatedtype ViewPresenter
+
+    /// A strong reference to the router.
+
+    var router: RouterPresenter! { get set }
```
typfel, line 95: Why do we need to define this as force unwrapped?

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// A strong reference to the interactor.
+
+    var interactor: InteractorPresenter! { get set }
+
+    /// A weak reference to the view.
+
+    var view: ViewPresenter! { get set }
```
typfel, line 103: Why not define this as optional if we require the property to be weak.

typfel:
Are you planning on also creating the template files as part of this PR?

typfel:
Added some questions but other than that I like this approach.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: This is implicitly unwrapped because the property isn't set during initialization, rather when the whole module is assembled.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: The weakness is to ensure no retain cycles between components, but we shouldn't ever expect to have a partially connected module, mostly because it doesn't make that much sense to have half a module. As long as there is a reference to the view controller the module as a whole will be retained. As such, despite the weakness of certain relationships we should still think of them as non-optional.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
typfel: Ok, and making the initialiser part of the interface would be too restrictive?

johnxnguyen:
> Are you planning on also creating the template files as part of this PR?

@typfel I have defined templates already but haven't yet included them in this PR. I'm not sure what's the best way to include them actually, since the template files need to be moved to a location within the Xcode application package.

Perhaps commit them in this repo together with a script that installs it to the right location?

typfel:
> > Are you planning on also creating the template files as part of this PR?
> 
> @typfel I have defined templates already but haven't yet included them in this PR. I'm not sure what's the best way to include them actually, since the template files need to be moved to a location within the Xcode application package.
> 
> Perhaps commit them in this repo together with a script that installs it to the right location?

Ah I see, then yes that sounds like a good approach.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: I guess if we used initializers then then creating the module would look something like:

```swift
func build() -> UIViewController {
  let router = Router()
  let interactor = Interactor()

  // Strong (non optional) references are set here.
  let presenter = Presenter(interactor: interactor, router: router)
  let view = View(presenter: presenter)

  // The remaining weak relationships are set here.
  assemble(router: router, Interactor: interactor, presenter: presenter, view: view)

  return view
}
```

I liked that before the `assemble` method was responsible for all the wiring up of the modules, but it also seems nice to remove the force unwrapping. Is there any particular reason we should avoid force unwrapping?

johnxnguyen: Regarding whether including initializers in the interfaces is too restrictive:

I can imagine that it could be annoying to require a mock presenter to implement an initializer accepting both the interactor and router. Suppose we want to test the interactor, then we would mock the presenter and call methods on the interactor and assert the invocations on the mock presenter. Here we're not testing the router at all (interactor knows nothing about it), yet we would have to pass one in to the mock presenter.

A similar thing happens when testing the router (and needing to pass in an interactor to the mock presenter).

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
```diff
@@ -0,0 +1,120 @@
...
+
+    /// Assembles the module by connecting each component together.
+
+    func assemble(router: Router, interactor: Interactor, presenter: Presenter, view: View)
+
+}
+
+extension ModuleInterface {
+
+    func assemble(router: Router, interactor: Interactor, presenter: Presenter, view: View) {
```
marcomaddy1982, line 42: @johnxnguyen Instead of passing he the `Router`, `Interactor`,  `Presenter`, `View` why we don't initialise and build all the dependencies of the module in a single method?

johnxnguyen:
@typfel I've added the template files and a script that runs in `setup.sh` to install them.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: Do you mean in such as in the `build` method? The `assemble` method has a default implementation that so that every module would be able to assemble easily rather than having to type check that each component fits together.

But we can easily include this assembling code as part of the template, such as:

```swift
func build() -> UIViewController {
  let router = Router()
  let interactor = Interactor()
  let presenter = Presenter(interactor: interactor, router: router)
  let view = View(presenter: presenter)

  interactor.presenter = (presenter as! Interactor.PresenterInteractor)
  presenter.interactor = (interactor as! Presenter.InteractorPresenter)
  presenter.router = (router as! Presenter.RouterPresenter)
  presenter.view = (view as! Presenter.ViewPresenter)
  view.presenter = (presenter as! View.PresenterView)

  router.viewController = view

  return view
}
```

The downside to this is having duplicated wiring code for each module. The upside though is that the compiler can report type errors directly in Xcode rather than at run time (the reason for this is that the default `assemble` method only knows about the base relationship protocols and not the specialized ones).

marcomaddy1982: ok ok make sense have them separated.  understand your point. Thanks for making it clear. 👍

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
```diff
@@ -0,0 +1,51 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Interface of the interactor from the perspective of the presenter.
+///
+/// Typically contains methods fetch data and perform business logic.
+
+protocol InteractorPresenterInterface: class { }
```
David-Henner, line 25: So these protocols only represent the relationships between VIPER components ? Or do they serve some additional purpose ?

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
typfel: I don't think having to pass in all the mocks in the initialiser is such a burden when writing tests. We could start using mock generators like https://github.com/seanhenry/SwiftMockGeneratorForXcode 

Wouldn't it be a nice if the `build()` method was included as a protocol extension.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
typfel: Couldn't the build method be a protocol extension if we include the initialisers as part of the protocols? Would remove a lot of code duplication.

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: > Wouldn't it be a nice if the build() method was included as a protocol extension.

It would, but all module dependencies should be injected in the `build` method and then passed on to the individual components (typically the interactor). So while we could include a default `build` method, it would only be useful for modules without any dependencies.

The idea of the `assemble` method was to extract out the common code needed in all build methods so that the build method need only to focus on initializing the components (with dependencies) and nothing else.

As I mentioned in another comment to Marco M, we could forego the assemble method completely and simply provide a generated build method containing everything needed to construct the module. (Note though there is already a build method in the template, but it uses the assemble method.)

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: Mentioned in another comment, the build method doesn't have a fixed function signature because it is the injection point for dependencies.

In my opinion, I think we should try to remove code duplication as much as possible and leave the module implementer to only need to think about defining the relationships and implementing them, rather than worry about connecting the module together. The only exception to this is modifying a generated build method to inject dependencies.

typfel: Ok, then it's probably not worth it.

typfel (approved):

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
johnxnguyen: Yes they are the base protocols representing directional relationships. Your specific module would need to define specialized versions of these relationships that declare their requirements. But they also help leverage the type system to explicitly define the shape of a particular viper component. All of these base protocols (for relationships and components) are primarily to have a precise definition of what a module looks like. You can think of these base protocols as defining the skeleton of the module, and the specialized protocols (and their conformances) as being the muscle. 

Perhaps this might make it clearer:

```swift
// An concrete interactor must full the presenter -> interactor relationship.
protocol InteractorInterface: InteractorPresenterInterface {
  associatedtype PresenterInteractor

  // It also refers to a presenter of some not yet defined type.
  var presenter: PresenterInteractor! { get set }
}

// We declare a concrete interactor, but it doesn't yet fulfill the relationship.
class AppLockInteractor: InteractorInterface {
  // Now we give the presenter a type, stating that it must conform to the inverse
  // relationship (whether that be AppLockPresenter or a mock).
  weak var presenter: AppLockPresenterInteractorInterface!
}

// We declare what the relationship requires by specializing the base protocol.
protocol AppLockInteractorPresenterInterface: InteractorPresenterInterface {
  func evaluate(passcode: String)
}

// Finally we fulfill the specialized relationship, thus fulfilling conformance to the
// base protocol and the interactor interface.
extension AppLockInteractor: AppLockInteractorPresenterInterface {
  func evalute(passcode: String) {
    // ...
  }
}
```

marcomaddy1982 (approved):

Templates/Module.xctemplate/___FILEBASENAME___Interactor.swift
```diff
@@ -0,0 +1,29 @@
...
+
+import Foundation
+
+final class ___FILEBASENAMEASIDENTIFIER___: InteractorInterface {
+
+    weak var presenter: ___VARIABLE_productName:identifier___PresenterInteractorInterface!
+
+}
+
+// MARK: - API for prsenter
```
johnxnguyen, line 27:
```suggestion
// MARK: - API for presenter
```

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
```diff
@@ -0,0 +1,120 @@
...
+protocol ModuleInterface {
+
+    associatedtype Router where Router: RouterInterface
+    associatedtype Interactor where Interactor: InteractorInterface
+    associatedtype Presenter where Presenter: PresenterInterface
+    associatedtype View where View: UIViewController & ViewInterface
+
+    /// Assembles the module by connecting each component together.
+
+    func assemble(router: Router, interactor: Interactor, presenter: Presenter, view: View)
```
billypchan, line 36: This method can be static and then we no need to create an instance of `ModuleInterface`

Templates/Module.xctemplate/___FILEBASENAME___Module.swift
```diff
@@ -0,0 +1,77 @@
...
+import Foundation
+
+final class ___FILEBASENAMEASIDENTIFIER___: ModuleInterface {
+
+    typealias Router = ___VARIABLE_productName:identifier___Router
+    typealias Interactor = ___VARIABLE_productName:identifier___Interactor
+    typealias Presenter = ___VARIABLE_productName:identifier___Presenter
+    typealias View = ___VARIABLE_productName:identifier___View
+
+    func build() -> View {
```
billypchan, line 28: This method can be static.

billypchan (commented):
looks nice! I think we can add some example or sample to try it out.

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
David-Henner: Thanks for clarifying! I like this approach :)
My only concern is that I'm getting mixed up by the naming of these relationship protocols. Would you mind including a detailed explanation of the naming convention ?

KaterinaWire (approved):

Wire-iOS/Sources/Viper/Viper+Interfaces.swift
johnxnguyen: Good idea!

johnxnguyen: Done

Templates/Module.xctemplate/___FILEBASENAME___Module.swift
johnxnguyen: Done

billypchan (approved):

johnxnguyen:
> looks nice! I think we can add some example or sample to try it out.

Probably the first few modules can serve as an example. For instance, once this is merged I'd like to convert the app lock module using these templates.

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
johnxnguyen: Of course! The naming of the relationship protocols follows the format `<ModuleName><Implementer><Caller>Interface`.

 For instance the name `AppLockInteractorPresenterInterface` indicates that the protocol relationship is for the `AppLock` module, implemented by the `Interactor` and the methods in this protocol are called by the `Presenter`. If we use the "input/output" terminology, then you could say that this protocol would be named `AppLockInteractorInput`.

On the other hand, `AppLockPresenterInteractorInterface` is the API implemented by the `Presenter` and called by the `Interactor`. Again, if we used "input/output" terminology, we would name this `AppLockInteractorOutput`.

The choice of this naming convention, as mentioned in the PR description, is to make it easy to search for a particular protocol. If you want to search in the `AppLock` module for the `Interactor` implementation for the `Presenter` relationship, you just type "AppLock" + "Interactor" + "Presenter".

Or suppose you search "AppLock" + "Presenter", then you would see a list of all relationship implementations for that presenter (since there are three of them).

Does that clarify things?

David-Henner: It's clear now! thanks!

Somehow I thought that it followed this format `<ModuleName><Caller><Implementer>Interface`. That's where my confusion came from.

It would be great to have a comment in `VIPER+Relationships.swift` as a reminder of the chosen format. I think that would prevent this misunderstanding

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
```diff
@@ -0,0 +1,51 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
```
johnxnguyen, line 20:
```suggestion

/// Naming convention:
///
/// The names of all relationship protocols follow the format:
///
///  <module name><implementer><caller>Interface
/// 
/// For example, the name FooInteractorPresenterInterface is the relationship
/// implemented by the interactor, called by the presenter, and located in the
/// "Foo" module.
///
/// This format is designed to make it easy to search for a particular relationship protocol.
/// If you are searching for a particular protocol, type first the module name, then the
/// component that implements the methods, then the component that calls the methods.

```

Wire-iOS/Sources/Viper/VIPER+Relationships.swift
johnxnguyen: Good idea 👍🏻 Done

David-Henner (approved):

PR #4811, opened 2021-01-07
[Refactor] Use markdown formatting in localized string

johnxnguyen:
## What's new in this PR?

**Jira:** https://wearezeta.atlassian.net/browse/SQSERVICES-182

### Issues

Some copy is constructed manually from two localized strings. This is problematic for translators.

### Causes

One of the strings needs bold formatting, so each string is styled differently and then joined together.

### Solutions

Use markdown formatting in a single localized string.


KaterinaWire (approved):

Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1802,10 +1802,8 @@
 // Custom app lock - create passcode
 "create_passcode.textfield.placeholder" = "";
 "create_passcode.title_label" = "Create a passcode";
-"create_passcode.info_label" = "It will be used to unlock Wire. If you forget this passcode ";
-"create_passcode.info_label.highlighted" = "it can not be recovered.";
-"create_passcode.info_label_forced_applock" = "Create a passcode to unlock Wire. Please remember it, as ";
-"create_passcode.info_label_forced_applock.highlighted" = "it can not be recovered.";
+"create_passcode.info_label" = "It will be used to unlock Wire. If you forget this passcode **it can not be recovered.**";
+"create_passcode.info_label_forced_applock" = "Create a passcode to unlock Wire. Please remember it, as **it can not be recovered.**";
```
billypchan, line 1806: We should jog some note for translater with comments?

billypchan (approved):

Wire-iOS/Resources/Base.lproj/Localizable.strings
johnxnguyen: Where would we put the comment?

PR #4814, opened 2021-01-07
[Feature] active speaker frame

David-Henner:
## What's new in this PR?

- Show a border around the video tile when the video participant is actively speaking


billypchan (approved):
It would be better to create a snapshot for this change.

David-Henner:
> It would be better to create a snapshot for this change.

I've created one, but I think that GitHub is not rendering it. You can find it [there](https://github.com/wireapp/wire-ios/blob/32f60ffb5df066025cab4d236b89d6f3f4596ba7/Wire-iOS%20Tests/ReferenceImages/VideoPreviewViewTests/testActiveState.1.png)

KaterinaWire:
Do we have tests for active speaker?

David-Henner:
> Do we have tests for active speaker?

There's a snapshot test [here](https://github.com/wireapp/wire-ios/blob/32f60ffb5df066025cab4d236b89d6f3f4596ba7/Wire-iOS%20Tests/ReferenceImages/VideoPreviewViewTests/testActiveState.1.png)

KaterinaWire (approved):

KaterinaWire:
> > Do we have tests for active speaker?
> 
> There's a snapshot test [here](https://github.com/wireapp/wire-ios/blob/32f60ffb5df066025cab4d236b89d6f3f4596ba7/Wire-iOS%20Tests/ReferenceImages/VideoPreviewViewTests/testActiveState.1.png)

I approved

PR #4815, opened 2021-01-08
[Feature] show spinner while the resetting session

typfel:
## What's new in this PR?

- Show spinner while the session is being reset
- Fix decryption error system messages not updating after resetting the session.

### Dependencies

Needs releases with:

- [ ]  https://github.com/wireapp/wire-ios-data-model/pull/1106


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -132,7 +132,7 @@ class LinkConversationSystemMessageCell: ConversationIconBasedCell, Conversation
         let attributedText: NSAttributedString?
         let showLine: Bool
 //        let url: URL
```
marcomaddy1982, line 134: Don't forgot to remove the comment if you don't need it anymore.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -132,7 +132,7 @@ class LinkConversationSystemMessageCell: ConversationIconBasedCell, Conversation
...
-        let urlHandler: ((_ url: URL) -> Void)?
+        let urlHandler: ((_ url: URL, _ cell: LinkConversationSystemMessageCell) -> Void)?
```
marcomaddy1982, line 134: Just an thought, instead to pass UI component around and so the entire `cell: LinkConversationSystemMessageCell)` here we could pass a `VoidClosure` or directly a void function (that actually is the same).

```swift
{
     cell.delegate?.perform(action: .resetSession, for: message, view: cell)
}
```

marcomaddy1982: So basically a reset handler of the cell.

marcomaddy1982: Anyway It is up to you. You can also leave as it is if you prefer.

marcomaddy1982:
@typfel also if I have made a comment I will approve the PR cause that comment doesn't involve any logic. Thanks 👍

marcomaddy1982 (approved):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
typfel: 👍 

The cell is a generic `LinkConversationSystemMessageCell` so that's why I had the handler in the cell description but it's actually only used in the case so yes it would be simpler if I moved everything to the cell.

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
marcomaddy1982: Great!! I saw the changes and are fine for me.

PR #4828, opened 2021-01-12
fix: non released ZMUser when running ConversationSystemMessageTests

billypchan:
### Issues
On Xcode 12, after running `ConversationSystemMessageTests`, `ZMUser` is unable to release

### Causes

`ZMUser.selfUser()` is called 

### Solutions

~~Inject mock self user to `ConversationNewDeviceSystemMessageCellDescription` to prevent creating `ZMUser`.~~
Replace `MockUser` factory method with `SwiftMockLoader`. 
Replace `ZMUser.selfUser()` with `SelfUser.current` for the code called in this test.


David-Henner (approved):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
```diff
@@ -844,7 +850,7 @@ class ConversationNewDeviceSystemMessageCellDescription: ConversationMessageCell
         if !systemMessage.addedUserTypes.isEmpty {
             return configureForAddedUsers(in: conversation, attributes: textAttributes)
         } else if systemMessage.systemMessageType == .reactivatedDevice {
-            return configureForReactivatedSelfClient(ZMUser.selfUser(), attributes: textAttributes)
+            return configureForReactivatedSelfClient(selfUser, attributes: textAttributes)
```
typfel, line 853: Why not use `SelfUser.current` instead on injecting the self user. `SelfUser.current` can be mocked in the tests by providing a `SelfUserProvider`.

typfel (changes requested):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/ConfigurationMessageCell/Components/ConversationSystemMessageCell.swift
billypchan: remove the injection and it works. Thanks!

typfel (approved):

PR #4835, opened 2021-01-14
[Chore] Restore app lock

johnxnguyen:
## What's new in this PR?

The feature configuration feature (for app lock) is not ready for a synchronized release, so "reverting" changes made for this feature. The changes have been significant and through different repositories so it's not feasible to revert the feature branch commit. Instead, we just modify that implementation to restore existing behavior.

### Changes

**The custom passcode has been avoided completely, we will never ask to set it or input it.** This restores previous behavior because although it was previously released, it has never been enabled in the build configuration.

**The warning screens of app lock config changes will never be shown.** This is not necessarily required because no team should be able to modify the config in team settings yet, but it's a nice safeguard.

**The app lock timeout is 10 seconds.** This was the behavior before, for both personal and pro users.

**Hide the app lock setting if no device passcode is set.** This is the behavior before, since we didn't off custom passcodes.

**Don't fetch the app lock config.** This is not strictly required and is another safeguard.

### Notes

One change that couldn't be reverted is that app lock is associated with individual accounts. This means that the setting can be enabled independently for each account (before it was global) and the user may need to authenticate when switching accounts.

Indeed some clients are relying on using the Wire account password as a fallback to failed biometrics. The changes here are incompatible with this, but those clients won't receive this update because they belong to a separate white label build.


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -339,11 +339,7 @@ final class SettingsPropertyFactory {
                 setAction: { _, value in
                     switch value {
                     case .number(value: let lockApp):
-                        self.delegate?.appLockOptionDidChange(self,
-                                                              newValue: lockApp.boolValue,
-                                                              callback: { result in
-                        self.isAppLockActive = result
-                        })
```
johnxnguyen, line 346: By not informing the delegate, we don't ask to set a custom passcode.

Wire-iOS/Sources/UserInterface/Overlay/AppLockInteractor.swift
```diff
@@ -111,16 +111,21 @@ extension AppLockInteractor: AppLockInteractorInput {
...
-        appLock?.evaluateAuthentication(scenario: authenticationScenario,
-                                        description: description.localized) { [weak self] result, context in
+        appLock?.evaluateAuthentication(scenario: authenticationScenario, description: description.localized) { [weak self] result, context in
             guard let `self` = self else { return }
+
+            var newResult = result
+
+            if case .needCustomPasscode = newResult {
+                newResult = .unavailable
+            }
```
johnxnguyen, line 121: This ensures that we never pass the result `needCustomPasscode` to the presenter, which means we'd never ask to create or input the custom passcode there.

KaterinaWire (approved):

billypchan (approved):

PR #4836, opened 2021-01-14
[Fix] Persist isAppLockActive value

KaterinaWire:
### Issues

If change `isAppLockActive` state in the settings and terminate the app, the new value is not saved.

### Causes

`isAppLockActive` is a `managedObjectContext` property and we didn't persist the changes.

### Solutions

Save `managedObjectContext`.

### DIFF
--- a/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
+++ b/Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
@@ -342,11 +342,9 @@ final class SettingsPropertyFactory {
                         self.delegate?.appLockOptionDidChange(self,
                                                               newValue: lockApp.boolValue,
                                                               callback: { result in
-                         if let userSession = self.userSession as? ZMUserSession {
-                             userSession.perform {
-                                 self.isAppLockActive = result
-                             }
-                         }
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -342,8 +342,12 @@ final class SettingsPropertyFactory {
                         self.delegate?.appLockOptionDidChange(self,
                                                               newValue: lockApp.boolValue,
                                                               callback: { result in
-                        self.isAppLockActive = result
-                        })                        
+                         if let userSession = self.userSession as? ZMUserSession {
```
johnxnguyen, line 345: I don't think you even need to cast this.

johnxnguyen (changes requested):

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
KaterinaWire: Fixed

David-Henner (approved):

johnxnguyen (approved):

PR #4837, opened 2021-01-14
[Fix] Warning screen text color

johnxnguyen:
### Issues

The app lock warning screen text color is dark in dark mode.

### Causes

The color variant is hard coded to `light`.

### Solutions

Use the default (dynamic) variant.

### Notes

I added some missing tests and needed to do some clean up.


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
```diff
@@ -26,7 +26,7 @@ extension ZClientViewController {
             return
         }
         if appLock.needsToNotifyUser && !appLock.isActive {
-            let warningVC = AppLockChangeWarningViewController()
+            let warningVC = AppLockChangeWarningViewController(isAppLockActive: false)
```
KaterinaWire, line 29: We cat set here `AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)`

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
johnxnguyen, line 29:
```suggestion
            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)
```

johnxnguyen: Done

billypchan (approved):

PR #4838, opened 2021-01-14
[Fix] Remove active speaker frame in 1:1 calls & when maximized

David-Henner:
## What's new in this PR?

Update the video grid's behaviour so that it **does not** show the active speaker frame in 1:1 calls and when the video tile is maximized.


johnxnguyen (approved):

Wire-iOS Tests/VideoGridViewControllerSnapshotTests.swift
```diff
@@ -20,6 +20,8 @@ import XCTest
 @testable import Wire
 
 final class MockVideoGridConfiguration: VideoGridConfiguration {
+    var isCallOneToOne: Bool = false
```
KaterinaWire, line 23: Do we have tests when it's one to one call?

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/BaseVideoPreviewView.swift
```diff
@@ -49,11 +49,21 @@ class BaseVideoPreviewView: OrientableView, AVSIdentifierProvider {
...
     }
     
-    var isMaximized: Bool = false
+    var shouldShowActiveSpeakerFrame: Bool = true {
+        didSet {
+            updateActiveSpeakerFrame()
+        }
+    }
+    
+    var isMaximized: Bool = false {
```
KaterinaWire, line 62: Where we change this property? And it might make sense to add a description of why we are using it.

David-Henner: This property is changed from the `VideoGridViewController`. It's toggled when a double tap happens at the view's location. I will add a description

Wire-iOS Tests/VideoGridViewControllerSnapshotTests.swift
David-Henner: There are none. The snapshot tests for this class are very slim, I will add some.

KaterinaWire (approved):

PR #4839, opened 2021-01-14
[Fix] App lock warning is displayed twice

johnxnguyen:
## What's new in this PR?

**JIRA:** https://wearezeta.atlassian.net/browse/SQSERVICES-201

### Issues

If a user who has no device passcode set, when when app lock becomes mandatory, then when they open the app they are warned twice about the configuration change. First they see the warning screen, then they see the same warning but on the create passcode screen.

### Causes

There are two screens that contain the warning. One should be shown when the user has at least a device passcode set, and the other is for the case when they need to create a custom passcode. The current code path first displays the first warning, then proceeds to attempt to authenticate. The result from this attempt indicates that the custom passcode is needed, then we see that the custom passcode needs to be set, in which case we display the warning again.

### Solutions

Before we attempt to show any warning, split the control path into three cases:

1. We need to set a custom passcode. Then show the create passcode screen with the warning.
2. There is already a custom passcode, then present the warning, then the custom passcode unlock screen.
3. There is at least the device passcode, then present the warning, then present the native OS authentication UI.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
+++ b/Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
@@ -110,9 +110,23 @@ final class AppLockPresenter {
         case .needed, .authenticated:
             authenticationState = .needed
             setContents(dimmed: true)
-            presentWarningIfNeeded {
-                self.appLockInteractorInput.evaluateAuthentication(description: AuthenticationMessageKey.deviceAuthentication)
+
+            if appLockInteractorInput.needsToCreateCustomPasscode {
+                userInterface?.presentCreatePasscodeScreen(callback: { _ in
+                    // User needs to enter the newly created passcode after creation.
+                    self.setContents(dimmed: true, withReauth: true)
+                    self.appLockInteractorInput.needsToNotifyUser = false
+                })
+            } else if appLockInteractorInput.shouldUseCustomPasscode {
+                presentWarningIfNeeded {
+                    self.requestAccountPassword(with: AuthenticationMessageKey.accountPassword)
+                }
+            } else {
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
```diff
@@ -110,9 +110,23 @@ final class AppLockPresenter {
...
+            if appLockInteractorInput.needsToCreateCustomPasscode {
+                userInterface?.presentCreatePasscodeScreen(callback: { _ in
+                    // User needs to enter the newly created passcode after creation.
+                    self.setContents(dimmed: true, withReauth: true)
+                    self.appLockInteractorInput.needsToNotifyUser = false
+                })
+            } else if appLockInteractorInput.shouldUseCustomPasscode {
+                presentWarningIfNeeded {
+                    self.requestAccountPassword(with: AuthenticationMessageKey.accountPassword)
+                }
```
KaterinaWire, line 123: I think we don't need this part 
```
} else if appLockInteractorInput.shouldUseCustomPasscode {
                 presentWarningIfNeeded {
                     self.requestAccountPassword(with: AuthenticationMessageKey.accountPassword)
}
```
Because the part below calls `evaluateAuthentication(description: ...)` which will later call `requestAccountPassword(with: AuthenticationMessageKey.accountPassword)`

billypchan (approved):

Wire-iOS/Sources/UserInterface/Overlay/AppLockPresenter.swift
johnxnguyen: Removed the branch and test.

KaterinaWire (approved):

PR #4850, opened 2021-01-18
[Fix] Delete passcode if app lock is disabled in team settings

KaterinaWire:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/SQSERVICES-203

### Issues
Disabling app lock in team settings doesn't delete the custom passcode in keychain. 

### Solutions

Delete the passcode when the client receives a notification that the team admin has disabled mandatory app lock.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
+++ b/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
@@ -36,10 +36,11 @@ extension ZClientViewController {
     }
     
     func notifyUserOfDisabledAppLockIfNeeded() {
-        guard let appLock = appLock,
-            let warningVC = changeWarningViewController,
+        guard 
+            let appLock = appLock,
             appLock.needsToNotifyUser,
-            !appLock.isActive else {
+            !appLock.isActive
+            let warningVC = changeWarningViewController, else {
                 return
         }
         
--- a/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
+++ b/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
@@ -39,8 +39,8 @@ extension ZClientViewController {
         guard 
             let appLock = appLock,
             appLock.needsToNotifyUser,
-            !appLock.isActive
-            let warningVC = changeWarningViewController, else {
+            !appLock.isActive,
+            let warningVC = changeWarningViewController else {
                 return
         }
         
--- a/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
+++ b/Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
@@ -26,7 +26,11 @@ extension ZClientViewController {
             return
         }
         if appLock.needsToNotifyUser && !appLock.isActive {
-            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)
+            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive) {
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
```diff
@@ -26,7 +26,11 @@ extension ZClientViewController {
             return
         }
         if appLock.needsToNotifyUser && !appLock.isActive {
-            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)
+            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive) {
+                if !appLock.isCustomPasscodeNotSet {
+                    try? appLock.deletePasscode()
+                }
+            }
```
johnxnguyen, line 33: I think we can solve this more generally by trying to delete the passcode whenever the app lock is turned off (either explicitly by the user or implicitly when the config changes). This would likely happen in the data model.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
```diff
@@ -26,7 +26,11 @@ extension ZClientViewController {
             return
         }
         if appLock.needsToNotifyUser && !appLock.isActive {
-            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)
+            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive) {
+                if !appLock.isCustomPasscodeNotSet {
+                    try? appLock.deletePasscode()
+                }
```
johnxnguyen, line 32:
```suggestion
                try? appLock.deletePasscode()
```

johnxnguyen (changes requested):
Please add a test.

KaterinaWire:
> Please add a test.

I added a test, but it required additional changes.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
```diff
@@ -53,15 +53,19 @@ final class ZClientViewController: UIViewController {
     private var networkAvailabilityObserverToken: Any?
     private var pendingInitialStateRestore = false
     
+    // For tests
```
typfel, line 56: I would remove this comment. We want to reach the point where we only interact with the user session through protocols and this is one step in that direction not only something we do for tests.

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController+FeatureChange.swift
```diff
@@ -21,15 +21,30 @@ import WireSyncEngine
...
-            let warningVC = AppLockChangeWarningViewController(isAppLockActive: appLock.isActive)
-            warningVC.modalPresentationStyle = .fullScreen
-            present(warningVC, animated: false)
+    }
+    
+    func notifyUserOfDisabledAppLockIfNeeded() {
+        guard let appLock = appLock,
+            let warningVC = changeWarningViewController,
+            appLock.needsToNotifyUser,
+            !appLock.isActive else {
```
typfel, line 42: To me it's nicer if we create the VC only after all the checks have been passed like this:

```suggestion
        guard 
            let appLock = appLock,
            appLock.needsToNotifyUser,
            !appLock.isActive
            let warningVC = changeWarningViewController, else {
```

typfel (changes requested):
Just minor nitpicks otherwise it looks good 👍

Wire-iOS/Sources/UserInterface/MainController/ZClientViewController.swift
KaterinaWire: Removed

typfel (approved):

johnxnguyen (approved):

PR #4853, opened 2021-01-20
Fix: MessageDetailsActionTests without using managed objects

billypchan:
## What's new in this PR?

Refactor `MessageDetailsActionTests` to prevent creating `NSMangedObject`s.


billypchan:
restart in https://github.com/wireapp/wire-ios/pull/4868

PR #4854, opened 2021-01-21
[Fix] Password validation

KaterinaWire:
### Issues

During password validation, we separately received validation error that password is tooShort, or a list of other password mismatches (e.g. uppercase, lowercase, digits). This is an issue in UI project, because we can update theCreate password screen with information that password satisfies minimum length or other requirements.

E.g. User A is entering the part of the password: Qwe1

**Expectation**
To see that this part satisfies 3 rules: uppercase, lowercase and digits.

**Outcome**
The password does not satisfy any requirements, because it's less than the minimum number of characters (error = tooShort).

https://wearezeta.atlassian.net/browse/SQSERVICES-208

### Solutions

Modify the structure of `PasswordValidationResult` to be able to contain a variety of rule violations:

```
public enum PasswordValidationResult: Equatable {

    /// The password is valid.
    case valid

	/// The password is invalid due to the violations.
	case invalid(violations: [Violation])

	public enum Violation {

	    /// The password is too short.
	    case tooShort
	
	    /// The password is too long.
	    case tooLong
	
	    /// The password contains a disallowed character.
	    case disallowedCharacter(Unicode.Scalar)
	
	    /// The password does not satisfy a requirement for a character class.
	    case missingRequiredClasses(Set<PasswordCharacterClass>)
	}
}
```
Needs releases with:

- [ wireapp/wire-ios-utilities/pull/104 ]

### Notes

`PasscodeValidationResult`, `PasscodeRuleSet`, `PasscodeCharacterClass` have been moved from `WireUtilities`.


Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
```diff
@@ -129,12 +129,14 @@ extension TextFieldValidator.ValidationError: LocalizedError {
...
-            default:
-                return PasswordRuleSet.localizedErrorMessage
+        case .invalidPassword(let validationResult):
+            switch validationResult {
+            case .invalid(let violations):
+                return violations.contains(.tooLong)
+                    ? "password.guidance.toolong".localized
+                    : PasscodeRuleSet.localizedErrorMessage
+            case .valid:
+                return ""
```
johnxnguyen, line 139: If you're only interested in the `invalid` case you can simplify this with:

```swift
case .invalidPassword(.invalid(let violations)):
  return ...
```

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeSetupInteractor.swift
```diff
@@ -48,52 +43,20 @@ extension PasscodeSetupInteractor: PasscodeSetupInteractorInput {
...
-                result = .error([.tooShort])
-            case .missingRequiredClasses(let passwordCharacterClass):
-                result = .error(passcodeError(from: passwordCharacterClass))
-            default:
-                result = .error([])
-            }
+            interactorOutput?.passcodeValidated(result: passwordValidationResult)
         default:
-            result = .error([])
+            interactorOutput?.passcodeValidated(result: .invalid(violations: []))
```
johnxnguyen, line 58: This doesn't really make any sense, I would expect that there is always at least one violation. Which situations lead to this case?

Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
```diff
@@ -21,23 +21,23 @@ import WireUtilities
 import FormatterKit
 import WireUtilities
 
-extension PasswordRuleSet {
+extension PasscodeRuleSet {
```
typfel, line 24: Aren't this password rules also used during registration?

Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
KaterinaWire: Yes, I'm interested only in the invalid case, but I should provide default value for all cases. I rewrited this part

```
case .invalidPassword(let validationResult):
            switch validationResult {
            case .invalid(let violations) where violations.contains(.tooLong):
                return "password.guidance.toolong".localized
            default:
                return PasscodeRuleSet.localizedErrorMessage
            }
```

Wire-iOS/Sources/UserInterface/Overlay/CustomAppLock/Setup/PasscodeSetupInteractor.swift
KaterinaWire: Yes, you are right. We will get `error = .invalidPassword(result) `or `error = nil` according to this code:
```
let result = PasscodeRuleSet.shared.validatePasscode(text)
return result != .valid ? .invalidPassword(result) : nil
```

Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
KaterinaWire: I moved `PasscodeRuleSet` class from `Utilities`. In the `Utilities` we use the word `Password` to name the classes, but in the UI we use the `Passcode`. I thought that it would be better if the names will match.

typfel: IMO it's better to keep the `Password` prefix since it's used for all our password validation: backup protection, registration and applock passcode, since that's more generic.

typfel (changes requested):
Please move the `PasscodeRulesSet` and related to classes out of the `AppLock` folder into a non feature related folder. (for example: `Sources/Helpers/PasswordRules/`) and change the prefix back to `Password` since these rules are not only related to the passcode.

Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
```diff
@@ -29,7 +29,7 @@ final class TextFieldValidator {
         case tooLong(kind: AccessoryTextField.Kind)
         case invalidEmail
         case invalidPhoneNumber
-        case invalidPassword(PasswordValidationResult)
+        case invalidPassword(PasscodeValidationResult)
```
johnxnguyen, line 32: Now that we have all violations held in a single `invalid` case, it would make more sense to changes this case to `invalidPassword(Violations)`.

Wire-iOS/Sources/Authentication/Helpers/PasscodeRules+Shared.swift
johnxnguyen: I think the terms "password" and "passcode" are interchangeable since both refer to a string of characters (regardless of what kind of characters).

typfel: I think in the context of mobile phones most people will think of an n-digit code when they read "passcode", might not be correct put that's what passcode means in the iOS world. I would prefer to keep using the word password at least for the generic code related to passwords.

KaterinaWire:
> Please move the `PasscodeRulesSet` and related to classes out of the `AppLock` folder into a non feature related folder. (for example: `Sources/Helpers/PasswordRules/`) and change the prefix back to `Password` since these rules are not only related to the passcode.

@typfel  Done

typfel (approved):

Wire-iOS/Sources/Authentication/Interface/Helpers/TextFieldValidator.swift
KaterinaWire: Done

johnxnguyen (approved):

PR #4855, opened 2021-01-21
[Feature] Add "Speaker / All" switch to call overlay

David-Henner:
## What's new in this PR?

- Creates a new UI component `RoundedSegmentedView`. It behaves similarly to `UISegmentedControl`, which was too rigid to satisfy design requirements. 
- Adds new component to the `CallActionsView` 
- Removes unused code


Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
```diff
@@ -0,0 +1,104 @@
...
+            let index = buttons.firstIndex(of: sender)
+        else { return }
+        
+        selectedButton = sender
+        setSelected(true, forItemAt: index)
+        delegate?.roundedSegmentedView(self, didSelectSegmentAtIndex: index)
+    }
+}
+
+private extension UIButton {
```
marcomaddy1982, line 99: Having a `private extension` is a good idea but another idea could be use a custom UIButton which `override` the `isSelect: Bool` instead to create a new method on the UIButton (also if private) :-) . Basically something like: 

```
class CustomButton: UIButton {
    override var isSelected: Bool {
        didSet {
            backgroundColor = isSelected ? .white : .clear
            setTitleColor(isSelected ? .black : .white, for: .normal)
        }
    }
}
```swift

Anyway, it is up to also in this manner is perfectly fine.

File: Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+
+protocol RoundedSegmentedViewDelegate: class {
+    func roundedSegmentedView(_ view: RoundedSegmentedView, didSelectSegmentAtIndex index: Int)
```
marcomaddy1982, line 23: This is just preference I have: in this case I like more to use a `actionBlock` (for instance you could add to the `init` method) than use the delegate pattern. The reason is the following: when I initialise the object I know directly which action is associated to it, on the contrary using the delegate pattern the logic related to the action is place where the delegate is implemented that sometimes for big class is not always super easy to find :-).
Anyway, it is just a thought, feel free to use your favourite approach :-).

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
```diff
@@ -0,0 +1,104 @@
...
+        setupViews(with: items)
+        setupConstraints()
+    }
+    
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    func setSelected(_ selected: Bool, forItemAt index: Int) {
```
marcomaddy1982, line 50: why don't just do:

```swift
for (i, button) in buttons.enumerated() {
  button.set(selected: i == index)
}
```
Am I missing something, i honestly don't understand the guard at the beginning :-)

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
David-Henner: Good point. I like the idea of making the action clearer and easier to find on the consumer side.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
David-Henner: If `selected = false`, then the `guard` ensures we're only unselecting the button at the given index, and not selecting all other buttons.

Your suggestion to use `buttons.enumerated()` is better than fetching the index each time though, so I'll update it.
And I'll see if I can find a clearer way to write this method.

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
```diff
@@ -0,0 +1,104 @@
...
+        selectedButton = sender
+        setSelected(true, forItemAt: index)
+        delegate?.roundedSegmentedView(self, didSelectSegmentAtIndex: index)
+    }
+}
+
+private extension UIButton {
+    func set(selected: Bool) {
+        backgroundColor = selected ? .white : .clear
+        setTitleColor(selected ? .black : .white, for: .normal)
```
typfel, line 102: You don't need to update this every time the selection changes since a UIButton can have different colours depending on its state. Instead you can do the following once during the button setup:

```
setTitleColor(.white, for: .normal)
setTitleColor(.black, for .selected)
setBackgroundImage(UIImage.singlePixelImage(with: .clear), for: .normal)
setBackgroundImage(UIImage.singlePixelImage(with: .white), for: .selectedl)
```

Wire-iOS/Sources/UserInterface/Calling/CallActionsView/RoundedSegmentedView.swift
typfel: A variant of this is to expose something like:
```
`func addButton(title: String, actionHandler: () -> Void)`
```

A benefit of having one handler per button is that the user of the `RoundedSegmentedView` doesn't need to look at the `index` to figure out which action he/she should perform.

typfel (changes requested):

marcomaddy1982 (approved):

marcomaddy1982:
Seems good for me!! Thanks 👍 @David-Henner

typfel (approved):

PR #4856, opened 2021-01-22
[Fix] unblocking not working from search results

typfel:
### Issues

Attempting to unblock a user found through search had no effect.

### Causes

The `ProfileViewControllerView` gets initialised with a  `ZMSearchUser` when a profile is opened from search. When performing the unblock action (see `toggleBlocked()`) we try to cast it to a `ZMuser` which fails and we end up doing nothing.

### Solutions

Avoid the casting to `ZMuser` by exposing the necessary methods/properties on the `UserType` protocol.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/1120


billypchan (approved):

typfel:
@billypchan please assign second reviewer

Wire-iOS/Sources/Helpers/syncengine/ZMUser+Additions.swift
```diff
@@ -19,12 +19,10 @@
 import WireDataModel
 
 extension UserType {
+    
     var nameAccentColor: UIColor? {
         return UIColor.nameColor(for: accentColorValue, variant: ColorScheme.default.variant)
     }
-}
-
-extension ZMUser {
```
KaterinaWire, line 27: I think it makes sense to change the file name as it's no longer  a `ZMUser` extension.

typfel: Done

KaterinaWire (approved):

PR #4871, opened 2021-01-28
[Refactor] Introduce locked app state

johnxnguyen:
### Issues

Currently the app lock sits in a window above the main window so that it can "lock" the app by hiding all content underneath it. This leads to several issues:

1. The app isn't really locked, event processing can happen underneath.
2. The app lock view controller and window always exists
3. Because the of this, the app lock view controller isn't strongly associated with a particular user session, it just takes the value of `ZMUserSession.shared()` (not nice for testing)
4. The app lock controller (and related classes) take on too much responsibility, such as observing app state changes to know when to show the app lock.

### Solutions

Introduce a `locked` app state to signify that the app is locked and can not be used. In this state, only the app lock UI is loaded. When the user successfully authenticates, then the app transitions to the `authenticated` state in which the conversation list is loaded.

This has several benefits:

1. The app is actually locked. No other event processing happens until it becomes unlocked.
2. The app lock UI only exists during the locked state and is torn down when it is no longer needed.
3. This means the app lock doesn't need to know when it needs to be shown or not. That logics in moved outside and is encoded in the state transitions.
4. Because we set up and tear down the app lock when needed, we can easily inject all dependencies that it needs, such an instance to the current user session. This makes testing much easier.

A summary of the flow is as follows:

1. Each session has an instance of `AppLockController` (defined in data model).
2. When a session is activated, the `SessionManager` reports to it's delegate (`AppStateCalculator`) that it activated the session.
3. The app state calculator checks to see if the session is locked, if it is, it transitions to the `locked` state, otherwise it transitions to the `authenticated` state.
4. The `AppLockModule` is created with the session, the user authenticates successfully and unlocks the session.
5. The session manager is informed the session unlocked and then informs the app state calculator that the active session changed again.
6. Go to step 3

### Viper module

These changes mean that the app lock module has lot less responsibility that before. Because of this, I decided to rewrite the module using the new Viper templates. Some of the code from the old module moved to the new module, some was moved to other components (such as the app lock controller) or the user session. The responsibility of the app lock module is now to **only provide a UI to allow the user to unlock the app**.

### Screen curtain

One consequence of moving the app lock out of the `NotificationWindow` is that it no longer serves as a screen curtain when the app is minimized. For preserve this feature, I've introduced the `ScreenCurtain` window. The `NotificationWindow` is no longer needed so it is deleted.

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-share-engine/pull/199
- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/1382
- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/276
- [x] https://github.com/wireapp/wire-ios-data-model/pull/1121

### Notes

I'm sorry for the size of this pull request, next time I'll try to plan to split it up better.


Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Router.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    func presentInputPasscodeModule(onGranted: @escaping () -> Void) {
+        // TODO: [John] Clean this up.
+        // TODO: [John] Inject these arguments.
+        let unlockViewController = UnlockViewController(selfUser: ZMUser.selfUser(), userSession: ZMUserSession.shared())
+        let keyboardAvoidingViewController = KeyboardAvoidingViewController(viewController: unlockViewController)
+        let navigationController = keyboardAvoidingViewController.wrapInNavigationController(navigationBarClass: TransparentNavigationBar.self)
+        navigationController.modalPresentationStyle = .fullScreen
+        unlockViewController.onGranted = onGranted
+        viewController?.present(navigationController, animated: false)
```
johnxnguyen, line 55: I'll address these todos when I refactor the `UnlockViewController`.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.LockView.swift
```diff
@@ -0,0 +1,172 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+import WireSystem
+
+extension AppLockModule.View {
+
+    final class LockView: UIView {
```
johnxnguyen, line 25: The UI code here has remained unchanged from the original `LockView`.

Wire-iOS/Sources/UserInterface/ConversationList/Container/ConversationListViewController.swift
```diff
@@ -166,10 +166,8 @@ final class ConversationListViewController: UIViewController {
         if !viewDidAppearCalled {
             viewDidAppearCalled = true
 
-            ZClientViewController.shared?.setUpCustomPasscodeIfNeeded(then: {
-                ZClientViewController.shared?.showDataUsagePermissionDialogIfNeeded()
-                ZClientViewController.shared?.showAvailabilityBehaviourChangeAlertIfNeeded()
-            })
```
johnxnguyen, line 172: We no longer need to set up the custom passcode here because when we transition from the unauthenticated state we will go through the locked state, and there we will set up the custom passcode.

Wire-iOS/Sources/UserInterface/CustomAppLock/WipeDatabase/WipeDatabaseWireframe.swift
```diff
@@ -34,11 +34,4 @@ final class WipeDatabaseWireframe {
         return viewController
     }
     
-    func presentWipeCompletion(from presentingViewController: WipeDatabaseUserInterface?) {
-        let wipeCompletionViewController = WipeCompletionViewController()
-        wipeCompletionViewController.modalPresentationStyle = .fullScreen
-        
-        AppDelegate.shared.notificationsWindow?.isHidden = false
-        (presentingViewController as? UIViewController)?.present(wipeCompletionViewController, animated: true)
-    }
```
johnxnguyen, line 43: We can't present this view here anymore (it was being presented in the notification window). Instead, it will be shown when we transition to the login screen.

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        interactor.needsToInformUserOfConfigurationChange = false
+
+        // When
+        sut.requestAuthentication()
+
+        // Then
+        XCTAssertEqual(router.methodCalls.presentCreatePasscodeModule.count, 1)
+        XCTAssertEqual(router.methodCalls.presentCreatePasscodeModule[0].shouldInform, false)
+
+        router.methodCalls.presentCreatePasscodeModule[0].completion()
```
KaterinaWire, line 64: This can be moved out in a separate test. The name of the test says that we should check that we have presented the CreatePasscode module.

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        interactor.currentAuthenticationType = .faceID
+        
+        // When
+        sut.authenticationEvaluated(with: .denied)
+
+        // Then
+        XCTAssertEqual(view.propertyCalls.state, [.locked(authenticationType: .faceID)])
+    }
+
+    func test_itRequestsToInputPasscode_WhenPasscodeIsNeeded() {
```
KaterinaWire, line 138: Typo
`It...`

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
```diff
@@ -0,0 +1,160 @@
...
+        XCTAssertEqual(view.propertyCalls.state, [.locked(authenticationType: .passcode)])
+        XCTAssertEqual(router.methodCalls.presentInputPasscodeModule.count, 1)
+
+        let onGranted = router.methodCalls.presentInputPasscodeModule[0]
+        onGranted()
+
+        XCTAssertEqual(interactor.methodCalls.openAppLock.count, 1)
+    }
+
+    func test_itUpdatesViewState_WhenAuthenticationMethodUnavailable() {
```
KaterinaWire, line 152: Typo
`_It...`

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -140,22 +140,20 @@ extension AppStateCalculator: SessionManagerDelegate {
         transition(to: appState,
                    completion: userSessionCanBeTornDown)
     }
-    
-    func sessionManagerDidReportDatabaseLockChange(isLocked: Bool) {
```
KaterinaWire, line 144: There are test that still use this method.

johnxnguyen: Yeah I know, I need to figure out how to fix those tests but might need to mock the user session.

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
johnxnguyen: Do you mean the assert below about the `openAppLock`?

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+}
+
+// MARK: - API for view
+
+extension AppLockModule.Presenter: AppLockPresenterViewInterface {
+
+    func requestAuthentication() {
+        if interactor.needsToCreateCustomPasscode {
```
marcomaddy1982, line 62: IMHO If we wanna stay sticky to the theory I guess that this `if` check should be made in the `Interactor`, then the `Presenter` should expose other methods depending on all the checks in the `Interactor`.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    }
+
+}
+
+// MARK: - API for presenter
+
+extension AppLockModule.Interactor: AppLockInteractorPresenterInterface {
+
+    var needsToInformUserOfConfigurationChange: Bool {
```
marcomaddy1982, line 76: IMHO the Interactor should not expose these properties. All the check done in the `Presenter` using these properties should be moved in the Interactor. 
Are you agree with me?

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
```diff
@@ -0,0 +1,83 @@
...
+        } else {
+            warnUserOfConfigurationChangeIfNeeded {
+                self.view.state = .authenticating
+                self.interactor.evaluateAuthentication()
+            }
+        }
+    }
+
+    private func warnUserOfConfigurationChangeIfNeeded(then block: @escaping () -> Void) {
+        guard interactor.needsToInformUserOfConfigurationChange else {
```
marcomaddy1982, line 75: Also here I guess this guard should be in the `Interactor`.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    func openAppLock()
+
+}
+
+// MARK: - View / Presenter
+
+protocol AppLockViewPresenterInterface: ViewPresenterInterface {
+
+    var state: AppLockModule.ViewState { get set }
```
marcomaddy1982, line 85: IMHO here the view should not expose the `state` property but should expose an `refresh` method to update the `View`. The `refresh` takes in input the `ViewState` or a AppLockViewModel object that contains all the info needed to the `View` to update itself.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
```diff
@@ -0,0 +1,113 @@
...
+
+        let lockView = LockView()
+
+        // MARK: - Life cycle
+
+        override func viewDidLoad() {
+            super.viewDidLoad()
+            setUpViews()
+            refresh()
+            presenter.requestAuthentication()
```
marcomaddy1982, line 48: IMHO the `Presenter` doesn't really know what we are going to do next, so the it should not expose a method called `requestAuthentication()` but `viewDidLoad()` . In the `viewDidLoad` method of the `Presenter` two methods will be called: 

1) `view.refresh()`, so you can remove the refresh on top of this 
2) `Interactor.requestAuthentication()`

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
johnxnguyen: Right, I wasn't sure about that. In my head, I see read only properties as if they were methods, but I guess then we still need to go back through another presenter method.

So should it look something like this?

```swift
// In presenter
func requestAuthentication() {
  interactor.evaluateAuthenticationState()
}

// In interactor
func evaluateAuthenticationState() {
  if needsToCreateCustomPasscode {
    presenter.authenticationStateEvaluated(.needsToCreateCustomPasscode)
  } else {
    presenter.authenticationStateEvaluated(.ready)
}

// In presenter
func authenticationStateEvaluated(_ state: State) {
  switch state {
  case .needsToCreateCustomPasscode:
    // Present passcode creation module
  case .ready:
    // Start authentication
  }
}
```

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift
johnxnguyen: Yeah I guess this is in the same vein as the previous comment. I can move them over to the interactor.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.swift
johnxnguyen: Yep makes sense.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
johnxnguyen: Right, I initially named it `start` to indicate that the presenter should "start". I was preferring this over calling something like `presenter.viewDidLoad()` because I feel that the method should be orientated toward the receiver and not the caller. So `presenter.start()` would make more sense in that regard.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
marcomaddy1982: Honestly I wouldn't pass the `State` object to the `Presenter` but directly call from the `Interactor` two more descriptive methods exposed from the `Presenter`.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
marcomaddy1982: Ok, I understand your point. it totally fine 👍 .

Wire-iOS/Sources/AppStateCalculator.swift
johnxnguyen: Fixed those tests by using a mock user session.

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
johnxnguyen: Fixed

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
johnxnguyen: Fixed

Wire-iOS Tests/App Lock/Tests/AppLockModulePresenterTests.swift
johnxnguyen: Fixed in c522cca311f2abff586bb80b1496779cdacb1ce1

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.swift
johnxnguyen: Fixed in 793b1bcee16f2b4f466ace44135e695e5a073394

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
johnxnguyen: I decided to try solve this in a slightly different way by defining various events (such as `viewDidLoad` and `unlockButtonTapped`) and sending these events to the presenter for processing. Changes are in 210f106cf8eb3cda6328c43b479aa0e2d4739d44

First I remembered why I had renamed the method from `start` to `requestAuthentication`, because I was calling this method when the view loaded but also when the unlock button was tapped. It sounded weird to call `start` when tapping the unlock button.

Then I went back to your original suggestion to add methods on the presenter `viewDidLoad()` and also `unlockButtonTapped()`, but then I realized we can fold all these events into a single process method. I think the approach is a nice solution because it satisfies both our wishes, namely that the view only knows about the events that it sends to the presenter, and the method `presenter.processEvent(.viewDIdLoad)` makes more sense than `presenter.viewDidLoad()`.

What do you think?

marcomaddy1982 (approved):

marcomaddy1982:
@johnxnguyen great job!! Really beautiful 👍 💯

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Interactor.swift
```diff
@@ -33,6 +33,8 @@ extension AppLockModule {
 
         let dispatchGroup = DispatchGroup()
 
+        private let isReadyForAuthentication = DispatchSemaphore(value: 1)
```
marcomaddy1982, line 36: Nice solution :-).  Anyway why don't use a `Bool`? Do you have a particular reason to use a `DispatchSemaphore` or you wanted just to use them :-).

johnxnguyen: It seemed like a a good idea but I realize now that it's probably not needed in this branch (I copied it over from develop so as not to cause a regression).

PR #4878, opened 2021-02-01
[Feature] update presentation mode when toggled

David-Henner:
## What's new in this PR?

In [this PR](https://github.com/wireapp/wire-ios/pull/4855), we introduced a new UI component: `RoundedSegmentedView`. Which was added to `CallActionsView`. (see screenshot)
It offers a selection between **Active speakers** and **All.** 
**All** being all current participants that have their video enabled. 

In [an other PR](https://github.com/wireapp/wire-ios-sync-engine/pull/1384), we created a `VideoGridPresentationMode` type and added it as a property of `CallSnapshot`. Allowing us to keep track of the selected presentation mode for a given call.

The PR you're looking at is connecting the **Active speaker / All** button to the presentation mode property. Allowing the user to change presentation mode by clicking on the button.

Note that this interaction only changes the `VideoGirdPresentationMode` property but is not reflected in the video grid. Updating the video grid will be done in a separate PR.


marcomaddy1982 (approved):

marcomaddy1982:
@David-Henner look fine for me :+1 . Anyway there are some snapshot tests failing, but probably cause this PR  still not reflect in the video grid. Right?

KaterinaWire (approved):

PR #4879, opened 2021-02-01
[Fix] video capture state

David-Henner:
## What's new in this PR?

start self preview capture only when video state is `.started`


johnxnguyen (changes requested):
It would be good at tests.

KaterinaWire (approved):

johnxnguyen (approved):

PR #4885, opened 2021-02-02
Fix: remove CoreDataFixtureTestHelper from GroupParticipantsDetailViewControllerTests

billypchan:
## What's new in this PR?

Remove `CoreDataFixtureTestHelper` from `GroupParticipantsDetailViewControllerTests` to prevent using `NSMangedObject` in tests


Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
```diff
@@ -35,70 +36,64 @@ extension ZMConversation {
...
     }
-    
+
     func testThatItRendersALotOfUsers() {
         // given
-        let users = (0..<20).map { createUser(name: "User #\($0)") }
+        let users: [MockUserType] = (0..<20).map {
+            let user = MockUserType.createUser(name: "User #\($0)")
+            return user
+        }
```
KaterinaWire, line 70: Could we write it like this
`let users: [MockUserType] = (0..<20).map { MockUserType.createUser(name: "User #\($0)") }`
?

Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
```diff
@@ -20,6 +20,7 @@ import XCTest
 @testable import Wire
 import SnapshotTesting
 
+/// TODO: retire this extension
```
typfel, line 23: What is necessary to remove this extension? Can it be done in this PR?

Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
```diff
@@ -35,70 +36,64 @@ extension ZMConversation {
...
-        coreDataFixture = CoreDataFixture()
-    }
-    
-    override func tearDown() {
-        resetColorScheme()
-        coreDataFixture = nil
 
-        /// restore to default light scheme
-        ColorScheme.default.variant = .light
+        SelfUser.setupMockSelfUser()
```
KaterinaWire, line 56: Why do we need it?

Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
billypchan: There are still some tests using this extension, it will be done after those tests are refactored.

Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
billypchan: Self user is called in `GroupParticipantsDetailViewController` so we need to set up the mock self user.

Wire-iOS Tests/GroupParticipantsDetailViewControllerTests.swift
billypchan: fixed

typfel (approved):

KaterinaWire (approved):

PR #4893, opened 2021-02-04
[Fix] Unlock share extension with a custom passcode

KaterinaWire:
## What's new in this PR?
https://wearezeta.atlassian.net/browse/SQSERVICES-225

### Issues

The share extension has no user interface to allow entering a custom passcode, which means that if the users rely on this custom passcode (i.e they can’t use biometrics) then they will not be able to unlock the share extension.

### Causes

This hasn't been implemented before.

### Solutions
1. Added a simplified version  of `UnlockViewController` to the Share extension. This viewController required a passcode field (`AccessoryTextField`) that is implemented in the UI target and contains a heavy logic. For this reason, I created `BaseAccessoryTextField` with basic logic that could be used for both the UI and the shareExtension targets.

2. Some extensions were opened for the Share extension.

### Notes

Working on adding tests now.


Wire-iOS Share Extension/UIButton+SetIcon.swift
```diff
@@ -0,0 +1,44 @@
...
+                 for state: UIControl.State) {
+        guard let iconType = iconType else {
+            setImage(nil, for: state)
+            return
+        }
+        
+        let image = UIImage.imageForIcon(iconType, size: size.rawValue, color: .black)
+        let renderingMode: UIImage.RenderingMode = UIImage.RenderingMode.alwaysTemplate
+        
+        setImage(image.withRenderingMode(renderingMode), for: state)
```
johnxnguyen, line 42:
```suggestion        
        setImage(image.withRenderingMode(.alwaysTemplate), for: state)
```

Wire-iOS/Sources/UserInterface/Components/UIColor+Const.swift
```diff
@@ -20,6 +20,8 @@ import UIKit
 extension UIColor {
     enum PasscodeUnlock {
         static let error: UIColor = UIColor(rgb: (236, 121, 121))
+        static var buttonGraphite: UIColor = UIColor(rgb: (51, 55, 58))
+        static var buttonLightGraphite: UIColor = UIColor(rgb:(141, 152, 159))
```
johnxnguyen, line 24: We should be using the colors from the color scheme. Are these new colors?

Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
```diff
@@ -34,7 +34,7 @@ protocol AccessoryTextFieldDelegate: class {
     func buttonPressed(_ sender: UIButton)
 }
 
-final class AccessoryTextField: UITextField, TextContainer, Themeable {
+final class AccessoryTextField: BaseAccessoryTextField, TextContainer, Themeable {
```
johnxnguyen, line 37: I think the `Base` prefix isn't very descriptive, so I would suggest to rename these types to something like:

```swift
final class ValidatedTextField: AccessoryTextField
```

Wire-iOS Share Extension/AccessoryTextField.swift
```diff
@@ -0,0 +1,137 @@
...
+    var revealButtonIcon: StyleKitIcon? {
+        didSet {
+            updateButtonIcon()
+        }
+    }
+    
+    lazy var revealButton: UIButton = {
+        let iconButton = UIButton()
+        
+        iconButton.tintColor = UIColor.Team.textColor
```
johnxnguyen, line 46: A team color seems a bit out of place here.

Wire-iOS Share Extension/AccessoryTextField.swift
```diff
@@ -0,0 +1,137 @@
...
+        iconButton.setBackgroundImage(UIImage.singlePixelImage(with: .clear), for: state)
+        
+        iconButton.adjustsImageWhenDisabled = false
+        
+        iconButton.accessibilityIdentifier = "passcode_text_field.button.reveal"
+        iconButton.accessibilityLabel = "Reveal passcode".localized
```
johnxnguyen, line 52: Is this accessibility label a key or content?

Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -433,11 +439,13 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
     
     private func presentChooseConversation() {
-        requireLocalAuthenticationIfNeeded(with: { [weak self] (status) in
-            if let status = status, status != .denied {
-                self?.showChooseConversation()
+        requireLocalAuthenticationIfNeeded { [weak self] in
+            guard let `self` = self else { return }
+            
+            if self.localAuthenticationStatus != .denied {
+                self.showChooseConversation()
```
johnxnguyen, line 446: Probably we should change this to

```swift
guard self.localAuthenticationStatus == .granted else { return }
self.showChooseConversation()
```

johnxnguyen (commented):
So which types did you need to move to common components?

Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
KaterinaWire: Yes, it makes sense to have `ValidatedTextField` in the UI. But for Share extension, I would give the name: 
`final class PasscodeTextField: AccessoryTextField`
because it's for passcode only and we are not doing any validation.

Wire-iOS Share Extension/AccessoryTextField.swift
KaterinaWire: It was a content. I changed it to the key.

Wire-iOS Share Extension/ShareExtensionViewController.swift
KaterinaWire: Fixed

Wire-iOS Share Extension/AccessoryTextField.swift
KaterinaWire: I used the same color as we are using in a current   `AccessoryTextField`.

Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
johnxnguyen: Sounds good to me 👍🏻

Wire-iOS/Sources/UserInterface/Components/UIColor+Const.swift
KaterinaWire: These colors are using by UnlockViewController in the `Share extension`. This colours are using by `UnlockViewController` in  Share extension. For this screen, we always use dark mode.

Wire-iOS/Sources/Authentication/Interface/Views/AccessoryTextField.swift
KaterinaWire: Done

KaterinaWire:
> So which types did you need to move to common components?

`ColorScheme.swift`.

Wire-iOS/Sources/UserInterface/Components/UIColor+Const.swift
KaterinaWire: I copied these colors from the `ColorScheme`. So, I think it's better to move `ColorScheme` to the Common components.

Wire-iOS Share Extension/ShareExtensionViewController.swift
```diff
@@ -501,43 +538,63 @@ final class ShareExtensionViewController: SLComposeServiceViewController {
...
+                let alert = UIAlertController(title: "share_extension.unlock.alert.title".localized, message: "share_extension.unlock.alert.message".localized, alertAction: .ok(style: .cancel))
+                self.present(alert, animated: true, completion: nil)
+                
+                localAuthenticationStatus = .denied
+            } else {
+                requestCustomPasscode { [weak self] status in
+                    guard let `self` = self else { return }
+                    
+                    self.localAuthenticationStatus = status
+                    completion()
```
David-Henner, line 570: doesn't this mean that `completion()` will be called twice because of the `defer` statement?

johnxnguyen (changes requested):
You should probably delete the old `AccessoryTextFieldTests/` snapshot directory.

KaterinaWire:
> `AccessoryTextFieldTests/` snapshot directory

Deleted

Wire-iOS Share Extension/ShareExtensionViewController.swift
KaterinaWire: Yes, you are right. I removed `defer` statement.

KaterinaWire:
@johnxnguyen Could you please help with the copies in the popup? Eva asked, because Lis is on vacation now.
"share_extension.unlock.alert.title"
"share_extension.unlock.alert.message"

KaterinaWire:
@johnxnguyen @David-Henner I will add tests in the separate PR, because it's not a trivial task. I can't just add Wire-Share extension Tests or add a test to the Wire-iOS Tests, that requires more research.

David-Henner (approved):

johnxnguyen:
> Could you please help with the copies in the popup? Eva asked, because Lis is on vacation now.

@KaterinaNikolaenko  I think the title is unnecessary here. Then I would change the message to:
```
"share_extension.unlock.alert.message" = "Open Wire to create a passcode";
```

johnxnguyen (approved):

PR #4905, opened 2021-02-08
refactor UserSearchResultsViewControllerTests

billypchan:
## What's new in this PR?

Refactor `UserSearchResultsViewControllerTests` with Swift snapshot testing and prevent creating `ZMUser`s


Wire-iOS Tests/UserSearchResultsViewControllerTests.swift
```diff
@@ -19,77 +19,69 @@
...
-
-        for name in usernames {
-            let user = ZMUser.insertNewObject(in: uiMOC)
-            user.remoteIdentifier = UUID()
-            user.teamIdentifier = nil
-            user.name = name
-            user.setHandle(name.lowercased())
+        var allUsers: [UserType] = []
+
+		for name in XCTestCase.usernames {
```
johnxnguyen, line 81: Where is `XCTestCase.usernames` defined? Why is it an extension on `XCTestCase`?

johnxnguyen (changes requested):
There are quite a few issues with indentation, looks like your editor is using 2 spaces.

Wire-iOS Tests/UserSearchResultsViewControllerTests.swift
billypchan: `usernames` was a property `CoreDataSnapshotTestCase`. I put it on its super class `XCTestCase` after refactoring. We can also put it in another class like `MockUserType`

johnxnguyen: I think `MockUserType` is a more suitable place to put it.

johnxnguyen (approved):

marcomaddy1982 (approved):

PR #4911, opened 2021-02-09
[Chore] Add type safe localization keys

johnxnguyen:
### Issues

Localized string keys are stringly typed which is error prone for several reasons:

1. A typo in the localization key is not caught at compile time and the result is that the misspelt key becomes the string value at runtime.
2. Argument placeholders accept specific types and will crash at runtime if an invalid value is passed into it.
3. It is tempting to construct long localized keys by joining together parts of the key. This makes search for both the usage and definitions of localized keys in code and the base string file respectively.

### Solutions

Add type safety to localization keys by generating Swift types to mimic the inherent tree structure of localized keys. This is achieved through the [SwiftGen](https://github.com/SwiftGen/SwiftGen) tool, which will read the `Resources/Base.lproj` directory to generate the code.

Keys can now be constructed as follows:

```swift
// "content.system.cannot_decrypt.error_details" = "(Fixed error: %d ID: %@)"
L10n.Localizable.Content.System.CannotDecrypt.errorDetails(p1: Int>, <p2: Any)
```

The generated definition is:

```swift
/// (Fixed error: %d ID: %@)
internal static func errorDetails(_ p1: Int, _ p2: Any) -> String {
    return L10n.tr("Localizable", "content.system.cannot_decrypt.error_details", p1, String(describing: p2))
}
```

Since this is a swift type, you can cmd click it to go to the definition. There you can see the full localization key and the localized string value as a doc comment. 

### Dependencies

In order for this to work, the `swiftgen` tool must be run whenever the base localization files change. The tool can be installed locally via homebrew.

### Notes

Do we want to add script to run whenever the localization files change?

### DIFF
--- a/Wire-iOS.xcodeproj/project.pbxproj
+++ b/Wire-iOS.xcodeproj/project.pbxproj
@@ -7920,6 +7920,7 @@
 				5E35F77C2182124E00D3F4FE /* PerformanceDebugger.swift in Sources */,
 				636AD79A23A916CB004A3050 /* AppLockPresenter.swift in Sources */,
 				5E8FFC0A21ECE3920052DF03 /* BackupRestoreStepDescription.swift in Sources */,
+				EEA65C2725D27A8300AA7519 /* Strings+Generated.swift in Sources */,
 				87DCF49C1D34E9E600BB420F /* LocationSendViewController.swift in Sources */,
 				F1199D421FC849450070FAC3 /* VerificationCodeFieldDescription.swift in Sources */,
 				BF808B4D1DE5AD1600718076 /* SettingsCellDescriptorFactory+Account.swift in Sources */,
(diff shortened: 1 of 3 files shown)


typfel:
Yes, I think we definitely want to add a script which we can run whenever the localization strings changes. The question is also how we guarantee that the scripts always gets run when it needs to.

Alternatives I can think of:
1. Trigger the script from setup.sh. 
2. Add a build phase which triggers the script

johnxnguyen:
@typfel I guess it depends on whether we commit the generated file or not? If we do, then I think we'd only need a build phase script. If we don't commit the generated code then we'd have to add it to the `setup.sh` so that CI would have access to the types.

I guess the workflow would be something like: add, remove of change a key or value, build, then use the generated types.

typfel:
>@typfel I guess it depends on whether we commit the generated file or not? If we do, then I think we'd only need a build phase script. If we don't commit the generated code then we'd have to add it to the setup.sh so that CI would have access to the types.

I'm in favour of committing the generated file because tools can in theory become incompatible with old versions of our code.

johnxnguyen:
@typfel I've added the build phase script in b535ae82996491f218b6225f7c5beb595bae988f

We don't need to specify input files because `swiftgen` already checks if there are changes to make.

typfel (approved):

David-Henner (approved):
🚀

PR #4913, opened 2021-02-09
Update createBumpPR.yml

billypchan:
## What's new in this PR?

expand script in `createBumpPR.yml` to debug get latest release failure.


billypchan:
confirm that the random release version  request is caused by api limit, providing access token can solve the issue.

PR #4925, opened 2021-02-10
Update createBumpPR.yml

billypchan:
## What's new in this PR?

rewrite update cartfile code in Swift for better readability and easier to fix bugs.


billypchan:
tested Swift script works and the performance is no huge difference with shell script

PR #4929, opened 2021-02-11
[Feature]: SQCORE-212: Hide Search Icon when EAR is enabled.

marcomaddy1982:
## What's new in this PR?

Currently when EAR is being enable the search icon in a conversation is disabled, this behaviour could confuse the user cause the icon is visible but he/she couldn’t use it.  

To avoid this when EAR is being enabled the search icon in a conversation will be no longer visible.


Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
```diff
@@ -132,6 +131,13 @@ extension ConversationViewController {
     }
 
     private func shouldShowCollectionsButton() -> Bool {
+        guard
+            SecurityFlags.forceEncryptionAtRest.isEnabled == false,
```
typfel, line 135: You can remove this since `session.encryptMessagesAtRest` is always true if `SecurityFlags.forceEncryptionAtRest.isEnabled` is true.

typfel (changes requested):

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
marcomaddy1982: @typfel I know this,  but I added also it cause we should hide he search icon when `SecurityFlags.forceEncryptionAtRest.isEnabled` is `false`,  and `session.encryptMessagesAtRest` is `true` cause enabled from the user.

typfel: Right, but you achieve this behaviour by only checking  `session.encryptMessagesAtRest == false`. If the user toggles EAR `session.encryptMessagesAtRest` will be `true`.

marcomaddy1982: Now I remember why i put both checks in the `guard` statements,  I tried again and correct me if I'm wrong. The scenariois the following:
1) in the `Security-Flags.xcconfig` file  `FORCE_ENCRYPTION_AT_REST_ENABLED=1`
and 
2) user disabled EAR

having just `session.encryptMessagesAtRest` in the `guard` the result is that the search icon is visible.

johnxnguyen: How is it possible that the user disables EAR when it's being forced?

johnxnguyen: So we found that `SecurityFlags` exists only in the UI project and is not used to influence the value of the `encryptMessagesAtRest` property on the user session. The old flag `encryptionAtRestEnabledByDefault` in `session_manager.json` is used for this purpose.

So, the logic here is correct, but I would say it's dangerous to have two different flags for the same purpose. We should remove one of them.

johnxnguyen (approved):

KaterinaWire (approved):

Wire-iOS/Sources/UserInterface/Conversation/ConversationViewController+NavigationBar.swift
marcomaddy1982: @johnxnguyen thanks for helping clarify

PR #4931, opened 2021-02-11
[Feature] Display smoothed active speakers in Speakers view

David-Henner:
## What's new in this PR?

This PR introduces the display of active speakers in the **Speakers** view.

Call participants qualify for the **Speakers** view when their audio level is positive for some time. They can remain in the view for some time after their audio level goes down, but will eventually be disqualified after being inactive for an arbitrary time period.

Qualified active speakers will be referred to as **smoothed** active speakers.

---

A video call can have two presentation modes:
- `allVideoStreams`. All the call participants sharing video are shown in the video grid. Active speakers are indicated with a coloured border and a pulsing microphone.
- `active speakers`. Only qualified active speakers are shown. Active speakers do not have a coloured border but do have a pulsing microphone.

The call overlay allows the user to switch between the two presentation modes. 
PR: [[Feature] add `videoGridPresentationMode` to call snapshots](https://github.com/wireapp/wire-ios-sync-engine/pull/1384) 
PR: [[Feature] update presentation mode when toggled](https://github.com/wireapp/wire-ios-sync-engine/pull/1384)

---

The Sync Engine provides an endpoint for the UI to get a list of call participants. 
There are two kinds of lists:
- the `all` list, composed of all call participants. 
- the `smoothedActiveSpeakers` list, composed only of qualified active speakers.

PR: [[Feature] integrate 'smoothed' active speakers to calling logic](https://github.com/wireapp/wire-ios-sync-engine/pull/1391)

---

### Changes

This PR update the logic for the creation of a `VideoConfiguration` object. `VideoConfiguration` objects are passed to the video grid and contain data about which video streams to present in the grid and how to present them.

The updated logic takes into account the presentation mode to create and arrange the video streams accordingly. 

Mainly, if the presentation mode is `allVideoStreams`, it will use a participant list of kind `all` to create the video streams, if the mode is `activeSpeakers`, it will use a list of kind `smoothedActiveSpeakers`. It also handles some minor differences between the two modes.


marcomaddy1982:
@David-Henner the comment i have made doesn't make much difference so for me the PR is good 👍 . Thanks

marcomaddy1982 (approved):

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -45,14 +45,78 @@ extension CallParticipant {
 }
 
 extension VoiceChannel {
+    
+    // MARK: - Video Stream Arrangment
+    
+    typealias VideoStreamArrangment = (preview: VideoStream?, grid: [VideoStream])
+        
+    fileprivate func videoStreamArrangment() -> VideoStreamArrangment {
```
KaterinaWire, line 53: Is there a reason to declare this method as `fileprivate`? Because others methods in this file are `private`.

KaterinaWire, line 53: Why did you decide to make `videoStreamArrangment` as a method and not a property as it was?

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
```diff
@@ -82,62 +146,55 @@ extension VoiceChannel {
...
-    fileprivate var videoStreamArrangment: (preview: VideoStream?, grid: [VideoStream]) {
-        guard isEstablished else { return (nil, selfStream.map { [$0] } ?? [] ) }
 
-        let videoStreams = Array(sortedActiveVideoStreams.prefix(VideoConfiguration.maxVideoStreams))
+    private func selfStream(from videoStreams: [VideoStream], createIfNeeded: Bool) -> VideoStream? {
         let selfStream = videoStreams.first(where: { $0.stream.streamId == selfStreamId })
-
-        return arrangeVideoStreams(for: selfStream ?? self.selfStream, participantsStreams: videoStreams)
+        
+        return selfStream ?? (createIfNeeded ? self.selfStream : nil)
```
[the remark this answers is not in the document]
KaterinaWire: I would suggest to write like this:
```
guard let selfStream = videoStreams.first(where: { $0.stream.streamId == selfStreamId }) else {
            return createIfNeeded ? self.selfStream : nil
        }
        return selfStream
```
But feel free to keep it, if you like.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
David-Henner: this method is defined in an extension of `VoiceChannel` but needs to be accessed from outside the extension

David-Henner: I thought `videoStreamArrangment` was starting to become too complex to be a computed property, so I decided to make it a function in order to reflect that.

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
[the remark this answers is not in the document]
David-Henner: > I guess here could be just `selfStream ?? nil`. Or am I wrong?

`selfStream` is defined in the function's scope, whereas `self.selfStream` is a computed property from the `VoiceChannel` extension. Renaming it might help clarify the code

Wire-iOS/Sources/UserInterface/Calling/CallViewController/VideoConfiguration.swift
marcomaddy1982: Maybe you could rename the the method `videoStreamArrangment` in `buildVideoStreamArrangment` or `createVideoStreamArrangment`. It could be a bit confusing having an `videoStreamArrangment` method and an 
`arrangeVideoStreams(for ...` method.

KaterinaWire (approved):

PR #4936, opened 2021-02-12
[Fix] App Lock Screen copy when biometrics are disabled

KaterinaWire:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/SQSERVICES-198

### Issues

When EAR is enabled and there're no biometrics on the device, we show an `Unlock` button on the Applock screen, but this button does nothing (we can't unlock Wire with a custom passcode).

### Solutions

Rename the button and open the device settings when the user press it.


Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
```diff
@@ -155,8 +166,20 @@ final class AppLockView: UIView {
...
+extension AppLockView {
+    @objc
+    private func onReauthenticatePressed(_ sender: AnyObject!) {
         self.onReauthRequested?()
     }
+    
+    @objc
+    private func openSettings(_ sender: Any?) {
+        if let url = URL(string: UIApplication.openSettingsURLString) {
+            UIApplication.shared.open(url, options: [:], completionHandler: nil)
```
johnxnguyen, line 182: Do this just open the settings? Can we open to the passcode page?

Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
```diff
@@ -46,13 +47,15 @@ final class AppLockView: UIView {
...
+            self.setupActionButton()
             self.authenticateLabel.isHidden = !showReauth
-            self.authenticateButton.isHidden = !showReauth
+            self.actionButton.isHidden = !showReauth
         }
     }
 
     init(authenticationType: AuthenticationType = .current) {
-
+        self.authenticationType = authenticationType
```
johnxnguyen, line 57: You don't need to store this as a property if you pass it to the `setupActionButton` method.

johnxnguyen (commented):
What happens when the user sets a passcode and comes back to the app? Will the view refresh?

Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
KaterinaWire: Yes, it opens the settings. I've made some research on how to open specific settings. There are many articles that suggest using URL schemes (e.g.` UIApplication.shared.open(URL(string: "App-prefs:Bluetooth"))`), but Apple could reject the app if we take this approach.
Here are a couple of links explaining it:
https://stackoverflow.com/questions/28152526/how-do-i-open-phone-settings-when-a-button-is-clicked
https://developer.apple.com/forums/thread/100471
https://developer.apple.com/forums/thread/65007

Wire-iOS/Sources/UserInterface/Overlay/AppLockView.swift
KaterinaWire: I merged this PR with the `develop` branch. There are new changes.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
```diff
@@ -93,7 +93,28 @@ extension AppLockModule {
                 return "self.settings.privacy_security.lock_cancelled.description_passcode_unavailable".localized
             }
         }
+        
+        var buttonTitle: String {
+            guard case let .locked(authenticationType) = self else { return "" }
+
+            switch authenticationType {
+            case .unavailable:
+                return "registration.push_access_denied.settings_button.title".localized
```
johnxnguyen, line 102: I think we should start to scope localized strings to the module so we avoid to use strings from other parts of the app. This is something that I will do for the other strings in the module, but I think it's worth pointing out here to keep it in mind for the future.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.swift
```diff
@@ -48,6 +48,7 @@ extension AppLockModule {
 
         case viewDidLoad
         case unlockButtonTapped
+        case openSettingsTapped
```
johnxnguyen, line 51:
```suggestion
        case openSettingsButtonTapped
```

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.LockView.swift
```diff
@@ -166,7 +171,7 @@ extension AppLockModule.View {
         @objc func onReauthenticatePressed(_ sender: AnyObject!) {
             onReauthRequested?()
```
johnxnguyen, line 172: Since the name of the button changed, I'd change the names here too.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
```diff
@@ -93,7 +93,28 @@ extension AppLockModule {
...
+        var onEvent: AppLockModule.Event {
+            guard case let .locked(authenticationType) = self else { return .unlockButtonTapped }
+            
+            switch authenticationType {
+            case .unavailable:
+                return .openSettingsTapped
+            default:
+                return .unlockButtonTapped
+            }
+        }
```
johnxnguyen, line 117: I would expect something called `onEvent` to be a function. I think a more suitable name would be `buttonEvent` or `buttonAction`.

johnxnguyen: You could simplify this with:

```swift
switch self {
case .locked(.unavailable):
    return openSettingsTapped
default:
    return .unlockButtonTapped
}
```

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
```diff
@@ -105,6 +126,10 @@ extension AppLockModule.View: AppLockViewPresenterInterface {
     func refresh(withModel model: AppLockModule.ViewModel) {
         lockView.showReauth = model.showReauth
         lockView.message = model.message
+        lockView.buttonTitle = model.buttonTitle
+        lockView.onReauthRequested = { [weak self] in
+            self?.presenter.processEvent(model.onEvent)
+        }
```
johnxnguyen, line 132: Now that you send the event from here, you can remove it from the `setUpViews` method.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
```diff
@@ -90,4 +94,9 @@ extension AppLockModule.Presenter {
         interactor.executeRequest(.evaluateAuthentication)
     }
 
+    private func openSettings() {
+        if let url = URL(string: UIApplication.openSettingsURLString) {
+            UIApplication.shared.open(url, options: [:], completionHandler: nil)
+        }
+    }
```
johnxnguyen, line 101: This is navigation so I would say it's the responsibility of the router. I would add a new action case `presentDeviceSettings`, ask the router to perform it, then have the router call this code.

johnxnguyen (changes requested):
I think I replaced the old lock view tests with the app lock module view tests. I think you can delete the snapshots in the `AppLockViewSnapshotTests/` folder.

Also, please add tests for the event.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
marcomaddy1982: I would also add the control `UIApplication.shared.canOpenURL`.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.LockView.swift
marcomaddy1982: At this point I would also change the naming of the label `authenticateLabel`

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
marcomaddy1982: Shouldn’t we merge SwiftGen? if yes, why don't start using it?

johnxnguyen: It is merged, and yes, we should start to use it. But I was thinking that this key should be replaced anyway, it could be included in: https://github.com/wireapp/wire-ios/pull/4939

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.LockView.swift
KaterinaWire: Done

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
KaterinaWire: Should I copy this string to the `App lock module section` and create a key like "appLockModule.button.open_settings.title"? And it also means we should rename the current key for the button title.

KaterinaWire:
> I think I replaced the old lock view tests with the app lock module view tests. I think you can delete the snapshots in the `AppLockViewSnapshotTests/` folder.
> 
> Also, please add tests for the event.

The snapshots were removed and tests added.

KaterinaWire:
> What happens when the user sets a passcode and comes back to the app? Will the view refresh?

Unfortunately no

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
```diff
@@ -95,23 +91,18 @@ extension AppLockModule {
...
         var buttonTitle: String {
-            
-            guard case let .locked(authenticationType) = self else { return "" }
-            
-            switch authenticationType {
-            case .unavailable:
-                return L10n.Localizable.Registration.PushAccessDenied.SettingsButton.title 
+            switch self {
+            case .locked(.unavailable):
+                return L10n.Localizable.Registration.PushAccessDenied.SettingsButton.title
```
johnxnguyen, line 96: I think now we should add a new localized string for the settings button, and change the key for the unlock button:

```
"appLockModule.unlockButton.title" = "Unlock";
"appLockModule.goToSettingsButton.title" = "Go to settings";
```

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Router.swift
```diff
@@ -75,5 +78,11 @@ extension AppLockModule.Router: AppLockRouterPresenterInterface {
         warningViewController.delegate = view
         view.present(warningViewController, animated: false)
     }
+    
+    private func presentDeviceSettings() {
+        if let url = URL(string: UIApplication.openSettingsURLString), UIApplication.shared.canOpenURL(url) {
+            UIApplication.shared.open(url, options: [:], completionHandler: nil)
+        }
```
johnxnguyen, line 85: If we make this all an extension on `UIApplication`, then we can hide away all details and make it reusable:

```swift
extension UIApplication {

    func openSettings() {
        if let url = URL(string: openSettingsURLString), canOpen(url) {
            open(url, options: [:], completionHandler: nil)
        }
    }

}
```

johnxnguyen (changes requested):

johnxnguyen:
> Unfortunately no

So I would expect that when coming back to the app after setting a passcode, the user would be asked to enter the passcode.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Router.swift
marcomaddy1982: Make sense 👍

KaterinaWire: Done

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
johnxnguyen: I see two options:

1. Add a new key in the app lock module section (see other comment about this)
2. Or add a new key in a "general" section, and use that key here in app lock, but also wherever the existing key is used.

I would go with option 1. I don't think it's so bad to have some duplicate localized strings since it's low effort to translate them (just duplicate translations).

marcomaddy1982 (approved):

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
KaterinaWire: I added those strings

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.View.swift
KaterinaWire: Added a new key

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
```diff
@@ -72,6 +72,9 @@ extension AppLockModule.Presenter: AppLockPresenterViewInterface {
         case .viewDidLoad, .unlockButtonTapped:
             interactor.executeRequest(.initiateAuthentication)
 
+        case .applicationWillEnterForeground:
+            interactor.executeRequest(.evaluateAuthentication)
+
```
johnxnguyen, line 77: I think it might be better to use `.initiateAuthentication` since that is the same flow that happens when `.viewDidLoad` is processed.

johnxnguyen (changes requested):
I would also add a test that the view sends the `applicationDidEnterForeground` event.

Wire-iOS/Sources/UserInterface/App Lock/AppLockModule.Presenter.swift
KaterinaWire: Done

KaterinaWire:
@johnxnguyen I added the `isAuthenticationNeeded` property to the Interactor as we discussed this morning.

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
```diff
@@ -107,11 +111,27 @@ final class AppLockModuleInteractorTests: XCTestCase {
...
         // Then
         XCTAssertEqual(presenter.results, [.readyForAuthentication(shouldInform: true)])
     }
+    
+    func test_InitiateAuthentication_SessionIsAlreadyUnlocked() {
+        // Given
+        session.lock = .none
+
+        // When
+        XCTAssertNil(session.lock)
```
johnxnguyen, line 126: I would prefer to avoid adding assertions in the **when**, it seems unnecessary and clutters the test a bit.

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
```diff
@@ -107,11 +111,27 @@ final class AppLockModuleInteractorTests: XCTestCase {
         appLock.needsToNotifyUser = true
 
         // When
+        XCTAssertEqual(session.lock, .screen)
```
johnxnguyen, line 114: I would also avoid assertions here too (and in previous tests). If you want to express that lock is of type screen, then I would explicitly set it in  **given**.

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
```diff
@@ -107,11 +111,27 @@ final class AppLockModuleInteractorTests: XCTestCase {
...
     }
+    
+    func test_InitiateAuthentication_SessionIsAlreadyUnlocked() {
+        // Given
+        session.lock = .none
+
+        // When
+        XCTAssertNil(session.lock)
+        sut.executeRequest(.initiateAuthentication)
+        XCTAssertTrue(waitForGroupsToBeEmpty([sut.dispatchGroup]))
```
johnxnguyen, line 128: There's no need to wait because we're not doing any async operations.

johnxnguyen (changes requested):
We should still write a test that the view sends the `applicationWillEnterForeground` event to the presenter.

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
KaterinaWire: Fixed

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
KaterinaWire: Fixed

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
KaterinaWire: Fixed

KaterinaWire:
> We should still write a test that the view sends the `applicationWillEnterForeground` event to the presenter.

@johnxnguyen I added a test

johnxnguyen (approved):

PR #4937, opened 2021-02-15
[Fix]: Hide Edit Button When One User Client In List

marcomaddy1982:
### Issues

When we have just one client in the client list the edit button is shown

### Causes

not set correctly to false `editList` when in the client list is empty 

NOTA BENE: `clients` represent all the client registered not being selfClient

```swift
self.clients = userClients.filter { !$0.isSelfClient() }
```

I'm highlighting this to you cause it confuse me a lot but then i discovered it :-) 

### Solutions

Set  `editList` to `false` when `clients.count == 0` (See NOTA BENE)

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
@@ -66,7 +66,7 @@ final class ClientListViewController: UIViewController,
             if clients.count > 0 {
                 createRightBarButtonItem()
             } else {
-                self.navigationItem.rightBarButtonItem = nil
+                self.editingList = false
             }
         }
     }
--- a/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
@@ -44,7 +44,7 @@ final class ClientListViewController: UIViewController,
 
     var editingList: Bool = false {
         didSet {
-            guard clients.count > 0 else {
+            guard !clients.isEmpty else {
                 self.navigationItem.rightBarButtonItem = nil
                 self.navigationItem.setHidesBackButton(false, animated: true)
                 return
@@ -63,7 +63,7 @@ final class ClientListViewController: UIViewController,
             self.sortedClients = self.clients.filter(clientFilter).sorted(by: clientSorter)
             self.clientsTableView?.reloadData()
 
-            if clients.count > 0 {
+            if !clients.isEmpty {
                 createRightBarButtonItem()
             } else {
                 self.editingList = false


Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
```diff
@@ -66,7 +66,7 @@ final class ClientListViewController: UIViewController,
             if clients.count > 0 {
```
KaterinaWire, line 66: You can use `!clients.isEmpty`

KaterinaWire:
@marcomaddy1982 Do we have tests to check this case?

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
marcomaddy1982: I can do but it was not part of my changes :-)

KaterinaWire: Yes, you are right. You can ignore this comment :)

marcomaddy1982:
> @marcomaddy1982 Do we have tests to check this case?

I think that this is the test `testForOneDeviceWithNoEditButton()`. So the answer is yes.

Wire-iOS/Sources/UserInterface/Settings/DeviceManagement/ClientListViewController.swift
marcomaddy1982: It is ok, i can do that 👍. No worries.

KaterinaWire:
> > @marcomaddy1982 Do we have tests to check this case?
> 
> I think that this is the test `testForOneDeviceWithNoEditButton()`. So the answer is yes.

If we have this test, why is there this issue?

marcomaddy1982:
> > > @marcomaddy1982 Do we have tests to check this case?
> > 
> > 
> > I think that this is the test `testForOneDeviceWithNoEditButton()`. So the answer is yes.
> 
> If we have this test, why is there this issue?

They are snapshot tests.

KaterinaWire:
> > > > @marcomaddy1982 Do we have tests to check this case?
> > > 
> > > 
> > > I think that this is the test `testForOneDeviceWithNoEditButton()`. So the answer is yes.
> > 
> > 
> > If we have this test, why is there this issue?
> 
> They are snapshot tests.

If the test is correct, it shouldn't pass without this PR, right?

David-Henner (approved):

marcomaddy1982:
> > > > > @marcomaddy1982 Do we have tests to check this case?
> > > > 
> > > > 
> > > > I think that this is the test `testForOneDeviceWithNoEditButton()`. So the answer is yes.
> > > 
> > > 
> > > If we have this test, why is there this issue?
> > 
> > 
> > They are snapshot tests.
> 
> If the test is correct, it shouldn't pass without this PR, right?

I think that the answer to this question is the following:  the issue was that the button was not updated after a user delete a client. Not when the view is loaded at the beginning.

KaterinaWire (approved):

PR #4951, opened 2021-02-18
[Fix] video capture stops on presentation mode change

David-Henner:
### Issues

Video capture stop when the user changes the presentation mode from **all video streams** to **active speakers**.
note that this happens only if the user is not qualified for the active speakers mode.

### Causes

Since the user doesn't qualify for the active speakers mode, the `SelfVideoPreviewView` is removed from `VideoGridViewController`'s view cache, which causes the capturer to be stopped. 

#### 1. Prevent the `SelfVideoPreviewView` to be removed from the view cache. 
 
Pruning the view cache is good for the remote video views, but not critical for the self preview. 
Keeping the instance also improves user experience, because the view doesn't have to be recreated each time the user enables the video.

#### 2. Sync the self user's video state with the capturer state, instead of using the video state provided by `VideoStream`

The `VideoConfiguration` passed to `VideoGridViewController` contains a list of `VideoStream` to be displayed. In the use case where the self user's `VideoStream` is not in the list, we want the user's video state to be provided by the configuration in order to update the capturer state.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
@@ -207,7 +207,9 @@ final class VideoGridViewController: SpinnerCapableViewController {
     private func updateSelfPreview() {
         guard let selfStreamId = ZMUser.selfUser()?.selfStreamId else { return }
 
+        // No stream to show. Update the capture state.
         guard let selfStream = stream(with: selfStreamId) else {
+            Log.calling.debug("updating capture state to \(configuration.videoState)")
             selfPreviewView?.updateCaptureState(with: configuration.videoState)
             return
         }
--- a/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
@@ -205,7 +205,7 @@ final class VideoGridViewController: SpinnerCapableViewController {
     }
     
     private func updateSelfPreview() {
-        guard let selfStreamId = ZMUser.selfUser()?.selfStreamId else { return }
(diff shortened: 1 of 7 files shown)


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
```diff
@@ -268,6 +270,7 @@ final class VideoGridViewController: SpinnerCapableViewController {
         let currentStreamsIds = configuration.allStreamIds
 
         for deletedStreamId in existingStreamsIds.subtracting(currentStreamsIds) {
+            guard deletedStreamId != ZMUser.selfUser()?.selfStreamId else { return }
```
billypchan, line 273:
```suggestion
            guard deletedStreamId != SelfUser.current?.selfStreamId else { return }
```

billypchan: better replace with SelfUser.current

billypchan (changes requested):

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoGridViewController.swift
David-Henner: done b8bfb80, thanks for the suggestion

billypchan (approved):

billypchan:
test failed: `CallViewControllerTests.testThatItDeallocates()`

David-Henner:
> test failed: `CallViewControllerTests.testThatItDeallocates()`

due to using `SelfUser.current` because `SelfUser.provider` needs to be set explicitly

David-Henner:
I'm reverting the changes to `SelfUser.current` because it causes multiple crashes in the tests. I will solve this in a separate PR.

PR #4952, opened 2021-02-18
[Fix] Video grid presentation mode stuck on active speakers

David-Henner:
### Issues

Video grid presentation mode gets stuck on active speakers view when the amount of connected participants in a call goes under 3

### Causes

By design, the presentation mode toggle isn't shown in this scenario

### Solutions

Set the presentation mode to `allVideoStreams` when the connected participants amount goes under 3.


johnxnguyen (commented):
Could we add a test for this?

johnxnguyen (approved):

KaterinaWire (approved):

PR #4978, opened 2021-02-23
[Fix] Do not request to create a custom passcode if EAR is enabled

KaterinaWire:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/SQSERVICES-300

### Issues
The App asks to create a custom passcode when it's not needed (**EAR** and Applock are enabled).

### Causes
We didn't check if EAR is enabled when a custom passcode is required.

### Solutions
Ask to create a custom passcode when it's needed and EAR isn't enabled.

### DIFF
--- a/Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
+++ b/Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
@@ -113,6 +113,19 @@ final class AppLockModuleInteractorTests: XCTestCase {
         XCTAssertEqual(presenter.results, [.readyForAuthentication(shouldInform: true)])
     }
     
+    func test_InitiateAuthentication_DoesNotNeedToCreateCustomPasscode_WhenDatabaseIsLocked() {
+        // Given
+        session.lock = .database
+        appLock.isCustomPasscodeSet = false
+        appLock.requireCustomPasscode = true
+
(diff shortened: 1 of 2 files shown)


Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
```diff
@@ -113,6 +113,19 @@ final class AppLockModuleInteractorTests: XCTestCase {
         XCTAssertEqual(presenter.results, [.readyForAuthentication(shouldInform: true)])
     }
     
+    func test_InitiateAuthentication_DoesNotNeedToCreateCustomPasscode_WhenDatabaseIsLocked() {
+        // Given
+        session.lock = .database
+        appLock.isCustomPasscodeSet = false
+        appLock.requireCustomPasscode = true
```
johnxnguyen, line 120: I think it would be more expressive replace this line with

```swift
authenticationType.current = .unavailable
```

johnxnguyen (approved):

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
KaterinaWire: Done

David-Henner (approved):

PR #4983, opened 2021-02-24
[Fix] Device authentication requested when app is minimized

johnxnguyen:
## What's new in this PR?

**JIRA:** https://wearezeta.atlassian.net/browse/SQSERVICES-294

### Issues

On iOS 14, if the encryption at rest is enabled, when the app is minimized, then the native device authentication (faceID, touchID or passcode) UI is presented, even though Wire is not visible.

### Causes

When EAR is enabled, when the app becomes inactive (moves to the background) the user session is locked. This triggers a change to the Wire app state, and we transitioned to the `locked` state and load the `AppLockModule`. When the view is loaded, we initiate the authentication flow, which results in asking for user authentication.

This only appears on iOS 14, and is probably due to a change that `viewDidLoad` and other lifecycle methods may be invoked even when the app is in the background.

### Solutions

While it makes sense to inform the presenter to initiate authentication when the `viewDidAppear` (and not when `viewDidLoad`), and I have made this change, it is not enough to solve the problem because also the `viewDidAppear` method is invoked when the app is in the background.

Instead we place a guard statement in the interactor before it asks the app lock controller to evaluate authentication. We only want to evaluate authenticate if the app is in an `active` state.

There's one caveat with this however, and that is when the app is brought to the foreground, we re-trigger the authentication flow as convenience to the user so they do not have to explicitly tap a button to authenticate. This is triggered by a `applicationWillEnterForeground` notification, but unfortunately this usually happens before the application transitions to the `active` state. So in this situation, the user must tap the unlock button explicitly. But this only affects EAR users and not normal app lock users.

### DIFF
--- a/Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
+++ b/Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
@@ -42,7 +42,7 @@ final class AppLockModuleInteractorTests: XCTestCase {
         sut = .init(session: session,
                     authenticationType: authenticationType,
                     applicationStateProvider: applicationStateProvider)
-        
+
         sut.presenter = presenter
     }
     
@@ -147,9 +147,23 @@ final class AppLockModuleInteractorTests: XCTestCase {
 
     // MARK: - Evaluate authentication
 
-    func test_EvaluateAuthentication_OnlyIfApplicationIsActive() {
+    func test_EvaluateAuthentication_ReturnsDeniedIfAppIsInactive() {
         // Given
         applicationStateProvider.applicationState = .inactive
+        authenticationType.current = .faceID
+
+        // When
+        sut.executeRequest(.evaluateAuthentication)
+        XCTAssertTrue(waitForGroupsToBeEmpty([sut.dispatchGroup]))
+
+        // Then
+        XCTAssertEqual(appLock.methodCalls.evaluateAuthentication.count, 0)
+        XCTAssertEqual(presenter.results, [.authenticationDenied(.faceID)])
+    }
+
+    func test_EvaluateAuthentication_ReturnsDeniedIfAppIsInBackground() {
+        // Given
+        applicationStateProvider.applicationState = .background
         authenticationType.current = .faceID
 
         // When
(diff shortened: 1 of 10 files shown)


typfel:
>There's one caveat with this however, and that is when the app is brought to the foreground, we re-trigger the authentication flow as convenience to the user so they do not have to explicitly tap a button to authenticate. This is triggered by a applicationWillEnterForeground notification, but unfortunately this usually happens before the application transitions to the active state. So in this situation, the user must tap the unlock button explicitly. But this only affects EAR users and not normal app lock users.

Why not trigger the authentication on `applicationDidEnterForeground` or `applicationDidBecomeActive`?

johnxnguyen:
> Why not trigger the authentication on `applicationDidEnterForeground` or `applicationDidBecomeActive`?

@typfel there is no `applicationDedEnterForeground`, and the problem with `applicationDidBecomeActive` is that it is called after the native authentication UI is dismissed (even after a successful completion) because we technically leave the app. The result is an infinite loop of authentication.

typfel:
> @typfel there is no `applicationDidEnterForeground`, and the problem with `applicationDidBecomeActive` is that it is called after the native authentication UI is dismissed (even after a successful completion) because we technically leave the app. The result is an infinite loop of authentication.

I see, seems like it should be possible to solve though by adding some more state or querying if we actually need to authenticate. But I guess it's better tackle this issue separately.

typfel (approved):

marcomaddy1982 (approved):

Wire-iOS Tests/App Lock/Tests/AppLockModuleInteractorTests.swift
```diff
@@ -141,6 +147,20 @@ final class AppLockModuleInteractorTests: XCTestCase {
 
     // MARK: - Evaluate authentication
 
+    func test_EvaluateAuthentication_OnlyIfApplicationIsActive() {
```
KaterinaWire, line 150: I think the name of the test doesn't match what we are testing in the test. We check that we don't execute an `evaluateAuthentication` when the applicationState is `.inactive`. We also have a `background` state, it might make sense to add a test for this state as well.

johnxnguyen: True, it's a typo.

KaterinaWire (approved):

johnxnguyen:
@typfel Yes I'll try to solve it separately.

PR #4989, opened 2021-02-25
[Fix] SearchBar background in share extension

KaterinaWire:
## What's new in this PR?
https://wearezeta.atlassian.net/browse/SQSERVICES-63

### Issues
Conversation search field is transparent and conversation list is visible through the field in share extension.

### Solutions
Set the background color for the searchBar.


billypchan (approved):
We can create snapshot tests for the share extension later.

David-Henner (commented):
Is white color fine to use in both light and dark UI ?

KaterinaWire:
> Is white color fine to use in both light and dark UI ?

@David-Henner No, it's not good. You're right. I fixed it by setting `isTranslucent` for searchBar.

David-Henner (approved):

PR #4999, opened 2021-03-01
[Fix]: Handle gracefully when device runs out of disk space

marcomaddy1982:
### Issues

See the [issue](https://wearezeta.atlassian.net/browse/SQCORE-429)

When there was a problem to load the database in case we run out of disk space (or for any other error), the database was dropped. This means that the user was losing all the previous conversation

NOTA BENE: The copies added in this PR are still not official. 

### Solutions

Now a screen similar to the `blacklisted` or `jailbroken` is shown but with a different message. Moreover the alert contains two buttons:

- Device Settings -> Which will show the application settings
- Delete Database -> When the user will choose it another confirmation alert will be show. If the user confirms the database will be removed and the login screen will be shown.

### Dependencies

data-model: https://github.com/wireapp/wire-ios-data-model/pull/1144
sync-engince:  https://github.com/wireapp/wire-ios-sync-engine/pull/1411


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -1626,6 +1626,10 @@
 "jailbrokendevice.alert.title" = "Jailbreak detected";
 "jailbrokendevice.alert.message" = "For security reasons, Wire can't be used on this device. Any existing Wire data has been erased.";
 
+// Database Loading Failure alert
+"databaseloadingfailure.alert.title" = "Database loading failure";
+"databaseloadingfailure.alert.message" = "The database could be loaded.";
```
typfel, line 1631:
```suggestion
"databaseloadingfailure.alert.message" = "The database could not be loaded.";
```

typfel: This copy is pending a review?

typfel (changes requested):
Look like some linters warnings was introduced in AppStateCalculator

Wire-iOS/Resources/Base.lproj/Localizable.strings
marcomaddy1982: Yes, this is just temp copy. Better if I mention on the ticket description.

KaterinaWire:
@marcomaddy1982 Everything is fine for me, I don't have comments. Could you notify when the copies will be approved by the design team?

KaterinaWire (approved):

marcomaddy1982:
> Look like some linters warnings was introduced in AppStateCalculator

Solved

Wire-iOS/Sources/UserInterface/BlockerViewController.swift
```diff
@@ -47,17 +51,77 @@ final class BlockerViewController: LaunchImageViewController {
...
 
     func showJailbrokenMessage() {
-        presentAlertWithOKButton(title: "jailbrokendevice.alert.title".localized, message: "jailbrokendevice.alert.message".localized)
+        presentAlertWithOKButton(title: "jailbrokendevice.alert.title".localized,
+                                 message: "jailbrokendevice.alert.message".localized)
     }
 
+    func showDatabaseFailureMessage() {
+        let databaseFailureAlert = UIAlertController(
+            title: "databaseloadingfailure.alert.title".localized,
```
typfel, line 73: What about using the type safe localization John introduced in https://github.com/wireapp/wire-ios/pull/4911

marcomaddy1982: Done

typfel (approved):

typfel (approved):

KaterinaWire (approved):

PR #5001, opened 2021-03-01
[Feature] Toast notification

David-Henner:
## What's new in this PR?

- Toast notification component according to new design system

### Design

Screenshot 2021-03-01 at 17 22 39

### Note

You can compare the design specs with the snapshots


marcomaddy1982 (approved):

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
```diff
@@ -0,0 +1,190 @@
...
+    }
+}
+
+// MARK: - Helpers
+
+private extension ColorSchemeVariant {
+    var moreInfoButtonColor: UIColor {
+        switch self {
+        case .dark:
+            return UIColor(rgba: (23, 24, 26, 0.12))
```
KaterinaWire, line 176: Don't we need this color in our constants?

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
```diff
@@ -0,0 +1,190 @@
...
+        case .light:
+            return .whiteAlpha24
+        }
+    }
+
+    var textColor: UIColor {
+        switch self {
```
KaterinaWire, line 183: I think we can use something like this `UIColor.from(scheme: .iconSelected, variant: variant).` What do you think?

KaterinaWire, line 183: I think we can use something like this `UIColor.from(scheme: .iconSelected, variant: variant)`. What do you think?

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
```diff
@@ -0,0 +1,190 @@
...
+    var moreInfoButtonColor: UIColor {
+        switch self {
+        case .dark:
+            return UIColor(rgba: (23, 24, 26, 0.12))
+        case .light:
+            return .whiteAlpha24
+        }
+    }
+
+    var textColor: UIColor {
```
KaterinaWire, line 182: I think we can use something like this `UIColor.from(scheme: .iconSelected, variant: variant)`. What do you think?

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
```diff
@@ -0,0 +1,190 @@
...
+            moreInfoButton.heightAnchor.constraint(equalToConstant: moreInfoButtonHeight),
+            moreInfoButton.centerYAnchor.constraint(equalTo: bottomView.centerYAnchor),
+            moreInfoButton.trailingAnchor.constraint(equalTo: bottomView.trailingAnchor),
+            bottomView.heightAnchor.constraint(equalTo: moreInfoButton.heightAnchor)
+        ])
+    }
+
+    // MARK: - Events
+
+    @objc func closeButtonTapHandler() {
```
KaterinaWire, line 161: Can we define event methods as `private`?

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
David-Henner: I don't see any other use cases for this one yet, but I've added it to the constants as `backgroundGraphiteAlpha12`.
It has slightly different values: `(22, 24, 25, 0.12)`. I made it to be consistent with the other similar colour constants

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
David-Henner: Good point. I used a different scheme because I noticed I wasn't using the right color for dark variant

Wire-iOS/Sources/UserInterface/Components/Toast/ToastView.swift
David-Henner: done

KaterinaWire (approved):

PR #5011, opened 2021-03-04
[Refactor] Remove a create passcode step in authentication flow

KaterinaWire:
### Issues
We have a create passcode step in authentication flow that is no longer used.

### Causes
This is handled by the app lock module now.

### Solutions
Remove a create passcode step in authentication flow.

### Notes

https://wearezeta.atlassian.net/browse/SQSERVICES-233


Wire-iOS/Sources/UserInterface/CustomAppLock/Setup/PasscodeSetupViewController.swift
```diff
@@ -239,7 +239,6 @@ final class PasscodeSetupViewController: UIViewController {
         guard let passcode = passcodeTextField.text else { return }
         presenter.storePasscode(passcode: passcode, callback: callback)
 
-        authenticationCoordinator?.passcodeSetupControllerDidFinish()
```
johnxnguyen, line 242: I think we can also remove the reference to the authentication coordinator.

KaterinaWire: Done

johnxnguyen (approved):

billypchan (approved):

PR #5026, opened 2021-03-18
[Fix] Tapping cell show wrong conversation

KaterinaWire:
### Issues
When a user taps on a conversation, the wrong conversation opens.

### Causes
Seems like the the "standard" reloadData() for the collectionView (conversation list) conflicts with incremental reloading of the conversation list.

### Solutions
Always use an incremental reload of the collectionView when we need to update the list of conversations.

### Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-308

### DIFF
--- a/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
+++ b/Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
@@ -705,11 +705,13 @@ extension ConversationListViewModel: ConversationDirectoryObserver {
             /// animation is missing since it started after the top section update animation started. To fix this
             /// we should calculate the change set in one batch.
             /// TODO: wait for SE update for returning multiple items in changeInfo.updatedLists
-            for updatedList in changeInfo.updatedLists {
-                if let kind = self.kind(of: updatedList) {
-                    update(for: kind)
+            DispatchQueue.main.async(execute: {
+                for updatedList in changeInfo.updatedLists {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/ConversationList/ListContent/ConversationListContentController/ViewModel/ConversationListViewModel.swift
```diff
@@ -699,7 +702,7 @@ extension ConversationListViewModel: ConversationDirectoryObserver {
         if changeInfo.reloaded {
             // If the section was empty in certain cases collection view breaks down on the big amount of conversations,
             // so we prefer to do the simple reload instead.
-            reload()
```
typfel, line 702: Let's delete this code path if it's now unused.

KaterinaWire: Do we need to check
```
if changeInfo.updatedLists.isEmpty {
            update()
 }
```
 ?

typfel: I meant let's delete the `func reload()` method if it's now unused.

I don't think we need to add something like:
```
if changeInfo.updatedLists.isEmpty {
            update()
 }
 ```

KaterinaWire: `reload()` is deleted

typfel (approved):

johnxnguyen (approved):

PR #5028, opened 2021-03-22
[Fix] Overlapping app state transitions

johnxnguyen:
### Issues

We have been experiencing several crashes lately when the trying to access the `SelfUser.current`, typically from the conversation list or in the conversation content.

### Causes

`SelfUser.current` intentionally crashes when `SelfUser.provider == nil`. The self user provider is configured during the app state transition:

- The provider is set when `applicationWillTransition(to: .authenticated)` is called, and
- The provider is reset when `applicationDidTransition(to: appState)`, where `appState != .authenticated`.

The problem however is that we can repeatedly call these methods in any order, for example:

```swift
// The self user isn't set yet, so SelfUser.current will crash.
applicationWillTransition(to: .migrating)

// This will set the provider, SelfUser.current won't crash.
applicationWillTransition(to: .authenticated)

// This will reset the provider, SelfUser.current will crash
applicationDidTransition(to: .migrating)

// Here we typically cause a load of the conversation list, which will
// result in an eventual call to SelfUser.current. But it's not set,
// so it will crash.
applicationDidTransition(to: .authenticated)
```

This exact scenario is currently happening due to recent changes that put the app through the migrating state each time the app launches.

### Solutions

Consider transitions (delimited by `applicationWillTransition` and `applicationDidTransition`) as atomic, meaning that we only begin a transition once the previous one (if any) has completed.

### Note

This is just a draft implementation, I'm looking for feedback whether there is a cleaner/simpler way to achieve this.

### DIFF
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -172,6 +172,8 @@ public class AppRootRouter: NSObject {
     ///     - completion: A block executed after the transition has completed.
 
     private func enqueueTransition(to appState: AppState, completion: @escaping () -> Void = {}) {
+        // Perform the wait on a background queue to we don't cause
+        // a deadlock on the main queue.
         appStateTransitionQueue.async { [weak self] in
             guard let `self` = self else { return }
 
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -172,8 +172,8 @@ public class AppRootRouter: NSObject {
     ///     - completion: A block executed after the transition has completed.
 
     private func enqueueTransition(to appState: AppState, completion: @escaping () -> Void = {}) {
-        // Perform the wait on a background queue to we don't cause
-        // a deadlock on the main queue.
+        // Perform the wait on a background queue so we don't cause a
+        // deadlock on the main queue.
         appStateTransitionQueue.async { [weak self] in
             guard let `self` = self else { return }
 
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -312,6 +312,7 @@ extension AppRootRouter {
                 error?.userSessionErrorCode == .accountDeleted,
             let sessionManager = SessionManager.shared
         else {
+            completion()
             return
         }
 
@@ -324,6 +325,7 @@ extension AppRootRouter {
                                                               statusProvider: AuthenticationStatusProvider())
 
         guard let authenticationCoordinator = authenticationCoordinator else {
+            completion()
             return
         }
 
@@ -333,8 +335,6 @@ extension AppRootRouter {
 
         rootViewController.set(childViewController: navigationController,
                                completion: completion)
-
-        presentAlertForDeletedAccountIfNeeded(error)
     }
 
     private func showAuthenticated(isComingFromRegistration: Bool, completion: @escaping () -> Void) {
@@ -343,6 +343,7 @@ extension AppRootRouter {
             let authenticatedRouter = buildAuthenticatedRouter(account: selectedAccount,
                                                                isComingFromRegistration: isComingFromRegistration)
         else {
+            completion()
             return
         }
 
@@ -418,7 +419,9 @@ extension AppRootRouter {
     }
 
     private func applicationDidTransition(to appState: AppState) {
-        if case .authenticated = appState {
+        if case .unauthenticated(let error) = appState {
+           presentAlertForDeletedAccountIfNeeded(error)
+        } else if case .authenticated = appState {
             authenticatedRouter?.updateActiveCallPresentationState()
             urlActionRouter.openDeepLink(needsAuthentication: true)
 
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -234,7 +234,7 @@ extension AppRootRouter: AppStateCalculatorDelegate {
         case .locked:
             // TODO: [John] Avoid singleton.
             screenCurtain.delegate = ZMUserSession.shared()
-            showAppLock()
+            showAppLock(completion: completionBlock)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -20,11 +20,6 @@ import UIKit
 import WireSyncEngine
 import avs
 
-extension AppRootRouter {
-    static let appStateDidTransition = Notification.Name(rawValue: "appStateDidTransition")
-    static let appStateKey = "AppState"
-}
-
```
johnxnguyen, line 27: This wasn't being used anywhere.

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -157,25 +152,50 @@ public class AppRootRouter: NSObject {
...
-    }
-
-    private func notifyTransition(for appState: AppState) {
-        NotificationCenter.default.post(name: AppRootRouter.appStateDidTransition,
-                                        object: nil,
-                                        userInfo: [AppRootRouter.appStateKey: appState])
+        enqueueTransition(to: appState, completion: completion)
     }
 
     private func transition(to appState: AppState, completion: @escaping () -> Void) {
```
johnxnguyen, line 196: I want to make this inaccessible so that we avoid to call it directly rather than through `enqueueTransition(to:completion)`

marcomaddy1982:
The approach is correct in my opinion> Now i don't remember in detail but I guess the idea you have implemented is something I have tried to implement in the in the `RootViewController`. But what i have done is more for the animation transition from a state to another. See the method:

```swift
private func transition(from fromViewController: UIViewController,
                            to toViewController: UIViewController,
                            animated: Bool = false,
                            completion: (() -> Void)?)
```

At this point it seems that it was not enough.

marcomaddy1982 (approved):

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -157,25 +152,52 @@ public class AppRootRouter: NSObject {
...
+        applicationWillTransition(to: appState)
```
typfel, line 199: So it's very important that we always call  `applicationWillTransition(to: appState)` followed by `applicationDidTransition(to: appState)` now, since otherwise the dispatch group will never be unlocked. Looking at the switch it seems like we don't call `applicationDidTransition(to: appState)` in the `.locked` case.

It's quite easy to introduce such an error so how can we structure the code to make such errors impossible?

Wire-iOS/Sources/AppRootRouter.swift
typfel: The options I can I think of:
1. Move the state transition methods to a separate file
1. Move the state transition methods to separate extension inside this file 
1. Merge `transition(to:completion:)` into `enqueueTransition(to:completion)`

Maybe option 2 is nicer since you could then at least still keep the `appStateTransitionGroup` and `appStateTransitionQueue` `fileprivate`.

Wire-iOS/Sources/AppRootRouter.swift
typfel: Was used by AppLock previously I think.

typfel (changes requested):

Wire-iOS/Sources/AppRootRouter.swift
johnxnguyen: Good catch, I didn't noticed that we weren't always calling the completion. The completion needs to be passed when we set the view controller, and I see that we duplicate this code for every switch case in separate methods.
One thing we could do is to extract the code that creates view controllers out to another method, so then we would only have a single place where we call 

 ```swift
rootViewController.set(childViewController: childViewController, completion: completion)
```

This would likely reduce the size of `transition(to:completion:)` so much so that it would be fine to merge it into the `enqueueTransition(to:completion:)` method body.

File: Wire-iOS/Sources/AppRootRouter.swift (hunk 2 above)
Comment (marcomaddy1982) [reply Wire-iOS/Sources/AppRootRouter.swift:198]: I like more then the other the solution number 2  too.

File: Wire-iOS/Sources/AppRootRouter.swift (hunk 3 above)
Comment (marcomaddy1982) [reply Wire-iOS/Sources/AppRootRouter.swift:199]: @johnxnguyen @typfel: IMHO here I would just add the `completion` also to the `showAppLock`  method. The reason is because really verbose to extract the code that creates view controllers out to another method. Moreover there is a special case in the `showUnauthenticatedFlow` where after `set(childViewController: childViewController, completion: completion)  ` we do another operation:

```swift
rootViewController.set(childViewController: navigationController,
                               completion: completion)

```
and I guess this particular scenario complicate a bit the solution you wanna to adopt in the comment.

typfel: @marcomaddy1982 I think John idea solves the problem in simple way. Only adding a the completion block to to the `showAppLock` method still makes easy to re-introduce this error the next time someone adds a new  app state.

Couldn't we remove the special case for `showUnauthenticatedFlow` by moving the `presentAlertForDeletedAccountIfNeeded(error)` to `applicationDidTransition(to appState: AppState)`.

johnxnguyen: I found that indeed it's not so easy to extract out view controller creation, specifically for the unauthenticated and authenticated states. There's a bit of decoupling needed and consideration about how to handle cases where we return early (i.e abort a transition so to speak).

One thing though that it's not enough to simply add the completion for the `locked` state. You'd also need to add it to any `guard` that returns early, as I mentioned above.

typfel: Ok, it's more complicated. Then maybe I agree to first do the fix this with the current system and do the re-factoring separately since the code freeze is today.

johnxnguyen: I agree, but I would also say that refactoring should take a high priority for the subsequent release.

marcomaddy1982: the suggestion made by @johnxnguyen is now pushed.

marcomaddy1982:
@johnxnguyen  I added you as assigner anyway if you approve the changes I have made add a comment cause since you created you can not approve it.

typfel (approved):

johnxnguyen (commented):
@marcomaddy1982  I approve 👍🏻

PR #5029, opened 2021-03-24
[Fix] Stuck on migration screen after restoring backup

typfel:
### Issues

Application gets stuck on the migration screen after restoring a backup.

### Causes

When the application transitions away from the `.unauthenticated` state we release the `AuthenticationCoordinator`, which is responsible for signalling when the login flow is completed. After https://github.com/wireapp/wire-ios-data-model/pull/1144 we started to attempt a migration every time we load the database. We still need to register a new client after restoring a backup so we don't immediately  transition to the `.authenticated` state and later when the client is registered we don't the state changed because `AuthenticationCoordinator` is gone.

The flow now looks like this:
- Restoring backup
- Migrating
- Registering client & perform slow sync
- Authenticated

### Solutions

Don't release the `AuthenticationCoordinator` when transitioning to the `.migrating` state.

### DIFF
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -240,7 +240,7 @@ extension AppRootRouter: AppStateCalculatorDelegate {
 
     private func resetAuthenticationCoordinatorIfNeeded(for state: AppState) {
         switch state {
-        case .unauthenticated:
+        case .unauthenticated, .migrating:
             break // do not reset the authentication coordinator for unauthenticated state
         default:
             authenticationCoordinator = nil // reset the authentication coordinator when we no longer need it
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -241,7 +241,6 @@ extension AppRootRouter: AppStateCalculatorDelegate {
     private func resetAuthenticationCoordinatorIfNeeded(for state: AppState) {
         switch state {
         case .authenticated:
-            // reset the authentication coordinator when we no longer need it
             authenticationCoordinator = nil
             break
         default:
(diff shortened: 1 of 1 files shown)


marcomaddy1982 (approved):

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -240,7 +240,7 @@ extension AppRootRouter: AppStateCalculatorDelegate {
 
     private func resetAuthenticationCoordinatorIfNeeded(for state: AppState) {
         switch state {
-        case .unauthenticated:
+        case .unauthenticated, .migrating:
             break // do not reset the authentication coordinator for unauthenticated state
         default:
             authenticationCoordinator = nil // reset the authentication coordinator when we no longer need it
```
johnxnguyen, line 246: Perhaps it's worth removing these comments since they're a bit redundant and now also a little outdated.

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -240,7 +240,7 @@ extension AppRootRouter: AppStateCalculatorDelegate {
 
     private func resetAuthenticationCoordinatorIfNeeded(for state: AppState) {
         switch state {
-        case .unauthenticated:
+        case .unauthenticated, .migrating:
```
johnxnguyen, line 243: Do you think it would be more expressive to define in which states we should don't need the authentication coordinator? For example, perhaps it's sufficient to simply to reset the coordination when we end up in the authenticated state.

typfel: Yes, good point!That way we avoid a similar bug the next time we add a new app state.

typfel: Updated

Wire-iOS/Sources/AppRootRouter.swift
typfel: Gone 💥

johnxnguyen (approved):

typfel:
Test failure are the flaky tests I disabled in the other PR.

PR #5035, opened 2021-03-24
[Fix]: SQCORE-275: Custom Backend Deep Link Fix

marcomaddy1982:
### Issues

The app were remaining stuck at the launch screen when the custom backend deep link were triggered and the app where not launched.

### Causes

 There is a race condition between the launch of the application and processing deep links.

### Solutions

Deep links are now processed once the app has arrived at an operational state, i.e either the unauthenticated or authenticated state (It doesn't make sense to process a deep link while we're in a migrating, or jailbroken or locked state)

### Dependencies

[JIRA](https://wearezeta.atlassian.net/browse/SQCORE-549)


Wire-iOS Tests/Routers/URLActionRouterTests.swift
```diff
@@ -36,26 +37,15 @@ final class URLActionRouterTests: XCTestCase {
...
-        // GIVEN
-        let validDeepLink = URL(string: "wire://user/user_id")!
-        let router =  TestableURLActionRouter(viewController: RootViewController(), url: validDeepLink)
-
-        // WHEN
-        router.openDeepLink(needsAuthentication: true)
-
-        // THEN
-        XCTAssertFalse(router.wasDeepLinkOpened)
-    }
```
johnxnguyen, line 58: Why don't we need this anymore?

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -30,7 +30,11 @@ public class AppRootRouter: NSObject {
     private let navigator: NavigatorProtocol
     private var appStateCalculator: AppStateCalculator
     private var urlActionRouter: URLActionRouter
-    private var deepLinkURL: URL?
+    private var deepLinkURL: URL? {
+        didSet {
+            urlActionRouter.url = deepLinkURL
+        }
+    }
```
johnxnguyen, line 37: What do you think about using only a custom setter? This way we wouldn't be storing the link twice. For instance:

```swift
private var deepLinkURL: URL? {
    set {
        urlActionRouter.url = deepLinkURL
    }
}
```

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -46,6 +50,10 @@ public class AppRootRouter: NSObject {
     private var observerTokens: [NSObjectProtocol] = []
     private var authenticatedBlocks: [() -> Void] = []
     private let teamMetadataRefresher = TeamMetadataRefresher()
+    private var canOpenDeepLink: Bool {
+        return appStateCalculator.isAuthenticated
+            || appStateCalculator.isUnauthenticated
+    }
```
johnxnguyen, line 56: Name suggestion: `isReadyToOpenDeepLink`, to be a little bit more precise.

Wire-iOS/Sources/AppStateCalculator.swift
```diff
@@ -81,6 +81,20 @@ class AppStateCalculator {
...
+        return true
+    }
+
+    var isAuthenticated: Bool {
+        guard case .authenticated = appState else {
+            return false
+        }
+        return true
+    }
+
```
johnxnguyen, line 97: Just an idea: we could add an extension to `AppState` to determine whether a state `canProcessDeepLinks`. It may reduce the duplication. What do you think?

```swift
extension AppState {

  var canProcessDeepLinks: Bool {
    switch self {
    case .unauthenticated, .authenticated:
      return true
    default:
      return false
    }
  }

}
```

Wire-iOS/Sources/URLActionRouter.swift
```diff
@@ -74,16 +72,10 @@ class URLActionRouter: URLActionRouterProtocol {
         }
     }
 
-    func openDeepLink(needsAuthentication: Bool = false) {
-        do {
-            guard let deeplink = url else { return }
-            guard let action = try URLAction(url: deeplink) else { return }
-            guard action.requiresAuthentication == needsAuthentication else { return }
```
johnxnguyen, line 81: Where do we check now whether a deep link can b opened in the current state? Or is it not needed anymore?

johnxnguyen (changes requested):
It would be great to add some tests, at least to assert that 

- if we're in unauthenticated or authenticated state, then we can open the links immediately
- otherwise, we store the link and open it as soon as we transition to one of those valid states.

Wire-iOS Tests/Routers/URLActionRouterTests.swift
marcomaddy1982: the method  old `openDeepLink(needsAuthentication: )` is now `openDeepLink()`, which it doesn't care if we need or not authentication cause now we wait that the natural transition of the app is completed before perform the deep link. Anymore this kind of check (if a deep link could be performed or not depending if the user is logged in or not) are also made in the `SE`, which then inform the UI that the deep link could not be executed. So now that kind of test it is not necessary anymore in `UI`.

johnxnguyen: So if we receive a deep link requiring authentication, and we're logged out, then we ignore the link?

Wire-iOS/Sources/URLActionRouter.swift
marcomaddy1982: See the first comment. I think we don't need anymore to have this kind of check in the `UI` anymore. Before it was done cause the loading of the app state and the triggering of the deep link were not in sync. Now that they are in sync this kind of check are made in the SE (they were always there).

johnxnguyen:
Also, I noticed there are some commits from me in this PR. Is that intentional?

marcomaddy1982:
> Also, I noticed there are some commits from me in this PR. Is that intentional?

Ohhhh yep, I think I have just reused the same brach (just changing the name) for the crash fix you have implemented and I have completed. Cause I need to use the new way we perform the transition using the `enqueueTransition` and at that time that part was still not merged.

marcomaddy1982:
> * otherwise, we store the link and open it as soon as we transition to one of those valid states.

So I think that this tests are not necessary anymore here in the UI.

johnxnguyen:
> So I think that this tests are not necessary anymore here in the UI.

@marcomaddy1982 how so? The test I'm imagining is something like:

```
// Given
appState = .migrating

// When
appRootRouter.openDeepLink(someLink)

// Then
urlActionRouter.didOpenDeepLink is FALSE

// When
appRootRouter.didTransitionTo(.authenticated) (or unauthenticated)

// Then
urlActionRouter.didOpenDeepLink is TRUE
```

Wire-iOS/Sources/URLActionRouter.swift
```diff
@@ -73,9 +73,14 @@ class URLActionRouter: URLActionRouterProtocol {
...
-        guard let deeplink = url else { return }
-        open(url: deeplink)
-        resetDeepLinkURL()
+        do {
+            guard let deeplink = url else { return }
+            guard let _ = try URLAction(url: deeplink) else { return }
+            open(url: deeplink)
+            resetDeepLinkURL()
+        } catch {
+            print("Cuold not open deepLink for url: \(String(describing: url?.absoluteString))")
```
johnxnguyen: I think it would be better to log the error, otherwise we won't know about the print statement.

johnxnguyen (changes requested):
There's a few lint warnings that need fixing up.

Wire-iOS/Sources/URLActionRouter.swift
```diff
@@ -74,15 +76,15 @@ class URLActionRouter: URLActionRouterProtocol {
...
             guard let deeplink = url else { return }
-            guard let action = try URLAction(url: deeplink) else { return }
-            guard action.requiresAuthentication == needsAuthentication else { return }
+            guard appState.canProcessDeepLinks else { return }
+            guard try (URLAction(url: deeplink) != nil) else { return }
             open(url: deeplink)
             resetDeepLinkURL()
         } catch {
-            print("Cuold not open deepLink for url: \(String(describing: url?.absoluteString))")
+            zmLog.error("Cuold not open deepLink for url: \(String(describing: url?.absoluteString))")
```
David-Henner, line 87:
```suggestion
            zmLog.error("Could not open deepLink for url: \(String(describing: url?.absoluteString))")
```

Wire-iOS Tests/Routers/URLActionRouterTests.swift
marcomaddy1982: Yep the deep link is ignored but an alert is shown to the user.

David-Henner (approved):

johnxnguyen (approved):

PR #5039, opened 2021-03-29
[Fix] Crash after picking credentials from keychain when adding second account

typfel:
### Issues

1. User A logs in to Wire App User A taps on self profile 
2. User A taps on Add Account 
3. User A sees suggested accounts on top of keyboard
4.  User A taps on suggested account 
5. User A enters passcode/biometrics

App crashes

### Causes

When the app returns the password picker it triggers `AppRootRouter.applicationDidBecomeActive()` and calls `teamMetadataRefresher.triggerRefreshIfNeeded()`, which end up accessing the `SelfUser.provider` which is not `nil` as expected in this scenario. This causes the app to crash when force unwrapping the `managedObjectContext` when it tries to retrieve the self user.

The `SelfUser.provider` is not nil because when the user taps "Add account" we log the user out with an error, which causes the app to take the first if branch:

https://github.com/wireapp/wire-ios/blob/cce055353f63a5d50834e11df3b7cb2cbd4ef38e/Wire-iOS/Sources/AppRootRouter.swift#L406

And the SelfUser.provider from the previous session therefore remains.

### Solutions

Break out the the first if branch into a separate if statement since we always want to un-set the `SelfUser.provider`

### DIFF
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -391,7 +391,7 @@ extension AppRootRouter {
 extension AppRootRouter {
     private func applicationWillTransition(to appState: AppState) {
         appStateTransitionGroup.enter()
-        configureSelfUserProvider(for: appState)
+        configureSelfUserProviderIfNeeded(for: appState)
         configureColorScheme()
     }
 
@@ -406,12 +406,12 @@ extension AppRootRouter {
             ZClientViewController.shared?.legalHoldDisclosureController?.discloseCurrentState(cause: .appOpen)
         }
 
-        resetSelfUserProvider(for: appState)
+        resetSelfUserProviderIfNeeded(for: appState)
         urlActionRouter.openDeepLink(for: appState)
         appStateTransitionGroup.leave()
     }
 
-    private func resetSelfUserProvider(for appState: AppState) {
+    private func resetSelfUserProviderIfNeeded(for appState: AppState) {
         guard AppDelegate.shared.shouldConfigureSelfUserProvider else { return }
 
         switch appState {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -405,7 +405,9 @@ extension AppRootRouter {
     private func applicationDidTransition(to appState: AppState) {
         if case .unauthenticated(let error) = appState {
             presentAlertForDeletedAccountIfNeeded(error)
-        } else if case .authenticated = appState {
+        }
+
+        if case .authenticated = appState {
```
johnxnguyen, line 410: Seeing as the branch to unset the self user provider is critical, what do you think about moving it out into a separate method to make its importance obvious and less likely to be skipped? Something like:

```swift
private func resetSelUserProvider(for appState: AppState) {
    switch appState {
        ...
    }
}
```

It might also be worthwhile doing the same for when we set the provider.

typfel: Updated

johnxnguyen (approved):

Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -391,32 +391,50 @@ extension AppRootRouter {
...
-        } else if AppDelegate.shared.shouldConfigureSelfUserProvider {
-            SelfUser.provider = nil
         }
 
+        resetSelfUserProvider(for: appState)
         urlActionRouter.openDeepLink(for: appState)
         appStateTransitionGroup.leave()
     }
 
+    private func resetSelfUserProvider(for appState: AppState) {
```
David-Henner, line 414: Suggestion to name this `resetSelfUserProviderIfNeeded`

typfel: Updated

David-Henner (approved):

PR #5044, opened 2021-04-01
[Fix] pinch-to-zoom should not zoom out nor pan un-zoomed

David-Henner:
### Issues

Pinch-to-zoom should not allow:
- Zooming out. Only scaling the view to identity and upwards should be allowed.
- Panning when the view is not zoomed.

### Causes

--

### Solutions

Updated gesture handlers logic to conform to the rules.


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/Video Previews/VideoPreviewView.swift
```diff
@@ -134,12 +134,20 @@ final class VideoPreviewView: BaseVideoPreviewView, UIGestureRecognizerDelegate
 
         // reset scale
         gestureRecognizer.scale = 1
+
```
billypchan, line 137: to standardize the folder name, we should not have a space in the path ` Wire-iOS/Sources/UserInterface/Calling/VideoGridView/Video Previews/VideoPreviewView.swift `

billypchan (changes requested):
remove space in `Video Previews` folder

` Wire-iOS/Sources/UserInterface/Calling/VideoGridView/Video Previews/VideoPreviewView.swift `

marcomaddy1982 (approved):

billypchan (approved):

PR #5045, opened 2021-04-06
[Chore] disable pinch to zoom

David-Henner:
## What's new in this PR?

- Disabling pinch-to-zoom feature


marcomaddy1982:
@David-Henner maybe we should fix also the tests here.

PR #5055, opened 2021-04-15
[Fix] Crash when disabling Allow guests and Services

KaterinaWire:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/SQSERVICES-367

### Issues
There's a crash when user disables `Allow guests and Services` on **iPad**.

### Causes
On iPad, `UIAlertController` must be presented in a popover and the popover's source view (the view which the popover points to) must be set.

### Solutions
Set the source view for the `UIAlertController`. This fix required some additional changes that shouldn't affect the current behavior.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
+++ b/Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
@@ -238,7 +238,7 @@ final class ProfileDetailsContentController: NSObject,
 
             cell.configure(with: CellConfiguration.groupAdminToogle(get: {
                 return groupAdminEnabled
-            }, set: {_,_  in
+            }, set: {_,_ in
                 self.isAdminState.toggle()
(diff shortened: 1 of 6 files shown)


billypchan (approved):

johnxnguyen (approved):
Don't forget to address the lint comments.

Wire-iOS/Sources/UserInterface/UserProfile/Details/ProfileDetailsContentController.swift
```diff
@@ -238,7 +238,7 @@ final class ProfileDetailsContentController: NSObject,
 
             cell.configure(with: CellConfiguration.groupAdminToogle(get: {
                 return groupAdminEnabled
-            }, set: {_ in
+            }, set: {_,_ in
```
johnxnguyen, line 241: Still one more lint issue here.

KaterinaWire: Yes, I noticed. No issues now

PR #5058, opened 2021-04-19
[Feature] disable join & leave haptics for group calls

David-Henner:
## What's new in this PR?

This PR disables the join and leave haptics for group calls

### Changes
- Updated `CallHapticsController` logic
- Added an `Array+CallParticipant` helper
- Updated and renamed some tests for `CallHapticsControllerTests`

### DIFF
--- a/Wire-iOS Tests/Calling/CallHapticsControllerTests.swift
+++ b/Wire-iOS Tests/Calling/CallHapticsControllerTests.swift
@@ -37,8 +37,10 @@ final class CallHapticsControllerTests: ZMSnapshotTestCase {
     private var generator: MockCallHapticsGenerator!
     private var firstUser: ZMUser!
     private var secondUser: ZMUser!
+    private var thirdUser: ZMUser!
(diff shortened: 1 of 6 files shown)


Wire-iOS Tests/Calling/CallHapticsControllerTests.swift
```diff
@@ -110,7 +115,26 @@ final class CallHapticsControllerTests: ZMSnapshotTestCase {
         XCTAssertEqual(generator.triggeredEvents, [.join])
     }
 
-    func testThatItTriggersCorrectEventWhenAParticipantLeaves() {
+    func testThat_ItDoesNotTriggerAnEvent_WhenAUserJoins_GroupCall() {
```
KaterinaWire, line 118: Do we have a test when it triggers an event? For example, there was one participant, and the second joined the call.

KaterinaWire (approved):

typfel (approved):

PR #5065, opened 2021-04-22
[Feature] Update indication of External and Guest users

KaterinaWire:
## What's new in this PR?

https://wearezeta.atlassian.net/browse/SQSERVICES-387

This is a new feature that required updating the indication for `external` and `guest` users. We need to add an icon in the conversation next to the name.

### Attachments
Draft design:
https://wearezeta.atlassian.net/wiki/spaces/SER/pages/349536366/Design+specs+draft


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -21,34 +21,69 @@ import UIKit
 import WireSyncEngine
 import WireCommonComponents
 
-private enum TextKind {
-    case userName(accent: UIColor)
-    case botName
-    case botSuffix
+private enum SenderType {
```
johnxnguyen, line 24: Is there a particular reason for `SenderType`? It seems a little unnecessary to me and could be replaced with some private extensions to the `UserType` or simply helper methods in `SenderCellComponent`.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -21,34 +21,69 @@ import UIKit
...
+            return .bot
+        case .guest:
+            return .guest
+        }
+    }
+
+    var accessibilityString: String {
+        switch self {
+        case .member:
+            return "Team member"
```
David-Henner, line 57: should this be localised?

Wire-iOS Tests/ConversationMessageCell/ConversationMessageSenderSnapshotTests.swift
```diff
@@ -0,0 +1,162 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import SnapshotTesting
+import XCTest
+@testable import Wire
+
+final class ConversationMessageSenderSnapshotTests: XCTestCase {
```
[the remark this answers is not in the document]
KaterinaWire: The background color of the `ConversationMessageSender` is transparent, so the icon isn't visible in dark mode. For tests, I set the background color and recreated snapshots.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: `accessibilityString` was used in this `SenderCellComponent` for `accessibilityLabel`.  Since the `accessibilityLabel` is meant to be read by VoiceOve and we don't fully support this feature, I think translating this string is not urgent. That's why I decided to keep it as a string. In the future, when we take care about VoiceOve, all strings should be translated.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: The reason was to group a list of possible items and display each item in the UI using the associated values. And in the future, if necessary, it can be expanded.
I think helper methods could be like this:

```
extension SenderCellComponent {
    func icon() -> StyleKitIcon {
        if user.isServiceUser {
            return .bot
        } else if user.isExternalPartner {
            return .externalPartner
        } else if let conversation = conversation,
                  user.isGuest(in: conversation) {
            return .guest
        } else {
            return .none
        }
    }

    func colorText() -> UIColor {
              if user.isServiceUser {
            return .from(scheme: .textForeground)
        } else {
            return ColorScheme.default.nameAccent(for: user.accentColorValue, variant: ColorScheme.default.variant)
        }
    }
}
```
But if we need  additional associated value in the future, we should create a new helper method and check again `isServiceUser`, `isExternalPartner`, `isGuest`

johnxnguyen: Ok makes sense. The only thing that I feel is a bit clunky is that we're storing the same accent color for three out of four cases. This suggests to me that perhaps a struct is more suited for this purpose,  something like: 

```swift
struct SenderCellConfiguration {
  
  init(user: UserType, in conversation: ConversationLike) {
    if user.isServiceUser {
      textColor = user.accentColor
      icon = .bot
    } else if user.isPartner {
      textColor = user.accentColor
      icon = .partner
    } else {
      ...
    }
  }

}
```

What do you think?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -21,34 +21,69 @@ import UIKit
...
+        } else {
+            self = .member(accent: accentColor)
         }
     }
 }
 
 final class SenderCellComponent: UIView {
 
+    private var type: SenderType?
+
```
johnxnguyen, line 86: Is it necessary to store this? Or perhaps it is sufficient to pass it along as an argument?

David-Henner (approved):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: Yes, I agree, the struct looks better in this scenario.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: I removed that, but it turned out that I need to save a conversation. We have a `userChange` observer, and at this point we need to recreate the configuration struct.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -21,30 +21,30 @@ import UIKit
...
-        switch self {
-        case .userName, .botName:
-            return FontSpec(.medium, .semibold).font!
-        case .botSuffix:
-            return FontSpec(.medium, .regular).font!
+    let fullName: String
+    let textColor: UIColor
+    let icon: StyleKitIcon?
+
+    init(user: UserType, in conversation: ConversationLike?) {
```
johnxnguyen, line 30: If you're always passing in a conversation, then it doesn't need to be optional.

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: We create and configure this view in the `ConversationSenderMessageCell`. To configure the current `SenderCellComponent` view, we use the `Config` structure, which contains the `conversation` as optional (which is strange to me).

johnxnguyen: I see now that the conversation passed in to the configure method is also optional, which is why the argument here is optional.

KaterinaWire:
@David-Henner @johnxnguyen I added a new commit after your approvals because there were failed tests. As I found there is a color difference between `user.accentColor` and  `UIColor.nameColor(for: user.accentColorValue, variant: ColorScheme.default.variant)`. Previously, `UIColor.nameColor(for: user.accentColorValue, variant: ColorScheme.default.variant)` was used for this view

PR #5071, opened 2021-04-29
[Fix] Call router post call action

David-Henner:
### Issues

The degradation alert is not showing up after the call 

### Causes

In `ActiveCallRouter`, the post call action is executed directly if the active call is shown. It should be the opposite.

### Solutions

Execute post call action directly only when the active call is not shown, otherwise, save the action for after the call.


johnxnguyen (commented):
Are you able to add a test to assert the correct behavior now?

David-Henner:
> Are you able to add a test to assert the correct behavior now?

done here ba973e0

typfel (approved):

johnxnguyen (approved):

PR #5076, opened 2021-05-03
[Fix] Navigation from push notification when switching accounts

typfel:
### Issues

App doesn't navigate to the conversation when opening the application via a push notification in all cases.

### Causes

The method call for navigating to a specific conversation sometimes happen before the app finished transitioning to the `.authenticated` state which causes the action not be performed at all. One case when this happens is when the account needs to change on multi account setup.

### Solutions

Delay the method call for navigating until the `.authenticated` state has been reached.

### DIFF
--- a/Wire-iOS Tests/Routers/URLActionRouterTests.swift
+++ b/Wire-iOS Tests/Routers/URLActionRouterTests.swift
@@ -59,6 +59,51 @@ final class URLActionRouterTests: XCTestCase {
         // THEN
         XCTAssertFalse(router.wasDeepLinkOpened)
     }
+
+    // MARK: Navigation
+
+    func testThatNavigationPerformed_WhenAuthenticatedRouterIsAvailable() {
+        // GIVEN
+        let authenticatedRouter = MockAuthenticatedRouter()
+        let router = TestableURLActionRouter(viewController: RootViewController())
+        router.authenticatedRouter = authenticatedRouter
+
+        // WHEN
+        router.navigate(to: .converationList)
+
+        // THEN
+        if case .converationList = authenticatedRouter.didNavigateToDestination { } else {
+            XCTFail()
+        }
+    }
+
+    func testThatNavigationPerformed_WhenAuthenticatedRouterBecomesAvailable() {
+        // GIVEN
+        let authenticatedRouter = MockAuthenticatedRouter()
+        let router = TestableURLActionRouter(viewController: RootViewController())
+        router.navigate(to: .converationList)
+
+        // WHEN
+        router.authenticatedRouter = authenticatedRouter
+
+        // THEN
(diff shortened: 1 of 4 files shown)


Wire-iOS Tests/Routers/URLActionRouterTests.swift
```diff
@@ -59,6 +59,51 @@ final class URLActionRouterTests: XCTestCase {
...
+        // GIVEN
+        let authenticatedRouter = MockAuthenticatedRouter()
+        let router = TestableURLActionRouter(viewController: RootViewController())
+        router.authenticatedRouter = authenticatedRouter
+
+        // WHEN
+        router.navigate(to: .converationList)
+
+        // THEN
+        if case .converationList = authenticatedRouter.didNavigateToDestination { } else {
```
johnxnguyen, line 75: Are these braces after `didNavigateToDestination` necessary?

Wire-iOS Tests/Routers/URLActionRouterTests.swift
```diff
@@ -59,6 +59,51 @@ final class URLActionRouterTests: XCTestCase {
...
+        // GIVEN
+        let authenticatedRouter = MockAuthenticatedRouter()
+        let router = TestableURLActionRouter(viewController: RootViewController())
+        router.navigate(to: .converationList)
+
+        // WHEN
+        router.authenticatedRouter = authenticatedRouter
+
+        // THEN
+        if case .converationList = authenticatedRouter.didNavigateToDestination { } else {
```
johnxnguyen, line 90: Same as above.

johnxnguyen (approved):

Wire-iOS Tests/Routers/URLActionRouterTests.swift
typfel: Yes. This assert could be much simplified if `NavigationDestination` implemented `Equatable` but that's problematic due to it having `UserType` as an associated type.

johnxnguyen: Ah I see now, it's an empty if block. I think it would be more readable to use a guard:

```swift
guard let case .conversationList = authenticatedRouter.didNavigateToDestination else { 
    XCTFail()
}
```

Wire-iOS/Sources/AuthenticatedRouter.swift
```diff
@@ -19,9 +19,17 @@
 import UIKit
 import WireSyncEngine
 
+public enum NavigationDestination {
+    case conversation(ZMConversation, ZMConversationMessage?)
+    case userProfile(UserType)
+    case connectionRequest(UUID)
+    case converationList
```
johnxnguyen, line 26: Typo: conversationList

David-Henner (approved):

PR #5082, opened 2021-05-05
[Fix] Resize the role indicator icon

KaterinaWire:
### Issues
According to the design specs, the team role icon should be 16X16. 

https://zpl.io/bo36JLG
or
zpl://screen?sid=6089665acc5a5134e260362a&pid=60882917e2efdeae03a21512

### Causes
The existing `service icon` was 12x12, so the new icons for `external` and `guest` users are the same size.

### Solutions
Move the role indicator to a separate view and set the correct size.


typfel:
According to the Zeplin document should the guest indicator icon be 14x14 and the external indicator icon should be 16x12

KaterinaWire:
> According to the Zeplin document should the guest indicator icon be 14x14 and the external indicator icon should be 16x12

@typfel Yes and no. The external indicator icon should be 16x16 (https://zpl.io/aR4Xd3E or zpl://screen?pid=60882917e2efdeae03a21512&sid=60884ac23f3d5135548f69c8), this's a size of SVG image.
But you are right about the guest icon, it should be 14x14. I added a helper method to define the icon size. But later I'll move it to some public place, because Wolfgang mentioned that we should revise the size of the icons on the profile page.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
```diff
@@ -122,30 +110,31 @@ final class SenderCellComponent: UIView {
             avatarSpacer.heightAnchor.constraint(equalTo: avatar.heightAnchor),
             avatarSpacer.centerXAnchor.constraint(equalTo: avatar.centerXAnchor),
             avatarSpacer.centerYAnchor.constraint(equalTo: avatar.centerYAnchor),
+
+            authorLabel.trailingAnchor.constraint(equalTo: teamRoleIndicator.leadingAnchor, constant: -8),
```
johnxnguyen, line 114: Is this necessary, even when both the label and image view are in the same stack?

KaterinaWire: I put three elements in one stack: avatar, label and indicator. Avatar has own container `avatarSpacer` (with a space)

typfel: I guess normally you don't put constraints between elements in a stack but only self constraint which define the size.

typfel:
> > According to the Zeplin document should the guest indicator icon be 14x14 and the external indicator icon should be 16x12
> 
> @typfel Yes and no. The external indicator icon should be 16x16 (https://zpl.io/aR4Xd3E or zpl://screen?pid=60882917e2efdeae03a21512&sid=60884ac23f3d5135548f69c8), this's a size of SVG image.
> But you are right about the guest icon, it should be 14x14. I added a helper method to define the icon size. But later I'll move it to some public place, because Wolfgang mentioned that we should revise the size of the icons on the profile page.

Ok all clear now! I think you should mention to design that it is not desired to have these special icon size cases depending on which icon we are displaying.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/CellComponents/SenderCellComponent.swift
KaterinaWire: This was done because there is no equal space between the elements on the stack.

KaterinaWire: I added an additional stackView and moved the title and indicator to this view. Thus, the main container no longer needs spaces.

johnxnguyen: So you want to add extra space between the label and the icon? `UIStackView` has a method for that.

KaterinaWire: Oh, I didn't know about that method. But now it seems clearer to have the label and icon in a separate stack view. I mean, at least the constraints are clearer. What do you think?

johnxnguyen (approved):

typfel (approved):

PR #5084, opened 2021-05-06
Slather code coverage support

billypchan:
## What's new in this PR?

Add Slather config file `.slather.yml` to config code coverage report generation on CI server


.slather.yml
```diff
@@ -0,0 +1,5 @@
+coverage_service: cobertura_xml
+xcodeproj: ./Wire-iOS.xcodeproj
+scheme: Wire-iOS
```
johnxnguyen, line 3: Will we also collect coverage for share extension? Do we even do this currently?

johnxnguyen (approved):

PR #5092, opened 2021-05-10
[Fix] Switch backend alert shown twice when switching accounts

typfel:
### Issues

When opening a deep link to switch backend the alert would be shown again after switching to your second account.

### Causes

The `URLActionRouter` stores the link in a `url` property for later processing when the application has reached a valid app state. For the backend switch case though it was processing the deep link immediately and storing it in the `url` property for later processing which gets triggered when the app enters  the `.authenticated` state after switching accounts.

### Solutions

Remove the `url` property and show or delay showing alerts in a similar way as we do for app navigation.

### DIFF
--- a/Wire-iOS Tests/Routers/URLActionRouterTests.swift
+++ b/Wire-iOS Tests/Routers/URLActionRouterTests.swift
@@ -21,43 +21,55 @@ import XCTest
 
 final class URLActionRouterTests: XCTestCase {
 
-    func testThatDeepLinkIsNotOpened_WhenDeepLinkIsNotValid() {
+    // MARK: Presenting Alerts
+
+    func testThatAlertIsPresented_WhenCanDisplayAlertsReturnsTrue() {
         // GIVEN
-        let invalidDeepLinkUrl = URL(string: "wire://invalidDeepLinkUrl")!
-        let router =  TestableURLActionRouter(viewController: RootViewController())
-        router.testHelper_setUrl(invalidDeepLinkUrl)
+        let alert = UIAlertController.alertWithOKButton(message: "Hello World")
+        let viewController = RootViewController()
+        let delegate = MockURLActionRouterDelegate()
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/URLActionRouter.swift
```diff
@@ -116,6 +103,28 @@ class URLActionRouter: URLActionRouterProtocol {
...
+            pendingAlert = alert
+            return
+        }
+
+        internalPresentAlert(alert)
+    }
+
+    func internalPresentAlert(_ alert: UIAlertController) {
+        rootViewController.present(alert, animated: true, completion: nil)
+    }
```
johnxnguyen, line 127: I only see this used in the method `presentAlert`. Perhaps we don't need this internal method?

johnxnguyen (approved):

Wire-iOS Tests/Routers/URLActionRouterTests.swift
```diff
@@ -21,43 +21,55 @@ import XCTest
...
-    func testThatDeepLinkIsNotOpened_WhenDeepLinkIsValidAndAppStateInvalid() {
+    func testThatPendingAlertIsPresented_WhenPerformPendingActionsIsCalled() {
         // GIVEN
-        let validDeepLink = URL(string: "wire://start-sso/wire-5977c2d2-aa60-4657-bad8-4e4ed08e483a")!
-        let router =  TestableURLActionRouter(viewController: RootViewController())
-        router.testHelper_setUrl(validDeepLink)
+        let alert = UIAlertController.alertWithOKButton(message: "Hello World")
+        let viewController = RootViewController()
+        let delegate = MockURLActionRouterDelegate()
+        delegate.canDisplayAlerts = false
```
KaterinaWire, line 62: Do we need to set the value to `False` if it changes to `True` in the "When" section?

PR #5093, opened 2021-05-10
[Refactor] Self and remote video preview views

David-Henner:
## What's new in this PR?

This PR refactors the views used to show video tiles in the video grid. 

There are three classes: 
- `BaseVideoPreviewView` - the parent class
- `VideoPreviewView` - a child class specialised in showing the remote videos
- `SelfVideoPreviewView` - a child class specialised in showing the self video

The following functionalities of `VideoPreviewView` are now supported by `BaseVideoPreviewView`:
- Scaling of video via pinch-to-zoom
- Toggling between fill / fit mode

This is great because it also enables the functionalities for `SelfVideoPreviewView`.

Additionally, I extracted the scaling logic to a separate component `ScalableView` and cleaned up the classes files.


Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoPreviews/SelfVideoPreviewView.swift
```diff
@@ -42,17 +47,26 @@ final class SelfVideoPreviewView: BaseVideoPreviewView {
...
+        insertSubview(scalableView, belowSubview: userDetailsView)
+        self.scalableView = scalableView
     }
 
     override func createConstraints() {
         super.createConstraints()
-        previewView.fitInSuperview()
+        [previewView, scalableView].forEach {
+            $0?.translatesAutoresizingMaskIntoConstraints = false
+            $0?.fitInSuperview()
```
billypchan, line 61: I would suggest use 4 constraints instead of this deprecated method in the future.

billypchan (approved):

Wire-iOS/Sources/UserInterface/Calling/VideoGridView/VideoPreviews/SelfVideoPreviewView.swift
johnxnguyen: Why is it deprecated? Personally I find this single line much cleaner that the min 4 that is needed to write the layout constraints explicitly.

johnxnguyen (approved):

PR #5103, opened 2021-05-24
[Feature] Showing an alert when new connections can't be added

KaterinaWire:
## What's new in this PR?

When we receive `error code 412` in response to posting a new connection, an alert will be shown with an error message.

Needs releases with:

- [ wireapp/wire-ios-sync-engine/pull/1441 ] 

### Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-445


Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -232,6 +235,17 @@ final class SearchResultsViewController: UIViewController {
         sectionController.collectionView?.collectionViewLayout.invalidateLayout()
     }
 
+    @objc
+    func showError() {
+        typealias ConnectionError = L10n.Localizable.Error.Connection
+
+        let alertController = UIAlertController(title: "error.conversation.title".localized,
+                                                message: ConnectionError.missingLegalholdConsent,
+                                                alertAction: .ok(style: .cancel))
```
johnxnguyen, line 244:
```suggestion
        typealias ErrorString = L10n.Localizable.Error

        let alertController = UIAlertController(title: ErrorString.Conversation.title,
                                                message: ErrorString.Connection.missingLegalholdConsent,
                                                alertAction: .ok(style: .cancel))
```

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -215,6 +216,8 @@ final class SearchResultsViewController: UIViewController {
         servicesSection.delegate = self
         createGroupSection.delegate = self
         inviteTeamMemberSection.delegate = self
+
+        NotificationCenter.default.addObserver(self, selector: #selector(showError), name: ZMConnectionLimitNotification.missingLegalHoldConsent, object: nil)
```
johnxnguyen, line 220: Does `ZMConnectionLimitNotification` need updating or renaming?

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
```diff
@@ -232,6 +235,17 @@ final class SearchResultsViewController: UIViewController {
         sectionController.collectionView?.collectionViewLayout.invalidateLayout()
     }
 
+    @objc
+    func showError() {
```
typfel, line 239: A more detailed methods name since it's not clear from the context which kind we'll show.

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
typfel: You added a an `addObserver` method in SE which you don't use here.

typfel:
Note that it's also possible to send a connection request by navigating to a user from  the conversations details screen.

typfel (changes requested):

Wire-iOS/Sources/UserInterface/StartUI/StartUI/SearchResultsViewController.swift
KaterinaWire: I guess I do not need addObserver method in `SE`, I can remove it there. @typfel  What do you think?

KaterinaWire:
> Note that it's also possible to send a connection request by navigating to a user from the conversations details screen.

@typfel Would it be a good solution to add the same observer to the  conversations details screen? I don't want to add it to the Profile screen because the user can close it before the BE response.

typfel:
> > Note that it's also possible to send a connection request by navigating to a user from the conversations details screen.
> 
> @typfel Would it be a good solution to add the same observer to the conversations details screen? I don't want to add it to the Profile screen because the user can close it before the BE response.

Yes, I think that would work. It's not 100% foolproof since the conversation details can also be closed.

KaterinaWire:
> > > Note that it's also possible to send a connection request by navigating to a user from the conversations details screen.
> > 
> > 
> > @typfel Would it be a good solution to add the same observer to the conversations details screen? I don't want to add it to the Profile screen because the user can close it before the BE response.
> 
> Yes, I think that would work. It's not 100% foolproof since the conversation details can also be closed.

@typfel I added this observer to the Profile screen. Anyway, it's a solution for this release. In the future, the connect button will be hidden and the user will have a chance to see a reason why it's not possible to make connect request  before sending the request.

PR #5127, opened 2021-06-15
Alert user when failing to join conversation

johnxnguyen:
## What's new in this PR?

Joining a conversation via a link may fail in the following cases:

- the link is invalid (it may be incorrect, or revoked),
- the conversation is already full, or
- some unknown network error occurs.

When one of these errors occur, we present an error alert to the user.


Wire-iOS/Sources/URLActionRouter.swift
```diff
@@ -215,18 +216,81 @@ extension URLActionRouter: PresentationDelegate {
...
         }
-        let alertMessage = error.failureReason ?? "error.user.unkown_error".localized
-        let alertController = UIAlertController.alertWithOKButton(title: error.errorDescription,
-                                                                  message: alertMessage)
-        presentAlert(alertController)
+
+        var errorDescription: String? {
+            switch self {
+            case .conversationIsFull, .conversationLinkIsInvalid:
+                fatalError("Not implemented")
```
johnxnguyen, line 262: I'll update this as soon as we have copy.

KaterinaWire (approved):

David-Henner (approved):

PR #5129, opened 2021-06-16
[Feature] audio tiles border

David-Henner:
## What's new in this PR?

This PR adds a border to call grid tiles when there is no video being shared in the tile.
The border shouldn't appear when there's only one tile in the grid, when the tile is maximised (full screen) or when it's the self participant tile being displayed in the thumbnail view.

## Changes
- Added border to `BaseCallParticipantView` and merged presentation logic with the active speaker frame border.
- Updated `CallGridViewController` to setup the participant tiles accordingly to requirements
- Updated tests & Added new snapshots


Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
```diff
@@ -367,6 +370,10 @@ final class CallGridViewController: SpinnerCapableViewController {
 
     // MARK: - Helpers
 
+    private func shouldShowBorderWhenVideoIsStopped(isSelfCallParticipantView: Bool = false) -> Bool {
+        return dataSource.count > 1 && (isSelfCallParticipantView ? configuration.floatingStream == nil : true)
```
johnxnguyen, line 374: I think it always helps to give names to components of a condition to make it more readable. For instance, this could become (I think...):

```swift
let existMoreThanOneTile = dataSource.count > 1
let existsFloatingTileForSelfUser = isSelfCallParticipantView && configuration.floatingStream != nil
return existsMoreThanOneTile && !existsFloatingTileForSelfUser
```

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
David-Henner: Good point! I don't think I can do it the way you suggested, but I will break it down into something clearer

David-Henner: This might be bit repetitive but I think it's pretty clear
```
let gridHasMoreThanOneParticipant = dataSource.count > 1
let gridHasFloatingParticipant = configuration.floatingStream != nil

let shouldShowBorderForSelfParticipant = gridHasMoreThanOneParticipant && !gridHasFloatingParticipant
let shouldShowBorderForOtherParticipant = gridHasMoreThanOneParticipant

return isSelfCallParticipantView ? shouldShowBorderForSelfParticipant : shouldShowBorderForOtherParticipant
```

David-Henner: Good point! What would you say about this?
```swift
let gridHasMoreThanOneParticipant = dataSource.count > 1
let gridHasFloatingParticipant = configuration.floatingStream != nil

let shouldShowBorderForSelfParticipant = gridHasMoreThanOneParticipant && !gridHasFloatingParticipant
let shouldShowBorderForOtherParticipant = gridHasMoreThanOneParticipant

return isSelfCallParticipantView ? shouldShowBorderForSelfParticipant : shouldShowBorderForOtherParticipant
```

johnxnguyen: Maybe a little too much? Perhaps just skipping the last two lets would be a nice balance. Or, would something like this also work?

```swift
let participantCount = dataSource.count
let gridHasOnlyOneParticipant = participantCount == 1
let gridIsOneToOneWithFloatingStream = participantCount == 2 && configuration.floatingStream != nil
return !gridHasOnlyOneParticipant && !gridIsOneToOneWithFloatingStream
```

David-Henner: I've updated it based off of your suggestion. Note that `dataSource.count` only counts the number of tiles in the CollectionView (GridView) but it doesn't take into account the floating tile.

KaterinaWire:
@David-Henner `SwiftLint` detected a whitespace violation.

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
johnxnguyen: The end result is quite nice!

David-Henner:
> @David-Henner `SwiftLint` detected a whitespace violation.

@KaterinaWire It's shown as successful to me 🤔

David-Henner:
Fixed it by running swiftlint autocorrect c37709f

KaterinaWire:
> > @David-Henner `SwiftLint` detected a whitespace violation.
> 
> @KaterinaWire It's shown as successful to me 🤔

Yes. Now it's fine

KaterinaWire (approved):

PR #5134, opened 2021-06-18
[Feature] Update hint notification behaviour for tiles without video

David-Henner:
## What's new in this PR?

In the calling grid, we use hint labels to notify the user about the actions available to them. The current hints are:
- `Double tap on a tile for fullscreen` - showed when the calling grid first appears
- `Double tap to go back, pinch to zoom` - showed when a user maximises a tile
- `Pinch to zoom` - showed when there's a screenshare in 1:1 calls

The calling grid now allows for non-video tiles, for which pinch-to-zoom is disabled. 
So, for non-video tiles, I changed the hint `Double tap to go back, pinch to zoom` into `Double tap to go back`

Maximising a non-video tile - when it's the only one in the grid - shows the `Double tap to go back` hint, but has no other effect since the tile already appears fullscreen. Thus I disabled maximisation in this case.

### Todo

- [ ] Update tests

### DIFF
--- a/Wire-iOS Tests/Calling/CallGridViewControllerSnapshotTests.swift
+++ b/Wire-iOS Tests/Calling/CallGridViewControllerSnapshotTests.swift
@@ -193,110 +193,179 @@ final class CallGridViewControllerSnapshotTests: XCTestCase {
 
     // MARK: - Hint update
 
-    func testThat_HintIs_Fullscreen_ForViewDidLoad() {
-        // given
+    func testThat_ItUpdatesHint() {
+
+        // Maximization
+
+        assertHint(
+            input: .maximizationChanged(state: .maximized(isSharingVideo: true)),
+            output: .show(hint: .goBackOrZoom)
+        )
+
+        assertHint(
+            input: .maximizationChanged(state: .maximized(isSharingVideo: false)),
+            output: .show(hint: .goBack)
+        )
+
+        assertHint(
+            input: .maximizationChanged(state: .notMaximized),
+            output: .hideHintAndStopTimer
+        )
+
+        // Configuration Changed
+
+        assertHint(
+            input: .configurationChanged(participants: .moreThanTwo),
+            output: .showNothing
+        )
+
+        assertHint(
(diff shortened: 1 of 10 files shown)


johnxnguyen (commented):
Looks good to me, but I'll wait for the tests before approving.

Wire-iOS Tests/Calling/CallGridViewControllerSnapshotTests.swift
```diff
@@ -185,4 +190,113 @@ final class CallGridViewControllerSnapshotTests: XCTestCase {
         // then
         verify(matching: sut)
     }
+
+    // MARK: - Hint update
```
KaterinaWire, line 194: To be honest, it feels like some test has been missed because there are a lot of parameters and, maybe, there are no tests for some combination of parameters. We had a similar case with `AppLock` in the `DataModel`. We needed to check the results, which depend on different combinations of parameters. You can find it in `test_ItEvaluatesAuthentication()`(data model).
I suggest to do something like this:
    ```
    typealias Input = (hasTwoParticipants: Bool, isSharingVideo: Bool, isScreensharing: Bool, isMaximized: Bool, ...)
    typealias Output = CallGridHintKind?

    private func assert(input: Input, output: Output, file: StaticString = #file, line: UInt = #line) {
        configuration.callHasTwoParticipants = input.hasTwoParticipants
        createSut()
        sut.hintView = mockHintView

        sut.updateHint(for: .configurationChanged)// can be a parameter

        XCTAssertEqual(mockHintView.hint, output, file: file, line: line)
    }
```
It might be more clear... If you don't like this approach, you can ignore this message

David-Henner: Thanks! I ended up using your suggestion and adapted it to make sure that we can only test valid combinations of parameters

KaterinaWire: Looks great!

KaterinaWire (approved):

johnxnguyen (approved):

PR #5150, opened 2021-07-02
[Feature] Request video streams page by page

David-Henner:
## What's new in this PR?

This PR optimises how video streams are loaded during a conference call. 
The calling grid can now support a large number of participants distributed across different pages. 
Since we only care about seeing videos for the participants visible on screen, AVS will only load video streams for the participants we request.

### Changes 

- Updated `CallGridViewController` to request video streams for clients in the current grid page, filtering out those with video disabled.
- Created `CallGridViewControllerDelegate` to delegate actions (`CallGridAction`) to. 
- Conformed `CallViewController` to `CallGridViewControllerDelegate`
- Minor changes to `GridView`
- Updated tests

### Dependencies

Needs releases with:

- https://github.com/wireapp/wire-ios-sync-engine/pull/1465


Wire-iOS Tests/Calling/CallGridViewControllerSnapshotTests.swift
```diff
@@ -246,6 +246,45 @@ final class CallGridViewControllerSnapshotTests: XCTestCase {
...
+        for _ in half..<CallGridViewController.maxItemsPerPage {
+            let client = AVSClient(userId: UUID(), clientId: UUID().transportString())
+            configuration.streams += [stubProvider.stream(client: client, videoState: .started)]
+            expectedClients += [client]
+        }
+
+        sut.configuration = configuration
+
+        // when
+        sut.requestVideoStreamsIfNeeded(forPage: 1)
```
johnxnguyen, line 282: Is this supposed to be for page 2?

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
```diff
@@ -42,8 +47,9 @@ final class CallGridViewController: SpinnerCapableViewController {
         PinchToZoomRule(isOneToOneCall: configuration.callHasTwoParticipants)
     }
 
+    private var visibleClientsSharingVideo: [AVSClient] = []
     private var dataSource: [Stream] = []
-    private let gridView = GridView(maxItemsPerPage: 8)
+    private let gridView = GridView(maxItemsPerPage: CallGridViewController.maxItemsPerPage)
```
johnxnguyen, line 52: If you want to make this a bit shorter, you can write:

```swift
Self.maxItemsPerPage
```

Wire-iOS/Sources/UserInterface/Calling/CallGridView/Grid/GridView.swift
```diff
@@ -19,7 +19,7 @@
 import UIKit
 
 protocol GridViewDelegate: class {
-    func gridViewPageDidChange(to index: Int)
+    func gridView(_ gridView: GridView, pageDidChangeTo page: Int)
```
johnxnguyen, line 22: Suggestion:

```swift
func gridView(_ gridView: GridView, didChangePageTo page: Int)
```

Wire-iOS/Sources/Helpers/Array+Prefix.swift
```diff
@@ -0,0 +1,25 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension Array {
+    func prefixed(by amount: Int) -> [Element] {
+        return Array(prefix(amount))
+    }
+}
```
johnxnguyen, line 25: Is this only used for tests? I'd be weary of making this widely accessible because it obscures the fact that it may be more expensive than simply using `prefix` (which doesn't allocate a new array).

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
```diff
@@ -353,6 +362,24 @@ final class CallGridViewController: SpinnerCapableViewController {
...
+        let startIndex = page * gridView.maxItemsPerPage
+        let endIndex = isLastPage ? dataSource.count : startIndex + gridView.maxItemsPerPage
+
+        guard endIndex <= dataSource.count, startIndex >= 0 else { return }
+
+        let clients = dataSource[startIndex..<endIndex]
+            .filter(\.isSharingVideo)
+            .map(\.streamId)
+
+        guard clients != visibleClientsSharingVideo else { return }
```
typfel, line 377: We can assume that the order is always the same and it's therefore valid to this on an Array and not a Set?

typfel (approved):

Wire-iOS/Sources/Helpers/Array+Prefix.swift
David-Henner: It's only used in tests. Good point, I'll move it to the test target

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
David-Henner: Good question! I wouldn't assume the order to always be the same. It would be better to use a Set

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
```diff
@@ -363,10 +363,9 @@ final class CallGridViewController: SpinnerCapableViewController {
     }
 
     func requestVideoStreamsIfNeeded(forPage page: Int) {
-        let isLastPage = page == gridView.numberOfPages
-
         let startIndex = page * gridView.maxItemsPerPage
-        let endIndex = isLastPage ? dataSource.count : startIndex + gridView.maxItemsPerPage
+        var endIndex = startIndex + gridView.maxItemsPerPage
+        endIndex = endIndex <= dataSource.count ? endIndex : dataSource.count
```
johnxnguyen, line 368:
```suggestion
        endIndex = min(endIndex, dataSource.count)
```

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
David-Henner: `Self` can't be referenced from a stored property initializer

johnxnguyen (approved):

PR #5162, opened 2021-07-08
[Refactor] IconLabelButton tests

David-Henner:
## What's new in this PR?

This refactor changes the way `IconLabelButtonTests` are structured. If it's deemed useful, we can consider applying this structure to other tests. It can also be kept as an example for future reference.

### Initial observation
- Setting up the view for testing is redundant. The same properties are set with different values. There is a lot of manual setup
- The test functions names are just describing the parameters being tested. We need to read through all the tests to understand what is being tested.

### Solution
By defining enums for each parameter, we get a better overview of what the parameters are, and what values they can have.
It also allows us to define the possible combination of values so that there is no invalid combination.
We can also leverage the `CaseIterable` protocol to compute a list of all test cases, which we can then use to set up each test, instead of doing it manually.


typfel:
Interesting idea David! 

Here's some thoughts from looking at your solution:
- Iterating through all cases works well when you deal with boolean properties or finite enums but becomes more problematic when dealing with Integers, colors, text.
- Continuing on that theme I don't think it makes sense to this kinda of enum enumeration for types of views since some of them will have a very large number of configurations it wouldn't make sense to try to test all of them when we would probably in practice only use a few variations. 
- Have you seen the new way of styling buttons introduced in iOS 15 https://wwdcbysundell.com/2021/uikits-new-button-configuration-api the concept of a separate configuration struct which can be re-used and extended would also be useful in tests.
- This testing strategy maybe makes the most sense for base UI components which are simple.

Wire-iOS Tests/IconLabelButtonTests.swift
```diff
@@ -36,120 +36,107 @@ final class IconLabelButtonTests: XCTestCase {
...
+            return selectionState.isSelected
+        }
+
+        var isEnabled: Bool {
+            return interactionState.isEnabled
+        }
+
+        var description: String {
+            switch self {
+            case .dark(let blurState, let selectionState, let interactionState):
```
johnxnguyen, line 96: When you need to bind multiple associated values, you can do them all at once using a single let:

```suggestion
            case let .dark(blurState, selectionState, interactionState):
```

Wire-iOS Tests/IconLabelButtonTests.swift
```diff
@@ -36,120 +36,107 @@ final class IconLabelButtonTests: XCTestCase {
...
+}
 
-        // Then
-        verify(matching: button)
+struct IconLabelButtonTestCase {
+
+    enum Appearance: CaseIterable {
+        typealias AllCases = [Appearance]
+
+        static var allCases: AllCases {
```
johnxnguyen, line 61: The compile can infer the type for `AllCases` if you specify the type of `allCases` explicitly:

```swift
static var allCases: [Appearance] {
  ...
}
```

johnxnguyen (approved):
I'm writing my comments that I couldn't finish during our call here:

There are several different things here that are interesting from a readability point of view. First is that it's nice to separate the code related to setting up the test from the code that represents what is being tested. Abstracting away setup code means you can easily capture exactly what is being tested. 👍

Taking this idea further, it helps greatly to give names to state that may consist of multiple pieces. You achieve this by creating enum cases with intentional relationships to express what is a valid state and what is not, and by giving more meaningful names (eg. blurred, selected) than relying on a collection of properties. 👍

Using `CaseIterable` to generate all possible configurations to test is a handy trick, but as mentioned by Jacob, this may make sense in only certain cases, such as testing a generic component that can be used in any number of ways. Keep in mind that the total configuration space (the number of unique configurations) can grow quite fast depending on how many properties there are (in general, it would be `a x b x c x ...` where each letter is the total number of states for a particular property.

Where this exhaustive testing doesn't make much sense, in my opinion, is when you're no longer testing a generic component but rather a specific one (such as a button used specifically for the call ui). You probably only need to test what you actually use, otherwise you run the risk of wasting resources testing things aren't being used. In such a case, as Jacob mentioned earlier, it would make sense to define styles and test the component with those styles applied.

As a final thought, just on the topic of testing itself, I think it's important to avoid focusing solely on testing the implementation of a function (by testing all inputs and outputs) and instead write tests that assert the (abstract) behavior that you are trying to build. In the past I've written tests that assert all possible input-output pairs, and then found it really hard later on to see through the clutter and understand what behavior I wanted to test. I think sometimes a few well picked test cases is more expressive than an exhaustive list, even if the latter has more tests.

David-Henner:
Thank you @johnxnguyen for your detailed feedback! I agree with your points. I'd just like to expand on the subject of using `CaseIterable`. I agree that generating all possible configurations won't always make sense, especially when we only want to test a limited, specific set of combinations. One obvious solution would be to not use `CaseIterable` and define the combinations one by one. There's another solution that would allow us to continue using `CaseIterable`: we can define the enums in such a way that they would only cover the combinations we want to test.

Here are two simple examples to illustrate what I mean. We could also imagine more complex scenarios

```swift
// In this example, the blur state is restricted to .dark appearance
enum Appearance {
 case light
 case dark(BlurState)
}

enum BlurState {
 case blurred
 case notBlurred
}

---

// In this example, we could have had a single enum BlurState with all possible blur combinations, 
// but using all cases would have us end up with combinations we do not want to test. 
// Defining two separate enums helps us solve the issue.
enum Appearance {
 case light(LightBlurState)
 case dark(DarkBlurState)
}

enum LightBlurState {
 case yellowBlur
 case blueBlur
}

enum DarkBlurState {
 case blurred
 case notBlurred
}

```

[Here](https://github.com/wireapp/wire-ios/blob/4b9836a0f7c20c98ff359c418a792ca0e65572f4/Wire-iOS%20Tests/Calling/CallGridViewControllerSnapshotTests.swift#L285) you can also find an example that asserts the behaviour of a component.

KaterinaWire (approved):

PR #5169, opened 2021-07-19
Fix: application extension not work after using XCFramework

billypchan:
### Issues

~~App crashed on device when start after using XCFramework.~~ Warning is produces when using non extension allowed method in share extension.

### Causes

From the log, some method is not allow on app extension.

### Solutions

clean up `APPLICATION_EXTENSION_API_ONLY` in project and follow the flag in `.xcconfig`.
Refactor the code and do not using `UIApplication` in `common component`.
Using `effectiveUserInterfaceLayoutDirection` for checking text direction.

### DIFF
--- a/Wire-iOS/Sources/Helpers/UIEdgeInsets+directionAware.swift
+++ b/Wire-iOS/Sources/Helpers/UIEdgeInsets+directionAware.swift
@@ -21,8 +21,7 @@ import UIKit
 
 extension UIView {
     var isLeftToRight: Bool {
-        return UIView.userInterfaceLayoutDirection(
-            for: semanticContentAttribute) != .rightToLeft
+        return effectiveUserInterfaceLayoutDirection != .rightToLeft
     }
 }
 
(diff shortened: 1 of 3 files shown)


Wire-iOS/Sources/Helpers/UIEdgeInsets+directionAware.swift
```diff
@@ -0,0 +1,54 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import UIKit
+
+extension UIView {
+    var isLeftToRight: Bool {
+        return effectiveUserInterfaceLayoutDirection != .rightToLeft
```
KaterinaWire, line 24: Why not to use
`effectiveUserInterfaceLayoutDirection == .leftToRight`?
Isn't it clearer?

billypchan: Yes, it is clearer and I fixed it.

KaterinaWire (approved):

David-Henner (approved):

PR #5190, opened 2021-07-29
[Feature] Large conference calls

David-Henner:
## What's new in this PR?

This is the feature branch for large conference calls

### Needs release with 

- [x]  https://github.com/wireapp/wire-ios-sync-engine/pull/1465


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -682,7 +682,7 @@
 "call.status.constant_bitrate" = "Constant Bit Rate";
 "call.status.variable_bitrate" = "Variable Bit Rate";
 
-"call.participants.show_all" = "Show All (%@)";
+"call.participants.show_all" = "Participants (%@)";
```
johnxnguyen, line 685: We always inject a number in this placeholder right? Then I think we want `%d`, which results in the more type safe: `L10n.Localizable.Call.Paratipants.showAll(n: Int)`.

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
```diff
@@ -114,24 +130,30 @@ final class VideoGridViewController: SpinnerCapableViewController {
...
 
         NSLayoutConstraint.activate([
             topStack.centerXAnchor.constraint(equalTo: view.centerXAnchor),
             topStack.topAnchor.constraint(equalTo: view.safeTopAnchor, constant: 24),
             topStack.leadingAnchor.constraint(greaterThanOrEqualTo: view.leadingAnchor, constant: 20),
-            topStack.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor, constant: -20)
+            topStack.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor, constant: -20),
+            pageIndicator.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            pageIndicator.heightAnchor.constraint(equalToConstant: CGFloat.pageIndicatorHeight),
+            pageIndicator.centerXAnchor.constraint(equalTo: view.trailingAnchor, constant: -22) // (pageIndicatorHeight / 2 + 10)
```
johnxnguyen, line 153: Left over comment?

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -468,7 +477,7 @@ extension CallViewController {
 
     fileprivate var canHideOverlay: Bool {
         guard case .established = callInfoConfiguration.state else { return false }
-        return callInfoConfiguration.isVideoCall
+        return true
```
johnxnguyen, line 480: Maybe it could be simplified to `return callInfoConfiguration.state == .established`

Wire-iOS/Sources/UserInterface/Calling/CallGridView/CallGridViewController.swift
David-Henner: Bill suggested to use `-22` directly to save compile time, the comment is there to let us know that `-22` isn't arbitrary and comes from the `pageIndicatorHeight` property

johnxnguyen: If `pageIndicatorHeight` changes then I guess `-22` needs to change as well?

David-Henner: That's right

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
David-Henner: We need to use the `if case` pattern here because `.established` has an associated value

Wire-iOS/Resources/Base.lproj/Localizable.strings
David-Henner: done here e410da0. Thanks for the suggestion :)

Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
johnxnguyen: All good as it is 👍🏻

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Components/RoundedView/RoundedBlurView.swift
```diff
@@ -0,0 +1,44 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func setupViews() {
+        addSubview(blurView)
+        blurView.translatesAutoresizingMaskIntoConstraints = false
+    }
+
+    func createConstraints() {
+        blurView.fitIn(view: self)
```
KaterinaWire, line 42: Can't we use `fitInSuperview()`?

Wire-iOS Tests/Calling/GridViewTests.swift
```diff
@@ -26,9 +26,19 @@ class OrientableViewMock: OrientableView {
     func layout(forInterfaceOrientation interfaceOrientation: UIInterfaceOrientation, deviceOrientation: UIDeviceOrientation) {}
 }
 
+class GridViewDelegateMock: GridViewDelegate {
+
+    var page: Int = 0
```
KaterinaWire, line 31: Can it be as `private`?

KaterinaWire (approved):

KaterinaWire (approved):

Wire-iOS Tests/Calling/GridViewTests.swift
David-Henner: it's used further down in the tests :)

PR #5205, opened 2021-08-11
[Feature] Hide download, save and share buttons

KaterinaWire:
## What's new in this PR?

If the file sharing feature is disabled, we should hide the `download`, `save` and `share` buttons on the message action toolbar.

### DIFF
--- a/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
+++ b/Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
@@ -20,7 +20,6 @@ import Foundation
 
 enum SecurityFlags {
     case clipboard
-    case saveMessage
     case generateLinkPreviews
     case forceConstantBitRateCalls
     case openFilePreview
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/Helpers/Bundle+SecurityFlags.swift
```diff
@@ -20,7 +20,6 @@ import Foundation
 
 enum SecurityFlags {
     case clipboard
-    case saveMessage
```
johnxnguyen, line 23: Make sure to also remove these flags from the configuration files.

Wire-iOS/Sources/Helpers/syncengine/ZMMessage+Actions.swift
```diff
@@ -109,7 +109,10 @@ extension ZMConversationMessage {
         guard let fileMessageData = self.fileMessageData else {
             return false
         }
-        return isFile && fileMessageData.transferState == .uploaded && fileMessageData.downloadState == .remote
+        return isFile
+            && fileMessageData.transferState == .uploaded
+            && fileMessageData.downloadState == .remote
+            && !isRestricted
```
johnxnguyen, line 115: I think we could check if it is restricted first, rather than last. Even in the guard statement.

johnxnguyen (approved):

Wire-iOS/Sources/Helpers/syncengine/ZMMessage+Actions.swift
KaterinaWire: Done

typfel (approved):

PR #5215, opened 2021-08-17
[Chore] disable large conference call features

David-Henner:
## What's new in this PR?

This PR introduces a feature configuration mechanism in the calling code in order to disable key conferencing features that won't be included in the upcoming release.

Disabled / Enabled features:
- [Disabled] - audio tiles
- [Disabled] - pagination
- [Enabled] - swipe down to dismiss call
- [Enabled] - color schemed audio calls (light / dark)
- [Disabled] - toggling the overlay in an audio call
- [Updated] - stream limit in calling grid

### Changes
- Created a `CallingConfiguration` struct
- Enabled / disabled features according to the `CallingConfiguration`

### Notes
- The flags added in `CallingConfiguration` will likely be retired once the large conference calls features has been released
- Snapshot tests and unit tests were not updated to reflect the chosen configuration in order to avoid doing the work twice once the large conference calls feature is re-enabled.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallStatusView.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallStatusView.swift
@@ -163,11 +163,7 @@ extension CallStatusViewInputType {
     var effectiveColorVariant: ColorSchemeVariant {
         guard callingConfig.isAudioCallColorSchemable else { return .dark }
 
-        if isVideoCall {
-            return .dark
-        } else {
-            return variant
-        }
+        return isVideoCall ? .dark : variant
     }
 
 }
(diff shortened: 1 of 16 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallStatusView.swift
```diff
@@ -151,7 +161,13 @@ extension CallStatusViewInputType {
...
 
     var effectiveColorVariant: ColorSchemeVariant {
-        .dark
+        guard callingConfig.isAudioCallColorSchemable else { return .dark }
+
+        if isVideoCall {
+            return .dark
+        } else {
+            return variant
+        }
```
KaterinaWire, line 170: It can be made shorter:
`return isVideoCall ? .dark : variant`

typfel (approved):

Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallStatusView.swift
David-Henner: right! thanks :)

KaterinaWire (approved):

PR #5224, opened 2021-08-23
pin wire-ios-build-configuration version

billypchan:
## What's new in this PR?

Read wire-is version from `version.xcconfig`.  Use the version as an option `--branch` for `download-assets.sh`, to download files form `wire-ios-build-configuration`'s tag.

In this PR, the setup script pins the project to https://github.com/wireapp/wire-ios-build-configuration/releases/tag/3.87

When a new pull request is merged into wire-ios-build-configuration, a new tag matching coming release version should be created to match this project's version.

### TODO
- [x] handle the case that the version tag is not exists in `wire-ios-build-configuration` (use `master` if the tag not exists.)


setup.sh
```diff
@@ -53,8 +53,12 @@ echo "ℹ️  Downloading AVS library..."
 ./Scripts/download-avs.sh
 echo ""
 
+# get version from Wire-iOS/Resources/Configuration/Version.xcconfig
+VERSION=$(swift ./Scripts/readVersion.swift Wire-iOS/Resources/Configuration/Version.xcconfig)
+echo "ℹ️  version is $VERSION"
+
 echo "ℹ️  Downloading additional assets..."
-./Scripts/download-assets.sh "$@"
+./Scripts/download-assets.sh "$@" --branch release/$VERSION
```
billypchan, line 61:
```suggestion
./Scripts/download-assets.sh "$@" --branch $VERSION
```

PR #5238, opened 2021-09-02
Replace localisable string

KaterinaWire:
## What's new in this PR?

According to [design spec](https://zpl.io/aMX9v3N), we should rename the `Pictures` section to `Images`.
We only use this string for the title on the search screen, so it is safe to simply rename it.


billypchan (approved):

billypchan (changes requested):
Seems snapshots needs update after this change.

KaterinaWire:
> Seems snapshots needs update after this change.

Updated

billypchan (approved):

David-Henner (approved):

PR #5253, opened 2021-09-09
[Feature] Set the internal packaging feature flag

KaterinaWire:
## What's new in this PR?

Added the possibility to change the inner packaging feature flag, which can enable/disable the packaging feature.

Needs releases with:

- [ wireapp/wire-ios-sync-engine/pull/1519 ]


johnxnguyen (commented):
Will you use this flag to prevent feature change alerts from appearing?

KaterinaWire:
> Will you use this flag to prevent feature change alerts from appearing?

Actually, I didn't want to prevent feature change alerts as they shouldn't appear if no one changes status. But I added a part where we can check the value of the flag before posting the alert.

johnxnguyen (approved):

David-Henner (approved):

PR #5257, opened 2021-09-15
[Fix] Add restrictions to drag & drop files depends on SecurityFlags

KaterinaWire:
### Issues
There is **drag & drop** feature for iPhones in iOS 15 and this is an issue for **file sharing and copy / paste** restrictions for BK build.

### Solutions
1. Prevent file dropping if SecurityFlags.clipboard.isEnabled is false.
2. Disable the drag interaction if SecurityFlags.clipboard.isEnabled is false.

### Notes

https://wearezeta.atlassian.net/browse/SQSERVICES-755
The playground build for **iOS 15** that has a restriction for drag&drop feature: **3.87 (992928)**

### DIFF
--- a/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
+++ b/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
@@ -36,15 +36,27 @@ final class ConversationInputBarViewControllerDropInteractionTests: XCTestCase {
             output: .forbidden
         )
 
-        // Drop file and the file sharing feature is disabled.
+        // Drop file, the clipboard is disabled and the file sharing feature is disabled.
         assert(
             input: (isText: false, isClipboardEnabled: false, canFilesBeShared: false),
             output: .forbidden
         )
 
-        // Drop file and the file sharing feature is enabled.
+        // Drop file, the clipboard is disabled and the file sharing feature is enabled.
         assert(
             input: (isText: false, isClipboardEnabled: false, canFilesBeShared: true),
+            output: .forbidden
+        )
+
+        // Drop file, the clipboard is enabled and the file sharing feature is disabled.
+        assert(
+            input: (isText: false, isClipboardEnabled: true, canFilesBeShared: false),
+            output: .forbidden
+        )
+
+        // Drop file when the clipboard is enabled and the file sharing feature is enabled.
+        assert(
+            input: (isText: false, isClipboardEnabled: true, canFilesBeShared: true),
             output: .copy
         )
 
--- a/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
+++ b/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
@@ -79,5 +79,5 @@ extension ConversationInputBarViewControllerDropInteractionTests {
 
         XCTAssertEqual(dropProposal.operation, output, file: file, line: line)
     }
-    
+
 }
--- a/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
+++ b/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
@@ -22,7 +22,7 @@ import WireCommonComponents
 
 final class ConversationInputBarViewControllerDropInteractionTests: XCTestCase {
 
-    func test_ItEvaluatesAuthentication() {
+    func testThatItHandlesDroppingFiles() {
 
         // Drop text and the clipboard is enabled.
         assert(
--- a/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
+++ b/Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
@@ -1,0 +1,71 @@
+//
+// Wire
+// Copyright (C) 2021 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import Wire
+import WireCommonComponents
+
+final class ConversationInputBarViewControllerDropInteractionTests: XCTestCase {
+
+    func test_ItEvaluatesAuthentication() {
+
+        // Drop text and the clipboard is enabled.
+        assert(
+            input: (isText: true, isClipboardEnabled: true, canFilesBeShared: false),
+            output: .copy
+        )
+
+        // Drop text and the clipboard is disabled.
+        assert(
+            input: (isText: true, isClipboardEnabled: false, canFilesBeShared: false),
+            output: .forbidden
+        )
+
+        // Drop file and the file sharing feature is disabled.
+        assert(
+            input: (isText: false, isClipboardEnabled: false, canFilesBeShared: false),
+            output: .forbidden
+        )
+
+        // Drop file and the file sharing feature is enabled.
+        assert(
+            input: (isText: false, isClipboardEnabled: false, canFilesBeShared: true),
+            output: .copy
+        )
+
+    }
+}
+
+// MARK: - Helpers
+
+extension ConversationInputBarViewControllerDropInteractionTests {
+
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
```diff
@@ -59,11 +59,11 @@ extension ConversationInputBarViewController: UIDropInteractionDelegate {
     }
 
     func dropInteraction(_ interaction: UIDropInteraction, sessionDidUpdate session: UIDropSession) -> UIDropProposal {
-        return UIDropProposal(operation: .copy)
+        return SecurityFlags.clipboard.isEnabled ? UIDropProposal(operation: .copy) : UIDropProposal(operation: .forbidden)
     }
 
     func dropInteraction(_ interaction: UIDropInteraction, canHandle session: UIDropSession) -> Bool {
-        return session.canLoadObjects(ofClass: UIImage.self)
+        return true
```
billypchan, line 66: should be `return SecurityFlags.clipboard.isEnabled` ?

billypchan (commented):
seems good, but need more tests on real device.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
johnxnguyen: From the docs:

> Return Value: true if the interaction supports the drag items in the sessions; otherwise, false.

So returning `false` would mean that no drop interactions are supported. Isn't this what we want?

johnxnguyen (commented):
Dragging a file into the input bar would allow the user to send a file, so shouldn't the file sharing feature config be involved here?

The play between the file sharing and clipboard restrictions creates a bit more complexity, as it would be possible to enabled the clipboard but disable file sharing. This would mean that we enable drop interactions, but only for items that aren't considered files.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
johnxnguyen: ok so coming back to this comment, I see now how it make sense for us to say that we're interested in all drop interactions, and then to actually decide to forbid it or not in the second step.

KaterinaWire: It was my original idea to use this method and depend on `SecurityFlags.clipboard.isEnabled`, but if I return `false` it will restrict all files, but allow **text** to be dropped in iOS 15. I couldn't find any documentation on this.

KaterinaWire:
> Dragging a file into the input bar would allow the user to send a file, so shouldn't the file sharing feature config be involved here?
> 
> The play between the file sharing and clipboard restrictions creates a bit more complexity, as it would be possible to enabled the clipboard but disable file sharing. This would mean that we enable drop interactions, but only for items that aren't considered files.

I added the file sharing feature config to this method and tests.

Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
```diff
@@ -0,0 +1,71 @@
...
+        assert(
+            input: (isText: false, isClipboardEnabled: false, canFilesBeShared: false),
+            output: .forbidden
+        )
+
+        // Drop file and the file sharing feature is enabled.
+        assert(
+            input: (isText: false, isClipboardEnabled: false, canFilesBeShared: true),
+            output: .copy
+        )
```
johnxnguyen, line 49: I would expect that if the clipboard is disabled, then I can't "paste" a file by dropping it in the input bar.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
```diff
@@ -59,11 +59,24 @@ extension ConversationInputBarViewController: UIDropInteractionDelegate {
...
     }
 
+    func dropProposal(isText: Bool, isClipboardEnabled: Bool, canFilesBeShared: Bool) -> UIDropProposal {
+        switch (isText, isClipboardEnabled, canFilesBeShared) {
+        case (true, true, _),
+             (false, _, true):
+            return UIDropProposal(operation: .copy)
+        default:
+            return UIDropProposal(operation: .forbidden)
+        }
```
johnxnguyen, line 78: Pattern matching on boolean tuples is bit hard to read. One solution I propose is to have another helper method:

```swift
func dropProposal(isText: Bool, isClipboardEnabled: Bool, canFilesBeShared: Bool) -> UIDropProposal {
  return shouldAllowDropInteraction(isText: isText, isClipboardEnabled: isClipboardEnabled, canFilesBeShared: canFilesBeShared) 
    ? UIDropProposal(operation: .copy) 
    : UIDropProposal(operation: .forbidden)
}

private func shouldAllowDropInteraction(isText: Bool, isClipboardEnabled: Bool, canFilesBeShared: Bool) -> Bool {
  if isText {
    return isClipboardEnabled
  } else {
    return isClipboardEnabled && canFilesBeShared
  }
}
```

You could even maintain a single level of help methods use it directly in the delegate method:

```swift
func dropInteraction(_ interaction: UIDropInteraction, sessionDidUpdate session: UIDropSession) -> UIDropProposal {
  let shouldAllow = shouldAllowDropInteraction(
    isText: session.canLoadObjects(ofClass: NSString.self),
    isClipboardEnabled: SecurityFlags.clipboard.isEnabled,
    canFilesBeShared: canFilesBeShared
  ) 

  return shouldAllow
    ? UIDropProposal(operation: .copy) 
    : UIDropProposal(operation: .forbidden)
}
```

You could then use this new helper method in your tests instead of the current one. Just some ideas.

johnxnguyen (commented):
I think we should also assert that if a file is dropped, the clipboard is enabled, but files are disabled, then it should be forbidden.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
KaterinaWire: I tried to avoid using `func dropInteraction(_ interaction: UIDropInteraction, sessionDidUpdate session: UIDropSession) -> UIDropProposal `in tests because we cannot mock `SecurityFlags.clipboard.isEnabled` and canFilesBeShared (due to` ZMUserSession.shared()`). Also `UIDropSession` is a complex protocol that we really don't need

Wire-iOS Tests/ConversationInputBarViewController/ConversationInputBarViewController+DropInteractionTests.swift
KaterinaWire: Added tests

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
KaterinaWire: Applied your suggestion

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
```diff
@@ -59,11 +59,27 @@ extension ConversationInputBarViewController: UIDropInteractionDelegate {
...
+        ? UIDropProposal(operation: .copy)
+        : UIDropProposal(operation: .forbidden)
+    }
+
+    private func shouldAllowDropInteraction(isText: Bool, isClipboardEnabled: Bool, canFilesBeShared: Bool) -> Bool {
+      if isText {
+        return isClipboardEnabled
+      } else {
+        return isClipboardEnabled && canFilesBeShared
+      }
```
billypchan, line 82:
```suggestion
        guard isClipboardEnabled else { return false }

        return isText || canFilesBeShared
```

billypchan (commented):
syntax suggestion

billypchan:
@KaterinaWire did you tested drag and drop image from Safari or other app cases?

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
KaterinaWire: Your suggestion is shorter, but I think the old version
```
      if isText {
        return isClipboardEnabled
      } else {
        return isClipboardEnabled && canFilesBeShared
      }
```
     is clearer and more readable.

KaterinaWire:
> @KaterinaWire did you tested drag and drop image from Safari or other app cases?

@billypchan Yes, I tested drag&drop from Safari, Image gallery, gmail.

Wire-iOS/Sources/UserInterface/Conversation/InputBar/ConversationInputBarViewController+DragAndDrop.swift
billypchan: `isClipboardEnabled == true` is always required for both condition, so I would prefer extract it.

billypchan (approved):

PR #5267, opened 2021-09-21
Fix: always use landscape mode UI layout on macOS

billypchan:
### Issues

When the iOS app runs Mac OS, conversation list view is not show 

### Causes

The app runs on portrait mode

### Solutions

When the app is running on macOS, always use landscape mode

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.swift
@@ -322,22 +322,19 @@ final class SplitViewController: UIViewController, SplitLayoutObservable {
     }
 
     private var isiOSAppOnMac: Bool {
-        let isiOSAppOnMac: Bool
         if #available(iOS 14.0, *) {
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Components/Controllers/SplitViewController.swift
```diff
@@ -321,18 +321,36 @@ final class SplitViewController: UIViewController, SplitLayoutObservable {
...
     }
 
+    private var isiOSAppOnMac: Bool {
+        let isiOSAppOnMac: Bool
+        if #available(iOS 14.0, *) {
+            if ProcessInfo.processInfo.isiOSAppOnMac {
+                isiOSAppOnMac = true
+            } else {
+                isiOSAppOnMac = false
+            }
```
KaterinaWire, line 331: You can make it shorter:
`isiOSAppOnMac = ProcessInfo.processInfo.isiOSAppOnMac`

David-Henner (approved):

PR #5285, opened 2021-09-29
[Fix] Wrong indicator icon for selfUser in conversation

KaterinaWire:
### Issues
In the group conversation created by the member from another team, the `selfUser` has a guest icon in the conversation.

### Causes
We have displayed a guest icon based on the role in the current conversation. Since the conversation was created by a member of a different team, `selfUser` was considered a guest.

### Solutions
Display guest icon if `selfUser` is a team member.


typfel:
So the guest icon is now shown for all users not in your team. Which conversation we are in is irrelevant now?

typfel (approved):

KaterinaWire:
> Which conversation we are in is irrelevant now?
The conversation is now irrelevant to the display of the guest icon, but we are relying on it to calculate the permissions. For example, adding participants to a conversation.

PR #5373, opened 2021-11-09
Fix: iOS: Username color in the conversation messages view show black colour in dark mode 

agisilaos:
### Issues

This issue happens when you’ve a user from another backend inside a group conversation. e.g., you can look into a group conversation containing Bella and anta federated backend users. Then send messages between such users in a group. So for the federated users the accentColor was either set to `white` in `light mode` and `black` in `dark mode`.

### Causes

That's caused on two different levels. One level is this repo and it's the order of the Coefficients. By having 0.0 first, we were doing some calculations and getting back the wrong color. 

### Solutions

On this level and for this repo, I duplicated the value (0.8) of the Coefficients so we can make sure when color is Undefined, we always get back the .strongBlue color.

The next thing I'm planning to look at is at `wire-ios-request-strategy` and see why we don't get back an accentColor for federatedUsers

### Attachments:

Screenshot 2021-11-09 at 3 27 15 PM

Screenshot 2021-11-09 at 3 54 38 PM

### Notes

https://wearezeta.atlassian.net/browse/SQSERVICES-970


billypchan (commented):
Would you create a snapshot test to prove that this PR solves the issue?
Also we can create a iOS 13+ solution with dynamic color support.

agisilaos:
I don't think a snapshot test would be appropriate here because to simulate such environment would be difficult. I believe it would be more appropriate to write a unit test for the nameColor method as discussed with Jacob.

Wire-iOS Tests/Color/ColorSchemeTests.swift
```diff
@@ -45,4 +45,18 @@ final class ColorSchemeTests: XCTestCase {
         XCTAssertNotEqual(sut.accentColor, alphaBlack)
         XCTAssertFalse(sut.isCurrentAccentColor(alphaBlack))
     }
+
+    func testForIsStrongBlueWhenUndefinedInLightVariant() {
```
billypchan, line 49:
```suggestion
    func testForIsStrongBlue_WhenUndefinedInLightVariant() {
```

Wire-iOS Tests/Color/ColorSchemeTests.swift
```diff
@@ -45,4 +45,18 @@ final class ColorSchemeTests: XCTestCase {
...
+        let expectedAccentColor = UIColor.nameColor(for: .strongBlue, variant: .light)
+        let accentColor = UIColor.nameColor(for: .undefined, variant: .light)
+
+        XCTAssertEqual(accentColor, expectedAccentColor)
+    }
+
+    func testForIsStrongBlueWhenUndefinedInDarkVariant() {
```
billypchan, line 56:
```suggestion
    func testForIsStrongBlue_WhenUndefinedInDarkVariant() {
```

billypchan (commented):
comment

billypchan (approved):

Wire-iOS Tests/Color/ColorSchemeTests.swift
```diff
@@ -45,4 +45,22 @@ final class ColorSchemeTests: XCTestCase {
...
+    func testForIsStrongBlue_WhenUndefinedInLightVariant() {
+        // GIVEN / WHEN
+        let expectedAccentColor = UIColor.white.mix(UIColor(fromZMAccentColor: .strongBlue), amount: 0.8)
+        let accentColor = UIColor.nameColor(for: .undefined, variant: .light)
+
+        // THEN
+        XCTAssertEqual(accentColor, expectedAccentColor)
+    }
+
+    func testForIsStrongBlue_WhenUndefinedInDarkVariant() {
```
KaterinaWire, line 58:
```suggestion
    func testForIsStrongBlue_WhenUndefinedInDarkVariant() {
```

Wire-iOS Tests/Color/ColorSchemeTests.swift
```diff
@@ -45,4 +45,22 @@ final class ColorSchemeTests: XCTestCase {
         XCTAssertNotEqual(sut.accentColor, alphaBlack)
         XCTAssertFalse(sut.isCurrentAccentColor(alphaBlack))
     }
+
+    func testForIsStrongBlue_WhenUndefinedInLightVariant() {
```
KaterinaWire, line 49:
```suggestion
    func testForIsStrongBlue_WhenAccentColorIsUndefinedInLightVariant() {
```

KaterinaWire (approved):

PR #5378, opened 2021-11-10
Fix: xcode13 warning - Using 'class' keyword to define a class-constrained protocol is deprecated; use 'AnyObject' instead

billypchan:
## What's new in this PR?

Replace all protocol extends `class` with `AnyObject `


agisilaos (approved):
🚀

sb88k (approved):

PR #5379, opened 2021-11-10
fix: Empty markdown produces invisible messages SQSERVICES-293

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-293" title="SQSERVICES-293" target="_blank">BugSQSERVICES-293</a>  Empty markdown produces invisible messages
  </td></table>
  <br />
 

### Issues

It's possible to create invisible messages by sending empty markdown such as: {{# }}, {{> }} and ****. These messages create an empty heading, block quote and bold text. Additionally, if the user receives these messages, they will see nothing either.

### Solutions

The aim is to avoid displaying markdown messages that contain no visible characters, and instead falling back to the raw text. We simply check for an empty attributed string.

--

Added some unit tests to cover the cases as well. 

### Notes

https://wearezeta.atlassian.net/browse/SQSERVICES-293


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
```diff
@@ -21,13 +21,31 @@ import Down
...
+    enum Symbol: String, CaseIterable {
+        case empty = ""
+        // For when user types only either #,##,###, *** etc
+        case newline = "\n"
+        // For when user types only *
+        case tabWithDot = "•\t"
+        // For when user types only **
+        case newlineWithStar = "**\n"
+        // For when user types only { }
+        case bracket = "{ }\n"
```
typfel, line 33: Why do we need to these two cases? Doesn't rendered markdown and the original text render to the same string?

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
```diff
@@ -21,13 +21,31 @@ import Down
...
+    }
+
     @objc
     static func markdown(from text: String, style: DownStyle) -> NSMutableAttributedString {
         let down = Down(markdownString: text)
         let result: NSMutableAttributedString
 
         if let attrStr = try? down.toAttributedString(using: style) {
```
typfel, line 41: I think we can simplify this by removing the inner if statement.

something like:

```
if let attrStr = try? down.toAttributedString(using: style), !attrStr.isEmptyMarkdown {
     result = NSMutableAttributedString(attributedString: attrStr)
} else {
     result = NSMutableAttributedString(string: text)
}
```

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
```diff
@@ -21,13 +21,31 @@ import Down
...
-            result = NSMutableAttributedString(attributedString: attrStr)
+            let symbols = Symbol.allCases.map { $0.rawValue }
+
+            if symbols.contains(attrStr.string), !symbols.contains(text) {
```
typfel, line 44: Can you explain what `!symbols.contains(text)` solves/prevents?

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -9564,7 +9556,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = F156509721DCED3000210504 /* CommonComponents-Debug.xcconfig */;
 			buildSettings = {
-				CODE_SIGN_IDENTITY = "";
+				CODE_SIGN_IDENTITY = "Apple Development";
```
typfel, line 9559: You should revert this.

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -9578,7 +9570,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = F156509621DCED3000210504 /* CommonComponents-Release.xcconfig */;
 			buildSettings = {
-				CODE_SIGN_IDENTITY = "";
+				CODE_SIGN_IDENTITY = "Apple Development";
```
typfel, line 9573: You should revert this.

typfel (changes requested):

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
agisilaos: The newLineWithStar method is needed when user just types `**` and the bracket when user types { }

> Doesn't rendered markdown and the original text render to the same string?

Unfortunately it doesn't work like that, without these two cases then many symbols are invisible when sending a message.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
agisilaos: Which are you referring to as inner if statement? The one related to the symbols check?

billypchan:
I would recommend make a patch or extension to `Down` to handle this issue (@johnxnguyen)

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
agisilaos: I don't see how we can simplify the if statement in this case since we need to check for the symbols etc.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
agisilaos: Nothing really. Interesting catch! I left it there based on the thinking of my previous implementation and I thought it was still needed.

agisilaos: Nothing really, I left it there based on the previous implementation I had in mind but it is not needed anymore.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
typfel: In my example above the symbols check would be done inside the new computed property `isEmptyMarkdown`

johnxnguyen (commented):
I think it's worth considering adding this functionality to the Down framework. There you would have access to the syntax tree, which means to you avoid relying on string matching and instead query the tree nodes, and subsequently adjust the rendering output there. This change in behavior could be put behind a configurable setting.

johnxnguyen (commented):
If the aim is to avoid displaying markdown messages that contain no visible characters, and instead falling back to the raw text, then why can't we simply check for an empty attributed string?

```swift
static func markdown(from text: String, style: DownStyle) -> NSMutableAttributedString {
        let down = Down(markdownString: text)
        let result: NSMutableAttributedString

        if let attrStr = try? down.toAttributedString(using: style) {
            result = .init(attributedString: attrStr)
        } else {
            result = NSMutableAttributedString(string: text)
        }

        if result.string.last == "\n" {
            result.deleteCharacters(in: NSMakeRange(result.length - 1, 1))
        }

        guard !result.string.isEmpty else {
            return .init(string: text)
        }

        return result
    }
```

Wire-iOS Tests/NSAttributedString+DownTests.swift
```diff
@@ -0,0 +1,179 @@
...
+
+    func testThatItReturnsH1Character() {
+        // GIVEN
+        let plainTextSymbol = "#"
+
+        // WHEN
+        let sut  = NSMutableAttributedString.markdown(from: plainTextSymbol, style: NSAttributedString.style)
+
+        // THEN
+        XCTAssertEqual(sut, NSMutableAttributedString(string: plainTextSymbol))
```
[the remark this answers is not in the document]
agisilaos: @johnxnguyen Can you elaborate a bit more on that? After messing around and trying to understand what's the best way to test `NSMutableAttributedString`. I found out the the process above is the most accurate.

agisilaos:
@johnxnguyen Regarding your comment: (Since I can't reply directly)

> I don't quite understand how { }\n produces empty markdown, since I don't believe { } corresponds to any markdown syntax (at least that I'm aware of.

For some reason that's the value of the `attrString.string` when user types { }. That's why it's been added as a case. So as a result when you type { } it produces an invisible message.

johnxnguyen:
> For some reason that's the value of the attrString.string when user types { }.

@agisilaos Down may be adding a newline at the end of the string after rendering the document node, which is probably why we're deleting trailing newline characters. So if you type `{ }` it would produce `{ }\n`. But I still don't see how that produces an invisible message, you still see `{ }` (at least I do).

Wire-iOS Tests/NSAttributedString+DownTests.swift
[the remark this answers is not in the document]
johnxnguyen: What exactly do we want to test? The string content and the styling? Or just the string content?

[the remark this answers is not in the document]
agisilaos: content I guess since we're talking about raw output.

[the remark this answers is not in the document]
johnxnguyen: Ok, in that case equating the attributed string might lead to unexpected test results because the styling would also be considered.

Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/Utility/NSAttributedString+Down.swift
agisilaos: After John's comments we simplified it a lot more: 8dfc769

Wire-iOS Tests/NSAttributedString+DownTests.swift
```diff
@@ -173,7 +174,7 @@ final class NSAttributedStringDownTests: XCTestCase {
         let sut  = NSMutableAttributedString.markdown(from: plainTextSymbol, style: NSAttributedString.style)
 
         // THEN
-        XCTAssertEqual(sut, NSMutableAttributedString(string: plainTextSymbol))
+        XCTAssertEqual(sut.string, plainTextSymbol)
```
johnxnguyen, line 177: I don't think you need this test at all, since (I don't believe) it relates to any markdown.

johnxnguyen (changes requested):

johnxnguyen (approved):

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -2647,7 +2648,6 @@
 		A95DFDC024E6D2E20005DC0E /* CGSize+Const.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "CGSize+Const.swift"; sourceTree = "<group>"; };
 		A95E561023FBF0C0005F8E8D /* UIViewControllerContextTransitioning+Helper.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "UIViewControllerContextTransitioning+Helper.swift"; sourceTree = "<group>"; };
 		A961319F23E81CC300194608 /* ParticipantDeviceHeaderView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ParticipantDeviceHeaderView.swift; sourceTree = "<group>"; };
-		A961465F273C0B8D00986596 /* SearchGroupSelectorSnapshotTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SearchGroupSelectorSnapshotTests.swift; sourceTree = "<group>"; };
```
typfel, line 2650: Removing `SearchGroupSelectorSnapshotTests` from the project needs to be reverted?

typfel (changes requested):

typfel (approved):

PR #5380, opened 2021-11-10
Fix: search group selector has no tab bar

billypchan:
### Issues

`SearchGroupSelector` missing tab bar 

### Causes

After removing Cartography `Tabbar`, its height is zero

### Solutions

Add `Tabbar` itself to `prepareForLayout` to remove its auto layout constraints
added snapshot tests to confirm the changes
Update `SearchGroup` to allow mocking `canCreateService`


Wire-iOS Tests/SearchGroup/SearchGroupSelectorSnapshotTests.swift
```diff
@@ -0,0 +1,62 @@
...
+
+import SnapshotTesting
+import XCTest
+@testable import Wire
+
+final class SearchGroupSelectorSnapshotTests: XCTestCase {
+    var sut: SearchGroupSelector!
+
+    override func setUp() {
+        super.setUp()
```
agisilaos, line 27: Shouldn't we initialize the `SearchGroupSelector` here?

billypchan: We have to mock self user `canCreateService` before creating a`SearchGroupSelector` so have to create it in each test.

typfel (approved):

agisilaos (approved):

PR #5385, opened 2021-11-11
Fix: iOS 15 group detail view controller has clear navi bar background

billypchan:
### Issues

On iOS 15, group detail view has a clear background instead of white color

### Causes

iOS 15 changes the default navigation bar behaviour, related discussion
https://developer.apple.com/forums/thread/682420

setting the Navigation Bar's background image is no longer works

### Solutions

Set the background color of navigation view controller for the affected view controllers. I also tried set the UI appearance of navigator bar but it overrides the config of navigation bar(font, tint color...).

Regenerate all snapshot tests and clean up snapshots for removed tests.

### discussion 
We should think about refactor the navigation bar design for iOS 15?


Wire-iOS/Sources/UserInterface/Calling/CallInfoViewController/CallInfoRootViewController.swift
```diff
@@ -56,7 +56,7 @@ final class CallInfoRootViewController: UIViewController, UINavigationController
          selfUser: UserType) {
         self.configuration = configuration
         contentController = CallInfoViewController(configuration: configuration, selfUser: selfUser)
-        contentNavigationController = contentController.wrapInNavigationController()
+        contentNavigationController = contentController.wrapInNavigationController(setBackgroundColor: false)
```
billypchan, line 59:
```suggestion
        contentNavigationController = contentController.wrapInNavigationController()
```

typfel:
@billypchan David is not working today.

typfel:
Why were snapshots deleted?

billypchan:
@typfel I rerun the tests and cleanup the old snapshots.

KaterinaWire (approved):

typfel (approved):

PR #5389, opened 2021-11-11
fix: Image viewer has a double height toolbar

billypchan:
### Issues

Image viewer has a double height toolbar

### Causes

`buttonOuterContainer` height constraint is link to inner container

### Solutions

`buttonOuterContainer` should always has a constant height


Wire-iOS/Sources/UserInterface/Conversation/InputBar/InputBarButtonsView.swift
```diff
@@ -98,7 +98,7 @@ final class InputBarButtonsView: UIView {
             buttonInnerContainer.bottomAnchor.constraint(equalTo: buttonOuterContainer.bottomAnchor),
             buttonRowHeight,
 
-            buttonOuterContainer.heightAnchor.constraint(equalTo: buttonInnerContainer.heightAnchor),
+            buttonOuterContainer.heightAnchor.constraint(equalToConstant: constants.buttonsBarHeight),
```
billypchan, line 101: buttonOuterContainer should always having a constant height, not following buttonInnerContainer's height

typfel (approved):

KaterinaWire (approved):

PR #5388, opened 2021-11-11
Fix: Canvas view has transparent navi bar

billypchan:
## What's new in this PR?

Fix the issue that Canvas view has a transparent navi bar

PR #5395, opened 2021-11-16
Fix: Tab bar and search field missing in add participart view

billypchan:
### Issues

- Search bar and tab bar missing when the user can add guest
- Navi bar is transparent in add participart view

### Causes

Missing a test for this case and Cartograhy removal causes this issue

### Solutions
disable `searchGroupSelector`'s auto layout constraints
add snapshot test for the case user can add a guest
apply the fix from https://github.com/wireapp/wire-ios/pull/5388


David-Henner (commented):
some lint warnings

KaterinaWire (approved):

David-Henner (approved):

PR #5432, opened 2021-11-29
fix: "Wire pasted from..." alert continuously appearing on start screen SQSERVICES-307

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-307" title="SQSERVICES-307" target="_blank">BugSQSERVICES-307</a>  "Wire pasted from..." alert continuously appearing on start screen
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The "Wire pasted from Safari" dialog appears and disappears continuously, even on the login screen, when it detects something on the pasteboard.

### Causes (Optional)

Part of the issue was polling timer which we don't need to use anymore so I removed it entirely. 

### Solutions

We only see it maximum once, when we check the clipboard for an SSO login code, in company login.

### Notes

By working on this issue, I discovered that teamCreation isn't used anywhere in the app. I will open a seperate PR removing such code. I already created a ticket for it in JIRA. 

----
##### References


Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
```diff
@@ -731,7 +731,7 @@ extension AuthenticationCoordinator {
 
     var canStartCompanyLogin: Bool {
         switch stateController.currentStep {
-        case .landingScreen, .provideCredentials, .createCredentials, .reauthenticate, .teamCreation(.setTeamName):
+        case .companyLogin:
```
billypchan, line 734: should we keep these enum, or explain why we only keep `.companyLogin`  ?

agisilaos: we keep only `.companyLogin` because that's where we want to check for the SSO code from the Pasteboard.

billypchan: how about other cases? Do we have tests for this logic?

agisilaos: In other cases we don't/shouldn't check for SSO code.

agisilaos: The canStartCompanyLogin isn't used for any other type of work except of what I mentioned already. I checked if there would be any side effects removing such cases. I also discussed it with @johnxnguyen and he thought as well that this is the way to move forward.

billypchan (approved):

Wire-iOS/Sources/Authentication/Coordinator/AuthenticationCoordinator.swift
johnxnguyen: I'm not sure why we previously allowed starting the company login on steps other than `companyLogin`, maybe it was a way to preemptively start the process? In any case, that decision doesn't appear to be documented anywhere, and I think it make more sense the way it has been changed.

Hopefully any defects will be discovered when QA verifies this. It might be worth checking with QA if their regression test suite covers the SSO login flow.

johnxnguyen (approved):

PR #5434, opened 2021-11-30
refactor: ScreenShare event to work with the new analytics mechanism

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR I transition the screenShare event to work with the new mechanism. In addition to that I fixed an issue regarding a failing test from the previous PR's regarding refactoring analytics. We were calling the `tagEvent` method from the previous implementation before the refactoring. Now we're calling the correct (updated) `tagEvent` method and the failing test has been fixed.

### Notes
Remaining stuff to be done is to transition the `EndCall` event to work with the new mechanism and then clean things up by removing old code etc.

----
##### References

### DIFF
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -165,12 +165,10 @@ extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
         } else if let screenSharedParticipant = participants.first(where: { $0.state.videoState == .stopped && ($0.user as? ZMUser != selfUser) }),
                   let screenSharingDate = screenSharingStartTimes[screenSharedParticipant.clientId],
                   let conversationId = conversation.remoteIdentifier,
-                  let callInfo = callInfos[conversationId] {
+                  let _ = callInfos[conversationId] {
 
             // When videoState == .stopped from a remote participant, tag the event if we found a record in screenSharingInfos set with matching clientId
-            analytics.tag(callEvent: .screenSharing(duration: -screenSharingDate.timeIntervalSinceNow),
-                          in: conversation,
-                          callInfo: callInfo)
+            Analytics.shared.tagEvent(.screenShare(callDirection: .outgoing, duration:  -screenSharingDate.timeIntervalSinceNow, in: conversation))
 
             screenSharingStartTimes[screenSharedParticipant.clientId] = nil
         }
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -163,9 +163,7 @@ extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
            screenSharingStartTimes[participant.clientId] == nil {
             screenSharingStartTimes[participant.clientId] = Date()
         } else if let screenSharedParticipant = participants.first(where: { $0.state.videoState == .stopped && ($0.user as? ZMUser != selfUser) }),
-                  let screenSharingDate = screenSharingStartTimes[screenSharedParticipant.clientId],
-                  let conversationId = conversation.remoteIdentifier,
-                  let _ = callInfos[conversationId] {
+                  let screenSharingDate = screenSharingStartTimes[screenSharedParticipant.clientId] {
 
             // When videoState == .stopped from a remote participant, tag the event if we found a record in screenSharingInfos set with matching clientId
             Analytics.shared.tagEvent(.screenShare(callDirection: .incoming, duration: -screenSharingDate.timeIntervalSinceNow, in: conversation))
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -168,7 +168,7 @@ extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
                   let _ = callInfos[conversationId] {
 
             // When videoState == .stopped from a remote participant, tag the event if we found a record in screenSharingInfos set with matching clientId
-            Analytics.shared.tagEvent(.screenShare(callDirection: .outgoing, duration:  -screenSharingDate.timeIntervalSinceNow, in: conversation))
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -165,12 +165,10 @@ extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
         } else if let screenSharedParticipant = participants.first(where: { $0.state.videoState == .stopped && ($0.user as? ZMUser != selfUser) }),
                   let screenSharingDate = screenSharingStartTimes[screenSharedParticipant.clientId],
                   let conversationId = conversation.remoteIdentifier,
-                  let callInfo = callInfos[conversationId] {
+                  let _ = callInfos[conversationId] {
```
johnxnguyen, line 168: The call info doesn't seem to be needed so I think it's safe to delete this line.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -165,12 +165,10 @@ extension AnalyticsCallingTracker: WireCallCenterCallParticipantObserver {
...
 
             // When videoState == .stopped from a remote participant, tag the event if we found a record in screenSharingInfos set with matching clientId
-            analytics.tag(callEvent: .screenSharing(duration: -screenSharingDate.timeIntervalSinceNow),
-                          in: conversation,
-                          callInfo: callInfo)
+            Analytics.shared.tagEvent(.screenShare(callDirection: .outgoing, duration:  -screenSharingDate.timeIntervalSinceNow, in: conversation))
```
johnxnguyen, line 171: This appears to be screenshares from other participants (not the self user), in which case I would say the the direction would be `incoming`. What do you think?

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -80,8 +80,7 @@ extension Analytics {
             attributes.merge(attributesForVideoToogle(with: callInfo), strategy: .preferNew)
             attributes.merge(["reason": reason], strategy: .preferNew)
         case .screenSharing(let duration):
-            attributes["screen_share_direction"] = "incoming"
-            attributes["screen_share_duration"] = Int(round(duration / 5)) * 5
+            Analytics.shared.tagEvent(.screenShare(callDirection: .incoming, duration: duration, in: conversation))
```
johnxnguyen, line 83: @David-Henner is the `screenSharing` call event related only to the self user? If so, @agisilaos I think that the direction here should actually be `outgoing`.

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -70,4 +75,9 @@ private extension AnalyticsAttributeKey {
     /// Expected to refer to a value of type `AnalyticsCallDirectionType`.
     static let callDirection  = AnalyticsAttributeKey(rawValue: "call_direction")
 
+    /// The duration of the screen-sharing.
+    ///
+    /// Expected to refer to a value of type `AnalyticsScreenShareDurationType`.
```
johnxnguyen, line 80: Do you mean to say?:

"Expected to refer to a value of type `RoundedInt`."

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
agisilaos: Correct, good catch!

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: Looking forward to what @David-Henner has to say.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
agisilaos: But the comment a bit down the line says:

`// When videoState == .stopped from a remote participant, tag the event if we found a record in screenSharingInfos set with matching clientId` 

So we maybe need the callInfo and if all constants match their appropriate values we call the method to track the event.

agisilaos: If I remove this line then I get the same warning that the `let conversationId = conversation.remoteIdentifier`. Should we remove both of those lines?

johnxnguyen: Sure, but I think that info wasn't being stored in `callInfo`, but is what `screenSharingDate` is for. The call info was previously being passed into the tag method so some attributes could be recorded, but we don't need those anymore it seems. 

If we get rid of `callInfo`, then we also don't need `conversationId`, so we can delete them both.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
David-Henner: I'm not familiar with this but as far as I can see this event is only used in `AnalyticsCallingTracker.swift:171`, which is not at all related to the self user. In that case the direction should be `.incoming`

agisilaos: So in line `83` should we leave it as it is? In line `171` I already changed it to .`incoming`

agisilaos: @johnxnguyen But the direction here initially was `.incoming` That's why I left it as it is.

David-Henner: > So in line 83 should we leave it as it is? In line 171 I already changed it to .incoming.

Yes we can leave it

johnxnguyen: ok, sounds fine to me!

johnxnguyen (approved):

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: Perfect! 🤩

sb88k (approved):

PR #5438, opened 2021-11-30
refactor: `CallEnded` event to work with the new analytics event

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This is the last big refactor for a major analytics event to make it work with the new analytics mechanism. @johnxnguyen I will leave some comments for you.

----
##### References

### DIFF
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -130,7 +130,13 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
                 guard let establishedDate = callInfo.establishedDate else { return }
                 let duration = Double(-establishedDate.timeIntervalSinceNow)
 
-                Analytics.shared.tagEvent(.endedCall(asVideoCall: video, callDirection: .outgoing, callDuration: duration, callParticipants: participants, videoEnabled: toggleVideo, screenShareEnabled: screenShare, callEndedReason: .normal, conversation: conversation))
+                Analytics.shared.tagEvent(.endedCall(asVideoCall: video,
+                                                     callDirection: .outgoing,
+                                                     callDuration: duration,
+                                                     callParticipants: participants,
+                                                     videoEnabled: toggleVideo,
+                                                     screenShareEnabled: screenShare,
+                                                     callEndedReason: .normal, conversation: conversation))
 
             }
             callInfos[conversationId] = nil
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -122,7 +122,8 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId],
-            guard let establishedDate = callInfo.establishedDate else { return } {
+               let establishedDate = callInfo.establishedDate {
+                
                 let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
                 let toggleVideo = callInfo.toggledVideo
                 let maximumCallParticipants = callInfo.maximumCallParticipants
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -122,10 +122,10 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId] {
-                let video = conversation.voiceChannel?.isVideoCall ?? false
+                let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
                 let toggleVideo = callInfo.toggledVideo
-                let participants = Double(callInfo.maximumCallParticipants)
-                let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing
+                let maximumCallParticipants = Double(callInfo.maximumCallParticipants)
+                let isSharingScreen = conversation.voiceChannel?.videoState ==  .screenSharing
                 guard let establishedDate = callInfo.establishedDate else { return }
                 let duration = -establishedDate.timeIntervalSinceNow
 
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -121,8 +121,8 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
 
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
         case .terminating(let reason):
-            if let callInfo = callInfos[conversationId] {
-                guard let establishedDate = callInfo.establishedDate else { return }
+            if let callInfo = callInfos[conversationId],
+            guard let establishedDate = callInfo.establishedDate else { return } {
                 let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
                 let toggleVideo = callInfo.toggledVideo
                 let maximumCallParticipants = callInfo.maximumCallParticipants
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -123,7 +123,7 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId],
                let establishedDate = callInfo.establishedDate {
-                
+
                 let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
                 let toggleVideo = callInfo.toggledVideo
                 let maximumCallParticipants = callInfo.maximumCallParticipants
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -120,7 +120,6 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             }
 
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
-            // swiftlint:disable line_length
         case .terminating(reason: let reason):
             if let callInfo = callInfos[conversationId] {
                 let video = conversation.voiceChannel?.isVideoCall ?? false
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -135,7 +135,8 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
                                                      callParticipants: participants,
                                                      videoEnabled: toggleVideo,
                                                      screenShareEnabled: screenShare,
-                                                     callClosedReason: reason, conversation: conversation))
+                                                     callClosedReason: reason,
+                                                     conversation: conversation))
 
             }
             callInfos[conversationId] = nil
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -122,19 +122,19 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId] {
+                guard let establishedDate = callInfo.establishedDate else { return }
                 let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
                 let toggleVideo = callInfo.toggledVideo
-                let maximumCallParticipants = Double(callInfo.maximumCallParticipants)
+                let maximumCallParticipants = callInfo.maximumCallParticipants
                 let isSharingScreen = conversation.voiceChannel?.videoState ==  .screenSharing
-                guard let establishedDate = callInfo.establishedDate else { return }
                 let duration = -establishedDate.timeIntervalSinceNow
 
-                Analytics.shared.tagEvent(.endedCall(asVideoCall: video,
+                Analytics.shared.tagEvent(.endedCall(asVideoCall: isVideoCall,
                                                      callDirection: .outgoing,
                                                      callDuration: duration,
-                                                     callParticipants: participants,
+                                                     callParticipants: maximumCallParticipants,
                                                      videoEnabled: toggleVideo,
-                                                     screenShareEnabled: screenShare,
+                                                     screenShareEnabled: isSharingScreen,
                                                      callClosedReason: reason,
                                                      conversation: conversation))
 
--- a/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
+++ b/Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
@@ -123,9 +123,9 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId] {
                 let video = conversation.voiceChannel?.isVideoCall ?? false
-                let toggleVideo = callInfo.toggledVideo ? true : false
+                let toggleVideo = callInfo.toggledVideo
                 let participants = Double(callInfo.maximumCallParticipants)
-                let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing ? true : false
+                let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing
                 guard let establishedDate = callInfo.establishedDate else { return }
                 let duration = -establishedDate.timeIntervalSinceNow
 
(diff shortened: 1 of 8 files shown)


Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,16 @@ extension Analytics {
...
-            attributes.merge(["reason": reason], strategy: .preferNew)
+        case .ended(reason: _):
+            let video = conversation.voiceChannel?.isVideoCall ?? false
+            let toggleVideo = callInfo.toggledVideo ? true : false
+            let participants = Double(callInfo.maximumCallParticipants)
+            let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing ? true : false
+            guard let establishedDate = callInfo.establishedDate else { return [:] }
+            let duration = Double(-establishedDate.timeIntervalSinceNow)
+
+            Analytics.shared.tagEvent(.endedCall(asVideoCall: video, callDirection: .outgoing, callDuration: duration, callParticipants: participants, videoEnabled: toggleVideo, screenShareEnabled: screenShare, callEndedReason: .normal, conversation: conversation))
```
agisilaos, line 86: @johnxnguyen I don't think that we should provide a reason for why the call ended. I suppose that we get that from the system like we do with some other things like `video`, `toggle video`. That's why I mentioned my `callEndedReason` implementation is not right. What you think?

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -20,6 +20,7 @@ import Foundation
 import UIKit
 import WireDataModel
 
+// swiftlint:disable line_length
```
agisilaos, line 23: This is because our method regarding `endedCall` exceeds 250 characters

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -75,9 +110,33 @@ private extension AnalyticsAttributeKey {
...
 
-    /// The duration of the screen-sharing.
+    /// The duration of the call in seconds.
+    ///
+    /// Expected to refer to a value of type `RoundedInt`.
+    static let callDuration  = AnalyticsAttributeKey(rawValue: "call_duration")
+
+    /// The peak number of participants in the call.
+    ///
+    /// Expected to refer to a value of type `AnalyticsCallParticipantsType`.
```
johnxnguyen, line 120: This should refer to a `RoundedInt`.

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -51,6 +51,41 @@ extension AnalyticsEvent {
...
+        event.attributes[.callDirection] = callDirection
+        event.attributes[.callDuration] = RoundedInt(Int(callDuration), factor: 6)
+        event.attributes[.callParticipants] = RoundedInt(Int(callParticipants), factor: 6)
+        event.attributes[.videoEnabled] = videoEnabled
+        event.attributes[.screenShareEnabled] = screenShareEnabled
+        event.attributes[.callEndedReason] = callEndedReason
+        return event
+    }
+
+    enum ReasonCallEnded: String, AnalyticsAttributeValue {
```
johnxnguyen, line 68: I suggest to call this `CallEndReason`, I think it's simpler and more direct.

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -51,6 +51,41 @@ extension AnalyticsEvent {
         return event
     }
 
+    // swiftlint:disable line_length
+    static func endedCall(asVideoCall: Bool, callDirection: CallDirection, callDuration: Double, callParticipants: Double, videoEnabled: Bool, screenShareEnabled: Bool, callEndedReason: ReasonCallEnded, conversation: ZMConversation) -> AnalyticsEvent {
```
johnxnguyen, line 55: You can split long method signatures over multiple lines to stay within the line length limit. For example:

```swift
static func endedCall(asVideoCall: Bool,
                      callDirection: CallDirection,
                      callDuration: CallDuration,
                      ...,
                      conversation: ZMConversation) -> AnalyticsEvent {
  ...
} 
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: I think it would be better to split long lines over multiple lines. See the comment below.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,16 @@ extension Analytics {
...
 
         switch event {
-        case .ended(reason: let reason):
-            attributes.merge(attributesForSetupTime(with: callInfo), strategy: .preferNew)
-            attributes.merge(attributesForCallDuration(with: callInfo), strategy: .preferNew)
-            attributes.merge(attributesForVideoToogle(with: callInfo), strategy: .preferNew)
-            attributes.merge(["reason": reason], strategy: .preferNew)
+        case .ended(reason: _):
+            let video = conversation.voiceChannel?.isVideoCall ?? false
+            let toggleVideo = callInfo.toggledVideo ? true : false
```
johnxnguyen, line 80:
```suggestion
            let toggleVideo = callInfo.toggledVideo
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,16 @@ extension Analytics {
...
+            let participants = Double(callInfo.maximumCallParticipants)
```
johnxnguyen, line 81: I suggest adding an new init for `RoundedInt` that accepts an integer, so you wouldn't need to do the cast here.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,16 @@ extension Analytics {
...
+            let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing ? true : false
```
johnxnguyen, line 82: @David-Henner is the `videoState` the current video state? And of who, the self user?

We're supposed to be sending  bool indicating that screen sharing was enabled at least once by any user.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: Why shouldn't we provide a value? Or are you wondering, where do we get the reason from?

johnxnguyen: I also don't think this is the right place to tag the event. This enclosing method is for extracting the attributes (in the legacy way). Let's move this to a better place, such as the calling method `tag(callEvent: CallEvent, in conversation: ZMConversation, callInfo: CallInfo)`.

Actually, that goes for other events we're tagging here, such as screen sharing.

johnxnguyen (changes requested):

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: > Why shouldn't we provide a value? Or are you wondering, where do we get the reason from?

I think the latter. Where do we get the reason from? And if we provide a reason from where do we get it? @johnxnguyen

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: How that would look like? @johnxnguyen

agisilaos: If I do

`let duration = RoundedInt(Int(-establishedDate.timeIntervalSinceNow), factor: 6)`

I will have to change the type of my duration from `Double` to `RoundedInt` correct?

agisilaos:
> I also don't think this is the right place to tag the event. This enclosing method is for extracting the attributes (in the legacy way). Let's move this to a better place, such as the calling method tag(callEvent: CallEvent, in conversation: ZMConversation, callInfo: CallInfo). Actually, that goes for other events we're tagging here, such as screen sharing.

@johnxnguyen Let's tackle this one on a separate PR when we're going to clean things up. Good point though.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: The reason is provided as an associated value in the enum case `ended(reason: ...)`. In this switch, it's being discarded, but you could just bind it to a variable and convert it to the type you need.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: Given a the variable 'let isThisBoolTrue: Bool`, then all of these code blocks are equivalent:

```swift
let result = isThisBoolTrue ? true : false
```

```swift
let result: Bool
if isThisBoolTrue {
  result = true
} else {
  result = false
}
```

```swift
let result = isThisBoolTrue
```

In my opinion, the last one is most concise and readable.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: I'm not sure what the existing initializer for `RoundedInt` is, but I'm suggesting that `RoundedInt` should have different initializers with different numeric types so you don't need to cast a number in order to create the `RoundedInt`. For instance:

```swift
extension RoundedInt {
 
  init(_ value: Int) {
    init(Double(value))
  }

}
```

agisilaos: We already have a public extension for `RoundedInt` so I will add it there.

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: So this should be applied for all constants?
```swift
let video = conversation.voiceChannel?.isVideoCall ?? false
let toggleVideo = callInfo.toggledVideo ? true : false
let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing ? true : false
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
agisilaos: So in this case should we leave it as it is?

agisilaos: I would love to know how I can do that. 🤔

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
David-Henner: `videoState` is the current video state of the self user

David-Henner: If you need to observe screen sharing state for each participant it can be done using the `WireCallCenterCallParticipantObserver`

agisilaos: @johnxnguyen thoughts?

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -120,10 +120,23 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
...
+                guard let establishedDate = callInfo.establishedDate else { return }
+                let duration = -establishedDate.timeIntervalSinceNow
+
+                Analytics.shared.tagEvent(.endedCall(asVideoCall: video,
+                                                     callDirection: .outgoing,
+                                                     callDuration: duration,
+                                                     callParticipants: participants,
+                                                     videoEnabled: toggleVideo,
+                                                     screenShareEnabled: screenShare,
+                                                     callClosedReason: reason, conversation: conversation))
```
johnxnguyen, line 138:
```suggestion
                                                     callClosedReason: reason,
                                                     conversation: conversation))
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -51,6 +84,27 @@ extension AnalyticsEvent {
         return event
     }
 
+    static func endedCall(asVideoCall: Bool,
+                          callDirection: CallDirection,
+                          callDuration: Double,
+                          callParticipants: Double,
```
johnxnguyen, line 90:
```suggestion
                          callParticipants: Int,
```

I think an Int is more suitable here.

Wire-iOS/Sources/Analytics/Events/Analytics+CallStateEvents.swift
```diff
@@ -75,9 +129,33 @@ private extension AnalyticsAttributeKey {
...
+    static let screenShareEnabled  = AnalyticsAttributeKey(rawValue: "call_screen_share")
+
+    /// The duration of the screen share in seconds.
     ///
     /// Expected to refer to a value of type `RoundedInt`.
     static let screenShareDuration = AnalyticsAttributeKey(rawValue: "screen_share_duration")
 
+    /// The reason the call ended, according to AVS.
+    ///
+    /// Expected to refer to a value of type `AnalyticsCallEndedReasonType`.
```
johnxnguyen, line 159:
```suggestion
    /// Expected to refer to a value of type `CallClosedReason`.
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
johnxnguyen: Let's see do a pairing session to see if we can get not only the current video state, but if someone shared the screen at least once.

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -120,10 +120,24 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             }
 
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
-
-        case .terminating(reason: let reason):
+        case .terminating(let reason):
             if let callInfo = callInfos[conversationId] {
-                analytics.tag(callEvent: .ended(reason: reason.analyticsValue), in: conversation, callInfo: callInfo)
+                let video = conversation.voiceChannel?.isVideoCall ?? false
```
billypchan, line 125: suggests a clearer name
```suggestion
                let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
```

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -120,10 +120,24 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
...
+                let toggleVideo = callInfo.toggledVideo
+                let participants = Double(callInfo.maximumCallParticipants)
```
billypchan, line 127:
```suggestion
                let maximumCallParticipants = Double(callInfo.maximumCallParticipants)
```

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -120,10 +120,24 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
...
+                let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing
```
billypchan, line 128:
```suggestion
                let isSharingScreen = conversation.voiceChannel?.videoState ==  .screenSharing
```

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -120,10 +120,24 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
...
+                guard let establishedDate = callInfo.establishedDate else { return }
```
billypchan, line 129: put this guard outside the if let to return earlier

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,23 @@ extension Analytics {
         attributes.merge(attributesForDirection(with: callInfo), strategy: .preferNew)
 
         switch event {
-        case .ended(reason: let reason):
-            attributes.merge(attributesForSetupTime(with: callInfo), strategy: .preferNew)
-            attributes.merge(attributesForCallDuration(with: callInfo), strategy: .preferNew)
-            attributes.merge(attributesForVideoToogle(with: callInfo), strategy: .preferNew)
-            attributes.merge(["reason": reason], strategy: .preferNew)
+        case .ended(let reason):
+            let video = conversation.voiceChannel?.isVideoCall ?? false
```
billypchan, line 79:
```suggestion
            let isVideoCall = conversation.voiceChannel?.isVideoCall ?? false
```

Wire-iOS/Sources/Analytics/Events/Analytics+CallEvents.swift
```diff
@@ -74,11 +75,23 @@ extension Analytics {
...
+            let toggleVideo = callInfo.toggledVideo
+            let participants = Double(callInfo.maximumCallParticipants)
+            let screenShare = conversation.voiceChannel?.videoState ==  .screenSharing
+            guard let establishedDate = callInfo.establishedDate else { return [:] }
```
billypchan, line 83: move this to top

billypchan (commented):
code suggestion

Wire-iOS/Sources/Analytics/AnalyticsCallingTracker.swift
```diff
@@ -122,19 +122,19 @@ extension AnalyticsCallingTracker: WireCallCenterCallStateObserver {
             callParticipantObserverToken = WireCallCenterV3.addCallParticipantObserver(observer: self, for: conversation, userSession: userSession)
         case .terminating(let reason):
             if let callInfo = callInfos[conversationId] {
```
billypchan, line 124: It would be cleaner to merge the guard let below with this if let

agisilaos: Hmm how so? What would be the difference?

johnxnguyen (approved):

agisilaos:
@billypchan Took care of all your comments and suggestions. PR's ready for another check.

billypchan (approved):

PR #5472, opened 2021-12-20
fix: SwiftLint warnings for rule `notification_center_detachment`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR I fix warnings for rule and removed it from the disabled rules list:

- **Notification Center Detachment Violation**: An object should only remove itself as an observer in `deinit`. `(notification_center_detachment)`

In addition to that I removed two other rules from the list since by having them enabled don't produce any warnings or errors.

- **Discarded Notification Center Observer**: When registering for a notification using a block, the opaque observer that is returned should be stored so it can be removed later.
- **Legacy CGGeometry Functions**: Struct extension properties and methods are preferred over legacy functions

----
##### References


Wire-iOS/Sources/Authentication/Interface/ViewControllers/AuthenticationStepController.swift
```diff
@@ -111,7 +115,6 @@ class AuthenticationStepController: AuthenticationStepViewController {
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
-        removeObservers()
```
billypchan, line 114: did you tested this code with the original design spec? I think there are some reason for remove observers when view will disappear

agisilaos: > did you tested this code with the original design spec? 

Since I moved `NotificationCenter.default.removeObserver(self)` in `deinit` I got rid of the method because it was empty. Do you think it would be appropriate to bring it back in this case? And just disable the rule for this case?

David-Henner (approved):

Wire-iOS/Sources/Authentication/Interface/ViewControllers/AuthenticationStepController.swift
agisilaos: @billypchan I tested the flow! Works as expected. We don't have a regression. I compared that to how the flow works on the development branch and both have been identical (in terms of how they work). 

I think we're good to go!

agisilaos:
@billypchan Addressed all of your comments, let me know what you think.

billypchan (approved):

PR #5474, opened 2021-12-20
chore: remove iOS 11 or earlier dead code

billypchan:
# What's new in this PR?

### Issues

Since the project is support iOS 12+, the code specific for iOS 11 or lower becomes dead code

### Solutions

Remove code for iOS 11 or below and remove markers for `#available(iOS 11, *)` which is always true.


agisilaos (approved):
🚀 🚀 🚀

sb88k (approved):

PR #5476, opened 2021-12-21
fix: SwiftLint warnings for rule `multiple_closures_with_trailing_closure`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

In this PR I remove multiple_closures_with_trailing_closure rule from the disabled rules list and fix all warnings related to it. 

- **Multiple Closures with Trailing Closure Violation**: Trailing closure syntax should not be used when passing more than one closure argument. `(multiple_closures_with_trailing_closure)`

----
##### References


billypchan:
I would prefer disable this rule since it would save some typing and a close bracket at the end is not clear to read.

agisilaos:
@billypchan I will have to disagree with this. Personally, I found it much more clear regarding to what's happening with the code.

Wire-iOS/Sources/UserInterface/ConnectRequests/ConnectRequestsViewController.swift
```diff
@@ -85,8 +85,8 @@ final class ConnectRequestsViewController: UIViewController, UITableViewDataSour
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         coordinator.animate(alongsideTransition: { _ in
             self.tableView.reloadData()
-        }) { _ in
-        }
+        }, completion: { _ in
+        })
```
KaterinaWire, line 89: I think we can set completion as nil, what do you think?

agisilaos: I think so too. Good catch.

KaterinaWire:
> @billypchan I will have to disagree with this. Personally, I found it much more clear regarding to what's happening with the code.

@agisilaos @billypchan  I don’t have a strong opinion, but if we have disagreements, I think it makes sense to discuss with the whole team and get more opinions.

johnxnguyen (approved):

billypchan (approved):

PR #5479, opened 2021-12-23
refactor: notification service extension SERVICES-1107

billypchan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SERVICES-1107" title="SERVICES-1107" target="_blank">EpicSERVICES-1107</a>  Notification Service Extension
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This is the feature branch of the notification service extension project. With this project, we are transitioning away from using voip pushes and instead adopting standard apns pushes as a way for the application to receive updates from the backend. 

**Some context:** when the backend needs to push an update to the client app, it does so by sending a push containing an data that only hints to the app that it needs to fetch events from the backend. The app is woken in the background and responds by fetching events from the notification stream on the backend. These fetched events may contain encrypted payloads that first need decrypting in order to process them. After processing, that app updates its local state and, in the case of new message events, will schedule local user notifications for the new message.

Even though voip pushes are intended for voip applications, we were using them for non-voip related signalling (such as processing new message events).  Since iOS 13, Apple introduced the restriction that if a voip push is received, then it must result in a call being reported, otherwise iOS may terminate the application.

And so to move away from our dependence on voip pushes, we are implementing a notification service extension that will effectively perform the same behaviour previously mentioned. It works in the following way:

- the app now registers for a standard apns push token
- the backend sends a push
- iOS passes this push to the notification service extension, specifically to the `NotificationService`
- the extension will create a notification session, which is defined in the new wire-ios-notification-engine project
- the session will fetch events from the notification stream in the backend, decrypt the events, store the events for later processing (when the app comes to the foreground), and finally generate any local notification content
- if local notifications content is generated, it is passed back to the notification service, which will then call a content handler that causes iOS to display the notification

### Contained in this PR

- A feature flag to control whether we use legacy (voip) push notifications. We only want to test this internally for a while, so voip pushes will continue to be used in public builds.
- Migration to register for standard apns tokens.
- The notification extension and classes.
- The ability to receive pushes and display notification content returned from the notification session.

### Not contained in this PR

There are still a few things to fix, which will come in subsequent PRs. These include:

- Multi-account support: we still need to create the notification session based on the account id for the push notification.
- Handling some edge cases when we don't follow the happy path (e.g should we display a notification, or nothing?)
- Testing

### Dependencies

Needs releases with:

- https://github.com/wireapp/wire-ios-sync-engine/pull/1620
- https://github.com/wireapp/wire-ios-share-engine/pull/346
- https://github.com/wireapp/wire-ios-request-strategy/pull/313
- https://github.com/wireapp/wire-ios-data-model/pull/1026
- https://github.com/wireapp/wire-ios-mocktransport/pull/223
- https://github.com/wireapp/wire-ios-transport/pull/243
- https://github.com/wireapp/wire-ios-utilities/pull/136

----

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

 - [x] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References


agisilaos:
@johnxnguyen Can you please take care of the SwiftLint warnings before merging?

johnxnguyen (approved):

PR #5485, opened 2022-01-04
fix: support DISABLE_APPCENTER_CRASH_LOGGING flag SQPIT-806

billypchan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQPIT-806" title="SQPIT-806" target="_blank">TaskSQPIT-806</a>  bund app: add flag to disable app center crash logging
  </td></table>
  <br />
 

# What's new in this PR?

add `DISABLE_APPCENTER_CRASH_LOGGING` support. When this flag is set, do not enable app center crash logging feature.


Wire-iOS Share Extension/CrashReporter.swift
```diff
@@ -21,7 +21,10 @@ import WireCommonComponents
 
 import AppCenter
 import AppCenterAnalytics
+#if DISABLE_APPCENTER_CRASH_LOGGING
```
johnxnguyen, line 24: How does this compare to `ExtensionSettings.shared.disableCrashSharing`?

Also, this flag is only preventing `AppCenterCrashes` from being imported, but there are no other changes in this file. Is the import even being used?

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
```diff
@@ -18,7 +18,10 @@
 
 import AppCenter
 import AppCenterAnalytics
+#if DISABLE_APPCENTER_CRASH_LOGGING
+#else
```
johnxnguyen, line 22: Is `AppCenterCrashes` being used at all?

Wire-iOS/Sources/LaunchSequenceOperation.swift
```diff
@@ -171,6 +177,8 @@ extension AppCenterOperation: DistributeDelegate {
     }
 }
 
+#if DISABLE_APPCENTER_CRASH_LOGGING
+#else
```
johnxnguyen, line 181: Below in the delegate method we use `TrackingManager.shared.disableCrashSharing`. Why do we additionally need `DISABLE_APPCENTER_CRASH_LOGGING`?

Wire-iOS/Sources/Components/Settings/SettingsPropertyFactory.swift
billypchan: removed unnecessary import `AppCenterCrashes`

Wire-iOS/Sources/LaunchSequenceOperation.swift
billypchan: I would prefer disable app center crash logging in when building the app. `CrashesDelegate` is provide by appcenter SDK and not unnecessary when do not collecting crash log.

johnxnguyen: My concern is that we have several overlapping ways to disable some code. If we already have logic that disables crash reporting, why don't we continue to use that? This doesn't exclude the possibility for a compile time flag, just perhaps there is only one place where we need to use that compile time flag instead of several.

billypchan: I read the code write to `TrackingManager.shared.disableCrashSharing` and it is for the user to choose enable the crash log sharing in setting. There are several flows may change it (first time logging alert/account setting/AppRootRouter/...). To reduce the testing effort, I would prefer remove the code when building the app.

johnxnguyen (approved):

PR #5507, opened 2022-01-11
Fix: add build time disable file sharing flag SQPIT-847

billypchan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQPIT-847" title="SQPIT-847" target="_blank">TaskSQPIT-847</a>  Bund app: check why EXTERNAL_FILE_PICKER_ACTION_ENABLED/SHARE_EXTENSION_ENABLED are ignored
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

after https://github.com/wireapp/wire-ios/pull/5210, `FILE_GIF_ACTION_ENABLED`, `SAVE_MESSAGE_ENABLED`, `EXTERNAL_FILE_PICKER_ACTION_ENABLED`, `OPEN_FILE_PREVIEW_ENABLED`, `SHARE_EXTENSION_ENABLED` flags no longer works.

### Causes 
These cases are removed from `SecurityFlags`

### Solutions
Add a new ~`canFilesBeShared`~ `FileSharingDisabled` Flag to allow config file sharing related features. (When this flag not set, enable file sharing bu default)

### TODO
- [ ] clean up outdated flags in `Security-Flags.xcconfig`
- [ ] create snapshot tests for input bar with `canFilesBeShared` Flag enabled


johnxnguyen:
The idea with #5210 was that the compile time flag was no longer needed. There is a report that the file sharing restriction doesn't work for a particular customer, do we know why?

billypchan:
@johnxnguyen the last release for the particular customer is after https://github.com/wireapp/wire-ios/pull/5210 is merged, and the initial release is before that PR is merged, since their server does not notice that change and  file sharing restriction is gone.
I would suggest merge this PR as they want the file sharing restriction disabled permanently.

johnxnguyen:
@billypchan so the problem is that their backend hasn't been updated yet to include support for file sharing restrictions via team management? We want to simplify and reduce the number of customization points. The file sharing restrictions feature was built primarily for that customer, so I think we should be focusing on getting it to work for them rather than perform workarounds.

billypchan:
checked that this PR is not necessary, will open a new PR for clean up flags

PR #5512, opened 2022-01-12
feat: support DISABLE_CALL_QUALITY_SURVEY flag SQPIT-845

billypchan:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQPIT-845" title="SQPIT-845" target="_blank">TaskSQPIT-845</a>  Bund app: remove feedback dialog after each call
  </td></table>
  <br />
 

# What's new in this PR?

Support `DISABLE_CALL_QUALITY_SURVEY` flag, it is disabled when it is not defined.


Wire-iOS/Sources/UserInterface/Calling/CallQualityController/CallQualityController.swift
```diff
@@ -113,7 +113,9 @@ class CallQualityController: NSObject {
             return
         }
 
+        #if !DISABLE_CALL_QUALITY_SURVEY
```
johnxnguyen, line 116: I suggest doing this check within `canPresentCallQualitySurvey`, which is defined at line 60 of this file.

billypchan: added checking as your suggestion

johnxnguyen (approved):

agisilaos (approved):

PR #5565, opened 2022-02-16
feat: update classifications banner in call when participants changed FS-100

sb88k:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-100" title="FS-100" target="_blank">Sub-taskFS-100</a>  [iOS] Show classified banner in calls
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Update the classification banner in call status screen when the conversation participants changed.

### DIFF
--- a/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
+++ b/Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
@@ -38,7 +38,8 @@ final class CallViewController: UIViewController {
     fileprivate let hapticsController = CallHapticsController()
     fileprivate var classification: SecurityClassification = .none
 
-    private var observerTokens: [Any] = []
+    private var voiceChannelObserverTokens: [Any] = []
+    private var conversationObserverToken: Any?
     private var callGridConfiguration: CallGridConfiguration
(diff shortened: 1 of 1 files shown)


Wire-iOS/Sources/UserInterface/Calling/CallViewController/CallViewController.swift
```diff
@@ -312,6 +333,14 @@ final class CallViewController: UIViewController {
...
 }
 
+extension CallViewController: ZMConversationObserver {
+    func conversationDidChange(_ changeInfo: ConversationChangeInfo) {
+        guard changeInfo.participantsChanged else { return }
+
+        updateConfiguration()
+    }
+}
+
```
johnxnguyen, line 343: It's not immediately apparent here why we need to update the configuration when the participants change. One way to make the intent clearer would be to recalculate the classification here in this method, then add a `didSet` to `classification` in which you call 
`updateConfiguration()`. What do you think?

sb88k: That makes sense, let me try

johnxnguyen (approved):

David-Henner (approved):

PR #5581, opened 2022-03-01
fix: Syncing Issues between clients by introducing an allowServicesChangedHandler in ConversationServicesOptionsViewModel SQSERVICES-1052

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1052" title="SQSERVICES-1052" target="_blank">StorySQSERVICES-1052</a>  [iOS] Separate access controls for guests and services
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The issue that we had was that the value for `allowServices` didn't get updated when a user was in the `allowServices` Screen and someone else from another client turned on or off `AllowServices`. That issue was discovered during the QA review. 

In contrast: When the user was in Group Details Screen and someone turned on or off the allowServices feature, that value was synced correctly over to the iOS Client.

### Causes (Optional)

The cause for that issue was the fact that we didn't implement a handler in the` allowServices Screen` to make sure the value get's updated and it's in sync properly. 

### Solutions

The solution for this problem is to implement a handler in ConversationServicesOptionsViewModel so we can make sure the value of the `allowServices toggle` is getting updated and synced properly when changed from another client and the selfUser is in `allowServices` screen.

----
##### References


Wire-iOS Tests/ConversationOptionsServicesViewControllerTests.swift
```diff
@@ -84,6 +85,22 @@ final class ConversationServicesOptionsViewControllerTests: XCTestCase {
...
+        let config = MockServicesOptionsViewModelConfiguration(allowServices: false)
+        let viewModel = ConversationServicesOptionsViewModel(configuration: config)
+        let sut = ConversationServicesOptionsViewController(viewModel: viewModel, variant: .light)
+
+        XCTAssertNotNil(config.allowServicesChangedHandler)
+        config.allowServices = true
+        config.allowServicesChangedHandler?(true)
+
+        // THEN
+        verify(matching: sut)
```
johnxnguyen, line 101: Is a snapshot tests really necessary here? Can't we just assert that the toggle value has changed?

agisilaos: We had a snapshot test for the guests as well and I thought it would be a good idea to follow the same pattern.

Wire-iOS/Sources/UserInterface/Conversation Options/ConversationServicesOptionsViewModel.swift
```diff
@@ -63,7 +64,16 @@ final class ConversationServicesOptionsViewModel {
...
         updateRows()
-}
+
+        configuration.allowServicesChangedHandler = { [weak self] allowServices in
+            guard let `self` = self else { return }
+            if allowServices {
+                self.updateRows()
+            } else {
+                self.updateRows()
+            }
```
johnxnguyen, line 74: You can replace this with the equivalent:

```suggestion
            self?.updateRows()
            }
```

Wire-iOS Tests/ConversationOptionsServicesViewControllerTests.swift
johnxnguyen: ok I see now. We're testing that given the value of `config.allowservices`, when `config.allowServicesChangedHandler` is called then the view is updated. Then I suggest you verify the change of the screen by doing something like:

```swift
// Given
let config = ...(allowServices: false)
let sut = ...(config: config)

// Verify that the toggle should be off.
verify(matching: sut)

// When
config.allowServices = true
// confusingly, the value passed here has no affect
config.allowServicesChangedHandler?(true) 

// Then, verify the toggle is now on.
verify(matching: sut)
```

KaterinaWire (approved):

johnxnguyen (approved):

PR #5624, opened 2022-04-08
fix: Broken link to german legal page SQSERVICES-1389

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1389" title="SQSERVICES-1389" target="_blank">BugSQSERVICES-1389</a>  [iOS] Broken link to German legal page
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we aim to fix the issue with the broken German legal page. But that problem also lead us to think how to treat URL's for different languages. Currently on the Wire website we support German and English. 

So when the Wire iOS app is in German we should redirect users to the German version of the website when it comes to "About", "Legal" and many other pages. 

The initial solution was to do something like:

```swift
static var wr_termsOfServicesURL: URL {
    if Locale.autoupdatingCurrent.languageCode == "de" {
        return BackendEnvironment.websiteLink(path: "datenschutz")
    } else {
        return BackendEnvironment.websiteLink(path: "legal")
    }
}
```

But Katerina and I didn't like it that much. The solution implemented in this PR (initially) was to use Localizable.strings. In our case it works fine but it would lead to more issues down the road and it's a good to not mix localisation and URL's.

As the third and final option implemented in this PR, we ended up created a enum `WebsitePages` and a method called

```swift
func localizedWebsiteLink(forPage: WebsitePages) -> URL
```

In that method we check for user's locale and then we pass the appropriate path for our link.

##### References

### DIFF
--- a/Wire-iOS/Sources/Helpers/URL+Wire.swift
+++ b/Wire-iOS/Sources/Helpers/URL+Wire.swift
@@ -181,7 +181,7 @@ extension URL {
     }
 
     static var wr_termsOfServicesURL: URL {
-        return BackendEnvironment.websiteLink(path: "datenschutz")
+        return BackendEnvironment.websiteLink(path: L10n.Localizable.Link.legal)
     }
 
     static var wr_legalHoldLearnMore: URL {
--- a/Wire-iOS/Sources/Helpers/URL+Wire.swift
+++ b/Wire-iOS/Sources/Helpers/URL+Wire.swift
@@ -23,6 +23,11 @@ import WireDataModel
 
 private let zmLog = ZMSLog(tag: "URL")
 
+enum WebsitePages {
+    case termsOfServices
+    case privacyPolicy
+}
+
 enum TeamSource: Int {
     case onboarding, settings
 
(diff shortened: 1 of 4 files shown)


Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -181,7 +181,7 @@ extension URL {
     }
 
     static var wr_termsOfServicesURL: URL {
-        return BackendEnvironment.websiteLink(path: "legal")
+        return BackendEnvironment.websiteLink(path: "datenschutz")
```
KaterinaWire, line 184: L10n.Localizable.Link.legal ?

agisilaos: Just fixed it.

David-Henner (approved):

Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -89,6 +94,17 @@ extension BackendEnvironment {
         return shared.websiteURL.appendingPathComponent(path)
     }
 
+    fileprivate static func localizedWebsiteLink(forPage: WebsitePages) -> URL {
+        switch forPage {
```
KaterinaWire, line 98:
```suggestion
    fileprivate static func localizedWebsiteLink(forPage page: WebsitePages) -> URL {
        switch page {
```

Wire-iOS/Sources/Helpers/URL+Wire.swift
```diff
@@ -23,6 +23,11 @@ import WireDataModel
 
 private let zmLog = ZMSLog(tag: "URL")
 
+enum WebsitePages {
```
KaterinaWire, line 26: Can we make it as a private?

agisilaos: Done!

KaterinaWire (approved):

PR #5674, opened 2022-05-23
fix: The like icon in the toolbar when an image is not liked by the self user IC-54

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/IC-54" title="IC-54" target="_blank">TaskIC-54</a>  The like icon in the toolbar when an image is not liked isn't the correct one
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the icon in the tool bar when we view an image from inside the conversation and that image isn't liked by the self user.

| BEFORE  | AFTER |
|---|---|
| Screenshot 2022-05-23 at 10 51 51  | Screenshot 2022-05-23 at 11 01 25 |

----
##### References


Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/MessageAction.swift
```diff
@@ -109,9 +109,9 @@ enum MessageAction: CaseIterable {
         case .forward:
             return .export
         case .like:
-            return .like
-        case .unlike:
             return .liked
+        case .unlike:
+            return .like
```
johnxnguyen, line 114: The names are confusing... so if the action is to `like` (then the current message is not liked), the icon is `liked` (meaning that the message is liked) ?

Is it that the icons are themselves wrong? i.e an unliked state is an empty heart, and a liked state is a filled heart?

agisilaos: I think the main issue was the name of the icons. So in this case based on the changes I made this is the correct behaviour. But I agree we may need to think of some better names.

So unlike = is not liked

johnxnguyen: Yeah, seems like the icons are incorrectly names in the style kit. But that's a generated file (from PaintCode), so let's not get into that right now.

johnxnguyen (approved):

KaterinaWire (approved):

PR #5697, opened 2022-06-03
fix: color of the navigation bar for the ChangeHandleViewController IC-56

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/IC-56" title="IC-56" target="_blank">TaskIC-56</a>  The color of the ChangeHandleViewController is transparent instead of black
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the color of the navigation bar from transparent to black.

| BEFORE  | AFTER  |
|---|---|
| Screenshot 2022-06-03 at 10 53 06  |  Screenshot 2022-06-03 at 13 31 20  |
----
##### References


BrzozowskiPawel:
Code looks good to me 👍

Wire-iOS/Sources/UserInterface/Settings/SettingsStyleNavigationController.swift
```diff
@@ -18,19 +18,20 @@
...
 import UIKit
+import WireCommonComponents
 
 final class SettingsStyleNavigationController: UINavigationController {
     override func viewDidLoad() {
         super.viewDidLoad()
-
         self.navigationBar.setBackgroundImage(UIImage(color: .black, andSize: CGSize(width: 1, height: 1)), for: .default)
         self.navigationBar.isTranslucent = false
+        self.navigationBar.backgroundColor = .black
```
KaterinaWire, line 28: What about light mode?

agisilaos: There isn't a light mode for that Navigation controller. Only appears when initially want to choose your own handle.

KaterinaWire: I mean is that setting the color to black regardless of the application mode (light/dark) doesn't look correct. Even if the user does not see any changes in light mode, I think it's better to use the background color appreciated for the selected mode.

johnxnguyen: I agree, even if now it it's only in dark mode, doesn't mean it always will be.

PR #5723, opened 2022-06-17
feat: Override userInterfaceStyle ACC-138

KaterinaWire:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/ACC-138" title="ACC-138" target="_blank">StoryACC-138</a>  Contrast distance from text to background (1.4.3)
  </td></table>
  <br />
 

----
#### PR Submission Checklist for internal contributors

- The **PR Title**
  - [x] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [x] contains a reference JIRA issue number like `SQPIT-764`
  - [x] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [x] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

These changes are part of the implementation of dynamic colors. To have dynamic colors and get the current style, we need to use `traitCollection.userInterfaceStyle`. We can also override `userInterfaceStyle` when changing the theme in the app settings. This became available with iOS 13.

----
#### PR Post Submission Checklist for internal contributors (Optional)

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----
#### PR Post Merge Checklist for internal contributors

 - [ ] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

PR #5727, opened 2022-06-21
fix: Color Contrast for Conversation Details Screen ACC-168

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the color contrast issues in conversation details screen.

### TO-DO:
- [ ] Fix tests when #5723 is merged
- [ ] Ask Anette if we should also change the background color of the header/footer and the cells.

----
##### References


WireCommonComponents/SemanticColors.swift
```diff
@@ -19,9 +19,10 @@
 import UIKit
 
 enum SemanticColors {
-
-    static let buttonBackground = UIColor(light: Asset.red200Light, dark: Asset.green500Dark)
-
+    static let footerLabelConversationDetails = UIColor(light: Asset.gray90, dark: Asset.gray20)
+    static let headerLabelConversationDetails = UIColor(light: Asset.gray70, dark: Asset.gray50)
+    static let textLabelTitleActive = UIColor(light: Asset.black, dark: Asset.white)
+    static let textLabelSubtitleActive = UIColor(light: Asset.gray90, dark: Asset.white)
```
agisilaos, line 25: I've named them like that because they're being used all across the conversation details screen and more specifically in the cells. It wouldn't make a good case to have them with a more specific name. I'm open to suggestions.

Wire-iOS/Sources/UserInterface/Components/Views/DetailsCollectionViewCell.swift
```diff
@@ -61,9 +61,7 @@ class DetailsCollectionViewCell: SeparatorCollectionViewCell, DynamicTypeCapable
     }
 
     var disabled: Bool = false {
-        didSet {
-            updateDisabledState()
-        }
+        didSet { }
     }
```
agisilaos, line 65: This can't be deleted just yet. Because it's gets overriden in another place

WireCommonComponents/SemanticColors.swift
```diff
@@ -19,9 +19,10 @@
 import UIKit
 
 enum SemanticColors {
-
-    static let buttonBackground = UIColor(light: Asset.red200Light, dark: Asset.green500Dark)
-
+    static let footerLabelConversationDetails = UIColor(light: Asset.gray90, dark: Asset.gray20)
+    static let headerLabelConversationDetails = UIColor(light: Asset.gray70, dark: Asset.gray50)
+    static let textLabelTitleCellActive = UIColor(light: Asset.black, dark: Asset.white)
+    static let textLabelSubtitleCellActive = UIColor(light: Asset.gray90, dark: Asset.white)
```
johnxnguyen, line 25: What is the convention here regarding the name of a semantic color?

Wire-iOS/Sources/Components/Styles/LabelStyle.swift
```diff
@@ -22,6 +22,11 @@ import UIKit
 public struct LabelStyle {
     var backgroundColor: UIColor
     var textColor: UIColor
+
+    static let footerLabelStyle: Self = LabelStyle(backgroundColor: .clear, textColor: SemanticColors.footerLabelConversationDetails)
+    static let headerLabelStyle: Self = LabelStyle(backgroundColor: .clear, textColor: SemanticColors.headerLabelConversationDetails)
+    static let primaryLabelInCellStyle: Self  = LabelStyle(backgroundColor: .clear, textColor: SemanticColors.textLabelTitleCellActive)
+    static let secondaryLabelInCellStyle: Self  = LabelStyle(backgroundColor: .clear, textColor: SemanticColors.textLabelSubtitleCellActive)
```
johnxnguyen, line 29: Is there a convention here for style names?

WireCommonComponents/SemanticColors.swift
agisilaos: Regarding the semantic colors we're trying to follow this convention. Any feedback is much appreciated. 

[Usage][UI Element][Color role][State / Interaction]

- Usage:  where to use (background, text, border, etc.)
- UI Element: button, label, switch, link, alert, search, etc.
- Color role: placeholder, header, link, verifyDevice, etc. 
- State/Interaction: active, hightLighted, disabled, selected, etc.

Wire-iOS/Sources/Components/Styles/LabelStyle.swift
agisilaos: I don't think we discussed a specific name convention. Again any feedback is much appreciated. I want for us to have a clear convention from the start so the job can become easier as we go on.

WireCommonComponents/SemanticColors.swift
johnxnguyen: I don't quite see the link between the convention you've stated and the names you've chosen.

Wire-iOS/Sources/Components/Styles/LabelStyle.swift
johnxnguyen: In the same sense that there should be some rules about semantic color names, I think there should be some rules about how styles are named too.

I think it could also be much more readable, for instance:

```swift
let footerText: Self = ...
let headerText: Self = ...
let primaryCellLabel: Self = ...
let secondaryCellLabel: Self = ...
```

WireCommonComponents/SemanticColors.swift
agisilaos: Done! Fixed, please take a look and let me know what you think.

Wire-iOS/Sources/Components/Styles/LabelStyle.swift
agisilaos: Done! Fixed, please take a look and let me know what you think.

Wire-iOS Tests/LegalHoldDetailsViewControllerSnapshotTests.swift
```diff
@@ -20,7 +20,8 @@ import XCTest
 import SnapshotTesting
 @testable import Wire
 
-final class LegalHoldDetailsViewControllerSnapshotTests: XCTestCase {
+@available(iOS 13.0, *)
```
johnxnguyen, line 23: Did you try bumping the test target to iOS 13? You may not need to have all these availability annotations.

Wire-iOS/Sources/UserInterface/Components/Views/DetailsCollectionViewCell.swift
```diff
@@ -61,9 +61,7 @@ class DetailsCollectionViewCell: SeparatorCollectionViewCell, DynamicTypeCapable
...
 
     var disabled: Bool = false {
-        didSet {
-            updateDisabledState()
-        }
+        didSet { }
```
johnxnguyen, line 64: This does nothing so you might as well delete it.

WireCommonComponents/SemanticColors.swift
johnxnguyen: I'm still not sure that it's the right way to name things but it's fine for now.

johnxnguyen (approved):

ghuMustafa (approved):

PR #5776, opened 2022-07-12
chore: improve nse debugging

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

We still have issues with the NSE not displaying notifications for periods of time. It's hard to know where the failure is happening because 1) notifications are just difficult to debug without a debugger, 2) the NSE is a separate process and our logging tools don't work in it, 3) it not easy to reproduce the issue.

### Solutions

Add two developer flags:

1. `nseDebugEntryPoint`, if on, will make the NSE display the contents of the push it receives at its entry point immediately and then return. This is useful to check whether the problem is if the NSE is receiving the push at all (eg. due to bad network) or if it's to do with our processing of the push.2. 
2. `nseDebugging`, if on, will display notifications with debug messages in all the failure points of the push processing, which would have otherwise resulted in a silent failure. This will be helpful to isolate specific code paths.3. 

### Notes 

These are just some attempt to isolate the cause(s) of the issues we're experiencing. Even if the debug notifications aren't being displayed, but we're sure that the entry point is being hit, then it would suggest something else going wrong such as memory budget violations or crashes.

----

----

----
##### References

### DIFF
--- a/Wire-iOS.xcodeproj/project.pbxproj
+++ b/Wire-iOS.xcodeproj/project.pbxproj
@@ -1218,7 +1218,6 @@
 		DBF7235F1B5566B80036F4C5 /* a_lot_of_people.json in Resources */ = {isa = PBXBuildFile; fileRef = DBF7235E1B5566B80036F4C5 /* a_lot_of_people.json */; };
 		E52D5A7B2858C18600D0494E /* Colors+Generated.swift in Sources */ = {isa = PBXBuildFile; fileRef = E52D5A7A2858C18600D0494E /* Colors+Generated.swift */; };
 		E52F448B2844E916008E11CB /* UITableViewCell+DisclosureIndicator.swift in Sources */ = {isa = PBXBuildFile; fileRef = E52F448A2844E916008E11CB /* UITableViewCell+DisclosureIndicator.swift */; };
-		E5436FAE285A1C83004D95EB /* SemanticColors.swift in Sources */ = {isa = PBXBuildFile; fileRef = E56CA3762858A36500CD2045 /* SemanticColors.swift */; };
 		E5443393286192CF00F2E0ED /* SwitchStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = E5443392286192CF00F2E0ED /* SwitchStyle.swift */; };
 		E56CA37528588C4F00CD2045 /* Colors.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = E56CA37428588C4F00CD2045 /* Colors.xcassets */; };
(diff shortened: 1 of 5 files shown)


Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -8013,7 +8015,6 @@
 				5E144FFF213EA0EC00B65E14 /* PreviewImageView.swift in Sources */,
 				5EDD2538213FD5E200C0E254 /* SLComposeServiceViewController+Preview.swift in Sources */,
 				06F9864625CC146B005572C5 /* UIView+ConstraintFactory.swift in Sources */,
-				06717E5D2865183600AF442C /* Stylable.swift in Sources */,
```
agisilaos, line 8016: @johnxnguyen Not sure why this keeps happening, but we shouldn't delete those files. Happened in my PR too.

Wire-iOS.xcodeproj/project.pbxproj
```diff
@@ -1218,7 +1218,6 @@
 		DBF7235F1B5566B80036F4C5 /* a_lot_of_people.json in Resources */ = {isa = PBXBuildFile; fileRef = DBF7235E1B5566B80036F4C5 /* a_lot_of_people.json */; };
 		E52D5A7B2858C18600D0494E /* Colors+Generated.swift in Sources */ = {isa = PBXBuildFile; fileRef = E52D5A7A2858C18600D0494E /* Colors+Generated.swift */; };
 		E52F448B2844E916008E11CB /* UITableViewCell+DisclosureIndicator.swift in Sources */ = {isa = PBXBuildFile; fileRef = E52F448A2844E916008E11CB /* UITableViewCell+DisclosureIndicator.swift */; };
-		E5436FAE285A1C83004D95EB /* SemanticColors.swift in Sources */ = {isa = PBXBuildFile; fileRef = E56CA3762858A36500CD2045 /* SemanticColors.swift */; };
```
agisilaos, line 1221: @johnxnguyen Not sure why this keeps happening, but we shouldn't delete those files. Happened in my PR too.

johnxnguyen: I think this file was moved from Wire-iOS to WireCommonComponents. I checked the project and it exists in common components, so I don't think this is a problem.

agisilaos: I don't think that's correct. I can see the file in Wire-iOS and not in WireCommonComponents. 🤔

johnxnguyen: You're right, the files are in Wire-iOS. The it's likely the other way around.... that these are "deleted" from WireCommonComponent.

agisilaos (approved):

KaterinaWire (approved):

PR #5818, opened 2022-08-26
fix: Add participant to a group SQSERVICES-1690

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues
Some conversations created by private users have wrong access roles.

### Causes (Optional)
For private users we do not show the screen where they can turn on or off the option to add guests/ services. That's why we use the default value of accessRoles when create the group conversation. For some reason the default value is `[.teamMember]`. As a result, the conversation doesn't have a team ID, but only team members can be added according to accessRoles. This is the cause of the BE error when trying to add new members.

### Solutions
Change the  default value of `accessRoles` if team == nil.

Needs releases with:

- [ wire-ios-data-model/pull/1398 ] 
- [ wire-ios-request-strategy/pull/559 ]
- [ wireapp/wire-ios-sync-engine/pull/1774 ]

----

----

----
##### References


Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
```diff
@@ -217,8 +217,31 @@ enum DebugActions {
 
     /// Accepts a debug command
     static func enterDebugCommand(_ type: SettingsCellDescriptorType) {
-        askString(title: "Debug command") { _ in
-            alert("Command not recognized")
+        askString(title: "Debug command") { str in
+            if str == "432.1.0" {
+                DebugActions.updateInvalidAccessRolesForExistingConversations()
```
johnxnguyen, line 222: I suggest to have an enum of the possible commands, something like:

```swift
enum DebugCommand {

  // Comment explaining what this does.
  case repairInvalidAccessRoles

  init?(string: String) {
    // We may want to have commands that accept arguments, which means
   // we'd have to do the parsing of the command here.
    switch string {
      case "repairInvalidAccessRoles":
        self = .repairInvalidAccessRoles
      default:
        return nil
    }
  }
}
```

You could then just switch over the command:

```swift
guard let command = DebugCommand(string: string) else { 
  showErrorAlert()
  return
 }

switch command {
  case .repairInvalidAccssRoles:
    DebugActions.updateInvalidAccessRolesForExistingConversations()
}

```

KaterinaWire: Done

johnxnguyen (approved):

Wire-iOS/Sources/UserInterface/Settings/DebugActions.swift
```diff
@@ -217,8 +217,36 @@ enum DebugActions {
...
+            case .repairInvalidAccessRoles:
+                DebugActions.updateInvalidAccessRoles()
+            }
+
+        }
+    }
+
+    static func updateInvalidAccessRoles() {
+        guard let userSession = ZMUserSession.shared() else { return }
+        let predicate = NSPredicate(format: "team == nil AND accessRoleStringsV2 == %@",
```
David-Henner, line 236: If possible, using  `#keyPath(ZMConversation.team)` and `#keyPath(ZMConversation.accessRoleStringsV2)` should be a more robust way to write the keys

KaterinaWire: Done

David-Henner (approved):

PR #5875, opened 2022-09-26
chore: add flag to break notifications

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We suspect that people still suffering from notification issues is because their app is in a bad state where iOS is no longer delivering notifications. To test this hypothesis, we're adding a flag to intentionally break the notification service. If the flag is on, we will not do anything in the service, which hopefully after receiving and ignoring enough pushes, the app will no longer show any notifications.

----

----

----
##### References

### DIFF
--- a/Wire Notification Service Extension/NotificationService.swift
+++ b/Wire Notification Service Extension/NotificationService.swift
@@ -33,7 +33,11 @@ public class NotificationService: UNNotificationServiceExtension{
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) {
-        if DeveloperFlag.useSimpleNSE.isOn {
+        if DeveloperFlag.breakMyNotifications.isOn {
+            // By doing nothing, we hope to get in a state where iOS will no
+            // longer deliver pushes to us.
+            return
+        } else if DeveloperFlag.useSimpleNSE.isOn {
             simpleService.didReceive(
                 request,
                 withContentHandler: contentHandler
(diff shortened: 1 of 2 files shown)


Wire Notification Service Extension/NotificationService.swift
```diff
@@ -33,7 +33,11 @@ public class NotificationService: UNNotificationServiceExtension{
         _ request: UNNotificationRequest,
         withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void
     ) {
-        if DeveloperFlag.useSimpleNSE.isOn {
```
marcoconti83, line 36: The `breakMyNotifications.isOn` should come after the `useSimpleNSE`. If we turn `useSimpleNSE` on we should turn off `breakMyNotifications`, which we can do by changing the `if` to:

```
if DeveloperFlag.useSimpleNSE.isOn {
...
} else if DeveloperFlag.breakMyNotifications.isOn {
...
} else {
//legacy
}
```

marcoconti83 (changes requested):

marcoconti83:
On a second thought, the way that you implemented this works too... it's more steps for the tester to get right (so more chances for human error) but it also gives us the most flexibility.

marcoconti83 (approved):

Wire Notification Service Extension/NotificationService.swift
johnxnguyen: The way I imagine how it would be used:

- install app, legacy notifications work 
- turn on my notifications, let it run for a while.
- turn off my notifications, verify that notifications are broken.
- turn on the simple notification service, see if it works.

PR #5931, opened 2022-11-01
fix: Remove uppercase from the placeholder in `MessageTextView`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we remove the uppercase text transform from the placeholder in Message Text field and in addition to that we removed the expanded separator from `ConversationCellBurstTimestampView`

----
##### References


KaterinaWire (approved):

David-Henner (commented):
Are we ok with the text being truncated? (it happens in several snapshots)

Screenshot 2022-11-04 at 10 34 51

agisilaos:
@David-Henner Yeah bc we increased the font size of the text.

David-Henner:
> @David-Henner Yeah because we increased the font size of the text.

Right, but shouldn't we at least make sure it's truncated right? like `self-deleting mess...` instead of the text being cut off a bit randomly?

agisilaos:
An interesting idea. I will add it as a todo on the list I'm creating with the fixes we need to make after we finish with the project. But you're right we need to take that into consideration.

David-Henner (approved):

PR #5937, opened 2022-11-03
fix: Font size of navigation bar title in the contacts tab

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the font size of navigation bar title in the contacts tab.

----
##### References


Wire-iOS/Sources/Helpers/UINavigationItem+NavigationBarTitle.swift
```diff
@@ -20,7 +20,7 @@ import UIKit
 
 extension UINavigationItem {
 
-    func setupNavigationBarTitle(title: String) {
+    func setupNavigationBarTitle(title: String?) {
```
johnxnguyen, line 23: Why is this change needed?

agisilaos: @johnxnguyen If I tried to use a guard let statement in `StartUIViewController.swift` where we set the time then I would have problems in tests where it would generate nothing no title, no content for some reason. Not sure if that's the right approach and that was one of the things I wanted to discuss with you.

agisilaos: @johnxnguyen So If I do something like this in `StartUIViewController`, it doesn't work in tests

```
if let team = (selfUser as? ZMUser)?.team {
            guard let title = team.name?.capitalized else { return }
            navigationItem.setupNavigationBarTitle(title: title)
        } else {
            guard let title = selfUser.name?.capitalized else { return }
            navigationItem.setupNavigationBarTitle(title: title)
        }
```

Screenshot 2022-11-04 at 10 20 32

KaterinaWire (approved):

johnxnguyen (approved):

PR #5941, opened 2022-11-04
fix: Border Color of `InviteButton` when highlighted/isSelected. 

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the border color of invite button when highlighted/isSelected. Before the changes the border color would be black and that's because we wouldn't override the `isHighlighted`. Now that we do the `InviteButton`'s border  has the correct color.

----
##### References


Wire-iOS/Sources/UserInterface/Components/Views/InviteButton.swift
```diff
@@ -25,9 +25,15 @@ final class InviteButton: IconButton {
         super.init()
 
         clipsToBounds = true
-        titleLabel?.font = FontSpec(.normal, .semibold).font
+        titleLabel?.font = FontSpec.normalSemiboldFont.font!
```
ghuMustafa, line 28: Here are we sure that we will always get the font instead of an optional nil?
And can we write it as below ?
```suggestion
        titleLabel?.font = .normalSemiboldFont
```

agisilaos: Nope,  we can't because we're trying to get rid of the that file where that property exists. That's why we're using FontSpec.normalSemiboldFont.font!

ghuMustafa: Thanks I got it.

ghuMustafa (approved):

KaterinaWire (approved):

PR #5949, opened 2022-11-10
fix: Fonts in Account Screen - ACC - 259

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR we fix the fonts in the Account Screen. 

----
##### References


Wire-iOS/Sources/UserInterface/Availability/AvailabilityStringBuilder.swift
```diff
@@ -72,7 +72,7 @@ final class AvailabilityStringBuilder: NSObject {
         switch size {
         case .small:
             verticalCorrection = -1
-        case .medium, .large, .normal, .header:
+        default:
```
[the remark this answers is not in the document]
agisilaos: Done!

WireCommonComponents/FontScheme.swift
```diff
@@ -30,6 +30,8 @@ public enum FontSize: String {
     case medium
     case small
     case header
+    case accountName
+    case accountTeam
```
johnxnguyen, line 34: This goes against the design of the font size, which only contains abstract sizes, rather than specific uses. Why do you need to add two new sizes only for the account name and team name?

agisilaos: Maybe the naming is off, I agree but we didn't have that font size with the current set of font sizes and we needed to add them somehow. Any suggestions would be more than appreciated!

johnxnguyen: Where are these font sizes coming from exactly? Figma?

agisilaos: Yes!

KaterinaWire (approved):

Issue #5655, opened 2022-05-03
Modal transitions are based on percentage of completion, not on direction / momentum.


elland:
By relying exclusively on percentage of the screen traversed by the gesture, it results in an inferior interaction model, whereas the user would expect a dismissal to be completed, based on the direction and momentum of the gesture, and not on how much of the screen has been covered by the movement.

https://github.com/wireapp/wire-ios/blob/a7c106e71ed0191809b8343f3923a47a901e4f7d/Wire-iOS/Sources/UserInterface/Helpers/InteractiveModalTransition.swift#L134

For examples of this working correctly, swiping from a conversation into the chat list in the following apps I tested worked as expected:

- Telegram
- Messages
- Whatsapp

May I suggest using https://developer.apple.com/documentation/uikit/uiscreenedgepangesturerecognizer instead?

## Checklist (please check before submitting)

* [x] I checked the existing issues, and the issue was not reported before.

## The problem

Briefly describe the issue you are experiencing (or the feature you want to see added to Wire).
Tell us what you were trying to do and what happened instead.

## Environment

* Wire version that exhibits the issue:
* Last Wire version that did not exhibit the issue (if applicable):
* iOS version used to run Wire:
* iOS device type: 
* Mobile network type (EDGE/LTE/Wi-Fi/Offline):
* Exact date, time and timezone (only for the crash issues): 

## Details

Describe the steps necessary to reproduce the problem in more details. The detailed description
will help us to fix the issue faster.

## Calling Debug logs (only for calling issues)

0. You have to have the iOS email client set up.
1. Reproduce the issue.
2. Go to Settings -> Advanced -> Calling Debug Report.
3. Send the report.

## Screenshots

Please attach screenshots if you consider them as helpful to understand/reproduce the issue.

elland:
Fixed by #5994

PR #5994, opened 2022-12-13
fix: Account for velocity instead of percetange to account for user intention when dismissing conversation view.

elland:
Addresses #5655


agisilaos (approved):
LGTM. @elland :)

johnxnguyen (approved):

PR #6009, opened 2022-12-17
chore: Get rid of `smallFont` from `UIFont+FontSpec.swift`

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Since we started implementing accessibility changes into the app, we decided at some point to get rid of the `UIFont+FontSpec` file. With this PR we're one step closer to that direction. The first part is to remove the `smallFont` from that file and replace it with code from `FontSpec.swift`

----
##### References


David-Henner (approved):

Wire-iOS/Sources/UserInterface/CustomAppLock/Setup/PasscodeError.swift
```diff
@@ -59,10 +59,10 @@ enum PasscodeError: CaseIterable {
     }
 
     var descriptionWithInvalidIcon: NSAttributedString {
-        return description(icon: .circleCross, color: SemanticColors.Icon.backgroundPasswordRuleCheck, font: .smallRegularFont)
+        return description(icon: .circleCross, color: SemanticColors.Icon.backgroundPasswordRuleCheck, font: FontSpec.smallRegularFont.font!)
```
KaterinaWire, line 62: Why didn't you want to replace `.smallRegularFont` ?

class var smallRegularFont: UIFont {
         return  FontSpec.smallRegularFont.font!
 }

agisilaos: I did replace it with FontSpec.smallRegularFont.don't!

KaterinaWire: I meant why do we need to remove `smallRegularFont`?  Why not to keep smallRegularFont, but replace the return value.

KaterinaWire: Then we don't need to change the code in many files, just in one.

agisilaos: Yeah, we do because we try to use a fontSpec in many places as we can. We want to get rid of that file. And instead, we use properties from FontSpec.swift file. Makes sense?

johnxnguyen: The `UIFont` extension is designed to make it easier to select certain fonts, I'm not sure I see the benefit in making these changes, it's just creating more duplication, or am I missing something?

agisilaos: @johnxnguyen We discussed it a while ago together and we came to the conclusion that eventually we want to get rid of that file and use the FontSpec.swift instead. Why? Because eventually, we wanted to use fontSpec everywhere. If you think that would be a waste of time I would be more than happy to close that PR.

PR #6047, opened 2023-01-05
fix: crashes when opening app to answer a call FS-1344

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1344" title="FS-1344" target="_blank">BugFS-1344</a>  [iOS] [Bund Col1] App crash upon starting call
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Given CallKit is disabled and I receive an incoming call while the app is the background, when I bring the app to the foreground, the app will sometimes crash.

### Causes

There seem to be a couple different crashes:
- Trying to access the self user when it is not yet available (in analytic events).
- Trying to constraint a call top overlay to an unrelated view.

The underlying cause of the top overlay crash is because:

- we try to present it twice, 
- because we have two instances of `CallController` that are listening to call state changes, 
- because we have two instances of `ActiveCallRouter`, 
- because there is a retain cycle between the `ActiveCallRouter` and the `NotificationCenter`, 
- because when we transition to the authenticated state we recreate a new `ActiveCallRouter` each time.

### Solutions

We can fix the self user crashes by safely unwrapping values.

We fix the call overlay issue by removing the retain cycle in `ActiveCallRouter`.

#### How to Test

- Have EAR enabled (helps reproduce the issue)
- Disabled CallKit
- Put the app in the background
- Receive an incoming call
- Open the app

----

 - [x] Wire's Github Workflow has automatically linked the PR to a JIRA issue
----

----
##### References

### DIFF
--- a/Wire-iOS/Sources/AppRootRouter.swift
+++ b/Wire-iOS/Sources/AppRootRouter.swift
@@ -413,7 +413,6 @@ extension AppRootRouter {
             presentAlertForDeletedAccountIfNeeded(error)
             sessionManager.processPendingURLActionDoesNotRequireAuthentication()
         case .authenticated:
-            authenticatedRouter?.updateActiveCallPresentationState()
             urlActionRouter.authenticatedRouter = authenticatedRouter
             ZClientViewController.shared?.legalHoldDisclosureController?.discloseCurrentState(cause: .appOpen)
             sessionManager.processPendingURLActionRequiresAuthentication()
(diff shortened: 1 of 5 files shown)


Wire-iOS/Sources/AppRootRouter.swift
```diff
@@ -413,7 +413,6 @@ extension AppRootRouter {
             presentAlertForDeletedAccountIfNeeded(error)
             sessionManager.processPendingURLActionDoesNotRequireAuthentication()
         case .authenticated:
-            authenticatedRouter?.updateActiveCallPresentationState()
```
johnxnguyen, line 416: This is not needed because the `CallController` will invoke this method when it receives new call states. Not sure what the purpose of this was, but having it here causes double presentation of the call ui.

KaterinaWire (approved):

David-Henner (approved):

PR #6116, opened 2023-02-10
fix: calling UI layout issues FS-1501

MarcinR:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-1501" title="FS-1501" target="_blank">BugFS-1501</a>  [iOS] Calling: not matching avatar is shown when receiving non-UI Kit call with different display name
  </td></table>
  <br />
 

- The **PR Title**

- The **PR Description**

### DIFF
--- a/Wire-iOS/Resources/Base.lproj/Localizable.strings
+++ b/Wire-iOS/Resources/Base.lproj/Localizable.strings
@@ -650,7 +650,7 @@
 "voice.cancel_button.title" = "Cancel";
 "voice.call_button.title" = "Call";
 "voice.end_call_button.title" = "End Call";
-"voice.pick_up_button.title" = "Accept";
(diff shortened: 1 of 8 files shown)


Wire-iOS/Resources/Base.lproj/Localizable.strings
```diff
@@ -650,7 +650,7 @@
 "voice.cancel_button.title" = "Cancel";
 "voice.call_button.title" = "Call";
 "voice.end_call_button.title" = "End Call";
-"voice.pick_up_button.title" = "Accept";
+"voice.pick_up_button.title" = "Pick Up";
```
agisilaos, line 653: I think we intentionally replaced Pick Up with Accept after talking with Astrid and Anette. Please revert that change

agisilaos (approved):

